PR #35, opened 2024-04-30
Don't send events or service checks

blt:
This commit removes the emission of events and service-checks in the Regression Detector experiments. They're not appropriate for the targets.


blt:
Rebased with https://github.com/DataDog/saluki/pull/37 in the history.

tobz (approved):

PR #41, opened 2024-05-08
env: add workload provider based on Agent remote tagger

tobz:
## Context

In Saluki, we have the concept of an "environment provider" which is meant to be a pluggable way to provide information about the environment in which a Saluki-based application is running, such as getting information about the host or tags for a workload entity.

For origin enrichment, we depend on being able to query the workload for tags about the entity that sent us the metrics, if we can properly extract an entity ID from them. This is generally something like the container ID or the PID of the process _in_ a container that sent the metrics.

A workload provider is meant to take an entity ID and find all tags for that entity. The Datadog Agent handles this through a combination of components called the "tagger" and "workloadmeta". In order to provide the same functionality, Saluki needs a workload provider that can do the same.

## Solution

This PR adds a new workload provider based on the Datadog Agent's tagger API. While the Datadog Agent does the hard work of collecting information about containers, and Kubernetes resources, and other things, this provider simply streams entity updates from the Agent and then keeps a local copy for lookup during origin enrichment.

We've scaffolded out a significant amount of code which I'll list out as high-level categories below.

### Metadata collectors and the tag store

Originally based on work that we tried where we attempted to simply port over all of the functionality that exists in the Agent, we've introduced a notion of composable metadata "collectors" which are responsible for streaming events about changes to metadata for a given subsystem. You can think of these as being generally tied to a specific "source": Kubernetes, containerd, Docker, etc.

Each collector uses the most optimal way -- poll on an interval, stream from the source, etc -- to collect the metadata events, and then converts them to metadata "operations", which describe the action to be taken on the tag store.

The tag store is specifically designed for holding a mapping of tags for a given entity ID, with varying cardinality levels, but it does this with an event sourcing-based design and a number of operation types that allow for building an entity hierarchy. This lets us add tags for various entity IDs, and then add links between entities, such that the tags can be merged together and we end up with a combined set of tags for entties based on their ancestry.

The Datadog Agent already does some of this, but we use it in Saluki to let us optimize one particular part of origin enrichment: using the container PID directly instead of always having to try and resolve it to the container ID first. This means we use a cheap `u32` identifier instead of allocating a `String` for every single metric.

### `containerd` collector

While this PR is primarily about the Datadog Agent-based workload provider/metadata collector, we also had to add a metadata collector for `containerd`. Why?

Going back to origin enrichment, many times a feature called "origin detection" is used to automatically determine the container that sent the metric. This involves a neat trick of asking the Linux kernel to send us "socket credentials" when using Unix domain sockets. These credentials include the process ID of the peer on the other end, which we use to map to a container ID. We store this process ID as a tag on each metric.

In order to map the process ID to the container ID, we need to be able to query the container runtime. This is where needing a metadata collector for `containerd` comes in. This collector simply watches for container task (process) events, and updates the tag store with a link between the process ID and the container ID.

When handling origin enrichment, we use the process ID to look up the entity tags... and since we established an ancestry link between the process ID and container ID prior to this, it "just works".

### Local testing with Kubernetes and minikube

Testing all of this work involves a complicated setup:

- running in Kubernetes
- running a new container in the typical Datadog Agent pod (tagger API is localhost-only!)
- wiring up another container (to exercise origin detection/enrichment logic) to send metrics to ADP

In order to simplify this and make it repeatable, we've added a number of new Makefile targets that are used to spin up a local Kubernetes cluster via `minikube`, build/push a custom Datadog Agent image with Agent Data Plane built in, and finally deploy the converged Datadog Agent with a simple metrics generator workload pointed at it.

### Various code files touched outside of `saluki-env`

I tried to peel back unnecessary changes where possible, but some of the things changed were quality-of-life improvements for sharp edges that kept slowing me down during testing. I left them in because they were going to get done one way or another, and it was easier to do them now than keep suffering the incremental overhead/pain.


lib/datadog-protos/proto/datadog/api/README.md
```diff
@@ -0,0 +1,43 @@
...
+### Notes
+
+We are currently pinned to fairly old versions for some of the 
+protobuf/grpc dependencies and tooling. These are required as a
+consequence of third-party libraries (most notably etcd). Please
+see `go.mod` and `internal/tools/proto/go.mod` to understand the
+version requirements. 
+
+The tooling in place should help our protobuf versions be consistent
+across the board.
```
blt, line 43: Copypasta?

lib/saluki-env/src/workload/collectors/containerd.rs
```diff
@@ -0,0 +1,173 @@
...
+            }
+        }
+
+        Some(operations)
+    }
+
+    fn watch(self) -> impl Stream<Item = MetadataOperation> + Unpin {
+        // We watch the given namespace for all of the relevant events, and convert those into metadata operations that
+        // we pass back to be collected by the parent watcher task, which then forwards them to the metadata aggregator.
+        Box::pin(stream! {
```
blt, line 138: Ah `stream!`, don't see that around a lot in the ecosystem yet. Neat.

blt (dismissed):
Nice.

lib/datadog-protos/proto/datadog/api/README.md
tobz: Ah, good catch. Yeah, will remove.

tobz: Removed in 97fafef.

lib/datadog-protos/proto/datadog/api/v1/api.proto
```diff
@@ -0,0 +1,147 @@
...
+import "datadog/model/v1/model.proto";
+import "datadog/remoteconfig/remoteconfig.proto";
+import "datadog/workloadmeta/workloadmeta.proto";
+import "google/api/annotations.proto";
+import "google/protobuf/empty.proto";
+
+option go_package = "pkg/proto/pbgo/core"; // golang
+
+
+// The greeting service definition.
```
GeorgeHahn, line 14: Pasted doc string

PR #48, opened 2024-05-13
Support checks

scottopell:
To run:
```
DD_LOG_LEVEL=debug RUST_BACKTRACE=1 DD_ADP_USE_NOOP_WORKLOAD_PROVIDER=true  DD_API_KEY=<some api key> cargo run --
```

I included an example py check in the `conf.d` folder.

- [x] Find a python check on disk
- [x] Load and run python checks on an interval
- [x] Submit metrics from the python checks
- [x] Support `init_config` and arbitrary fields in the instance-config
- [x] Test HTTPS requests (they work, requires setting `ca_cert` via `init_config` though)
- [ ] Try out full default integration suite


scottopell:
continued in #277

PR #52, opened 2024-05-16
Try using jemalloc as the global allocator.

tobz:
Work in progress.


tobz:
Going to close this for now as there wasn't enough evidence early on that we could reasonably overcome the large jumps in memory growth when switching to something other than the system allocator.

We may revisit this in the future.

PR #55, opened 2024-05-17
Propagate metrics resulting from a check execution to the Event Buffer.

remeh:
Nearly have something which should do the link Python `submit_metric`  ->  saluki's event buffer

I did a few non-idiomatic/stupid things to move forward, feel free to just get inspiration/copy paste some code from the PR if you don't want to merge it:
* used a global queue (sigh) to transmit metrics resulting from the check execution into the saluki's SourceContext, I don't really know how this link, you or toby might know better
* used an intermediate CheckMetric struct
  * most likely better to directly use saluki::Event but I was already to deep into it I didn't look into what existed just yet
  * because of this I created an ugly fn event_from_check_metric and didn't even bother to impl TryFrom
* not fitting the serialiser/codec architecture of saluki
  
For some reasons the queue always seems empty so nothing is actually transiting up to the event buffer.

### DIFF
--- a/lib/saluki-components/src/sources/checks/aggregator.rs
+++ b/lib/saluki-components/src/sources/checks/aggregator.rs
@@ -1,12 +1,131 @@
 use super::*;
 use pyo3::prelude::*;
+use saluki_event::{metric::*, Event};
+use saluki_env::time::get_unix_timestamp;
 
+#[derive(Clone, Copy)]
+enum PyMetricType {
+    Gauge = 0,
+    Rate,
+    Count,
+    MonotonicCount,
+    Counter,
+    Histogram,
+    Historate,
+}
+
+#[derive(Debug, Snafu)]
+#[snafu(context(suffix(false)))]
+pub enum AggregatorError {
+    UnsupportedType{},
+}
+
+impl TryFrom<i32> for PyMetricType {
+    type Error = ();
+
+    fn try_from(v: i32) -> Result<Self, Self::Error> {
+        match v {
+            x if x == PyMetricType::Gauge as i32 => Ok(PyMetricType::Gauge),
+            x if x == PyMetricType::Rate as i32 => Ok(PyMetricType::Rate),
+            x if x == PyMetricType::Count as i32 => Ok(PyMetricType::Count),
+            x if x == PyMetricType::MonotonicCount as i32 => Ok(PyMetricType::MonotonicCount),
+            x if x == PyMetricType::Counter as i32 => Ok(PyMetricType::Counter),
+            x if x == PyMetricType::Histogram as i32 => Ok(PyMetricType::Histogram),
+            x if x == PyMetricType::Historate as i32 => Ok(PyMetricType::Historate),
+            _ => Err(()),
+        }
+    }
+}
+
+/// CheckMetric are used to transmit metrics from python check execution results
+/// to forward in the saluki's pipeline.
+pub struct CheckMetric {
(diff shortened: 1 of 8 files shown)


lib/saluki-components/src/sources/checks/aggregator.rs
```diff
@@ -1,12 +1,131 @@
...
+            tags: self.tags.clone(),
+        }
+    }
+}
+
+
+/// Global for Python checks execution to report the data
+pub static SUBMISSION_QUEUE: Lazy<Mutex<Queue<CheckMetric>>> = Lazy::new(|| {
+    Mutex::new(queue![])
+});
```
tobz, line 94: This seems like it could easily end up leading to multiple checks having their metrics interleaved in the queue. Is there any requirement that after triggering a check, we need to be able to extract only their specific metrics?

lib/saluki-components/src/sources/checks/aggregator.rs
```diff
@@ -1,12 +1,131 @@
...
+            metric_type: self.metric_type,
+            value: self.value.clone(),
+            tags: self.tags.clone(),
+        }
+    }
+}
+
+
+/// Global for Python checks execution to report the data
+pub static SUBMISSION_QUEUE: Lazy<Mutex<Queue<CheckMetric>>> = Lazy::new(|| {
```
tobz, line 92: You could skip using this whole `queues` crate and just use `std::collections::VecDeque` instead, as the `push_back`/`pop_front` methods let you achieve the same thing.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -315,7 +319,23 @@ async fn process_listener(source_context: SourceContext, listener_context: liste
                 info!("Running a check request: {check_request}");
 
                 match scheduler.run_check(check_request) {
-                    Ok(_) => {}
+                    Ok(_) => {
+                        // drain the global queue and send everything to source_context
+                        let mut event_buffer = source_context.event_buffer_pool().acquire().await;
```
tobz, line 324: I would move this outside of the call to `run_check`, since it's meant to act as a backpressure mechanism to avoid producing/receiving events when we don't have any free event buffers to store them in.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -315,7 +319,23 @@ async fn process_listener(source_context: SourceContext, listener_context: liste
...
+                        while q.size() > 0 {
+                            let check_metric = q.remove().expect("can't read from the global submission queue");
+
+                            // TODO(remy): fit the the deserializer/codec architecture instead of doing it here
+                            let event: Event = aggregator::check_metric_as_event(check_metric).expect("can't convert");
+                            event_buffer.push(event);
+                            debug!("one metric sent to the event buffer from check execution");
+                        }
+                        debug!("queue drained");
+                        drop(q); // we're in an infinite loop, explicitely drop q to unlock the mutex
```
tobz, line 337: Another benefit of switching to `VecDeque` instead of that queue type would be that you could reduce this code down to:

```suggestion
                        let mut q = aggregator::SUBMISSION_QUEUE.lock().unwrap();
                        event_buffer.extend(q.drain(..).filter_map(aggregator::check_metric_as_event));
                        drop(q);
```

You'd have to change `check_metric_to_event` to return `Option<Event>`, though.... but that might not be a bad change overall.

PR #64, opened 2024-05-20
DDSketch: Add allocation tests & microbenchmarks

GeorgeHahn:
- Uses [dhat-rs](https://github.com/nnethercote/dhat-rs) to assert on the allocations performed during DDSketch operations.

Caveat:
> Warning: This crate is experimental. It relies on implementation techniques that are hard to keep working for 100% of configurations. It may work fine for you, or it may crash, hang, or otherwise do the wrong thing. Its maintenance is not a high priority of the author. Support requests such as issues and pull requests may receive slow responses, or no response at all. Sorry!

- Uses criterion to add microbenchmarks for sketch insert operations.

### DIFF
--- a/lib/ddsketch-agent/tests/common/mod.rs
+++ b/lib/ddsketch-agent/tests/common/mod.rs
@@ -1,0 +1,82 @@
+#![allow(dead_code)]
+
+use datadog_protos::metrics::Dogsketch;
+use ddsketch_agent::DDSketch;
+use dhat::HeapStats;
+use rand::SeedableRng;
+use rand_distr::{Distribution, Pareto};
+
+pub fn insert_single_and_serialize(ns: &[f64]) {
+    let mut sketch = DDSketch::default();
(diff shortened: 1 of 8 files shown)


lib/ddsketch-agent/tests/common/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+            total_bytes: stats.total_bytes,
+            max_blocks: stats.max_blocks,
+            max_bytes: stats.max_bytes,
+            curr_blocks: stats.curr_blocks,
+            curr_bytes: stats.curr_bytes,
+        }
+    }
+}
+
+impl std::ops::Sub for MathableHeapStats {
```
GeorgeHahn, line 69: I sent a PR for this: https://github.com/nnethercote/dhat-rs/pull/34

tobz (approved):
👍🏻

PR #66, opened 2024-05-21
Add insertion tests and a faster single-value slow path

GeorgeHahn:
Further speed up single-value insertion by adding an optimized slow path. Adds a bunch of insertion tests to make sure I got insertion and overflow right.

### DIFF
--- a/lib/ddsketch-agent/src/lib.rs
+++ b/lib/ddsketch-agent/src/lib.rs
@@ -439,7 +439,7 @@ impl DDSketch {
         for (bin_idx, b) in self.bins.iter_mut().enumerate() {
             if b.k == key {
                 if b.n < MAX_BIN_WIDTH {
-                    // Fast path for adding to an existing bin.
+                    // Fast path for adding to an existing bin without overflow.
                     b.n += 1;
                     return;
                 } else {
@@ -1195,7 +1195,13 @@ mod tests {
                 expected: "0:1 1338:1",
             },
             Case {
-                description: "inserting a value into a sketch and causing an overflow",
+                description: "inserting a value into an existing bin and filling it",
+                start: "0:65534",
+                insert: Value::Float(0.0),
+                expected: "0:65535",
+            },
+            Case {
+                description: "inserting a value into an existing bin and causing an overflow",
                 start: "0:65535",
                 insert: Value::Float(0.0),
                 expected: "0:1 0:65535",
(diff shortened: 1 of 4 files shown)


lib/ddsketch-agent/tests/one_thousand_single_points_ddsketch.rs
```diff
@@ -21,8 +21,8 @@ fn test_one_thousand_single_points_ddsketch() {
     let after: MathableHeapStats = dhat::HeapStats::get().into();
 
     let diff = after - before;
-    dhat::assert_eq!(diff.total_blocks, 1426);
-    dhat::assert_eq!(diff.total_bytes, 249802);
+    dhat::assert_eq!(diff.total_blocks, 21);
+    dhat::assert_eq!(diff.total_bytes, 6096);
```
GeorgeHahn, line 25: This is extra cool because it now beats the allocation counts for batched inserts.

lib/ddsketch-agent/src/lib.rs
```diff
@@ -434,16 +434,31 @@ impl DDSketch {
...
+        let mut insert_at = None;
+
+        for (bin_idx, b) in self.bins.iter_mut().enumerate() {
+            if b.k == key {
+                if b.n < MAX_BIN_WIDTH {
+                    // Fast path for adding to an existing bin without overflow.
+                    b.n += 1;
+                    return;
+                } else {
+                    insert_at = Some(bin_idx);
```
tobz, line 446: This might be entirely irrelevant, but do you think it makes any difference if we insert the new bin _at_ the existing index, or _after_ the existing index?

My mind naturally wants to see `0:65535 0:1`, ordered largest to smallest, but then I wondered if it actually may _matter_ to downstream systems or not. Guessing not, but it feels like you're very up-to-speed here and might have a good intuition of if it does or doesn't matter.

tobz, line 446: This might be entirely irrelevant, but do you think it makes any difference if we insert the new bin at the existing index, or after the existing index?

My mind naturally wants to see 0:65535 0:1, ordered largest to smallest, but then I wondered if it actually may matter to downstream systems or not. Guessing not, but it feels like you're very up-to-speed here and might have a good intuition of if it does or doesn't matter.

GeorgeHahn: I'm not aware of any downstream requirement on the bin order, but we do depend on the incremental counts coming first in `insert_keys` and `trim_left`.

tobz: Ah, good point. 👍🏻

tobz (approved):
Looks good to me. 👍🏻

PR #77, opened 2024-05-31
context: try sorting tags up front

tobz:
Small test to see what the performance difference is (if any) by sorting tags when we resolve a context versus sorting them in the aggregate transform when doing equality checks in `AggregationContext`.


tobz:
No longer relevant since we handle equality checking with the order-oblivious hash of the tags.

PR #85, opened 2024-06-05
context: switch to pure hash-based equality

tobz:
## Context

Contexts represent the sum of both the name and tags of a metric, and are used to uniquely identify a metric. As such, contexts are touched often in the hot path of both parsing and aggregation.

Normally, as part of "resolving" a context (similar to string interning) or just looking them up in a map (such as during aggregation), the hash of the context is checked as a fast path, and if there are any collisions with existing items, the equality of those items is checked directly: this looks sort of like check that the name of each context is equal, and that the tags of each context are equal, which compares them _in order_.

This is slower than the hash checking (because comparing two integers is hard to beat) but also means that contexts may not present as equal if their tags are in different orders. Solving the problem naively -- sorting the tags -- involves additional allocations, and if we don't sort them, we risk treating identical sets of name/tags pairs as different, even if all that is different is the order of the tags.

## Solution

This PR unifies the hashing/equality checking of both `ContextRef<'a, T>` and `Context` (by way of `ContextInner`) to ensure that both hashing of the tags is order-oblivious (i.e. they can be in any order) and that equality is checked by comparing the hash of the given context/context reference.

There's a comment on `ContextRef` that explains why we as eschewing a "true" equality check here, as the likelihood of a collision is extremely low, even more so when considering the expected cardinality that a single process will see.

Additionally, we've slightly refactored the aggregate transform to get rid of its custom context newtype and use `Context` directly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1514,9 +1514,9 @@ dependencies = [
 
 [[package]]
 name = "metrics"
-version = "0.22.3"
+version = "0.23.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2be3cbd384d4e955b231c895ce10685e3d8260c5ccffae898c96c723b0772835"
+checksum = "884adb57038347dfbaf2d5065887b6cf4312330dc8e94bc30a1a839bd79d3261"
 dependencies = [
  "ahash",
  "portable-atomic",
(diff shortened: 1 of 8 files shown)


tobz:
This doesn't actually seem to fix a glitch I'm observing in the SMP benchmark runs where the active context count is higher in ADP, compared to DSD... but uh, it's way faster, so I'm going to merge it.

PR #88, opened 2024-06-11
Cleanup dhat-rs usage

GeorgeHahn:
Turns out there's no need to do `HeapStats` math if you simply setup the heap profiler later.


tobz (approved):
👌🏻

PR #100, opened 2024-06-25
smp: try out a realllllllly high context count to see what memory usage is like

tobz:


tobz:
Added this in 8d0f1e7f93e0ca3b6bf3723eb5d6a9c6ece57497.

PR #122, opened 2024-07-19
Reintroduce origin-based enrichment of metrics.

tobz:
## Context

Prior to #121, we had to disable origin enrichment as our work to improve context resolving left us with immutable tags, and so there was no way to actually enrich tags.

## Solution

With #121 merged, we can now reenable origin enrichment. This did involve some changes beyond uncomment the code, namely around refactoring to handle the use of `MetaString` and changes to struct accessors, but otherwise is pretty strightforward.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2478,6 +2478,7 @@ dependencies = [
  "serde_json",
  "snafu",
  "socket2",
+ "stringtheory",
  "tokio",
  "tokio-rustls",
  "tokio-util",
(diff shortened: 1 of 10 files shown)


tobz:
This _does_ represent a regression in memory usage, which I plan to write an issue for, in terms of how we plan to resolve it, but for the sake of posterity...

Essentially, with the reintroduction of the `origin_enrichment` transform, we're always calling `Context::tags_mut` when we go through and remove origin detection-specific tags... even if they don't exist and we end up leaving the tags exactly as we found them. This is due to how we blindly call `Arc::make_mut` in `Context::tags_mut`, which is always going to clone the tag set if it's otherwise shared.

Again, I plan to write an issue for this to describe ways I can think we can recover this, but I wanted to be clear we are knowingly and intentionally accepting this regression for the moment.

Issue #123, opened 2024-07-20
Add new `Event` variants to support holding events/service checks.


tobz:
We'll want to add two new variants to the core `Event` type, one for events and service checks.

The internal structures of the wrapped struct should likely match closely to their API data model ([here](https://docs.datadoghq.com/api/latest/events/#post-an-event) and [here](https://docs.datadoghq.com/api/latest/service-checks/#submit-a-service-check)). Naming is a little tricky since while service checks can just be `ServiceCheck`, events would then be `Event` if we followed that scheme... and `Event::Event` is very silly.

Gotta workshop the names a little more perhaps.

PR #129, opened 2024-07-22
Add new event variants

rayz:
# Context 
Currently the core `Event` type only has the `metric` variant.  We would like support for events and service checks. (issue #123)

# Solution
This pr adds the new `Event` variants `Eventd` (open to changing this name) and `ServiceCheck`. 

# Testing
This pr should not have an effect on any existing code. 
I ran `make test` and `make build-adp`. Unit tests to be added in next pr that adds support to the dogstatsd codec.


lib/saluki-components/src/destinations/prometheus/mod.rs
```diff
@@ -119,41 +119,46 @@ impl Destination for Prometheus {
...
-                            match existing_contexts.get_mut(&context) {
-                                Some(existing_value) => {
-                                    existing_value.merge(value);
-                                },
-                                None => {
-                                    if contexts >= CONTEXT_LIMIT {
-                                        debug!("Prometheus destination reached context limit. Skipping metric '{}'.", context.name());
-                                        continue
+                            match event {
+                                Event::Metric(metric)=>{
```
tobz, line 123: Since we only accept metrics here, we can make this cleaner by just doing something like:

```rust
if let Some(metric) = event.into_metric() {
   ...
}|
```

That lets is skip having to even deal with the match expression in the first place.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -425,6 +425,8 @@ async fn drive_stream(
                                             .set_origin_entity(OriginEntity::ProcessId(creds.pid as u32));
                                     }
                                 }
+                                Event::EventD(_) => todo!(),
+                                Event::ServiceCheck(_) => todo!(),
```
tobz, line 429: Similar comment as the one for the Prometheus destination around just using `Event::into_metric` instead... although here we're only mutating the event, so we would need to add something like:


```rust
impl Event {
  pub fn as_metric_mut(&mut self) -> Option<&mut Metric> {
    match self {
      Self::Metric(mut metric) => Some(metric),
      _ => None,
    }
  }
}
```

(Code might not exactly compile but I think it captures the intent.)

lib/saluki-components/src/transforms/origin_enrichment/mod.rs
```diff
@@ -249,6 +249,8 @@ where
         for event in event_buffer {
             match event {
                 Event::Metric(metric) => self.enrich_metric(metric),
+                Event::EventD(_) => todo!(),
```
tobz, line 252: Similar comment to the one above for the DogStatsD source around a mutable fallible conversion function (i.e. `Event::as_metric_mut`).

lib/saluki-components/src/transforms/host_enrichment/mod.rs
```diff
@@ -73,6 +73,8 @@ impl SynchronousTransform for HostEnrichment {
         for event in event_buffer {
             match event {
                 Event::Metric(metric) => self.enrich_metric(metric),
+                Event::EventD(_) => todo!(),
```
tobz, line 76: Similar comment to the one above for the DogStatsD source around a mutable fallible conversion function (i.e. `Event::as_metric_mut`).

lib/saluki-event/src/service_check/status.rs
```diff
@@ -0,0 +1,15 @@
+/// ServiceCheckStatus is the status type for service checks
```
tobz, line 1: I know that in Go it's common to start a doc comment with the type/method name -- perhaps even enforced by the linter? -- but in Rust, we _generally_ prefer human-readable descriptions, which also means we try to avoid "stuttering", where we repeat the name of the thing we're describing.

So in this case, we know from the type name that we're dealing with the status of the service check, so we really don't need to say much more than "Service status."

lib/saluki-event/src/eventd/priority.rs
```diff
@@ -0,0 +1,9 @@
+/// EventPriority is the event priority for events
```
tobz, line 1: Similar comment here as the one on `ServiceCheckStatus` about avoiding stuttering, etc.

lib/saluki-event/src/service_check/status.rs
```diff
@@ -0,0 +1,15 @@
...
+#[derive(Clone, Debug)]
+pub enum ServiceCheckStatus {
+    /// Ok is the "ok" ServiceCheck status
```
tobz, line 4: Similar comment to the above.

For example, you might say "The service is operating normally." for the `Ok` variant, and then say "The service is in a warning state." for `Warning` state, and so on.

Basically, if we're going to restate what is likely already obvious, we should just do it in a way we'd naturally say it.

tobz (changes requested):

lib/saluki-event/src/service_check/status.rs
rayz: gotcha, was just copying from the [go code ](https://github.com/DataDog/datadog-go/blob/1139efe797e4760daf3d30a60e3d30afc1687df8/statsd/service_check.go#L12) 😅

tobz (approved):

PR #130, opened 2024-07-22
chore: clean up eventd and service check types

tobz:
## Context

Just a little bit of cleanup to polish the code comments and organization of events/service checks, introduced in #129.


lib/saluki-event/src/eventd/mod.rs
```diff
@@ -1,9 +1,33 @@
-//! Event types
-mod priority;
-pub use self::priority::*;
+//! Events.
 
-mod alert;
-pub use self::alert::*;
+// TODO: Switch usages of `String` to `MetaString` since we should generally be able to intern these strings as they
+// originate in in the DogStatsD codec, where interning is already taking place.
```
tobz, line 4: Usually we'll try to keep TODO comments close to the area where they're most relevant, but in this case, this is a thing we'd want to apply to multiple fields on the `EventD` struct, so we just shove it up here as a general TODO.

lib/saluki-event/src/eventd/mod.rs
```diff
@@ -1,9 +1,33 @@
...
+
+/// Alert type.
+#[derive(Clone, Copy, Debug)]
```
tobz, line 7: While this was previously only `Clone`, enums are a prime candidate for being `Copy`. This means the value can be copied directly -- just copy the raw memory bytes and nothing else -- without any special logic.

Generally, integral types (signed and unsigned integers, floats, booleans) and "unit" enums (the variants don't have any fields) can stand to be `Copy` because their size in memory is no bigger than a _reference_ to this types.

This is to say that if you compare the size of this enum to a reference to this enum, the reference to it is much larger: 8 bytes versus 1 byte. References are a machine word (so 8 bytes on 64-bit architectures) while the compiler ends up bitpacking this enum in a single byte. The same thing holds if we use an integer type like `u32`: `&u32` is 8 bytes while `u32` itself is only 4 bytes.

lib/saluki-event/src/eventd/mod.rs
```diff
@@ -13,55 +37,57 @@ pub struct EventD {
     timestamp: u64,
     host: String,
     aggregation_key: String,
-    priority: EventPriority,
+    priority: Priority,
```
tobz, line 40: We've renamed `EventPriority` to `Priority` to avoid stuttering.

lib/saluki-event/src/eventd/mod.rs
```diff
@@ -13,55 +37,57 @@ pub struct EventD {
...
     source_type_name: String,
-    alert_type: EventAlertType,
+    alert_type: AlertType,
```
tobz, line 42: Similarly, we've renamed `EventAlertType` to `AlertType` to avoid stuttering.

lib/saluki-core/src/topology/graph.rs
```diff
@@ -447,9 +447,9 @@ mod test {
     fn invalid_input_id() {
         let mut graph = Graph::default();
         let result = graph
-            .with_source("in_log", DataType::Log)
-            .with_destination("out_log", DataType::Log)
-            .with_edge_fallible("in log", "out_log")
+            .with_source("in_eventd", DataType::EventD)
```
tobz, line 450: This, and all changes in this file, are just renaming to reflect the fact I removed `DataType::Log` and `DataType::Trace`.

lib/saluki-event/src/eventd/mod.rs
```diff
@@ -13,55 +37,57 @@ pub struct EventD {
...
     tags: Vec<String>,
 }
 
 impl EventD {
-    /// Gets a reference to the title.
+    /// Returns the title of the event.
```
tobz, line 47: This change in wording (like across the board for the fields in this type) is a little bit nit-picky, and I know the existing codebase has a lot of `Gets a reference to X`, but that's a pattern I want to clean up in the future.

In general, for accessors, this is the format I'd like to use for doc comments, since it better reflects the common wording for accessors, which is `Returns ...`.

We also do have some stuttering here -- the `of the event` suffix for each comment -- but this is generally the only where I think it's OK, because stuttering in types is always in your face since it's in the code, whereas in comments, it's hidden until you... look at the docs, or hover over a type and get intellisense, etc.

Not a hard and fast rule, but just trying to solidify the style here since we don't (yet) have a style guide.

lib/saluki-event/src/eventd/mod.rs
```diff
@@ -1,9 +1,33 @@
...
+    /// Indicates a warning event.
+    Warning,
+
+    /// Indicates a successful event.
+    Success,
+}
+
+/// Event priority.
+#[derive(Clone, Copy, Debug)]
+pub enum Priority {
```
tobz, line 24: For both `Priority` and `AlertType`, I've moved them to `mod.rs` because they're simple enough that it ends up being a little cleaner to have it all in the same file. If we had an impl block, or tests, then it might make sense to push it into a separate file.

lib/saluki-event/src/lib.rs
```diff
@@ -25,11 +25,11 @@ pub enum DataType {
...
 
-    /// Logs.
-    Log,
+    /// Datadog Events.
+    EventD,
 
-    /// Traces.
-    Trace,
+    /// Service checks.
+    ServiceCheck,
```
tobz, line 32: The old variants -- `Log` and `Trace` -- didn't actually map to any possible variant in `Event`, so I took this as a good chance to align it with `Event`.

lib/saluki-event/src/lib.rs
```diff
@@ -25,11 +25,11 @@ pub enum DataType {
     /// Metrics.
     Metric,
 
-    /// Logs.
-    Log,
+    /// Datadog Events.
```
tobz, line 28: In general, Saluki wants to be as agnostic as it can be: we're providing building blocks to making something like the Agent but we don't want to _necessarily_ write as if it's just the agent... so we'll keep names/descriptions generic where possible, and then specialize (like calling these "Datadog Events") where needed to disambiguate.

lib/saluki-event/src/lib.rs
```diff
@@ -64,48 +64,48 @@ pub enum Event {
...
     ///
-    /// If the underlying event is not a `Eventd`, this will return `None`.
-    pub fn into_eventd(self) -> Option<EventD> {
+    /// Otherwise, `None` is returned and the original event is consumed.
+    pub fn try_into_eventd(self) -> Option<EventD> {
         match self {
             Event::EventD(eventd) => Some(eventd),
             _ => None,
         }
     }
```
tobz, line 103: I updated the name of these methods because I gave you a bad suggestion in the original PR.

The common naming practice for conversion methods in Rust is:

- `into_foo` if it _consumes_ the value and the conversion isn't fallible (i.e. no `Option<T>` or `Result<T, E>`)
- `as_foo` if it provides an immutable reference conversion (i.e. `String::as_str` returns `&str`)
- `as_foo_mut` if it provides a mutable reference conversion (i.e. `Vec::as_mut_slice` returns `&mut [T]`)
- `try_into_foo`/`try_as_foo`/`try_as_foo_mut` if the conversion is fallible (i.e. how we're doing it because we don't know if an `Event` is actually a `Metric` or not)

lib/saluki-event/src/service_check/mod.rs
```diff
@@ -1,46 +1,68 @@
...
-    /// Gets a reference to the message
-    pub fn message(&self) -> &String {
+    /// Returns the message associated with the check.
+    pub fn message(&self) -> &str {
         &self.message
     }
 
-    /// Gets a reference to the tags
-    pub fn tags(&self) -> &Vec<String> {
+    /// Returns the tags associated with the check.
```
tobz, line 64: Also, super small style: we always write comments as full sentences, including ending with a period.

rayz (approved):

PR #132, opened 2024-07-24
[APR-205] dogstatsd: add support for a configurable tag interceptor for filtering/augmenting metadata via tags

tobz:
## Context

A lot of code exists at the moment surrounding context resolving and the sharing of tags, including code to then try and allow us to _modify_ tags in resolved contexts and minimize how much we have to clone/allocate, and so on and so forth.

The extent to which we work around this is due entirely to the need to modify tags after decoding. While we certainly aspire to allow for such free-form transformation of metrics, the current need to do so stems from origin enrichment, where we extract relevant tags that may or may not be present and use them to augment the metric metadata.

We consciously chose to do this in order to attempt to get the DogStatsD codec somewhat decoupled from being a one-to-one rewrite of the equivalent code in the Agent. In essence, though, our lives would be simpler if we could both exclude certain tags from the context resolving logic, as well as visit those tags as we build the `Metric` in order to update the metadata from them accordingly.

## Solution

This PR introduces a new trait, `TagMetadataInterceptor`, which serves two purposes:
- evaluate individual tags to determine whether they should be passed through, dropped, or "intercepted"
- when a tag is intercepted, define how we use that tag to augment the metadata for the metric

Most of the code involved here is boilerplate to fit how we've structured decoding to avoid allocations when building `ContextRef<'a, T>`... so we've like `TagSplitter<'a>`/`TagIter<'a>`, we now have `TagFilterer<I, TMI>`/`TagFiltererIter<I, TMI>`.

We handle this in two phases:

- wrap our normal `TagSplitter<'a>` to get `TagFilterer<TagSplitter<'a>, TMI>`, where `TMI` is the interceptor implementation to use (this can be cloned, and serves purely as a way to filter out intercepted/dropped tags when doing context stuff)
- call a new function, `update_metadata_from_tags`, which handles updating the `MetricMetadata` we got from decoding with any tags that get identified as needing to be intercepted

For now, we've pushed this into `DogstatsdCodec` as a new generic parameter (`TMI`) with a default of `()`. All the relevant traits implementations are in place _such_ that existing code works exactly as it did before, and no tags are filtered out or do anything to modify the metric metadata. In a subsequent PR, we'll add a Agent-like interceptor implementation specifically for handling some of the tags we currently handle in the `origin_enrichment` transform, such as `dd.internal.entity_id`, `dd.internal.jxm_check_name`, and `dd.internal.card`.

### DIFF
--- a/lib/saluki-io/src/deser/codec/dogstatsd.rs
+++ b/lib/saluki-io/src/deser/codec/dogstatsd.rs
@@ -1,6 +1,5 @@
 use std::collections::HashSet;
 
-use bytes::Buf;
 use nom::{
     branch::alt,
     bytes::complete::{tag, take_while1},
@@ -18,7 +17,7 @@ use snafu::Snafu;
 use saluki_core::topology::interconnect::EventBuffer;
 use saluki_event::{metric::*, Event};
 
-use crate::deser::Decoder;
+use crate::{buf::ReadIoBuffer, deser::Decoder};
 
 type NomParserError<'a> = nom::Err<nom::error::Error<&'a [u8]>>;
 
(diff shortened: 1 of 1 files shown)


lukesteensen (approved):

lib/saluki-io/src/deser/codec/dogstatsd.rs
```diff
@@ -551,20 +581,156 @@ impl<'a> Iterator for ValueIter<'a> {
...
+                InterceptAction::Drop
+            } else {
+                InterceptAction::Pass
+            }
+        }
+
+        fn intercept(&self, tag: &str, metadata: &mut MetricMetadata) {
+            if let Some((key, value)) = tag.split_once(':') {
+                if key == "host" {
+                    metadata.set_hostname(Arc::from(value));
```
rayz, line 728: unrelated question: why is `hostname` the only field the needs to be wrapped in an `Arc`for `MetricMetadata`?

rayz (approved):

lib/saluki-io/src/deser/codec/dogstatsd.rs
tobz: It's the only string field, and to make `MetricMetadata` cloneable, as least cheaply... we wrap the hostname in `Arc<T>`.

Realistically, it should actually be `MetaString` so that we can intern it or inline it where possible, skipping a heap allocation entirely.

PR #133, opened 2024-07-25
[APR-205] dogstatsd: switch to using tag interceptor for origin entity support

tobz:
## Context

As a follow-on to #132, we need to actually switch to using the tag interceptor mechanism to intercept relevant tags and update the metric metadata.

## Solution

This PR introduces a new interceptor implementation, `AgentLikeTagMetadataInterceptor`, which specifically focuses on `dd.internal.entity_id`, `dd.internal.jmx_check_name`, and `dd.internal.card`. With this, we're able to skip having to _remove_ any tags from metrics in the `origin_enrichment` transform.

We've gone through and tweaked `OriginEntity` from an enum into a struct that can hold all of the relevant entity ID information, similar to `OriginInfo` in the Agent codebase. This is somewhat suboptimal, but it allows us to track the Agent logic better by being able to incrementally enrich and not having to bake in a bunch of ordering/precedence logic into `OriginEntity`/`EntityId`.

## Notes

We have some struct size concerns now, give that `OriginEntity` isn't an enum anymore, but I plan to shrink `OriginEntity`, and `MetricMetadata`, in a follow-up PR.

PR #134, opened 2024-07-25
[APR-205] stringtheory/event: shrink MetaString and MetricMetadata struct sizes

tobz:
## Context

In #133, we ended up increasing the size of `MetricMetadata` by a significant amount due to tracking three potential entity IDs instead of just the latest/highest priority one we had decoded so far.

## Solution

While a long-term solution would ideally center on going back to using the highest priority entity ID found, for now we need to deal with the status quo... and so we've shrunk the size of the `MetricMetadata` struct to try and get back some of the memory efficiency we've lost by doing so.

We attacked this in two ways: shrink` `MetricMetadata` itself, and shrinking `MetaString`.

### `MetricMetadata`

We had some non-optimal field sizes for optional things like sample rate and timestamp. Rather than have them be optional, we've just switched to giving them specific default values. Sample rate always defaults to 1.0 now, which is what we used anyways when the field was otherwise empty. For timestamp, we treat zero as the "none" value, since given that it's the number of seconds since the Unix epoch, we know it can in fact never legitimately be zero. We maintained the accessors that do `Option<u64>`, though, since it makes downstream code a little nicer.

### `MetaString`

We ended up removing the `Shared` and `ProtoShared` variants, as we don't actually use them. Those variants were the largest in the enum, so removing them saves us a machine word (8 bytes on 64-bit) now. This does have the side effect of limiting the number of bytes we can inline, going from 31 bytes to 23 bytes.

This will certainly affect some tags, but isn't likely to affect a _ton_ of them.

### DIFF
--- a/lib/saluki-event/src/metric/metadata.rs
+++ b/lib/saluki-event/src/metric/metadata.rs
@@ -96,10 +96,10 @@ impl OriginEntity {
 /// Metadata includes all information that is not specifically related to the context or value of the metric itself,
 /// such as sample rate and timestamp.
 #[must_use]
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub struct MetricMetadata {
-    sample_rate: Option<f64>,
-    timestamp: Option<u64>,
+    sample_rate: f64,
+    timestamp: u64,
     hostname: Option<Arc<str>>,
     origin_entity: OriginEntity,
     origin: Option<MetricOrigin>,
@@ -109,13 +109,17 @@ impl MetricMetadata {
     /// Gets the sample rate.
     ///
     /// This value is between 0 and 1, inclusive.
-    pub fn sample_rate(&self) -> Option<f64> {
+    pub fn sample_rate(&self) -> f64 {
         self.sample_rate
     }
 
     /// Gets the timestamp.
     pub fn timestamp(&self) -> Option<u64> {
-        self.timestamp
+        if self.timestamp == 0 {
+            None
+        } else {
+            Some(self.timestamp)
+        }
     }
 
     /// Gets the hostname.
@@ -144,7 +148,7 @@ impl MetricMetadata {
     ///
     /// This variant is specifically for use in builder-style APIs.
     pub fn with_sample_rate(mut self, sample_rate: impl Into<Option<f64>>) -> Self {
-        self.sample_rate = sample_rate.into().map(|sr| sr.clamp(0.0, 1.0));
+        self.set_sample_rate(sample_rate);
         self
     }
 
@@ -152,7 +156,9 @@ impl MetricMetadata {
     ///
     /// This value must be between 0 and 1, inclusive. If the value is outside of this range, it will be clamped to fit.
     pub fn set_sample_rate(&mut self, sample_rate: impl Into<Option<f64>>) {
-        self.sample_rate = sample_rate.into().map(|sr| sr.clamp(0.0, 1.0));
+        if let Some(sample_rate) = sample_rate.into().map(|sr| sr.clamp(0.0, 1.0)) {
+            self.sample_rate = sample_rate;
+        }
     }
 
     /// Set the timestamp.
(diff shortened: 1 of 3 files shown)


tobz:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/saluki/134?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#134** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/134?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#133** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/133?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#132** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/132?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

lib/saluki-event/src/metric/metadata.rs
```diff
@@ -96,10 +96,10 @@ impl OriginEntity {
 /// Metadata includes all information that is not specifically related to the context or value of the metric itself,
 /// such as sample rate and timestamp.
 #[must_use]
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub struct MetricMetadata {
-    sample_rate: Option<f64>,
-    timestamp: Option<u64>,
+    sample_rate: f64,
+    timestamp: u64,
```
tobz, line 102: We've specifically made those both non-optional because:

- it shrinks them from 16 bytes to 8 bytes
- we can just set the default sample rate to 1.0, which is what we would do anyways when unwrapping it and finding it to be `None`
- timestamp being 0 can serve as the `None` value, since we know it's impossible for a real timestamp to actually be 0

lib/stringtheory/src/lib.rs
```diff
@@ -111,17 +97,17 @@ impl MetaString {
     /// This does not allocate.
     pub const fn empty() -> Self {
         Self {
-            inner: Inner::Shared(bytes::Bytes::new()),
+            inner: Inner::Owned(String::new()),
```
tobz, line 100: For the sake of reviewers: the change here is a no-op because `String` wraps over `Vec<u8>`, which in fact does not allocate when created via `Vec::new`, and only allocates when first trying to push something into the vector.

lib/stringtheory/src/lib.rs
```diff
@@ -111,17 +97,17 @@ impl MetaString {
...
 
     /// Creates a new `MetaString` from the given static string.
     ///
     /// This does not allocate.
     pub fn from_static(s: &'static str) -> Self {
-        Self::try_inline(s).unwrap_or_else(|| Self {
-            inner: Inner::Shared(bytes::Bytes::from_static(s.as_bytes())),
-        })
+        Self {
+            inner: Inner::Static(s),
```
tobz, line 109: I changed this because there's no good reason to inline it if we don't have to, as it should be equivalent to following the static reference directly.

lib/stringtheory/src/lib.rs
```diff
@@ -206,23 +188,12 @@ impl From<&str> for MetaString {
                 inner: Inner::Inlined(i),
             },
             None => Self {
-                inner: Inner::Shared(bytes::Bytes::copy_from_slice(s.as_bytes())),
+                inner: Inner::Owned(s.to_owned()),
```
tobz, line 191: These are functionally equivalent in that they both allocate the same exact amount.. although this change does mean overall that things are less efficient when we clone `MetaString` since there's no variant that can be shared cheaply like `Bytes` can be.

lib/stringtheory/src/lib.rs
```diff
@@ -290,23 +250,23 @@ impl fmt::Display for MetaString {
     }
 }
 
-/// A string that can be inlined up to 31 bytes.
+/// A string that can be inlined up to 23 bytes.
 #[derive(Clone)]
 struct InlinedString {
-    data: [u8; 32],
+    data: [u8; 24],
```
tobz, line 256: Note to self: we might want to make the array size a const generic parameter in the future, and that way some of the logic below -- knowing that we can encode the string size in `u8`, etc -- can be made entirely const, making it all a little more safe in the process.

rayz (approved):

PR #137, opened 2024-07-26
[APR-226] Add basic support for decoding events

rayz:
# Context 
Currently the codec does not have support for decoding dogstatsd events.

# Solution
This pr adds the implementation for decoding dogstatsd events in the newly created folder `dogstatsd` in `saluki-io/src/deser/codec`

### DIFF
--- a/lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
+++ b/lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
@@ -850,8 +850,7 @@ mod tests {
     use saluki_core::{pooling::helpers::get_pooled_object_via_default, topology::interconnect::EventBuffer};
     use saluki_event::{eventd::EventD, metric::*, Event};
 
-
-    use super::{parse_dogstatsd_metric, DogstatsdCodecConfiguration, parse_dogstatsd_event};
+    use super::{parse_dogstatsd_event, parse_dogstatsd_metric, DogstatsdCodecConfiguration};
     use crate::deser::{codec::DogstatsdCodec, Decoder};
 
     use super::{InterceptAction, TagMetadataInterceptor};
--- a/lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
+++ b/lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
@@ -258,7 +258,8 @@ fn parse_dogstatsd_metric<'a>(
 ) -> IResult<&'a [u8], (&'a str, TagSplitter<'a>, ValueIter<'a>, MetricMetadata)> {
     // We always parse the metric name and value(s) first, where value is both the kind (counter, gauge, etc) and the
     // actual value itself.
-    let (remaining, (metric_name, values_iter)) = separated_pair(metric_name, tag(":"), metric_value)(input)?;
+    let (remaining, (metric_name, values_iter)) =
+        separated_pair(ascii_alphanum_and_seps, tag(":"), metric_value)(input)?;
 
     // At this point, we may have some of this additional data, and if so, we also then would have a pipe separator at
     // the very front, which we'd want to consume before going further.
@@ -338,7 +339,7 @@ fn parse_dogstatsd_metric<'a>(
 }
 
 fn parse_dogstatsd_event<'a>(input: &'a [u8], config: &DogstatsdCodecConfiguration) -> IResult<&'a [u8], EventD> {
-    // We parse the title length and text length from ``` _e{<TITLE_UTF8_LENGTH>,<TEXT_UTF8_LENGTH>}: ```
+    // We parse the title length and text length from `_e{<TITLE_UTF8_LENGTH>,<TEXT_UTF8_LENGTH>}:`
     let (remaining, (title_len, text_len)) = delimited(
         tag(message::EVENT_PREFIX),
         separated_pair(parse_u32, tag(b","), parse_u32),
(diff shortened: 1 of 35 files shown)


lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
```diff
@@ -287,6 +310,36 @@ fn parse_dogstatsd<'a>(
...
 
+fn parse_dogstatsd_event<'a>(input: &'a [u8], _config: &DogstatsdCodecConfiguration) -> IResult<&'a [u8], EventD> {
+    let event_header_parser = map(take_while1(|c| c != b':'), |bytes: &[u8]| unsafe {
+        std::str::from_utf8_unchecked(bytes)
+    });
+
+    let event_data_parser = map(
+        take_while1(|c: u8| c.is_ascii_alphanumeric() || c == b' ' || c == b'|' || c == b'_' || c == b'-' || c == b'.'),
+        |bytes: &[u8]| unsafe { std::str::from_utf8_unchecked(bytes) },
+    );
```
tobz, line 321: I think you could boil this down heavily and get rid of `parse_dogstatsd_event_header` entirely. `nom` makes it trivial to handle this sort of thing, and you've already got some of the underpinning below.

Essentially, what do we want out of the header? The title and text length. Their structure is well-formed: a prefix, a length, a separator, a length, and a suffix... so we just match that directly:

```rust
// In the imports:
use nom::character::complete::u32 as parse_u32;

// Down in `parse_dogstatsd_event`:
let (remaining, (title_len, text_len)) = terminated(delimited(EVENT_PREFIX, separated_pair(parse_u32, tag(b":"), parse_u32), tag(b"}:"))?;
```

After that, you've got `<title>|<text>|......` left, which you can use `take` for:

```rust
// In the imports:
use nom::bytes::complete::take;

// Down in `parse_dogstatsd_event`:
let (remaining, (raw_title, raw_text)) = separated_pair(take(title_len), tag(b"|"), take(text_len))?
```

Now you've got the raw byte slices for title and text, which you can try converting to UTF-8: look at our usage of `simdutf8::basic::from_utf8` in `metric_value` for how to do that.

After that, you've just got your list of optional values, which you can parse by reusing the pattern of how we parse optional fields in the metrics payloads.

tobz (changes requested):

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
```diff
@@ -317,6 +337,127 @@ fn parse_dogstatsd<'a>(
...
+                    maybe_source_type = Some(source_type.to_string());
+                }
+                // Alert type: client-provided alert type of the event.
+                event::ALERT_TYPE_PREFIX => {
+                    let (_, alert_type) =
+                        all_consuming(preceded(tag(event::ALERT_TYPE_PREFIX), event_option_value))(chunk)?;
+                    match alert_type {
+                        event::ALERT_TYPE_ERROR => maybe_alert_type = Some(saluki_event::eventd::AlertType::Error),
+                        event::ALERT_TYPE_WARNING => maybe_alert_type = Some(saluki_event::eventd::AlertType::Warning),
+                        event::ALERT_TYPE_SUCCESS => maybe_alert_type = Some(saluki_event::eventd::AlertType::Success),
```
tobz, line 427: One way to clean this up, and just sort of as a general pattern to follow: any time you're trying to parse some structured type from unstructured data (like going from a raw string to `AlertType`), etc... you _generally_ want to colocate that conversion logic with the the type itself.

There's some traits in Rust around conversions, both fallible and infallible, but in our case, we know we're dealing with `Option<T>`, and there's no trait for fallible conversions that return `Option<T>`, so we might just write an inherent method for this, following naming best practices:

```rust
impl AlertType {
    pub fn try_from_string(raw_alert_type: &str) -> Option<Self> {
        ...
    }
}
```

This function would have the same logic that you have here, and this codeblock would then just become `maybe_alert_type = AlertType::try_from_string(alert_type)`. We'd ideally do this for `Priority` as well.

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
```diff
@@ -317,6 +337,127 @@ fn parse_dogstatsd<'a>(
...
+                        _ => maybe_alert_type = Some(saluki_event::eventd::AlertType::Info),
+                    }
+                }
+                // Tags: additional tags to be added to the metric.
+                event::TAGS_PREFIX => {
+                    let (_, tags) = all_consuming(preceded(tag(event::TAGS_PREFIX), metric_tags(config)))(chunk)?;
```
tobz, line 433: While it's probably not a bad idea to follow the same logic as metrics, it would be good to check, and maybe even add a link to the code, showing that tags in events follow the same restrictions as tags in metrics.

tobz (changes requested):

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
```diff
@@ -173,8 +187,14 @@ where
...
 
+        let message_type = parse_message_type(data);
+        if message_type == MessageType::Event {
+            return self.decode_event(buf, events);
+        }
+
         // Decode the payload and get the representative parts of the metric.
-        let (remaining, (metric_name, tags_iter, values_iter, mut metadata)) = parse_dogstatsd(data, &self.config)?;
+        let (remaining, (metric_name, tags_iter, values_iter, mut metadata)) =
+            parse_dogstatsd_metric(data, &self.config)?;
```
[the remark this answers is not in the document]
rayz: I think i did try that but I had to put `decode_metric` inside the `impl<TMI> DogstatsdCodec<TMI>` block (since I can't put it inside `impl<TMI> Decoder for DogstatsdCodec<TMI>` due to the trait I believe) and then I think I ran into some error with the new tag filterer stuff `        let filtered_tags_iter = TagFilterer::new(tags_iter.clone(), &self.tag_metadata_interceptor);` but I'll take a look again in the service checks PR.

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
rayz: Yea, will need to look to see if this is documented somewhere

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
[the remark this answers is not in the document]
tobz: Hmm, well, we can leave this for the service checks PR or even a follow-on PR after that, just to make the diff easier to review as we play around with it.

tobz (approved):

PR #143, opened 2024-07-31
[APR-205] app: eliminate per-call allocations for querying process RSS

tobz:
## Context

Currently, we use a crate called `memory-stats` to query the RSS used by the process. While this crate works well, one downside is that it constantly allocates buffers to read the process' memory mappings.

While the buffers aren't insanely big or anything, it's just an irony that we're constantly doing little allocations (with all the potential for fragmentation, etc, that they bring) in code related to limiting the global process memory usage.

## Solution

This PR introduces our own small crate for querying RSS, `process-memory`. It has its own lightweight `Querier` type that allows for amortizing any allocations needed for buffer space, which entirely eliminates any ongoing allocations when we query for the process' memory usage.

We've added basic support for Linux, macOS, and Windows (untested). Adding more cross-platform testing is on our radar overall but given that we're only targeting Linux at the moment, supporting both Linux and macOS -- our primary development environment -- is really the only bar we need to meet.


Makefile
```diff
@@ -27,7 +27,7 @@ export CARGO_BIN_DIR ?= $(shell echo "${HOME}/.cargo/bin")
 export GIT_COMMIT ?= $(shell git rev-parse --short HEAD)
 
 # Specific versions of various Rust tools we use.
-export CARGO_TOOL_VERSION_dd-rust-license_tool ?= 1.0.3
+export CARGO_TOOL_VERSION_dd-rust-license-tool ?= 1.0.3
```
tobz, line 30: The name structure was wrong so if `dd-rust-license-tool` wasn't already present, the fallback code in the Makefile to install it would fail to run.

tobz:
> how do we determine if this is an improvement? is it just via the smp experiment comment?

In terms of the intended improvement, we wanted to eliminate all of the runtime allocations. This doesn't really show up as a memory usage reduction, but only as a reduced number of allocations from this code path. We want this -- in a general sense -- because consistently allocating brings a larger chance of allocator fragmentation.

Overall, I verified that we eliminated the runtime allocations by running the branch locally under `ddprof`, and examining the Continuous Profiler output... but this is a good point that it'd be nice to actually quantify this in some way.

I'm going to add a unit test that will let us demonstrate that no additional allocations happen after `Querier` is first created.

rayz (approved):

Issue #124, opened 2024-07-20
Add support to the DogStatsD codec for decoding events/service checks.


tobz:
Pretty straightforward: add support to the DSD codec to decode events and service checks.

Tricky part will just be updating the decoder logic since right now it's very geared towards metrics, insofar as the top-level `decode` function expects the DSD parser function to give it to the pieces of a metric... so we'll potentially need to add another layer of indirection where it returns an enum with the relevant fully-formed data (events, service checks) or the necessary individual pieces (metrics).

## Open questions / thoughts

Even though the event and service check payloads start with `_e` and `_sc`, respectively, a metric name is allowed to have underscores, but the language around _where_ those underscores can live is not specified... which would imply that metrics could also start their name with an underscore, which would make decoding a lot more ambiguous.

We'll need to confirm whether or not leading underscores are _actually_ allowed or not for metrics. If they're not, we should also open a PR to update the DogStatsD docs [here](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell?tab=servicechecks#the-dogstatsd-protocol).

PR #144, opened 2024-07-31
[APR-227] Add basic codec support for service checks

rayz:
# Context
Currently the codec does not have support for decoding dogstatsd service checks.

# Solution
This pr adds the implementation for decoding dogstatsd service checks and completes #124 

`codec/dogstatsd/event.rs` file is deleted and the field prefixes are moved to `message.rs` since all saluki event variants have the same prefixes / non-conflicting. https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=metrics#the-dogstatsd-protocol


lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
```diff
@@ -473,7 +564,7 @@ fn split_at_delimiter(input: &[u8], delimiter: u8) -> Option<(&[u8], &[u8])> {
 
 #[inline]
 fn ascii_alphanum_and_seps(input: &[u8]) -> IResult<&[u8], &str> {
-    let valid_char = |c: u8| c.is_ascii_alphanumeric() || c == b'_' || c == b'-' || c == b'.';
+    let valid_char = |c: u8| c.is_ascii_alphanumeric() || c == b' ' || c == b'_' || c == b'-' || c == b'.';
```
rayz, line 567: I added `b' '` so that service check messages can contain spaces. Also, this function used to be called `metric_name` since it was only used for parsing metric names. Metric names can have spaces (they just get converted to spaces) `statsd.increment('raymond space') => raymond_space` in dd

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
```diff
@@ -457,6 +472,82 @@ fn parse_dogstatsd_event<'a>(input: &'a [u8], config: &DogstatsdCodecConfigurati
...
+
+    let remaining = if !remaining.is_empty() {
+        let (mut remaining, _) = tag("|")(remaining)?;
+        while let Some((chunk, tail)) = split_at_delimiter(remaining, b'|') {
+            if chunk.len() < 2 {
+                break;
+            }
+
+            // Message field must be positioned last among the metadata fields but it was already seen
+            if seen_message {
```
rayz, line 500: break if we already saw `m:` but are still parsing optional fields since `m:` [should be positioned last](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=servicechecks). should we error instead?

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
```diff
@@ -161,6 +161,16 @@ impl<TMI> DogstatsdCodec<TMI> {
         buf.advance(data.len() - remaining.len());
         Ok(1)
     }
+
+    fn decode_service_check<B: ReadIoBuffer>(
+        &mut self, buf: &mut B, events: &mut EventBuffer,
+    ) -> Result<usize, ParseError> {
+        let data = buf.chunk();
+        let (remaining, service_check) = parse_dogstatsd_service_check(data, &self.config)?;
+        events.push(saluki_event::Event::ServiceCheck(service_check));
```
tobz, line 170: No need for this to be fully qualified since `Event` is already imported and in scope.

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
tobz: Conceptually, we should error, yes... but I think we should put the error in the catch-all match arm (`_ => {...}`) and then actually move the message parsing to outside of the loop, to make it a little clearer that we _always_ expect the message to 1) be present and 2) come as the very last field.

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
tobz: Do spaces in metric names get converted in the Agent itself on ingest/decode? If so, seems like we might need to open an issue to handle that... 😅

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
```diff
@@ -457,6 +472,82 @@ fn parse_dogstatsd_event<'a>(input: &'a [u8], config: &DogstatsdCodecConfigurati
...
 
+fn parse_dogstatsd_service_check<'a>(
+    input: &'a [u8], config: &DogstatsdCodecConfiguration,
+) -> IResult<&'a [u8], ServiceCheck> {
+    let (remaining, (name, raw_check_status)) = preceded(
+        tag(message::SERVICE_CHECK_PREFIX),
+        separated_pair(ascii_alphanum_and_seps, tag(b"|"), parse_u8),
+    )(input)?;
+
+    let check_status = saluki_event::service_check::CheckStatus::try_from(raw_check_status)
```
tobz, line 483: Nit: might just want to import `CheckStatus` directly to avoid needing the qualified path.

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
rayz: Conversion happens in the intake side.

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
rayz: > we always expect the message to 1) be present and 2) come as the very last field.

message is an optional field though.

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
tobz: Phew. 😅

tobz (approved):
Looks good to me. 👍🏻

Issue #126, opened 2024-07-20
Add a new destination to support sending events/service checks to the Datadog platform


tobz:
For supporting events/service checks in the DogStatsD source, we'll need a new destination to send them.

Given that they should both end up being trivially serialized, without specific logic, we'll likely just want a single destination that can handle both of them.

One thing to note is that while events cannot be batched in a single API call (at least according to the API docs), service checks _can_ be. This means we may or may not want to consider aggregation of service check payloads, but it's not yet clear to me if we should try and offload that to the aggregate transform (or another more generic aggregation transform) or if the destination should just handle it directly.

PR #154, opened 2024-08-05
[APR-229] Add named outputs for Saluki events

rayz:
# Context
Currently, a single `EventBuffer` is used to hold metrics, events, and service checks. We would like each event type to have their own named outputs.

# Solution
We use 2 additional `EventBuffer` for events and service checks, leaving the original event buffer dedicated to metrics. 

We also add a new blackhole destination for events and service checks, which we will remove when we have a destination that supports them (next pr  #126)

# Testing
Should not change anything for metrics.
Run `make run-dsd-basic-udp` and see metrics.

[Example for sending service checks
](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=servicechecks)

[Example for sending events
](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=events)

Blackhole will log that it has received events when sending events/service checks.

### DIFF
--- a/lib/saluki-components/src/sources/dogstatsd/mod.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/mod.rs
@@ -10,13 +10,14 @@ use saluki_core::{
     pooling::{FixedSizeObjectPool, ObjectPool as _},
     spawn_traced,
     topology::{
+        interconnect::{is_eventd, is_service_check},
         shutdown::{DynamicShutdownCoordinator, DynamicShutdownHandle},
         OutputDefinition,
     },
 };
 
 use saluki_error::{generic_error, GenericError};
-use saluki_event::{DataType, Event};
+use saluki_event::DataType;
 use saluki_io::{
     buf::{get_fixed_bytes_buffer_pool, BytesBuffer},
     deser::{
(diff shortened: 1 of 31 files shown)


lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -257,9 +259,16 @@ impl SourceBuilder for DogStatsDConfiguration {
     }
 
     fn outputs(&self) -> &[OutputDefinition] {
-        static OUTPUTS: &[OutputDefinition] = &[OutputDefinition::default_output(DataType::Metric)];
-
-        OUTPUTS
+        static OUTPUTS: Lazy<Vec<OutputDefinition>> = Lazy::new(|| {
```
rayz, line 262: `named_output` is not const so I used `once_cell` per the editor's suggestion in order to update `OUTPUTS`

lib/saluki-core/src/topology/interconnect/event_buffer.rs
```diff
@@ -115,6 +141,22 @@ impl Iterator for IntoIter {
...
+/// Predicate function for extracting eventd events.
+pub fn is_eventd(event: &Event) -> bool {
+    matches!(event.data_type(), DataType::EventD)
+}
+
+/// Predicate function for extracting service check events.
+pub fn is_service_check(event: &Event) -> bool {
+    matches!(event.data_type(), DataType::ServiceCheck)
+}
+
```
tobz, line 159: I would just move these directly to `Event` and have them match on the structure itself (i.e. `matches!(self, Event::EventD(_))`).

Simple enough as a basic helper method and this module is kind of an obscure place for them to leave, even if we primarily use them here.

tobz (changes requested):

lib/saluki-components/src/sources/dogstatsd/mod.rs
tobz: Actually, thinking about it... I'd also be fine if we bumped to Rust 1.80.0 (just change the value in `rust-toolchain.toml`) and used `std::sync::LazyLock`, to keep the slightly cleaner closure-in-variable-declaration pattern that we're using `Lazy` to achieve.

lib/saluki-event/src/lib.rs
```diff
@@ -120,4 +120,20 @@ impl Event {
             _ => None,
         }
     }
+
+    #[allow(unused)]
+    /// Predicate function for extracting metric events.
```
tobz, line 125: I would reword these as ``Returns `true` if the event is a metric``.

tobz (changes requested):
Just one small change and then I think this is good to go.

tobz (approved):

PR #167, opened 2024-08-09
Add an OpenTelemetry receiver source.

tobz:
## Context

Adds a new source for receiving OpenTelemetry data.

Super WIP.


tobz:
This is so wildly out-of-date now that keeping it around does not bear much value, if any at all.

PR #180, opened 2024-08-12
[APR-231] Add support to Datadog Metrics destination for timing out requests.

rayz:
# Context

Currently, the Datadog Metrics destination does not time out requests.

# Solution

We add a new config option `forwarder_timeout` to match the current datadog-agent settings, with a default of 20 seconds. `Tower` is used to wrap the http client which already implements the `tower::Service` trait.

# Testing

Change `dd_url` to point to a local http server that sleeps for a duration greater than `forwarder_timeout` upon receiving a request. You should receive an error ` Failed to send request. error=request timed out `.


lib/saluki-io/src/net/client/http.rs
```diff
@@ -52,4 +53,8 @@ where
     pub async fn send(&self, req: Request<B>) -> Result<Response<Incoming>, Error> {
         self.inner.request(req).await
     }
+
+    pub fn call(&mut self, req: Request<B>) -> ResponseFuture {
+        self.inner.call(req)
+    }
```
tobz, line 59: I think the simpler thing to do here would just be to actually implement `Service<Request<B>>` for `HttpClient<C, B>` rather than create the new `DatadogMetricsService<O>`.

lib/saluki-components/src/destinations/datadog_metrics/mod.rs
```diff
@@ -382,8 +404,9 @@ async fn run_io_loop<O>(
                     }
                 }
             }
-            Err(e) => {
-                error!(error = %e, error_source = ?e.source(), "Failed to send request.");
+            Err(_e) => {
+                error!("Failed to send request");
+                // error!(error = %e, error_source = ?e.source(), "Failed to send request.");
```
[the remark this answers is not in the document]
tobz: Chatted behind-the-scenes about this and it's just related to some type constraint stuff, so resolving this for now.

lib/saluki-components/src/destinations/datadog_metrics/mod.rs
```diff
@@ -383,6 +407,7 @@ async fn run_io_loop<O>(
                 }
             }
             Err(e) => {
+                let e: tower::BoxError = e.into();
```
tobz, line 410: Does it not actually work without the `into` call? `e` should already be `Box<dyn Error + ...>`.

rayz: It does not. Compiler shows `e` as `<S as Service<Request<ChunkedBuffer<O>>>>::Error` when I hover over it.

tobz: Oh, right, duh.

Change line 378 to be `S::Error:: std::error::Error + Send` and then you can drop the `into` down here.

tobz: For posterity: we paired on this because doing my suggested change still results in the compiler getting very angry about `DatadogMetrics<O, S>` being unsized and not being able to box it and yadda yadda.

Not worth fighting with the compiler over this, but maybe something to try and poke at on a rainy day. :shrug:

tobz (approved):

tobz (approved):

Issue #166, opened 2024-08-08
Investigate a design for establishing/collecting readiness and liveness status for a topology and its components.


tobz:
## Context

In order to have a better red/green view of the state of ADP at any given moment, it would be useful to have a simple design for allowing a topology, and by extension its components, to be able to indicate both when they are ready and able to begin processing (applicable to sources and destinations) as well if they are currently responding: readiness and liveness.

Practically speaking, we need this for further integration with the Datadog Agent: we need to be able to tell the Agent, and an orchestration runtime (for containerized environments) when ADP is ready and that it's still alive.

We will likely end up with a two-part design here:

- a mechanism for exposing the readiness/liveness information (standalone HTTP endpoint, over an IPC channel, etc)
- building blocks for instrumenting the topology, and components, such that they can report readiness/liveness in the first place

## Work items

- [x] #181
- [x] #192
- [x] #193

PR #181, opened 2024-08-13
[APR-208] Add a basic health check system for exposing readiness/liveness.

tobz:
## Context

This PR provides the basics to start addressing #166, which is centered around being to expose readiness/liveness endpoints for container orchestration runtimes to determine the health of ADP.

We've introduced two new crates to support this:

- `saluki-health`: new crate that defines the "health registry", used for registering and tracking the health of components
- `saluki-api`: new crate that wraps over `axum` and provides an opinionated way to expose an API server based on registering different "handlers", which register specific routes and handler-specific state data for those routes to use

The `saluki-api` crate provides a break in the dependency chain when wanting to utilize the core `APIHandler` trait that it exposes, while the meat of constructing and serving an API lives in `saluki-app`, which is spiritually aligned with what `saluki-app` aims to provide.

We've opted to not wire up the health stuff just yet because the changes will be somewhat pervasive and we need to do some more cleanup/consolidation of component lifecycle handling first.


lib/saluki-io/src/net/addr.rs
```diff
@@ -178,3 +181,9 @@ impl From<ProcessCredentials> for ConnectionAddress {
         Self::ProcessLike(Some(creds))
     }
 }
+
+impl<'a> Connected<&'a Connection> for ConnectionAddress {
+    fn connect_info(target: &'a Connection) -> Self {
+        target.remote_addr()
+    }
+}
```
tobz, line 189: We had to add this so that we can use our listener stuff from `saluki-io` with `axum`.

lib/saluki-io/src/net/stream.rs
```diff
@@ -39,6 +39,14 @@ impl Connection {
...
         }
     }
+
+    pub(super) fn remote_addr(&self) -> ConnectionAddress {
+        match self {
+            Self::Tcp(_, addr) => ConnectionAddress::SocketLike(*addr),
+            #[cfg(unix)]
+            Self::Unix(_) => ConnectionAddress::ProcessLike(None),
+        }
+    }
```
tobz, line 49: Similar story here re: using our listener stuff from `saluki-io` with `axum`.

rayz (approved):

PR #187, opened 2024-08-15
[APR-205] experiment: refactor `MetaString` to shrink size by one machine word

tobz:
## Context

Currently, `MetaString` is four machine words (32 bytes on 64-bit) large, while a normal `String` is three machine words (24 bytes). This is due to the overhead of the enum discriminant field: while only a `u8` is needed to encode all the discriminants, alignment requirements round this up to `usize`, on top of the largest variant (`String`), thus leaving us with four machine words.

Overall, this is less than ideal because we're always paying just a little more in memory to store a `MetaString` than a `String`, and double and quadruple the size cost of an `Arc<str>` and `InternedString`, respectively.

## Solution

This PR introduces a refactored `MetaString` based on a hand-rolled enum pattern using unions. This design encodes the discriminant into the data fields themselves by virtue of valid/invalid bitpatterns for those fields, allowing us to decode the information necessary to extract the discriminant on access.

`MetaString` still supports all four primary variants -- owned (`String`), static (`&'static str`), interned (`InternedString`), and inlined -- allow the maximum size of an inlined string is now 23 bytes, instead of 31 bytes, as we're limited by the overall size of the structure.


lib/stringtheory/src/lib.rs
```diff
@@ -1,58 +1,478 @@
...
+            //                                                                                    │
+            //                       inlined last byte                                            │
+            //                       maximum of 23 (0x17)      [0 0 0 1 0 1 1 1] ◀────────────────┤
+            //                                                                                    │
+            //                       "owned" discriminant                                         │
+            //                       bitmask (any X bit)       [X X X ? ? ? ? ?] ◀────────────────┘
+            //
+            // As such, the length byte of an inlined string overlaps with the capacity field of an owned string.
+            // Additionally, due to the integer fields being written in little-endian order, the "upper" byte of the
+            // capacity field -- the eight highest bits -- as actually written in the same location as the length byte
```
acharles1533, line 195:
```suggestion
            // capacity field -- the eight highest bits -- is actually written in the same location as the length byte
```

lib/stringtheory/src/lib.rs
```diff
@@ -1,58 +1,478 @@
...
+    fn as_str(&self) -> &str {
+        let union_type = unsafe { self.discriminant.get_union_type() };
+        match union_type {
+            UnionType::Empty => "",
+            UnionType::Owned => {
+                let owned = unsafe { &self.owned };
+                owned.as_str()
+            }
+            UnionType::Static => {
+                let _static = unsafe { &self._static };
```
acharles1533, line 337: I finally realized that you're using `_static` because `static` is a keyword. I kept focusing on `_foo` being the way to silence the `dead_code` warning. I guess I'd recommend using `static_` or `r#static` instead?

lib/stringtheory/src/lib.rs
```diff
@@ -1,58 +1,478 @@
...
 ///
-/// `MetaString` supports the following "modes":
+/// `MetaString` supports the following "variants":
 ///
-/// - owned (`String`)
+/// - owned (`String` and non-inlineable `&str`)
 /// - static (`&'static str`)
 /// - interned (`InternedString`)
-/// - inlined (up to 31 bytes)
+/// - inlined (up to 23 bytes)
```
acharles1533, line 469: `InlinedString` didn't support 31 bytes, only 23. :)

lib/stringtheory/src/lib.rs
```diff
@@ -261,78 +650,162 @@ impl fmt::Display for MetaString {
...
+    }
+
+    #[test]
+    fn owned_string() {
+        let s = String::from("hello");
+        let meta = MetaString::from(s);
 
-        // SAFETY: We know the data is valid UTF-8 because we only ever write valid UTF-8 data into the buffer.
-        unsafe { std::str::from_utf8_unchecked(s) }
+        assert_eq!("hello", &*meta);
```
acharles1533, line 695:
```suggestion
        assert_eq!(&s, &*meta);
```
Just like above?

lib/stringtheory/src/lib.rs
```diff
@@ -261,78 +650,162 @@ impl fmt::Display for MetaString {
...
-    type Target = str;
+    #[test]
+    fn static_str_not_inlineable() {
+        let s = "hello there, world! it's me, margaret!";
+        let meta = MetaString::from_static(s);
 
-    fn deref(&self) -> &str {
-        let len = self.data[0] as usize;
-        let s = &self.data[1..len + 1];
+        assert_eq!(s, &*meta);
```
acharles1533, line 686:
```suggestion
        assert_eq!(s, &meta);
```
Verify AutoDeref?

lib/stringtheory/src/lib.rs
```diff
@@ -261,78 +650,162 @@ impl fmt::Display for MetaString {
...
-        let s = "hello, world!";
-        assert_eq!(s.len(), 13);
+    fn empty_string_owned() {
+        // When a string has capacity, we don't care if it's actually empty or not, because we want to preserve the
+        // allocation... so our string here is empty but _does_ have capacity.
+        let s = String::with_capacity(4);
+
+        let meta = MetaString::from(s);
+        assert_eq!("", &*meta);
+        assert_eq!(meta.inner.get_union_type(), UnionType::Owned);
```
acharles1533, line 760: None of the tests call into_owned? It'd be particularly interesting here, since you could verify that you get a string with 4 capacity at the end.

acharles1533 (approved):

lib/stringtheory/src/lib.rs
tobz: Ooh, that's a good point, in order to test the invariant that owned strings going in are the same coming out.

lib/stringtheory/src/lib.rs
tobz: Yeah, fair point. Definitely not idiomatic to use a leading whitespace for an actually-used variable.

lib/stringtheory/src/lib.rs
tobz: Yeah, this comment is old but it _used_ to be true!

Two versions ago, with the enum-based approach, we had a `Bytes` variant, which was 32 bytes, and we sized `InlinedString` to fit that.

lib/stringtheory/src/lib.rs
tobz: It doesn't seem like autoderef kicks in here, since it's just focusing on if both sides have a lined up `PartialEq<T>` implementation.

lib/stringtheory/src/lib.rs
tobz: Good call.

lib/stringtheory/src/lib.rs
tobz: Fixed in c483845.

lib/stringtheory/src/lib.rs
tobz: Fixed in c483845 to be `static_`.

lib/stringtheory/src/lib.rs
tobz: Fixed in c483845.

lib/stringtheory/src/lib.rs
```diff
@@ -261,78 +650,162 @@ impl fmt::Display for MetaString {
...
 
-#[cfg(test)]
-mod tests {
-    use super::*;
+    #[test]
+    fn inlined_string() {
+        let s = "hello";
+        let meta = MetaString::from(s);
+
+        assert_eq!("hello", &*meta);
```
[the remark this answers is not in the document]
tobz: Fixed in c483845.

lib/stringtheory/src/lib.rs
tobz: Addressed in c483845.

acharles1533: Just curious, is the minimum allocation 32?

PR #195, opened 2024-08-15
[APR-205] Start to flatten out framing and codecs

lukesteensen:
The goal here is just to start reducing some complexity so that subsequent changes are easier to make and simpler to review. Where previously we had a `Deserializer` wrapping a framer wrapping a codec, we now just have a framer and a codec that are used side-by-side. By avoiding the nesting, we're making it easier to introduce a semi-parsed result struct from the codec without needing to thread reference lifetimes through multiple layers of machinery. Actually implementing that change to the codec's parsing function will be the next step, and from there we will be able to rework the rest of the logic (enrichment, differentiating between message types, forwarding, etc) to be a more straightforward function from that semi-parsed representation to a full event, with fewer intermediate steps and less mutation, as well as our ultimate goal, to easily pass in a string/context interner specific to that socket handler.

There's likely to be a few bits that were dropped that need to be added back (as well as some things that ought to be simplified further given the new state), so don't hesitate to point anything out in review 😅


lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -392,108 +429,181 @@ async fn process_listener(source_context: SourceContext, listener_context: Liste
...
 
-                trace!(%listen_addr, %peer_addr, events_len = n, "Forwarding events.");
+        // If our buffer is full, we can't do any reads.
+        if !buffer.has_remaining_mut() {
+            // return Err(DeserializerError::BufferFull {
+            //     remaining: buffer.remaining(),
+            // });
+            // try to get a new buffer on the next iteration?
+            continue;
+        }
```
tobz, line 474: This should also throw an error log, because a fresh I/O buffer should always be cleared after acquiring it.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -392,108 +429,181 @@ async fn process_listener(source_context: SourceContext, listener_context: Liste
...
-
-                if let Some(service_checks_event_buffer) = maybe_service_checks_event_buffer {
-                    if let Err(e) = source_context
-                        .forwarder()
-                        .forward_named("service_checks", service_checks_event_buffer)
-                        .await
-                    {
-                        error!(%listen_addr, %peer_addr, error = %e, "Failed to forward service checks events.");
+                Ok(None) => {
+                    debug!("No more frames to decode.");
```
tobz, line 514: I think this is more "Not enough data to decode frame."

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -392,108 +429,181 @@ async fn process_listener(source_context: SourceContext, listener_context: Liste
...
+            // try to get a new buffer on the next iteration?
+            continue;
+        }
 
-                if let Err(e) = source_context.forwarder().forward_named("metrics", event_buffer).await {
-                    error!(%listen_addr, %peer_addr, error = %e, "Failed to forward metric events.");
-                }
+        // Try filling our buffer from the underlying reader first.
+        debug!("About to receive data from the stream.");
+        let (bytes_read, peer_addr) = stream.receive(&mut buffer).await.expect("TODO");
```
tobz, line 478: Can we just flesh this back out to emit an error log and then break out of the receive loop? (i.e. lift up the code from the previous lines 491-493)

lib/saluki-components/src/sources/dogstatsd/framer.rs
```diff
@@ -1,26 +1,31 @@
...
+    Newline(NewlineFramer),
+    LengthDelimited(LengthDelimitedFramer),
+}
 
-multi_framing!(name => DogStatsD, codec => DogstatsdCodec<AgentLikeTagMetadataInterceptor>, {
-    Newline => NewlineFramer,
-    LengthDelimited => LengthDelimitedFramer,
-});
+impl Framer for DsdFramer {
+    fn next_frame<'a, B: saluki_io::buf::ReadIoBuffer>(
```
tobz, line 12: Small nit: we should use imports rather than fully-qualified paths.

(Gotta make a mental note to see if I can make `rustfmt` check/fix this for us automatically)

tobz (changes requested):
Just a few small things to fix, really.

Overall, I'm fine with the direction, and I'd like to get this merged quickly so we can continue iterating and polishing it back up... since I don't _love_ having this much complexity in the source for common things (like having to directly emit basic metrics, etc).

lib/saluki-components/src/sources/dogstatsd/mod.rs
lukesteensen: Yeah this is a bit of an odd one. Part of me wants to panic because if it happens we have a logic bug somewhere, the other part thinks it's a non-issue and we can just clear the buffer 😅 .

For now I'll just add the error log.

lib/saluki-components/src/sources/dogstatsd/framer.rs
lukesteensen: Agreed, that was an editor oopsie.

PR #208, opened 2024-08-20
[APR-208] chore: add cgroups v2 metadata collector

### LINKED ISSUE #186: Add workload collector for querying containers by inode based on cgroupsv2 groups.
## Context

As part of fully supporting origin detection in DogStatsD, we need to be able to resolve the inode-based form of "container ID" used in the DogStatsD protocol -- `|c:in-1234` -- into an actual container ID. This inode form represents the literal inode number of the cgroupsv2 group that the originating container is scoped to, which can be used to query the cgroupsv2 filesystem and recover the container ID.

We need a Linux-only collector that can scrape the cgroupsv2 hierarchy and extract a mapping of inodes to container ID. We only need the most simplistic data extraction here, purely to be able to push an ancestry link into the tag store. Likewise, we just need the simplest logic that gets us valid container IDs, and not just other arbitrary things using cgroups.

From a cursory search, the [`cgroupfs`](https://docs.rs/cgroupfs) crate should be sufficient for our needs here.

tobz:
## Context

This PR addresses #186 by adding a new metadata collector for cgroups v2.

This metadata collector scrapes all discovered control groups under the unified v2 hierarchy that match a simple naming heuristic. It feeds the metadata aggregator, similar to the containerd collector, with ancestry links, to be able to translate from cgroup controller inode to container ID.

We've done some work overall in terms of beefing up the feature detector to support configuring the new collector, as well as some logging cleanup and unification around string interning and so on.

## Notes

This was tested locally on a Linux machine using cgroups v2 (Ubuntu 23.10), with a single Docker container running. It correctly traverses the unified v2 hierarchy and finds the control group for the container, extracting both the correct container ID and the associated controller inode.

Fixes #186.


lib/saluki-app/src/memory.rs
```diff
@@ -228,7 +228,7 @@ pub async fn initialize_allocator_telemetry() -> Result<(), GenericError> {
     // We can't enforce, at compile-time, that the tracking allocator must be installed if a caller is trying to
     // initialize the allocator's reporting infrastructure... but we can at least warn them if we detect it's not
     // installed here at runtime.
-    if AllocationGroupRegistry::allocator_installed() {
+    if !AllocationGroupRegistry::allocator_installed() {
```
tobz, line 231: This is entirely unrelated, but is a simple fix for a logic bug that leads to us always emitting a warning that the tracking allocator isn't installed when, in fact, it is.

lib/saluki-env/src/workload/collectors/cgroups_v2.rs
```diff
@@ -0,0 +1,153 @@
...
+                match interner.try_intern(name.as_str()) {
+                    Some(interned) => Some(MetaString::from(interned)),
+                    None => {
+                        error!(container_id = %name.as_str(), "Failed to intern container ID.");
+                        None
+                    }
+                }
+            }
+        }
+        None => None,
```
lukesteensen, line 151: You could probably use `Option::filter` in place of the outer outer match and first conditional to make this a little more concise.

lukesteensen (dismissed):

lib/saluki-env/src/workload/collectors/cgroups_v2.rs
tobz: Good call. Cleaned this up a little in 8b9038c.

PR #213, opened 2024-08-22
[APR-240] Add support for configuring the idle connection timeout for HTTP clients.

rayz:
##  Context

When building an HTTP client, we should be able to configure the maximum idle connection age before the connection is terminated. Open connections consume resources on the downstream target systems, and we should prefer to be sparing with how long we try to hold on to reusable resources before we declare the efficiency gains we stand to make not worth the imposition on the downstream target systems themselves.

## Solution

Use [Builder::pool_idle_timeout](https://docs.rs/hyper-util/latest/hyper_util/client/legacy/struct.Builder.html#method.pool_idle_timeout)

Value hard coded to `45s` in [datadog-agent](https://github.com/DataDog/datadog-agent/blob/main/pkg/util/http/transport.go#L119)

### DIFF
--- a/lib/saluki-io/src/net/client/http.rs
+++ b/lib/saluki-io/src/net/client/http.rs
@@ -1,6 +1,6 @@
 //! Basic HTTP client.
 
-use std::task::Poll;
+use std::{task::Poll, time::Duration};
 
 use http::{Request, Response};
 use hyper::body::{Body, Incoming};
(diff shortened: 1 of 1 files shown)


lib/saluki-io/src/net/client/http.rs
```diff
@@ -64,6 +64,7 @@ where
         HttpClient {
             inner: Client::builder(TokioExecutor::new())
                 .pool_max_idle_per_host(5)
+                .pool_idle_timeout(Duration::from_secs(45))
```
tobz, line 67: Looks like we're missing setting the pool timer as the docs [mention](https://docs.rs/hyper-util/latest/hyper_util/client/legacy/struct.Builder.html#method.pool_idle_timeout) needing to do.

tobz (changes requested):

tobz (approved):

PR #225, opened 2024-08-30
[APR-205] chore: allow for contexts to be expired from `ContextResolver`

tobz:
## Context

Work in progress.


tobz:
Just to jot down some notes here..

The two biggest problems are that what we really want to be able to do is:

- avoid having to allocate in order to signal that a context is no longer used
- leave the metric around for a little while (like a cache, with a TTL) so that we aren't just _immediately_ blowing away resolved contexts

We can solve the first problem with `Arc<T>`-like semantics, just tracking when no outstanding reference to a context exists (besides our reference in the resolver) and then triggering the removal of that context... but that means while we're very _precise_ about expiration, we actually expire too fast which means we spend gobs of time re-interning because of having to search through the interner.

If we made the interner `O(1)`-esque, then this might not be a problem... but doing so would also mean that it would be far less bounded than it currently is.

Likewise, we can trivially solve the second problem by just incrementally iterating over the resolved contexts, with sleeps in between, which isn't so much a true TTL as much as it simply introduces an inherently delay between a context becoming unused and being cleaned up. This, however, means that we either need to use a scheme that allows crawling the list in chunks (which will need locking) or crawling it in full, every time, which is naturally more and more expensive as the number of resolved contexts go up... and still isn't a true TTL.

I was trying to noodle around the idea of how to make the "signal that this context is now unused" bit super cheap, which would allow us to register it somewhere that could then try to do more of a true "has it been unused for more than X seconds?" check... but so far I haven't come up with something sufficiently simple and performant.

PR #232, opened 2024-09-03
[APR-206] refactor to add shim for per-tenant context resolving/interning

lukesteensen:
The rebase on #217 was a little hairy, so there may be some bits left in a weird state.


lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -550,60 +615,52 @@ async fn drive_stream(mut stream: Stream, source_context: SourceContext, handler
...
+    let metric_metadata = build_metric_metadata_from_packet(&packet, origin_detection, peer_addr);
+    let mut context_resolver = multitenant_strategy.get_context_resolver_for_origin(metric_metadata.origin_entity());
+
+    // Try resolving the context first, since we might need to bail if we can't.
+    let context_ref = context_resolver.create_context_ref(packet.metric_name, &packet.tags);
+    let context = match context_resolver.resolve(context_ref) {
+        Some(context) => context,
+        None => {
+            source_metrics.failed_context_resolve_total().increment(1);
+            // TODO: maybe be louder here, but could be high frequency
```
tobz, line 643: I think it's fine not to log/be louder.

To your point, this could be very high frequency and throwing a bunch of extra stuff at stdout is likely to only slow things down/consume more memory anyways.

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
```diff
@@ -288,23 +227,67 @@ fn parse_dogstatsd_metric<'a>(
...
-            metric_name,
-            maybe_tags.unwrap_or_else(TagSplitter::empty),
-            metric_values,
-            metric_metadata,
-        ),
-    ))
+    // We do one optional enrichment step here, which is to add the client's socket credentials as a tag
+    // on each metric, if they came over UDS. This would then be utilized downstream in the pipeline by
+    // origin enrichment, if present.
+    if origin_detection {
```
tobz, line 266: We can actually drop `origin_detection` as a parameter, and this conditional that uses it.

We basically won't end up with any peer address when using UDS _unless_ we configure the socket early on to do grab it... which means that we can't ever get `Some(creds)` unless `origin_detection` is already `true`.

One less conditional, and one less function parameter.

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
```diff
@@ -623,7 +606,7 @@ fn limit_str_to_len(s: &str, limit: usize) -> &str {
 /// `TagSplitter` can be cloned to create a new iterator with its own iteration state. The same underlying input byte
 /// slice is retained.
 #[derive(Clone)]
-struct TagSplitter<'a> {
+pub struct TagSplitter<'a> {
```
tobz, line 609: It's fine for now, but more as a note to myself that I don't love this being exposed, and it'd be nice to find a clean way to avoid needing to expose this to support resolving the context.

It's _just_ weird/confusing enough of an `IntoIterator` impl (and in that same vein, weird how it's used) that my brain prefers it being hidden entirely.

tobz (commented):
Overall, this is looking reasonable to me!

I think we just need to clean up the literal TODO comments and what not, and otherwise the code/naming/etc is generally fine as-is.

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
lukesteensen: Does that apply all the way up to the config option? Seems like that is its only use currently.

lukesteensen: Went ahead and removed it all the way up.

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
lukesteensen: I agree it is a bit odd. I initially thought about returning something more like an array of tags, but this representation is nicely compact and sidesteps the issue of knowing ahead of time how many tags there might be to avoid heap allocation.

tobz: Yeah, I have no problem with the implementation itself, just that I wish there was maybe a _little_ cleaner of an implementation that it could have given that it needs to be public now.

tobz (approved):

Issue #118, opened 2024-07-17
How do we emit internal telemetry that works with existing Datadog Agent operational tooling?


tobz:
At a high-level, both Datadog Agent and Agent Data Plane/Saluki emit internal telemetry used for debugging performance issues and understanding their operational state. However, the naming differs between the two by a large amount, even for metrics that are functionally identical. This makes it challenging to use ADP, as it currently exists, as a drop-in replacement for DSD support in the core Agent.

The metric prefix we use when emitting internal metrics is configurable at the tippity top when initializing the metrics subsystem via `saluki_app::metrics::initialize_metrics`, so that's fine... but how do we line up individual metrics with their spiritual equivalent in the Datadog Agent?

This is a problem we need to solve if we hope to have ADP replace DSD in the core Agent.

tobz:
One idea: metric remapping.

Conceptually, specific components in Saluki map to specific components in the core Agent. For example, the DogStatsD source in ADP is the `dogstatsd` component in the Datadog Agent, and the Datadog Metrics destination in ADP is the `defaultforwarder` component in the Datadog Agent.
 If we included the component type in internal metrics (e.g., metrics from the Datadog Metrics destination have a `component_type` tag with a value of `datadog_metrics`), we could conceivably use that to remap metrics to their Datadog Agent equivalent.

For example, `datadog.agent.transactions.errors` in the Datadog Agent is used to track "transaction errors", which occur when the default forwarder fails to send a request to the Datadog intake. The `error_type` tag indicates the specific type of error. Similarly, on the Saluki side, the Datadog Metrics destination emits a `component_errors_total` metric, with an `error_type` tag that has a value of `http_send`, when we fail to send a request.

Since we should expect to only have one Datadog Metrics destination running in ADP, we could conceivably map all instances of `component_errors_total`, where `component_type` was equal to `datadog_metrics`, to `agent.transactions.errors`.. and potentially map the `error_type` tag as well.

We could likely do this pretty simply with a dedicated transform that remaps metric names, perhaps one even designed _solely_ for remapping to Datadog Agent-equivalent metric names. Biggest downside, I think, is just the general aspect of us having to maintain this mapping in the first place rather than doing it by default.

tobz:
We've ended up taking an approach (in #240) which blends the two ideas previously mentioned: a dedicated component that "remaps" specific configured metrics, but does so by emitting _additional_ metrics.

[...]

PR #240, opened 2024-09-09
[APR-207] chore: experimental approach to doing internal telemetry remapping

tobz:
## Context

In #118, we talked about the disconnect between the internal telemetry emitted by the Datadog Agent and Agent Data Plane. This gap makes it harder to test Agent Data Plane as a replacement for the DogStatsD portion of the core Agent as it requires focusing in to new telemetry and eschewing existing dashboards and monitors, which is not great.

## Solution

This PR takes a simple approach of adding a new ADP-specific component, the Agent Telemetry Remapper, which is a transform that emits Datadog Agent-compatible versions of Saluki internal metrics.

The transform is fairly hardcoded, defining all of the metric mappings manually, and has a minimum of helper code for making those definitions a little easier to write. It emits _additional_ metrics, which means that for any metric with a mapping configured, we emit the original _and_ a remapped version.

### DIFF
--- a/lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
+++ b/lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
@@ -158,6 +158,7 @@ impl<TMI: TagMetadataInterceptor> DogstatsdCodec<TMI> {
     fn decode_metric(&mut self, data: &[u8], events: &mut EventBuffer) -> Result<usize, ParseError> {
         // Decode the payload and get the representative parts of the metric.
         let (_remaining, (metric_name, tags_iter, values, mut metadata)) = parse_dogstatsd_metric(data, &self.config)?;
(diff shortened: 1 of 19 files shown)


lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
```diff
@@ -180,7 +181,7 @@ impl<TMI: TagMetadataInterceptor> DogstatsdCodec<TMI> {
 
         events.push(Event::Metric(Metric::from_parts(context, values, metadata)));
 
-        Ok(1)
+        Ok(values_len)
```
tobz, line 184: We actually want to emit the number of samples as the number of events because, ultimately, "metrics received" for the DogStatsD source should be able to speak to single _and_ multi-value payloads.

PR #241, opened 2024-09-09
[APR-231] Add support to Datadog Metrics destination for retrying requests.

tobz:
## Context

This is based on @rayz's work in #197, mostly to work through the remaining errors they were encountering but also to make it all a little more generic and tucked away, to avoid polluting the destination source code so much.


lib/saluki-io/src/net/util/hyper.rs
```diff
@@ -0,0 +1,59 @@
...
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use pin_project_lite::pin_project;
+use tower::{util::Oneshot, ServiceExt};
+
+/// A Tower service converted into a Hyper service.
+#[derive(Debug, Copy, Clone)]
+pub struct TowerToHyperService<S> {
```
tobz, line 12: We're doing this here because `hyper-util` hasn't caught up to the release of `tower` 0.5, which we're using for the updated retry layer bits.

Issue #185, opened 2024-08-13
Memory usage when extending event buffers is attributed to components


tobz:
## Context

Currently, the global event buffer pool is shared across components in order to move events through a topology. At startup, we allocate a fixed number of these buffers, but crucially, their capacity is empty. As the buffers are used, they are extended as needed.

This causes two main issues:

- event buffers are a potential avenue of unbounded memory growth
- the component that happens to end up causing the buffers to be resized gets stuck with the attributed memory usage

While the growth of these buffers generally won't _actually_ be unbounded -- a lot of other bounds will come into play first -- their misattributed memory usage is problematic, because it can completely blow the budget of a component that otherwise _doesn't_ allocate at all at runtime.

tobz:
In terms of solutions, this would need to be two-fold:

- event buffers would need to be of fixed-size
- we would need a new object pool that was elastic and could drop unused items after some period of being idle/unused

## Fixed-size event buffers

This one is the most difficult because while writing such a buffer itself is trivial, _using_ them is less so. Having to add in the code to handle the buffer being full, and then sending it and acquiring a new buffer, would be non-trivial and ugly. At present time, it would be _very_ ugly/contrived with something like the DogStatsD source, based on how we thread through a pre-acquired `EventBuffer` to `Deserializer`. Some of the work @lukesteensen is doing might make that a non-issue if we lift up the code where we actually generate, and thus buffer, the events we decoded.

For the rest of the cases, we would potentially write helper adapters to encapsulate the logic of iterating over something that generates the events, and then manages the event buffer being written into, sending it out and acquiring a new one when full.

## Elastic object pool

This would be pretty trivial, and would just involve needing to write the implementation.

PR #242, opened 2024-09-10
experiment: add elastic-sized object pool implementation

tobz:
## Context

In #185, we explored the concept of an elastic object pool as a way to help combat the nearly-unbounded growth, over time, of the existing `FixedSizeObjectPool<EventBuffer>` pattern.

## Solution

This PR introduces a new object pool implementation, `ElasticObjectPool<T>`, that instead of allocating all items upfront, instead is configured with a minimum and maximum capacity, and initially allocates only the minimum necessary items to populate the pool.

When a caller attempts to acquire an item, and the pool is empty but not yet at maximum capacity, items are allocated on-demand and then eventually returned back to the pool. In the background, a "shrinker" task runs periodically and measures both the "active" pool size (number of items attached to the pool, whether available or in use) and the average number of available items (as an exponentially weighted moving average).

When the active size is greater than the minimum capacity, and the average number of available items is greater than one, the shrinker will attempt to remove a single item from the pool, thereby decreasing its active size. This process is repeated every time the shrinker task wakes up and evaluates the pool state (once per second) until the active size has reached the minimum capacity or the average number of available items is less than one.

In doing so, we always allow the pool to burst up to the maximum capacity if necessary, and then slowly downsize it in the background until we reach an equilibrium with the average pool demand.

## Notes

This is clearly something that can be very sensitive to tuning: how quickly should we downsize the pool? should we be removing _specific_ items rather than just whatever it as the front/back of the list of items? and so on.

Right now, it defaults to the simplest thing I could conceive as being useful, but this is certainly not the end of the story if we find this provides a meaningful foundation for memory reduction over time.

### DIFF
--- a/lib/saluki-core/src/components/sources/context.rs
+++ b/lib/saluki-core/src/components/sources/context.rs
@@ -5,7 +5,7 @@ use saluki_health::{Health, HealthRegistry};
 
 use crate::{
     components::ComponentContext,
-    pooling::FixedSizeObjectPool,
+    pooling::ElasticObjectPool,
     topology::{
         interconnect::{EventBuffer, Forwarder},
         shutdown::ComponentShutdownHandle,
@@ -15,7 +15,7 @@ use crate::{
 struct SourceContextInner {
     component_context: ComponentContext,
     forwarder: Forwarder,
-    event_buffer_pool: FixedSizeObjectPool<EventBuffer>,
+    event_buffer_pool: ElasticObjectPool<EventBuffer>,
     memory_limiter: MemoryLimiter,
     health_registry: HealthRegistry,
     component_registry: ComponentRegistry,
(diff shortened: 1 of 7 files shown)


tobz:
Note to self: one of the things in my mind that I want to quantify/handle better is compensating for on-demand bursts.

Essentially, right now, our heuristic is very much focused on "what does the average number of available items look like?" which isn't wrong, but it ignores that demand can be very bursty. The instantaneous number of available items could easily approach zero between measurements N and N+1 made by the shrinker. As the object pool, and really all object pools for that matter, strive to _avoid_ unnecessary allocation, what we really want to do is reach an equilibrium with the average _demand_ such that we don't shrink the pool any smaller than the peak demand over the last rolling, let's say, 30 seconds or whatever.

The ideal outcome of approaching it from a demand-based perspective is that the pool size would instead reflect the peak observed instantaneous demand (smoothed, to some extent) over the past N seconds, which should then lead to the "allocate on demand" codepath being taken rarely, if ever.

PR #261, opened 2024-10-08
[APR-205] enhancement: switch to fixed-size event buffers

### LINKED ISSUE #185: Memory usage when extending event buffers is attributed to components

tobz:
## Context

#185 details the trouble with the current approach to how event buffers (`EventBuffer`) are used in Saluki, in regards to their bounded-ness and attributing their memory usage incorrectly to individual components, based on how/when those buffers are used and expanded.

As mentioned in that issue, one part of the overall solution is to use fixed-size event buffers as a means to at least remove the need to resize event buffers at runtime.

## Solution

This PR introduces the concept of fixed-size event buffers through a new poolable type, `FixedSizeEventBuffer`. This new type is simple and not particular interesting, save for only exposing a fallible `try_push` operation. However, as now all components need to be aware of the fallible naturing of pushing into an event buffer, we had to do a large amount of work to make doing that ergonomic, which is the bulk of this PR.

We've primarily worked in `Forwarder`, where we've done a few things:

- expose a new "buffered" forwarder type/helper methods (more below)
- convert the existing `forward`/`forward_named` methods to use a buffered sender, and take in an iterator of `Event`s
- expose a new set of methods for directly sending event buffers (`forward_buffer`/`forward_buffer_named`)

At a high-level, the main problem with the new fixed-size event buffer approach is dealing with needing to flush/send an event buffer once it has been filled up, and then acquiring a new event buffer to continue writing the remaining events into, so on and so forth, until all input events have been flushed/sent. The work to `Forwarder` aims to support this paradigm, obviously, but also doing so in a way that allows for more flexibility at the callsite where forwarding happens.

For example, in some cases, like the Agent telemetry remapper transform, we potentially end up generating new transformed variants of specific metrics, while still forwarding all of the original metrics. With the new exposed methods for forwarding, we can optimally buffer the new transformed metrics that get generated, and still forward the original event buffer when we're done.

In other cases, like the aggregate transform, we were previously using an out-of-band event buffer object pool to avoid extending the capacity of event buffers in the global event buffer pool, leading to large of amounts of allocated memory that contributed to a ballooning RSS. This worked well, but meant that we couldn't share the memory used by the dedicated aggregate transform event buffers with other components, and vice versa, which wasn't great. By using the new buffered forwarder paradigm, we can participate in using the global event buffer pool without pathologically blowing up memory usage.

## Notes

This PR is large, and beyond the main work around fixed-size event buffers, we've done some other things related to the effort:

- fixed up the bounds calculations for the topology builder (since we've radically changed how they work now, and since they actually have a bound we can depend on!)
- cleaned up the SMP experiment configs, specifically around certain context/memory limits, to ensure tests are running as intended
- added support to buffer incoming DogStatsD metrics in terms of time (100ms) to add more efficiency around use of the new buffered forwarder paradigm, since otherwise we'd still be forwarding metrics one-by-one which is incredibly suboptimal when event buffers no longer start off so small

## Future Looking Notes

A lot of this work has spawned conversations around the nature of how components are built/constructed. In particular, this work has shown that, more and more, components generally end up looking a lot like simple bits of logic -- decode, parse, transform, etc -- that could likely stand to be decoupled from the IO/scheduling/concurrency aspects that are imposed in order to compose these components into a usable data plane binary.

In particular, the need to flush event buffers in the middle of an overall "forward" operation point heavily to the fact that a more coroutine/generator-esque approach could be extremely useful to better model some of these operations.

While we have not done any of that experimentation here, it is more relevant than ever after going through the exercise of adding support for fixed-size event buffers, and it behooves us to investigate it more seriously in the future.

Fixes #185.


lib/saluki-core/src/topology/interconnect/fixed_event_buffer.rs
```diff
@@ -0,0 +1,340 @@
...
+        I::IntoIter: Send,
+        F: Fn(FixedSizeEventBuffer) -> Fut + Send,
+        Fut: Future<Output = Result<(), E>> + Send,
+        E: Send;
+}
+
+#[async_trait]
+impl<T> FixedSizeEventBufferPoolExt for T
+where
+    T: ObjectPool<Item = FixedSizeEventBuffer>,
```
blt, line 181: Intentionally left?

tobz: Ah, yeah. I had followed shortly after this comment with a commit to both remove the now-unused `EventBuffer`, and then also remove the commented-out code here. 😅

lukesteensen (dismissed):

PR #277, opened 2024-10-21
Python Check Support

scottopell:
Continuation of #48 

To get started: 

```
# If you don't have any python checks defined in your local FS
# Also sets up a minimal virtualenv with check dependencies
make gen-testing-pychecks
# Activate the venv
source ./checks_venv/bin/activate
# Run ADP
DD_API_KEY="foo00001" make run-adp-standalone
```


lib/saluki-components/src/sources/checks/python_exposed_modules.rs
```diff
@@ -0,0 +1,162 @@
...
+        name,
+        metric_type: mtype.into(),
+        value,
+        tags,
+    };
+
+    match module.getattr("SUBMISSION_QUEUE") {
+        Ok(py_item) => match py_item.extract::<Py<python_scheduler::PythonSenderHolder>>() {
+            Ok(q) => {
+                let res = pyo3::Python::with_gil(|py| q.bind_borrowed(py).borrow_mut().sender.clone());
```
GeorgeHahn, line 35: Nit, I don't think the gil needs to be acquired here - see https://pyo3.rs/main/performance#access-to-bound-implies-access-to-gil-token

scottopell: ah, you're right, that lets me get rid of this `sender.clone` as well, good spot!

PR #283, opened 2024-10-24
[APR-248] chore: add support for configurable lifecycle events in retry policies to enable contextual logging

### LINKED ISSUE #267: Log messages around HTTP requests being retried kind of suck.
## Context

A few problems here, really:

- the log message that retries are even happening are at the DEBUG level, so you get no indication by default that retries are happening, so things sometimes appear stalled (pretty bad)
- no contextual information about _what_ is being retried (request URI, status code causing the retry, etc)
- the log message comes from the retry code itself, so you can't even tell what downstream code (i.e. the component that's using it) the retry is originating from

All in all, this makes for pretty poor operational UX. I don't want to get _too_ deep into it, really, but we might just want to consider adding support to the retry middleware -- unfortunately, by having to fork it, i think -- for configurable lifecycle hooks and ways to have the retry policy emit contextual data that the hooks could access.

Essentially, have a way to emit a log when we make a request, when a request failed and is going to be retried, and when the retried request succeeds... all while being able to be given some contextual data back from the retry policy related to why a request was marked as needing to be retried.

tobz:
## Context

In #267, we detailed the various ways in which logging (or the lack thereof) of retries, well, sucks. Overall, the experience when retries are firing is suboptimal because there's just no user-visible indication that it's even occurring.

## Solution

This PR adds support for what we call a retry "lifecycle", a new trait that defines callbacks that are used to provide hooks into specific parts of the "lifecycle" of a retry: when we've decided that a retry needs to happen, and when a request was executed successfully. This now provides the means to add customized lifecycles tailored to the request/response at hand, such as a custom lifecycle for HTTP request/response types, which can then emit logs that explain why a response was considered for retry (4xx/5xx response code, TLS failure, etc).

While the `RetryLifecycle` trait is the main bit of scaffolding to make this possible, we've added a batteries-included lifecycle for HTTP-based services, called `StandardHttpRetryLifecycle`, that is meant to be used together with `StandardHttpRetryClassifier`. This lifecycle implementation provides rich, contextual information when retries are occurring, such as the request URI itself, a human-readable version of the underlying reason a retry was made, and it actually logs at the warning level, so users can _see_ that retries are happening.

## Notes

A good chunk of the work in this PR is split into two categories: refactoring the structure of all of the retry-related code, and the error handling for `StandardHttpRetryLifecycle`.

As far as the code restructuring, hopefully it's self-evident: we had a lot of code in one file, and it was starting to get out of hand, and likely soon to get further out of hand with any additional implementations, such as the ones we'll likely end up needing as we bolster gRPC-based clients, with all of their gRPC-specific needs.

The error handling code itself is also a lot of boilerplate, unwrapping, introspection, and so on. Examining a `Box<dyn Error>` to get a rich, contextual underlying cause is, well... tricky: downcasting to known error types to try and craft the best error message, going through all the different variants to specialize the error message or maybe leave it as the default, handling the inherent nesting of "source" errors within one another. Long story short, it's a decent amount of code, but I think it's fairly straightforward to understand, and should still be able to grow to accommodate more specialization in the future without being too gnarly.

## Example

One poignant example is related to errors that come up _before_ a response is received. While debugging an issue with @rayz, we were a bit stumped due to initially seeing no log output: nothing at the INFO level or above. We had to explicitly add a log line to `StandardHttpRetryClassifier` to emit a message when the result of a call to `HttpClient` was `Err(...)`. Doing so gave us this:

```text
# Condensed for readability.
2024-10-18T19:19:50.749110Z ERROR: request go boom: hyper_util::client::legacy::Error(Connect, Custom { kind: Other, error: Custom { kind: InvalidData, error: InvalidCertificate(NotValidForName) } })
```

As you can see, the output (using the error's `Debug` implementation) is not the greatest. We can see there's a TLS issue, but it's hard to tell why. What's the host we're having the issue with? For anyone not decently familiar with TLS, what does `NotValidForName` indicate? You might also be thinking "you're using the `Debug` output of the error... what about the `Display` output?"

Turns out, it was even worse:

```text
# Condensed for readability.
2024-10-18T19:21:37.034739Z ERROR: request go boom: client error (Connect)
```

This is just one of many possible errors that might be encountered, and it's frustratingly impenetrable without dropping down to a `Debug` implementation, which is OK for developers but utterly useless for users. Here's what the same exact error looks like with this PR:

```text
# Condensed for readability.
2024-10-23T22:59:22.591335Z  WARN: Request failed due to a TLS error: peer certificate is invalid: certificate is not valid for the given entity name. Retrying after 4.725096926s. error_count=2 request_uri=https://api.ddstaging.datadoghq.com/api/v2/series
```

First off, it's at the WARN level so users will see it now: no futzing with log levels. We have a human-readable error message (`Request failed due to a TLS error: peer certificate is invalid: certificate is not valid for the given entity name`) that gives a little more detail... not really enough for the user to fully solve this problem on their own, but at least something they can easily read. We're telling them what URI the request was being sent to, which in the case of this error, can be used to arrive at the answer: `api.ddstaging.datadoghq.com` is a valid DNS record, pointing at a frontend that serves a TLS certificate for `*.datadoghq.com`, which isn't valid for `api.ddstaging.datadoghq.com`, hence the "not valid for the given entity name". Most importantly... we actually tell the user that we're retrying the request!

Closes #267.

### DIFF
--- a/lib/saluki-components/src/destinations/datadog_metrics/mod.rs
+++ b/lib/saluki-components/src/destinations/datadog_metrics/mod.rs
@@ -20,7 +20,10 @@ use saluki_io::{
     buf::{BytesBuffer, ChunkedBuffer, FixedSizeVec, ReadWriteIoBuffer},
     net::{
         client::{http::HttpClient, replay::ReplayBody},
-        util::retry::{ExponentialBackoff, RollingExponentialBackoffRetryPolicy, StandardHttpClassifier},
+        util::retry::{
+            ExponentialBackoff, RollingExponentialBackoffRetryPolicy, StandardHttpClassifier,
+            StandardHttpRetryLifecycle,
+        },
     },
 };
 use serde::Deserialize;
@@ -248,6 +251,7 @@ impl DestinationBuilder for DatadogMetricsConfiguration {
         let recovery_error_decrease_factor =
             (!self.request_recovery_reset).then_some(self.request_recovery_error_decrease_factor);
         let retry_policy = RollingExponentialBackoffRetryPolicy::new(StandardHttpClassifier, retry_backoff)
+            .with_retry_lifecycle(StandardHttpRetryLifecycle)
             .with_recovery_error_decrease_factor(recovery_error_decrease_factor);
 
         let service = ServiceBuilder::new()
--- a/lib/saluki-io/src/net/util/retry/backoff.rs
+++ b/lib/saluki-io/src/net/util/retry/backoff.rs
@@ -1,20 +1,10 @@
 use std::{
     fmt,
-    sync::{
-        atomic::{
-            AtomicU32,
-            Ordering::{AcqRel, Relaxed},
-        },
-        Arc, Mutex,
-    },
+    sync::{Arc, Mutex},
     time::Duration,
 };
 
-use http::StatusCode;
 use rand::{thread_rng, Rng as _, RngCore};
-use tokio::time::{sleep, Sleep};
-use tower::retry::Policy;
-use tracing::debug;
 
 #[derive(Clone)]
 pub enum BackoffRng {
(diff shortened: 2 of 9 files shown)


rayz (approved):
🔥

Issue #270, opened 2024-10-19
Add full histogram support for DogStatD.


tobz:
## Context

Currently, our DogStatsD parser, and indeed Saluki's event model, only allow for the possibility of storing distributions: a histogram-like data structure (in the mathematical sense) that can answer arbitrary quantile queries against a set of samples. This happens for all histogram-like input types: timers, histograms, and distributions. We did this because it was a reasonable approximation of the average behavior of these metric types: be able to answer questions about the distribution of values observed.

However, in the Datadog Agent, histograms are treated separately (by default) where pre-determined aggregations/quantiles  are calcuated and emitted as individual metrics, rather than sending the entire distribution to the Datadog backend for generic querying.

We need to be able to match this behavior such that we can emit histograms in the same way for compatibility with the Datadog Agent.

tobz:
In my mind, the most reasonable design is one that involves adding a new variant to `MetricValues`, called `Histogram`, that holds a `DDSketch`. I believe this is the most reasonable design for a few reasons:

- we already store histograms this way (effectively), so memory usage won't change
- `DDSketch` can give us all of the default aggregations (min, max, sum, count) and default percentile(s) (p95) needed for histograms, rather than having to write new/more code to do so over an array of `f64` samples

tobz:
Closed via #270.

PR #279, opened 2024-10-22
[APR-248] chore: add proper support for sample rate in DSD codec

tobz:
## Context

In #271, we detailed the fact that the DogStatsD codec/source currently don't _do_ anything with the sample rate we parse from DogStatsD payloads. This inevitably leads to inconsistencies when payloads have a non-`1.0` sample rate.

## Solution

We've done a small amount of refactoring to rearrange the codec to collect the raw values, then the sample rate, and then generate the `MetricValues` by combining the sample rate with the raw values. We've emulated what the Datadog Agent does, which is:

- for counts, we just multiple the value by the reciprocal of the sample rate (e.g., a sample rate of `0.25` and value of `10` leads to `40` based on `10 * (1 / 0.25) -> 10 * 4 -> 40`)
- for timers, histograms, and distribution, we record the value once if the sample rate wasn't specified, or is 1.0 (current behavior), and if it's not 1.0, we record the value `n` times, where `n` is the reciprocal of the sample rate (e.g., a sample rate of `0.25` leads to recording the value 4 times (`1 / 0.25 -> 4`))

Closes #271.

PR #285, opened 2024-10-25
[APR-248] chore: add full support for DSD histograms/timer

tobz:
## Context

As mentioned in #270, we currently lack "full" histogram support when compared to the Datadog Agent. This means that with default Datadog Agent settings, sending the same DogStatsD histogram to both the Datadog Agent and Agent Data Plane results in different metrics being emitted. No bueno.

## Solution

We've added a new metric value type, `MetricValues::Histogram`, that is used to hold the individual samples and allow calculating statistics over those samples. This metric type is now used for histograms and timers coming in through the DogStatsD source, while `MetricValues::Distribution` is used exclusively for distributions.

We've additionally added support to the aggregate transform to allow decomposing histograms into a set of calculated statistics, which allows us to match the behavior of the Datadog Agent when handling histograms. The same configuration settings are exposed, and default values used, which means that ADP should handle histograms identically to the Datadog Agent.

### Handling of histograms in other components

While we specifically updated the aggregate transform with the intention of allowing for decomposing histograms, we also had to handle the new metric type in other metrics-based components, specifically the Datadog Metrics destination.

We've chosen to take the path of treating histograms as sketches, as we can build a `DDSketch` from the raw samples trivially. Obviously this shouldn't matter for ADP, as the pipeline is configured to decompose histograms before they reach the Datadog Metrics destination... but for other usages of histograms or future pipeline changes, that's how we'll handle them by default.

### Updates to correctness tests

We had to also update the correctness tests in two ways: stop compensating for a lack of true histograms, and compensate for the indeterminism of how aggregation buckets are aligned to the current time. The changes to update the correctness analysis itself, now that we have true histogram support, speaks for itself. The changes are the indeterminism bits are more interesting.

In running the correctness tests locally, I kept hitting an issue where certain histogram metrics were off, and the metrics that were off -- both the literal contexts and the actual statistics (e.g., max vs min vs avg) -- kept changing between runs. The details are a little dense, but essentially what was happening is that if all of the metrics sent by `millstone` did not land within a single aggregation bucket, we'd see multiple data points, and usually the last data point we saw (since these are gauges, and we only track the last value) would be "wrong": the maximum value for the histogram may have been seen in a prior aggregation bucket, and so on.

How we've approached this for now is to make `millstone` wait before sending. Since we know that both DSD and ADP will align their aggregation buckets to the wall-clock time, using a multiple of the bucket width in seconds, we know that if the current time is, for example, 14, we can wait until a current time of 20, which means we're now at the very beginning of the current aggregation bucket.

This effectively maximizes our chances of sending all payloads within a single aggregation bucket, avoiding the issues described above. In local testing, this is very effective: I could not get the indeterminism originally observed to resurface after applying this delay logic. It's not a resilient long-term solution -- what if the aggregation bucket width changes? etc -- but it should be satisfactory in the medium-term.


tobz:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/saluki/285?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#285** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/285?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#279** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/279?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#278** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/278?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

tobz:
One update here is that while this approach does work, the current situation is that we end up with a roughly ~0.5% difference in the computed aggregates between Datadog Agent and ADP. This is trivially attributed to the fact we're using `DDSketch` on the ADP side, which will be inherently lossy (even if the error is relatively bounded).

I'm reaching out internally to see if various stakeholders are okay to keep this in the short-term, where we could then refactor down the road to reach exact parity, since there's a lot more work to get exact parity, and histograms are relatively rare compared to other metric types.

blt (approved):
Neat.

PR #291, opened 2024-10-29
chore: add new crate, `saluki-metadata`, for easier cross-crate build metadata access

tobz:
## Context

Whew, that PR title is a mouthful!

One recurring problem – albeit a small one – is that the splitting of the binary portion of ADP (`bin/agent-data-plane`) and the library portion (`lib/saluki-*`) means that it's difficult to pass "build metadata" from the ADP binary into the Saluki libraries.

As a refresher, the ADP binary (and software projects in general) will push "build metadata" into the main binary so that when you run the application, or do something like `./my-app-name --version`, you get some nice output like this:

```text
toby:~/src/some-app$ target/debug/some-app --version
some-app version 0.1.0 (c3acd1d, x86_64-unknown-linux-gnu, built 2024-10-29T10:06:58Z)
```

That build metadata -- the version, Git commit hash, target architecture, build time, etc -- usually gets passed in from either the outside (a build script) or from something like Cargo itself, which supplies environment variables with some of this information, and crates can read them at build time via `env!` and so on. The crucial part, though, is that this usually only gets pushed into the binary being built, and not the dependencies used by the binary.

Where this is a problem is if any of the code in your dependencies also wants to access this information. One example of this in Saluki is the Datadog Metrics destination, which wants to apply some of this information to the requests it send: a special URL depending on the application version, an HTTP header indicating the version, and so on.

If the build metadata is generally only able to be passed to the binary being built, but not the dependencies of that binary, what do you do?

## Solution

This PR introduces a new crate, `saluki-metadata`, to provide a reasonably ergonomic shim over the concept of cross-crate build metadata.

Our approach involves setting many of these items (but not all, more on that later) via environment variables, which `saluki-metadata` reads at build time (in a build script, `build.rs`) and then turns into static data which gets embedded in the crate itself. From there, we expose a free function, `saluki_metadata::get_app_details` which returns a simple struct that contains all of these values: application name/short name, version, Git commit hash, build time, target architecture, and so on.

Both the ADP binary and dependent crates (like `saluki-components`) can access this build metadata simply by calling `saluki_metadata::get_app_details`, knowing that they're both using the same data, and in a portable way. So long as the given data plane binary (ADP or any future data planes built on Saluki) both import and use `saluki-metadata`, and the build process sets the right environment variables, they'll both see the same build metadata.

## Notes

While almost all of the information comes from user-supplied environment variables, the one piece that doesn't is target architecture. Since we know that the target architecture is going to be the same for any dependency of the data plane binary and the data plane itself, we can just collect that from Cargo and avoid having to calculate and pass yet another environment variable.

We've also updated our Makefile to set these environment variables globally for any target, and the values it uses are roughly derived from the ADP `Cargo.toml` (version, specifically) while everything else comes from the environment (Git commit hash, build time) or is set manually (app name/short name).

## Outstanding Items

- [ ] Update the build steps in CI to use the same environment variables that the Makefile does, perhaps by adding a Make target that emits them such that they can be exported into the environment before building ADP.


tobz:
Example of output when running ADP now:

```
❯ docker run --rm -it registry.ddbuild.io/saluki/agent-data-plane:0.1.0-beta-47764941-af955273
2024-10-29T19:37:46.607618Z  INFO agent_data_plane: Agent Data Plane starting... version="0.1.0" git_hash="af95527" target_arch="aarch64-unknown-linux-gnu" build_time="2024-10-29T19:19:28Z"
```

PR #292, opened 2024-10-29
[APR-233] Add support to Datadog Metrics destination for shipping to multiple endpoints.

### LINKED ISSUE #117: Add support to Datadog Metrics destination for shipping to multiple endpoints.
Currently, the Datadog Metrics destination does not support multiple endpoints, only the "main" endpoint as configured by either `site` or `dd_url`. The Datadog Agent _does_ support multiple endpoints through its general `additional_endpoints` configuration scheme, used across logs and metrics for configuring additional endpoints to dual (or triple! etc) ship to.

We should:

- add support for dual shipping (this should also use `tower`, but there's no existing `tower` middleware for fanning out a request, so we'll have to write that ourselves, and take care around making request cloning cheap)
- match the existing `additional_endpoints` configuration in the Datadog Agent's forwarder

rayz:
# Context

This PR adds support to Datadog Metrics destination for shipping to multiple endpoints in order to improve Datadog Metrics destination resiliency and functionality to match Agent forwarder.

# Approach

Parse `DD_ADDITIONAL_ENDPOINTS` and store the domains and api keys. In the `run_io_loop` we are able to clone the Request due to `ReplayBody`. We then set the appropriate uri and headers and send the request.

# Testing
1. Configure `DD_ADDITIONAL_ENDPOINTS`
2. Send a metric
3. Watch the metric show up across the different datadog websites.

Closes #117.


lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -0,0 +1,114 @@
...
+
+impl<'de> Deserialize<'de> for AdditionalEndpoints {
+    fn deserialize<D>(deserializer: D) -> Result<AdditionalEndpoints, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s: String = Deserialize::deserialize(deserializer)?;
+        Ok(AdditionalEndpoints::from_string(&s))
+    }
+}
```
tobz, line 83: With this implementation, we can _only_ support setting additional endpoints as a JSON-encoded string, so we wouldn't be able to use an existing `datadog.yaml` with `additional_endpoints` set.

I think what we should do here is:

- change `AdditionalEndpoints` to `AdditionalEndpoints(HashMap<String, Vec<String>>)`, and add `#[derive(Deserialize)]`, so that it can be natively deserialized from the existing structure that you would find in `datadog.yaml`
- add a `FromStr` impl for `AdditionalEndpoints` that calls into `serde_json::from_str`, to add the necessary support of parsing it from a JSON string
- finally, we use `serde-with` and the `PickFirst` helper to allow us to deserialize `AdditionalEndpoints` via either its native nested map form _or_ a JSON string: https://docs.rs/serde_with/latest/serde_with/struct.PickFirst.html#examples

Since we implemented `FromStr`, you could use their example with `DisplayFromStr` as-is as that's the codepath we'd want to utilize. We don't be about making this serializable, though.

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -0,0 +1,114 @@
...
+    endpoints: &AdditionalEndpoints,
+) -> Result<Vec<SingleDomainResolver>, EndpointError> {
+    let mut resolvers = Vec::new();
+    for (domain, api_keys) in &endpoints.endpoints {
+        let new_domain = add_adp_version_to_domain(domain.to_string(), "adp".to_string())?;
+        let mut seen = HashMap::new();
+        let mut resolver = SingleDomainResolver::default();
+        for api_key in api_keys {
+            let trimmed_api_key = api_key.trim();
+            if trimmed_api_key.is_empty() || seen.contains_key(trimmed_api_key) {
```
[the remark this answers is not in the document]
rayz: Intent is to disallow duplicate API keys for the same endpoint.

https://github.com/DataDog/datadog-agent/blob/main/pkg/config/utils/endpoints.go#L56-L74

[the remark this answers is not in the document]
tobz: Huh, weird... seems odd that people would have duplicates within a single endpoint, when the bigger problem is using the wrong API key with the wrong endpoint (where duplicates would really be problematic).

Anyways, no harm in following the Agent's logic. 👍🏻

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -0,0 +1,152 @@
...
+use regex::Regex;
+use saluki_error::GenericError;
+use saluki_metadata;
+use serde::Deserialize;
+use snafu::{ResultExt, Snafu};
+use url::Url;
+
+use crate::destinations::datadog_metrics::DEFAULT_SITE;
+
+/// AdditionalEndpoints holds the endpoints and api keys parsed
```
tobz, line 16: Super small nit: I know that the Go style guide pushes to do `// <function/struct name> ....` for comments, but we avoid this in Saluki (and in Rust, in general).

A more idiomatic doc comment for this might be:

```suggestion
/// A set of additional API endpoints to forward metrics to.
///
/// Each endpoint can be associated with multiple API keys. Requests will be forwarded to each unique endpoint/API key pair.
```

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -0,0 +1,152 @@
...
+
+/// Error type for invalid endpoints.
+#[derive(Debug, Snafu)]
+#[snafu(context(suffix(false)))]
+pub enum EndpointError {
+    Parse { source: url::ParseError, domain: String },
+}
+
+#[derive(Default)]
+/// Holds the mapping between domains and their api keys.
```
tobz, line 29: Two small nits, fixed as part of the suggestion:

- We want derives to be right above the struct, and doc comments to be on top.
- Capitalization/casing for common acronyms like `API`.

```suggestion
/// A single API endpoint and its associated API keys.
#[derive(Default)]
```

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -0,0 +1,152 @@
...
+        self.api_keys.push(api_key);
+    }
+
+    fn set_domain(&mut self, domain: String) {
+        self.domain = domain;
+    }
+}
+
+/// Converts AdditionalEndpoints into a list of SingleDomainResolvers with its
+/// destination domain and api keys.
```
tobz, line 56: Small nit: in Rust doc comments, since they're all Markdown-compatible, we try to surround type/function names in backticks so that they get rendered as nice monospaced inline code blocks.

In some cases, we also try to surround them in link brackets, since then rustdoc will autogenerate intra-doc links to let you navigate to the referenced types quickly.

```suggestion
/// Converts [`AdditionalEndpoints`] into a list of [`SingleDomainResolver`]s.
```

(Just a little extra context: you generally wouldn't need/want to do the link syntax if the doc comment is already on the type itself, or a method of that type... but since this is a free function, the link syntax is nicer because then we ensure that the person reading the docs in their IDE hover, or whatever, can quickly navigate to either of the referenced types.)

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    let mut new_domain = domain.clone();
+
+    // Url::Parse doesnt work without this
+    if !domain.starts_with("https://") && !domain.starts_with("http://") {
+        new_domain = format!("https://{}", domain);
+    }
+
+    let d = new_domain.clone();
+    let url = Url::parse(&new_domain).context(Parse { domain })?;
```
tobz, line 115: I think you can make this a little cleaner, and avoid the clones, by using `with_context` instead of `context`:

```suggestion
    let new_domain = if !domain.starts_with("https://") && !domain.starts_with("http://") {
        format!("https://{}", domain)
    } else {
        domain
    };
    
    let url = Url::parse(&new_domain).with_context(|_| Parse { domain.clone() })?;
```

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -0,0 +1,152 @@
...
+        let new_subdomain = get_domain_prefix("adp".to_string());
+
+        let new_host = &host.replacen(subdomain, &new_subdomain, 1);
+
+        return Ok(new_host.to_string());
+    }
+    Err(EndpointError::Parse {
+        source: url::ParseError::EmptyHost,
+        domain: d,
+    })
```
tobz, line 133: You can make this a little more idiomatic (and some of this also is meant to work in tandem with the above suggestion) by using a `match` block:

```suggestion
    match url.host_str() {
        Some(host) => {
            // Do not update unknown URLs
            if !dd_url_regexp.is_match(host) {
                return Ok(host.to_string());
            }

            let subdomain = host.split('.').next().unwrap_or("");
            let new_subdomain = get_domain_prefix("adp");
            let new_host = host.replacen(subdomain, &new_subdomain, 1);

            Ok(new_host)
        },
        None => Err(EndpointError::Parse {
            source: url::ParseError::EmptyHost,
            domain: new_domain,
        }),
    }
```

(I also removed some stuff like the reference on `&host.replacen(...)` and then the `new_host.to_string()` because you already get an owned string back from `replacen`)

lib/saluki-components/src/destinations/datadog_metrics/mod.rs
```diff
@@ -499,39 +522,56 @@ async fn run_io_loop<O, S>(
...
+        let clone = request.clone();
+        let mut requests = vec![request];
+
+        for resolver in &resolvers {
+            for api_key in resolver.api_keys() {
+                let mut clone_req = clone.clone();
+                let uri = clone_req.uri_mut();
+                let mut parts = uri.clone().into_parts();
+                parts.authority = Some(resolver.domain().parse().expect("invalid authority"));
+                *uri = Uri::from_parts(parts).expect("Failed to construct new URI");
```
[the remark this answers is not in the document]
tobz: Looks like we still haven't done this yet.

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -0,0 +1,152 @@
...
+    api_keys: Vec<String>,
+}
+
+impl SingleDomainResolver {
+    /// The domain of the resolver.
+    pub fn domain(&self) -> &String {
+        &self.domain
+    }
+
+    /// The api keys associated with the domain.
```
tobz, line 41:
```suggestion
/// Returns the API keys associated with the domain.
```

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -0,0 +1,152 @@
...
+
+#[derive(Default)]
+/// Holds the mapping between domains and their api keys.
+pub struct SingleDomainResolver {
+    domain: String,
+    api_keys: Vec<String>,
+}
+
+impl SingleDomainResolver {
+    /// The domain of the resolver.
```
tobz, line 36:
```suggestion
/// Returns the domain of the resolver.
```

lib/saluki-components/src/destinations/datadog_metrics/mod.rs
```diff
@@ -1,7 +1,9 @@
 use std::time::Duration;
 
 use async_trait::async_trait;
-use http::{Request, Uri};
+mod endpoint;
+use endpoint::endpoints::{create_single_domain_resolvers, determine_base, AdditionalEndpoints, SingleDomainResolver};
```
tobz, line 5: Style nit: we generally prefer to prefix imported modules with `self` to visually differentiate that they're our own modules rather than a crate dependency of the same name.

```suggestion
use self::endpoint::endpoints::{create_single_domain_resolvers, determine_base, AdditionalEndpoints, SingleDomainResolver};
```

tobz (changes requested):

tobz (approved):

tobz (changes requested):
Oh, woops: looks like this needs a `make sync-licenses` to pass CI.

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -0,0 +1,149 @@
...
+        "{}-{}-{}-{}.agent",
+        version.major(),
+        version.minor(),
+        version.patch(),
+        app
+    )
+}
+
+/// Prefixes the domain with the ADP version.
+pub fn add_adp_version_to_domain(domain: String) -> Result<String, EndpointError> {
```
tobz, line 103: I think we _probably_ just want some simple unit tests for this?

At some point, it's very likely we'll want to refactor this code to use it generically for handling all types of different DD intake endpoints, and right now, it's very specific to metrics (i.e. all endpoints need to be `app.<site>`) so having some simple unit tests will obviously help us avoid breaking the code in the future, but make it a little more clear what it is does and what it expects the input to look like.

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -0,0 +1,149 @@
...
+    let dd_url_regexp = Regex::new(r"^app(\.mrf)?(\.[a-z]{2}\d)?\.(datad(oghq|0g)\.(com|eu)|ddog-gov\.com)$").unwrap();
+
+    let new_domain = if !domain.starts_with("https://") && !domain.starts_with("http://") {
+        format!("https://{}", domain)
+    } else {
+        domain.clone()
+    };
+
+    let url = Url::parse(&new_domain).with_context(|_| Parse { domain: domain.clone() })?;
```
tobz, line 112: I think we want to return `new_domain` in the error, not `domain`, just so that it's fully indicative of what the parsed input was.

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    match url.host_str() {
+        Some(host) => {
+            // Do not update unknown URLs
+            if !dd_url_regexp.is_match(host) {
+                return Ok(host.to_string());
```
tobz, line 118: I think we may want a debug log here saying that we're avoiding modifying the domain since it appears to be custom.

Something like...

> Configured endpoint '{host}' appears to be a non-Datadog endpoint. Utilizing endpoint without modification.

tobz (changes requested):
A few small change requests.

tobz (approved):

PR #297, opened 2024-11-04
[APR-224] chore: add max connection age support + builder for `HttpClient`

### LINKED ISSUE #179: Add support for retiring old connections for HTTP clients.
## Context

When building an HTTP client, we should be able to configure a time-to-live for all pooled connections, such that an individual connection that is idle can never be older than the configured TTL. This is useful for ensuring that new connections are periodically established to the downstream target system, aiding in scenarios like switching over to newly provisioned IPs when load balancer nodes are being replaced, etc.

## Notes

This is not something currently exposed by `hyper-util` but I think, conceptually, could be easy enough to upstream.

tobz:
## Context

In many scenarios, it is advantageous to re-use connections for network-based clients, as this can allow avoiding paying the cost of DNS lookups, reconnecting, re-establishing TLS, and so on. However, it is also advantageous to sometimes allow retiring those reused connections so that brand-new connections can be established: if operators are attempting to, for example, roll out updates to a load balancer, they may want connections to the old load balancer to go away so that subsequent connections are established with the new load balancer instead.

## Solution

We've added support for setting a "max connection age" for connections made by `HttpClient`. We've done so by creating a new extension type, `MaxConnectionAge`, which is seeded into each established connection at the moment of creation. This allows us to check the age of the connection each time it is used, and if the age of the connection exceeds a configurable limit, we "poison" it such that it skips being re-added to the idle connection pool, and is instead replaced with an entirely new connection.

To better facilitate configuring this, along with other client settings, we've also introduced a new builder, `HttpClientBuilder`, for setting the relevant client settings such as timeouts, TLS configuration, and more. The intent here is to move more and more logic/middleware behavior into `HttpClient` and configure it all through ergonomic builder methods on `HttpClientBuilder`, rather than forcing the composition to need to happen at the point of usage (e.g., individual components).

Closes #179.


tobz:
* **#298** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/298?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#297** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/297?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

rayz (approved):

PR #305, opened 2024-11-04
chore: update Rust to 1.82.0

tobz:
## Context

I'm doing the Rust update separately from the deps update (#304) to try and track down why the `dsd_uds_100mb_250k_contexts` test got such a weird memory usage regression.

### DIFF
--- a/bin/correctness/ground-truth/src/analysis/metric.rs
+++ b/bin/correctness/ground-truth/src/analysis/metric.rs
@@ -30,11 +30,6 @@ impl NormalizedMetricContext {
     pub fn name(&self) -> &str {
         &self.name
     }
-
-    /// Returns the tags of the metric.
-    pub fn tags(&self) -> &[String] {
-        &self.tags
-    }
 }
 
 impl fmt::Display for NormalizedMetricContext {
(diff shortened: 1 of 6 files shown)


tobz:
This is in fact a legit regression in Rust 1.82 where the generated machine code for `miniz-oxide` is demonstrably worse, which causes us to compress requests slower... slow enough, in fact, that the Datadog Metrics destination pushes backpressure all the way up to the source causing the slowdowns/perturbations in the ingress throughput that we see in the SMP capture data.

We've opened an upstream issue (https://github.com/rust-lang/rust/issues/132636) to track this down.

tobz:
Looks like this comment on the aforementioned issue indicates that building with `opt-level=2` instead of `3` might solve the issue.

We should create a draft PR changing from `3` to `2` without changing the Rust version, and then make another draft PR based off of that with Rust updated to 1.82. This would give us a good idea of changing to `opt-level=2` loses us any performance (or a meaningful amount, at least) and then also if switch to `opt-level=2` with Rust 1.82 actually fixes the observed performance regression here.

tobz:
Closing this as we leapfrogged to 1.85.0 without observing the same regression in compression performance.

PR #306, opened 2024-11-06
[APR-270] Add support for splitting oversized payloads in the Datadog Metrics destination

### LINKED ISSUE #159: Add support for splitting oversized payloads in the Datadog Metrics destination.
## Context

Currently, the Datadog Metrics destination lacks the ability to split request payloads when we detect that finalizing a request payload has resulted in a payload size that exceeds the allowable limits of the Datadog metrics API. When this happens, all we currently do is discard the request and move on, which means dropped metrics.

We should add support to retry the finalization of a request payload by taking the list of metrics we've encoded so far, splitting it in half, and trying to create two smaller request payloads.

rayz:
# Context
This pr adds support for splitting oversized metric payloads. Previously, oversized payloads would just be dropped.

# Notes 
This pr introduces a change to the `scratch_buf` , allowing it to grow up to the endpoints `uncompressed_size_limit`. Instead of holding one singular encoded metric, `scratch_buf` will hold multiple and be cleared after `flush`.

Closes #159

### DIFF
--- a/lib/saluki-components/src/destinations/datadog_metrics/request_builder.rs
+++ b/lib/saluki-components/src/destinations/datadog_metrics/request_builder.rs
@@ -177,10 +177,10 @@ where
         // If not, we return the original metric, signaling to the caller that they need to flush the current request
         // payload before encoding additional metrics.
         let encoded_metric = encode_single_metric(&metric);
-        let before = self.scratch_buf.len();
+        let previous_len = self.scratch_buf.len();
         encoded_metric.write(&mut self.scratch_buf)?;
-        let delta = self.scratch_buf.len() - before;
-        self.scratch_buf_lens.push(delta);
+        let encoded_len = self.scratch_buf.len() - previous_len;
+        self.scratch_buf_lens.push(encoded_len);
 
         // If the metric can't fit into the current request payload based on the uncompressed size limit, or isn't
         // likely to fit into the current request payload based on the estimated compressed size limit, then return it
@@ -189,7 +189,6 @@ where
         // TODO: Use of the estimated compressed size limit is a bit of a stopgap to avoid having to do full incremental
         // request building. We can still improve it, but the only sure-fire way to not exceed the (un)compressed
         // payload size limits is to be able to re-do the encoding/compression process in smaller chunks.
-        let encoded_len = delta;
         let new_uncompressed_len = self.uncompressed_len + encoded_len;
         if new_uncompressed_len > self.endpoint.uncompressed_size_limit()
             || self
@@ -206,15 +205,15 @@ where
             return Ok(Some(metric));
         }
 
-        let start = self.scratch_buf.len().saturating_sub(delta);
+        let start = self.scratch_buf.len().saturating_sub(encoded_len);
 
         // Write the scratch buffer to the compressor.
         self.compressor
             .write_all(&self.scratch_buf[start..])
             .await
             .context(Io)?;
-        self.compression_estimator.track_write(&self.compressor, delta);
-        self.uncompressed_len += delta;
+        self.compression_estimator.track_write(&self.compressor, encoded_len);
+        self.uncompressed_len += encoded_len;
         self.metrics_written += 1;
 
         trace!(
@@ -256,13 +255,10 @@ where
 
         self.compression_estimator.reset();
 
-        // Clear scratch buf data.
-        let scratch_buf = std::mem::take(&mut self.scratch_buf);
-        let scratch_buf_lens = std::mem::take(&mut self.scratch_buf_lens);
-
         let new_compressor = create_compressor(&self.buffer_pool).await;
         let mut compressor = std::mem::replace(&mut self.compressor, new_compressor);
         if let Err(e) = compressor.shutdown().await.context(Io) {
+            self.clear_scratch_buffer();
             return vec![Err(e)];
         }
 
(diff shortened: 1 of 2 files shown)


lib/saluki-components/src/destinations/datadog_metrics/request_builder.rs
```diff
@@ -243,11 +256,79 @@ where
...
+        &self, scratch_buf: Vec<u8>, scratch_buf_lens: Vec<usize>,
+    ) -> Vec<Result<(usize, Request<ReplayBody<ChunkedBuffer<O>>>), RequestBuilderError>> {
+        let mut requests = Vec::new();
+
+        if scratch_buf_lens.is_empty() {
+            return requests;
+        }
+
+        let pivot = scratch_buf_lens.len() / 2;
+        let first_half: usize = scratch_buf_lens.iter().take(pivot + 1).sum();
```
tobz, line 292: I see what you're doing, but I think it could be a little simpler.

You have four possible scenarios:

- no metrics encoded (you handle this when checking `scratch_buf_lens.is_empty()`)
- single metric encoded
- even number of metrics encoded (two or more)
- odd number of metrics encode (three or more)

If you have a single metric, you're going to end up writing no bytes to the second compressor, because `first_half` will be a number that equals the total number of bytes in the scratch buffer. If you have two metrics, you're going to do the same thing, since you always take at least two items from `scratch_buf_lens` if it has two or more elements. Once we get to three metrics, it'll start biasing the larger half towards "first".

So we can solve the problem of "single metric" by just putting a check in `flush`, when we exceed our compressed size limit, to return an error directly: if a single metric can't be compressed within the limit, we have nothing to split, and thus nothing to send.

When it comes to two or more metrics, I think a small tweak in your logic would fix things up. Specifically, if we do something like this:

```rust
let lens_pivot = scratch_buf_lens.len() / 2;
let first_half_metrics_len = scratch_buf_lens.len() - lens_pivot;
let second_half_metrics_len = scratch_buf_lens.len() - first_half_metrics_len;

let scratch_buf_pivot = scratch_buf_lens.iter().take(first_half_metrics_len).sum();

let first_half_scratch_buf = scratch_buf[0..scratch_buf_pivot];
let second_half_scratch_buf = scratch_buf[scratch_buf_pivot..];
```

Instead of always adding one to the value given to `take`, you do it implicitly by capturing the rounded-down portion of `scratch_buf_lens.len() / 2`: when `scratch_buf_lens.len()` is 2, the pivot is 1, so we only take 1, and when `scratch_buf_lens.len() / 2` is 3, the pivot is also still 1, but since we do `scratch_buf_lens.len() - pivot`, we take 2 on the first half.

The rest of the variables are just then deriving from that, and in particular, generating the relevant slices for the first and second half. We also have `first_half_metrics_len` and `second_half_metrics_len` now, which we should be using when we build the requests: currently you're passing a hard-coded value of `1` which will mess up the metrics we emit.

lib/saluki-components/src/destinations/datadog_metrics/request_builder.rs
```diff
@@ -243,11 +256,79 @@ where
...
+
+        let mut compressor_half_one = create_compressor(&self.buffer_pool).await;
+
+        if first_half <= scratch_buf.len() {
```
tobz, line 296: I'm not actually sure you need these conditionals (this one and the similar one further down) unless you're incorrectly generating the length of each write way further up in `encode`.

You should do this once, perhaps with an `assert!`, right below my proposed `scratch_buf_pivot`, and just assert that `scratch_buf_pivot < scratch_buf.len()`, since that's all we need to do to ensure we never generate an invalid slice reference.

lib/saluki-components/src/destinations/datadog_metrics/request_builder.rs
```diff
@@ -243,11 +256,79 @@ where
 
         self.compression_estimator.reset();
 
+        // Clear scratch buf data.
+        let scratch_buf = std::mem::take(&mut self.scratch_buf);
```
tobz, line 260: Taking the existing scratch buffer requires that we replace it with a new `Vec<u8>`, which means we're always going to be allocating a new scratch buffer every time we flush.

I would probably just add a new method, `clear_scratch_buffer`, or something... and have it clear the buffer, and the lengths buffer. You'd then add a call to `clear_scratch_buffer` in the relevant spots:

- before we return if we fail to shutdown the compressor (right before line 266)
- right before we return the single request in the non-split scenario (right before line 279)
- right before we return our split requests (right before line 324)

lib/saluki-components/src/destinations/datadog_metrics/request_builder.rs
```diff
@@ -175,7 +177,10 @@ where
         // If not, we return the original metric, signaling to the caller that they need to flush the current request
         // payload before encoding additional metrics.
         let encoded_metric = encode_single_metric(&metric);
+        let before = self.scratch_buf.len();
         encoded_metric.write(&mut self.scratch_buf)?;
+        let delta = self.scratch_buf.len() - before;
```
tobz, line 182: I would just call this `encoded_len` and then skip declaring it down below, especially since you're mixing `encoded_len` and `delta` right now, which becomes confusing.

tobz (changes requested):

lib/saluki-components/src/destinations/datadog_metrics/request_builder.rs
```diff
@@ -201,11 +205,15 @@ where
...
 
+        let start = self.scratch_buf.len().saturating_sub(encoded_len);
+
         // Write the scratch buffer to the compressor.
-        self.compressor.write_all(&self.scratch_buf).await.context(Io)?;
-        self.compression_estimator
-            .track_write(&self.compressor, self.scratch_buf.len());
-        self.uncompressed_len += self.scratch_buf.len();
+        self.compressor
+            .write_all(&self.scratch_buf[start..])
```
tobz, line 212: Replace `start` with `previous_len`.

tobz (changes requested):

PR #315, opened 2024-11-08
chore: get internal stage out of normal pipeline + make release stage one-click

tobz:
## Context

Just some general cleanup that hides the `internal` stage (used for building helper images) from normal CI pipeline runs, as well as tweaks the `release` stage to make it so you don't have to manually click each job to start... just the initial image build job step.


tobz:
Note: the reported check status is kind of lying, since it's reporting on the _latest_ pipeline run for this branch which I triggered manually and then cancelled after ensuring the manual run made the `internal` stage show up. The original CI pipeline triggered when I pushed this branch completed all required checks.

PR #326, opened 2024-11-14
Sup3rmark patch 2

sup3rmark:
Test for @tobz


tobz:
Despite the cuteness of the Japanese raccoon dog, and the even cuter Mario variant that pays homage to it, I must unfortunately close this PR as we're not looking to switch mascots at this time.

image

PR #329, opened 2024-11-15
chore: try shrinking global event buffer pool to a minimum of 32 buffers

tobz:
## Context

Testing different changes to the global event buffer pool, specifically around the size of the buffers themselves as well as the minimum/maximum size of the object pool.

This PR tests changing the minimum size of the object pool to 32 items instead of 64.


tobz:
This seems to have a positive effect on idle RSS (not as much as the other variants) and has no noticeable impact on the 100MB/s x 250K contexts experiment (while the other two variants do) so I'm going to merge this one.

PR #330, opened 2024-11-15
chore: try shrinking global event buffer pool to a minimum of 32 buffers and 512 items per buffer

tobz:
## Context

Testing different changes to the global event buffer pool, specifically around the size of the buffers themselves as well as the minimum/maximum size of the object pool.

This PR tests changing the minimum size of the object pool to 32 items instead of 64, and the size of the buffers to 512 items instead of 1,024.


tobz:
Closing this as it ends up having a negative perturbing effect on high context count experiments as it oscillates between flushing metrics from the aggregate transform and accepting new metrics.

Issue #162, opened 2024-08-08
Add support for External Data-based origin detection in the DogStatsD codec.


tobz:
## Context

We currently support the more basic/generalized origin detection mechanisms, such as utilizing the peer process ID on the other end of a UDS-based DogStatsD client, or the "container ID" DogStatsD extension, or the `dd.internal.entity_id` tag.

However, we lack support for the newly-added ["External Data" origin detection](https://github.com/DataDog/datadog-agent/blob/main/comp/dogstatsd/server/parse.go#L49), which uses a stringified list of data attributes, passed in a new DogStatsD field, to provide richer origin data than just container ID alone.

We should add support for this in order to ensure our origin detection support is at parity with the Datadog Agent.

tobz:
This was completed in #389.

PR #336, opened 2024-11-18
[APR-248] chore: refactor metadata aggregator to support multiple stores

tobz:
## Context

This PR introduces some refactoring of the core workload primitives, namely `MetadataAggregator`, in order to better support multiple stores that subscribe to metadata operations.

This is in preparation of adding support for a Remote Agent "workload" collector (in addition to the existing Remote Agent "tag" collector) where ADP will be able to subscribe to a stream of `workloadmeta` events from the Datadog Agent and use them to collect the necessary information to map External Data entries to containers, necessary for #162.

## Notes

The PR primarily introduces the concept of "stores", which are added to `MetadataAggregator` and receive metadata operations to be processed. We've added support for two types of stores -- consuming and borrowing -- to support stores which really only need to look at the operation data and those which benefit from being able to consume it, perhaps due to wanting to avoid unnecessary string allocations, and so on. Most of the refactoring is centered around this.

We did some very small, additional cleanup mostly to make `RemoteAgentWorkloadProvider::from_configuration` less crappy to look at.

### DIFF
--- a/lib/saluki-env/src/workload/aggregator.rs
+++ b/lib/saluki-env/src/workload/aggregator.rs
@@ -113,7 +113,11 @@ impl MemoryBounds for MetadataAggregator {
     }
 }
 
+/// A store which receives a stream of metadata operations.
 pub trait MetadataStore: MemoryBounds {
+    /// Returns the name of the store.
     fn name(&self) -> &'static str;
+
+    /// Process a metadata operation.
     fn process_operation(&mut self, operation: MetadataOperation);
 }
--- a/lib/saluki-env/src/workload/aggregator.rs
+++ b/lib/saluki-env/src/workload/aggregator.rs
@@ -1,6 +1,3 @@
-use std::{num::NonZeroUsize, sync::Arc};
-
-use arc_swap::ArcSwap;
 use memory_accounting::{MemoryBounds, MemoryBoundsBuilder};
 use saluki_health::Health;
 use tokio::{select, sync::mpsc};
@@ -9,12 +6,9 @@ use tracing::debug;
 use super::{
     collectors::{MetadataCollector, MetadataCollectorWorker},
     metadata::MetadataOperation,
-    store::{TagSnapshot, TagStore},
 };
 
 // TODO: Make this configurable.
-// SAFETY: The value is demonstrably not zero.
-const DEFAULT_ENTITY_LIMIT: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(2000) };
 const OPERATIONS_CHANNEL_SIZE: usize = 128;
 
 /// Aggregates the metadata from multiple collectors.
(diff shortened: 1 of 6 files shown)


blt (approved):

lib/saluki-env/src/workload/store.rs
```diff
@@ -384,22 +400,29 @@ impl MemoryBounds for TagStore {
...
+#[derive(Default)]
+struct TagSnapshot {
     low_cardinality_entity_tags: AHashMap<EntityId, TagSet>,
     high_cardinality_entity_tags: AHashMap<EntityId, TagSet>,
 }
 
-impl TagSnapshot {
+/// A snapshotter that supports concurrently querying the tag store.
+#[derive(Clone)]
+pub struct TagSnapshotter {
```
lukesteensen, line 411: Nitpicking, but I don't love this name. IIUC this is a pretty simple read handle or view, while "snapshotter" alludes to some more complex logic that I don't think exists. This made the code more confusing to read, because it seemed like there should be something to this concept of snapshotting and I wasn't seeing it.

lib/saluki-env/src/workload/aggregator.rs
```diff
@@ -107,6 +126,22 @@ impl MemoryBounds for MetadataAggregator {
...
+
+pub trait ConsumingStore: MemoryBounds {
+    fn store_name(&self) -> &'static str;
+    fn process_operation(&mut self, operation: MetadataOperation);
+}
+
+pub trait BorrowingStore: MemoryBounds {
+    fn store_name(&self) -> &'static str;
+    fn process_operation(&mut self, operation: &MetadataOperation);
+}
```
lukesteensen, line 147: I know this is intended for use in an upcoming change, so this might be totally off the mark, but my initial reaction is skepticism that it's worth the conceptual overhead to have two different traits for this. I know it enables us to avoid extra cloning in some circumstances, but it's not obvious to me that we'll see actual benefit from that. Happy to wait and see though.

lukesteensen (dismissed):

lib/saluki-env/src/workload/store.rs
tobz: That's a fair point, considering I thought the same thing after renaming it.

I'll tweak it to be something closer to the intended usage.

lib/saluki-env/src/workload/aggregator.rs
tobz: Also a fair point.

I can collapse them for now and have a consuming variant only. I know that the upcoming usage _does_ benefit from only needing immutable access, but this codepath is fairly low volume. All of these optimizations are mostly just trying to preemptively stymie the growth of very small allocations over time... but maybe we can approach it differently in the future and do so in a lightweight way where we're not having to futz with the traits _here_, but with a global optimization somewhere else, etc.

I'll collapse them for now.

tobz:
* **#338** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/338?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#336** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/336?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/saluki/336?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

lib/saluki-env/src/workload/store.rs
tobz: I think/hope this sounds more reasonable now as `TagStoreQuerier`.

lib/saluki-env/src/workload/aggregator.rs
tobz: Alright, I collapsed them for now, so we just have `MetadataStore` which is a consuming variant.

blt (approved):

lukesteensen (approved):

PR #337, opened 2024-11-19
[APR-248] rearrange things to make sure we intern metadata strings

lukesteensen:
This is a bit rough around the edges still, but opening to get feedback. Things that are currently awkward:
1. We test metric metadata in the deserializer (parser) tests, but there are fields somewhat outside the scope of the parser (imo) like `process_id`, `origin`, etc. This means we're either making up data to fill the tests or duplicating logic from the source.
2. Origin is an optional field in the metadata but we set it unconditionally (this seems like it was for convenience in tests)
3. A few metadata fields were using empty strings to signal `None` (maybe an optimization? It can be put back).
4. Some of the builder-style APIs on metadata (e.g. `set_container_id`) very effectively hid places where we should have been interning strings via `Into<MetaString>`, which is a bit of a performance footgun (arguably correctness for bounding as well).
5. I probably overcorrected with `pub` fields, but it was the easiest way to get this working for now.

None of this should be terribly difficult to sort out, and there's room for some nice simplification once we do, but wanted to make sure folks are on the same page before making any real design decisions.


lib/saluki-context/src/tags.rs
```diff
@@ -105,17 +105,17 @@ impl<'a> BorrowedTag<'a> {
...
-    pub fn value(&self) -> Option<&str> {
+    pub fn value(&self) -> Option<&'a str> {
         self.0.split_once(':').map(|(_, value)| value)
     }
 
     /// Gets the name and value of the tag.
     ///
     /// For bare tags (e.g. `production`), this always returns `(Some(...), None)`.
-    pub fn name_and_value(&self) -> (Option<&str>, Option<&str>) {
+    pub fn name_and_value(&self) -> (&'a str, Option<&'a str>) {
```
tobz, line 115: 👍🏻

tobz:
> 1. We test metric metadata in the deserializer (parser) tests, but there are fields somewhat outside the scope of the parser (imo) like `process_id`, `origin`, etc. This means we're either making up data to fill the tests or duplicating logic from the source.

I agree that the parser really shouldn't care except for knowing how to adjust the raw input payload in order to exercise those codepaths. I do think the tests are valuable overall: for example, we have all the various permutations because of a real bug I caught where I broke how optional fields are parsed which didn't show up unless the optional fields were in a certain order.

I'm ambivalent to where you place the tests, but it does feel like we need to keep them in spirit to avoid regressing the codec.

> 2. Origin is an optional field in the metadata but we set it unconditionally (this seems like it was for convenience in tests)

I'm guessing I made this optional trying to futureproof for a scenario where we weren't deal with Agent metrics. I'm fine with making it required until that need actually arises.

> 3. A few metadata fields were using empty strings to signal `None` (maybe an optimization? It can be put back).

This was a hack to to have `Option<T>` semantics without the size increase of actually having the fields be `Option<MetaString>`. I'm not strongly tied to having accessors that return `Option<&str>`, though, so fine to eliminate these for now.

> 4. Some of the builder-style APIs on metadata (e.g. `set_container_id`) very effectively hid places where we should have been interning strings via `Into<MetaString>`, which is a bit of a performance footgun (arguably correctness for bounding as well).

> 5. I probably overcorrected with `pub` fields, but it was the easiest way to get this working for now.

This is a good point. My brain personally enjoys builder APIs, but I'm not strongly tied to them. A middleground could be that we just make them more restricted by requiring `MetaString` instead of transparently converting them... but again, I'm not _strongly_ tied to the builderAPI.

My inclination would be to have a builder API if we do end up with a good number of optional fields to avoid massive, multi-line struct initializers... that's it, really.

Overall, this approach seems fine to me. 👍🏻

lukesteensen:
@tobz 

> I do think the tests are valuable overall: for example, we have all the various permutations because of a real bug I caught where I broke how optional fields are parsed which didn't show up unless the optional fields were in a certain order.
>
> I'm ambivalent to where you place the tests, but it does feel like we need to keep them in spirit to avoid regressing the codec.

Strongly agree! I wouldn't get rid of any of the tests, was just trying to figure out a good way to split out the dirty external stuff (i.e. connection address) from the pure parsing concerns and organize the tests accordingly. But that's something we can maybe follow up on.

> I'm guessing I made this optional trying to futureproof for a scenario where we weren't deal with Agent metrics. I'm fine with making it required until that need actually arises.

Not a big deal, this makes sense.

> This was a hack to to have Option<T> semantics without the size increase of actually having the fields be Option<MetaString>. I'm not strongly tied to having accessors that return Option<&str>, though, so fine to eliminate these for now.

Noticed this regression and backed out the change.

> This is a good point. My brain personally enjoys builder APIs, but I'm not strongly tied to them. A middleground could be that we just make them more restricted by requiring MetaString instead of transparently converting them... but again, I'm not strongly tied to the builder API.
> 
> My inclination would be to have a builder API if we do end up with a good number of optional fields to avoid massive, multi-line struct initializers... that's it, really.

I'm good with that too, I just think the current implementation via setters/getters ends up encouraging/allowing mutation in a way that's limiting for sharing metadata, for example. Just having separate builder structs would fix that.

tobz (approved):
Looks good to me.

I think what we might want to consider in the future is simply changing the DSD codec tests to deal with `MetricPacket<'a>` instead of `Metric` when parsing.

As nice as it would be to just have codecs only return `Event`/`Event`-compatible types, that's clearly not the way to approach things when we need to obsess over performance/efficiency, so we can probably afford to be more sparse in the tests just to avoid polluting them with too much setup/boilerplate.

Nothing you need to do here unless you really want to.

PR #339, opened 2024-11-19
[APR-276] Add support to pull the configured memory limit from Cgroups

### LINKED ISSUE #310: Add support to pull the configured memory limit from the environment (a la cgroups).
## Context

Currently, we only support configuring a memory limit when it's passed directly to ADP via environment variable. In containerized environments, especially in Kubernetes, it's far more common to set resource "requests" and "limits" for CPU and memory.

We should add support for doing autodiscovery of any such memory limits such that, if ADP is running in a containerized environment and the environment has these limits in place, we can utilize them unless explicitly overridden (via `DD_MEMORY_LIMIT`, etc).

rayz:
# Context

We would like to pull configured memory limits from the environment (Cgroups).

# Solution

If the memory limit is not set, we check for `DOCKER_DD_AGENT` to see if we are running in a containerized environment. We pull the memory limit from the appropriate place depending on which Cgroup is enabled (v1 vs v2).

Closes #310

### DIFF
--- a/lib/saluki-app/src/memory.rs
+++ b/lib/saluki-app/src/memory.rs
@@ -79,6 +79,10 @@ impl MemoryBoundsConfiguration {
                 if !value.is_empty() {
                     let cgroup_memory_reader = CgroupMemoryParser;
                     if let Some(memory) = cgroup_memory_reader.parse() {
+                        info!(
+                            "Setting memory limit to {} based on detected cgroups limit.",
+                            memory.to_string_as(true)
+                        );
                         config.memory_limit = Some(memory);
                     }
                 }
@@ -273,9 +277,6 @@ struct CgroupMemoryParser;
 
 impl CgroupMemoryParser {
     /// Parse memory limit from memory controller.
-    ///
-    /// `parse_controller_v2` is called if a unified controller is found in /proc/self/cgroup.
-    /// `parse_controller_v1` is called on the controller with ":memory:" present.
     ///
     /// Returns `None` if memory limit is set to max or if an error is encountered while parsing.
     fn parse(self) -> Option<ByteSize> {
--- a/lib/saluki-app/src/memory.rs
+++ b/lib/saluki-app/src/memory.rs
@@ -74,22 +74,25 @@ impl MemoryBoundsConfiguration {
             .as_typed::<Self>()
             .error_context("Failed to parse memory bounds configuration.")?;
 
+        if config.memory_limit.is_none() {
+            // Try to pull configured memory limit from Cgroup if running in a containerized environment.
+            if let Ok(value) = env::var("DOCKER_DD_AGENT") {
+                if !value.is_empty() {
+                    let cgroup_memory_reader = CgroupMemoryParser {};
+                    let read_memory = cgroup_memory_reader.parse();
+                    if let Some(memory) = read_memory {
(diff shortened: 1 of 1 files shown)


lib/saluki-app/src/memory.rs
```diff
@@ -255,3 +270,68 @@ pub async fn initialize_allocator_telemetry() -> Result<(), GenericError> {
...
+    fn parse(self) -> Option<ByteSize> {
+        if let Ok(mut file) = fs::File::open("/proc/self/cgroup") {
+            let mut contents = String::new();
+            if file.read_to_string(&mut contents).is_ok() {
+                let parts: Vec<&str> = contents.trim().split("\n").collect();
+                // CgroupV2 has unified controllers.
+                if parts.len() == 1 {
+                    return self.parse_controller_v2(parts[0]);
+                }
+                for line in contents.lines() {
```
tobz, line 286: You already have `parts` where you've split the file into separate lines... I think you can just that here instead of `contents.lines()`.

lib/saluki-app/src/memory.rs
```diff
@@ -255,3 +270,68 @@ pub async fn initialize_allocator_telemetry() -> Result<(), GenericError> {
...
+            let memory_path = format!("/sys/fs/cgroup/memory{}/memory.limit_in_bytes", path);
+            let memory_file_maybe = fs::File::open(memory_path).ok();
+            if let Some(mut memory_file) = memory_file_maybe {
+                let mut s = String::new();
+                if memory_file.read_to_string(&mut s).is_ok() {
+                    return self.convert_to_bytesize(&s);
+                }
+            }
+        }
+        None
```
tobz, line 310: In Rust, you can also use the `?` operator for `Option<T>` return values in addition to `Result<T, E>`, which would let you get rid of almost all of the nesting here.

Taking advantage of that, plus some other simplifications, we could end up with something like this:

```rust
fn parse_controller_v1(self, controller: &str) -> Option<ByteSize> {
    let path = controller.split(":").nth(2)?;

    let memory_path = format!("/sys/fs/cgroup/memory{}/memory.limit_in_bytes", path);
    let raw_memory_limit = fs::read_to_string(memory_path).ok()?;
    self.convert_to_bytesize(&raw_memory_limit)
}
```

tobz (changes requested):

lib/saluki-app/src/memory.rs
```diff
@@ -68,10 +69,22 @@ impl MemoryBoundsConfiguration {
...
             .as_typed::<Self>()
             .error_context("Failed to parse memory bounds configuration.")?;
 
+        if config.memory_limit.is_none() {
+            // Try to pull configured memory limit from Cgroup if running in a containerized environment.
+            if let Ok(value) = env::var("DOCKER_DD_AGENT") {
+                if !value.is_empty() {
+                    let cgroup_memory_reader = CgroupMemoryParser;
+                    if let Some(memory) = cgroup_memory_reader.parse() {
+                        config.memory_limit = Some(memory);
```
tobz, line 82: I would probably add an info level log line here saying that we configured the memory limit based on the detected cgroup limit:

```rust
info!("Setting memory limit to {} based on detected cgroups limit.", memory.to_string_as(true));
```

tobz (approved):

PR #345, opened 2024-11-21
chore: coalesce tail entries with spare capacity during reclamation

tobz:
## Context

In `GenericMapInterner` (and `FixedSizeInterner`, as well), we support the reclamation of interned strings which are no longer referenced. This is achieved by storing markers -- `ReclaimedEntry` -- that denote the holes in the backing buffer where strings can be stored, before needing to fallback to simply writing into the "spare capacity" of the backing buffer.

Checking for reclaimed entries that can fit a string is, naturally, a little slower than just doing the equivalent of bumping a pointer. Storing reclaimed entries also requires allocations for the container that holds them. Overall, it's useful to avoid storing reclaimed entries when possible. One such case is when we're about to reclaim a "tail" entry that is adjacent to the spare capacity in the buffer.

Imagine a backing buffer split into three equal parts, where parts one and two are in use, and part three is free (spare capacity). When part two is reclaimed, we could add a reclaimed entry for it... or we could simply coalesce it with part three and adjust our state so parts two and three are merged together: you could imagine this as simply adjusting the "offset" field used to indicate the point in the buffer where spare capacity starts.

This reduces the size of the reclaimed entries list, which helps avoid it growing increasingly large over time, and it also means we can merge reclaimed entries with the spare capacity: we do this for reclaimed entries themselves, but we don't merge reclaimed entries with spare capacity when possible, which means we're not maximizing how much contiguous available capacity we have.

## Solution

This PR implements support for coalescing tail entries with adjacent spare capacity. The code to do this is simple, since we only need to check if a merged reclaimed entry is adjacent to the spare capacity, and if so, remove the reclaimed entry and adjust `self.len`.

Most of the code changed was related to the tests themselves, as they were written with the invariant of all dropped strings generating a reclaimed entry that would be available the next time an intern operation occurred.

## Notes

It's very unclear if this will have _any_ meaningful performance benefit or show up in benchmarks. I'm doing it because it's 1) possible and 2) it's the optimal thing to do.


rayz (approved):
lgtm

rayz (commented):
Oh, looks like we should remove the `TODO` comment in the top of the file of `fixed_size.rs` too?

PR #349, opened 2024-11-25
chore: add ability to convert `BytesBuffer` into shareable 'frozen' buffer

tobz:
## Context

This PR introduces the ability to convert an owned `BytesBuffer` into a shareable `FrozenBytesBuffer`, while still retaining the semantics of returning the underlying buffer (`FixedSizeVec`) to the object pool when all shared references are dropped. This will allow us to build a future abstraction where buffers can be written to by the Datadog Metrics destination, and then frozen to allow sending requests to multiple endpoints _concurrently_ where we want to cheaply reference the existing buffer(s) without cloning.

### Approach

I've taken a specific approach here which is to use the `triomphe` crate, which provides an enhanced version of `std::sync::Arc`. Specifically, `triomphe::UniqueArc` allows creating an `Arc` that can be used to mutate the inner value without any atomic loads since it encodes the invariants of having exclusive access to the inner value. After that, it can be converted to `triomphe::Arc` which allows sharing, by simply incrementing the relevant strong count field.

This is important because we need to allocate our `Vec<u8>` in `FixedSizeVec` within the `Arc` (`UniqueArc`, in this case) otherwise we would constantly be doing two allocations -- one to create `FrozenFixedSizeVec`, and one when converting it back to `FixedSizeVec` -- since `Arc` is added as a header to the value it holds. With `UniqueArc`, and its ability to go to `Arc` and then back to `UniqueArc`, we only pay this allocation cost once, and then losslessly convert back and forth between the two representations as the buffer is (re)used over time.

### Forked version of `triomphe`

Unfortunately, `triomphe` does not have support out of the box to convert its `Arc` into a `UniqueArc`, so we've forked the crate (https://github.com/tobz/triomphe) and [submitted a PR](https://github.com/Manishearth/triomphe/pull/103) adding the functionality. Given that the functionality very closely mirrors the functionality of `std::sync::Arc::into_inner`, I don't expect much pushback on adding it, but the crate is mostly in maintenance mode, so it might just take some time to get it reviewed and release.

### Changes to `ChunkedBuffer`

I also included some changes to `ChunkedBuffer` to take advantage of this new "frozen" buffer support, which overall were aimed to make it easier to do the conversion. I've tried to eschew some of the code in different areas where we were propagating a ton of information in the where bounds for anything interacting with the buffer object pool and `ChunkedBuffer` -- most of this is in the Datadog Metrics destination -- which I believe is a reasonable usability improvement but admittedly is tangential to the core changes of the PR.

### DIFF
--- a/lib/saluki-core/src/pooling/helpers.rs
+++ b/lib/saluki-core/src/pooling/helpers.rs
@@ -198,19 +198,19 @@ macro_rules! pooled_newtype {
             #[doc = "Poolable version of `" $inner_ty "`."]
             pub struct $name {
                 strategy_ref: ::std::sync::Arc<dyn $crate::pooling::ReclaimStrategy<$name> + Send + Sync>,
-                data: ::std::mem::ManuallyDrop<$inner_ty>,
+                data: ::std::option::Option<$inner_ty>,
             }
         }
 
         impl $name {
             /// Gets a reference to the inner data.
             pub fn data(&self) -> &$inner_ty {
-                &self.data
+                self.data.as_ref().unwrap()
             }
 
             /// Gets a mutable reference to the inner data.
             pub fn data_mut(&mut self) -> &mut $inner_ty {
-                &mut self.data
+                self.data.as_mut().unwrap()
             }
         }
 
@@ -223,7 +223,7 @@ macro_rules! pooled_newtype {
             ) -> Self {
                 Self {
                     strategy_ref,
-                    data: ::std::mem::ManuallyDrop::new(data),
+                    data: ::std::option::Option::Some(data),
                 }
             }
         }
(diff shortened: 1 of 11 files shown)


blt (approved):

lib/saluki-core/src/pooling/helpers.rs
```diff
@@ -223,16 +223,18 @@ macro_rules! pooled_newtype {
...
 
         impl Drop for $name {
             fn drop(&mut self) {
                 // SAFETY: We never use `self.data` again since we're already dropping `self`.
-                let data = unsafe { ::std::mem::ManuallyDrop::take(&mut self.data) };
-                self.strategy_ref.reclaim(data);
+
+                if let ::std::option::Option::Some(data) = self.data.take() {
+                    self.strategy_ref.reclaim(data);
+                }
```
tobz, line 237: For posterity: we had to do this because otherwise, trying to consume `self.data` (like we now do in `BytesBuffer::freeze`) becomes impossible as the compiler recognizes that we're emptying out `self.data` at some point, and then the drop logic here runs, and also tries to take the value... which is clearly not valid.

This is just a simpler, and safe, way to support `freeze` semantics with the lowest overhead we can reasonably hope for.

PR #351, opened 2024-11-25
[APR-278] Add config refresher to update API key

rayz:
WIP

### DIFF
--- a/bin/agent-data-plane/src/main.rs
+++ b/bin/agent-data-plane/src/main.rs
@@ -7,6 +7,7 @@
 #![deny(missing_docs)]
 use std::{
     future::pending,
+    sync::Arc,
     time::{Duration, Instant},
 };
 
@@ -19,7 +20,7 @@ use saluki_components::{
         AggregateConfiguration, ChainedConfiguration, HostEnrichmentConfiguration, OriginEnrichmentConfiguration,
     },
 };
-use saluki_config::{ConfigurationLoader, GenericConfiguration};
+use saluki_config::{ConfigRefresher, ConfigRefresherConfiguration, ConfigurationLoader, GenericConfiguration};
 use saluki_core::topology::TopologyBlueprint;
 use saluki_error::{ErrorContext as _, GenericError};
 use saluki_health::HealthRegistry;
@@ -174,8 +175,14 @@ fn create_topology(
         .with_transform_builder(host_enrichment_config)
         .with_transform_builder(origin_enrichment_config);
     let internal_metrics_remap_config = AgentTelemetryRemapperConfiguration::new();
-    let dd_metrics_config = DatadogMetricsConfiguration::from_configuration(configuration)
+
+    let config_refresher_configuration = ConfigRefresherConfiguration::from_configuration(&configuration)?;
+    let config_refresher: Arc<ConfigRefresher> = Arc::new(config_refresher_configuration.build()?);
+    config_refresher.clone().spawn_refresh_task();
+
+    let mut dd_metrics_config = DatadogMetricsConfiguration::from_configuration(configuration)
         .error_context("Failed to configure Datadog Metrics destination.")?;
+    dd_metrics_config.set_config_refresher(config_refresher);
     let events_service_checks_config = DatadogEventsServiceChecksConfiguration::from_configuration(configuration)
         .error_context("Failed to configure Datadog Events/Service Checks destination.")?;
 
--- a/lib/saluki-components/src/destinations/datadog_metrics/mod.rs
+++ b/lib/saluki-components/src/destinations/datadog_metrics/mod.rs
@@ -1,4 +1,4 @@
-use std::time::Duration;
+use std::{sync::Arc, time::Duration};
 
 use async_trait::async_trait;
 use http::{HeaderValue, Method, Request, Uri};
@@ -6,7 +6,7 @@ use http_body_util::BodyExt;
 use hyper::body::Incoming;
 use memory_accounting::{MemoryBounds, MemoryBoundsBuilder};
 use metrics::Counter;
-use saluki_config::GenericConfiguration;
+use saluki_config::{ConfigRefresher, GenericConfiguration};
 use saluki_core::{
     components::{destinations::*, ComponentContext, MetricsBuilder},
     pooling::{FixedSizeObjectPool, ObjectPool},
@@ -188,6 +188,9 @@ pub struct DatadogMetricsConfiguration {
     #[serde_as(as = "PickFirst<(DisplayFromStr, _)>")]
     #[serde(default, rename = "additional_endpoints")]
     endpoints: AdditionalEndpoints,
+
+    #[serde(skip)]
+    config_refresher: Arc<ConfigRefresher>,
 }
 
 fn default_request_timeout_secs() -> u64 {
@@ -223,6 +226,11 @@ impl DatadogMetricsConfiguration {
     fn api_base(&self) -> Result<Uri, GenericError> {
         calculate_api_endpoint(self.dd_url.as_deref(), &self.site)
     }
+
+    /// Add option to refresh config with a remote source
+    pub fn set_config_refresher(&mut self, refresher: Arc<ConfigRefresher>) {
+        self.config_refresher = refresher;
+    }
 }
 
 #[async_trait]
@@ -253,6 +261,7 @@ impl DestinationBuilder for DatadogMetricsConfiguration {
             api_base.clone(),
             MetricsEndpoint::Series,
             rb_buffer_pool.clone(),
+            self.config_refresher.clone(),
         )
         .await?;
         let sketches_request_builder = RequestBuilder::new(
@@ -260,6 +269,7 @@ impl DestinationBuilder for DatadogMetricsConfiguration {
             api_base,
             MetricsEndpoint::Sketches,
             rb_buffer_pool,
+            self.config_refresher.clone(),
         )
         .await?;
 
--- a/lib/saluki-components/src/destinations/datadog_metrics/request_builder.rs
+++ b/lib/saluki-components/src/destinations/datadog_metrics/request_builder.rs
@@ -1,9 +1,11 @@
+use std::sync::Arc;
 use std::{io, num::NonZeroU64};
 
 use datadog_protos::metrics::{self as proto, Resource};
 use ddsketch_agent::DDSketch;
 use http::{Method, Request, Uri};
 use protobuf::CodedOutputStream;
+use saluki_config::ConfigRefresher;
 use saluki_core::pooling::ObjectPool;
 use saluki_event::metric::*;
 use saluki_io::net::client::replay::ReplayBody;
@@ -110,6 +112,7 @@ impl MetricsEndpoint {
     }
 }
 
+#[allow(unused)]
 pub struct RequestBuilder<O>
 where
     O: ObjectPool + 'static,
@@ -125,6 +128,7 @@ where
     uncompressed_len: usize,
     metrics_written: usize,
     scratch_buf_lens: Vec<usize>,
+    refresher: Arc<ConfigRefresher>,
 }
 
 impl<O> RequestBuilder<O>
@@ -134,12 +138,11 @@ where
 {
     /// Creates a new `RequestBuilder` for the given endpoint, using the specified API key and base URI.
     pub async fn new(
-        api_key: String, api_base_uri: Uri, endpoint: MetricsEndpoint, buffer_pool: O,
+        api_key: String, api_base_uri: Uri, endpoint: MetricsEndpoint, buffer_pool: O, refresher: Arc<ConfigRefresher>,
     ) -> Result<Self, RequestBuilderError> {
         let chunked_buffer_pool = ChunkedBufferObjectPool::new(buffer_pool);
         let compressor = create_compressor(&chunked_buffer_pool).await;
         let api_uri = build_uri_for_endpoint(api_base_uri, endpoint)?;
-
         Ok(Self {
             api_key,
             api_uri,
@@ -151,6 +154,7 @@ where
             uncompressed_len: 0,
             metrics_written: 0,
             scratch_buf_lens: Vec::new(),
+            refresher,
         })
     }
 
@@ -354,7 +358,7 @@ where
                     .uri(self.api_uri.clone())
                     .header("Content-Type", "application/x-protobuf")
                     .header("Content-Encoding", "deflate")
-                    .header("DD-API-KEY", self.api_key.clone())
+                    .header("DD-API-KEY", self.refresher.api_key())
                     // TODO: We can't access the version number of the package being built that _includes_ this library, so
                     // using CARGO_PKG_VERSION or something like that would always be the version of `saluki-components`, which
                     // isn't what we want... maybe we can figure out some way to shove it in a global somewhere or something?
--- a/lib/saluki-config/Cargo.toml
+++ b/lib/saluki-config/Cargo.toml
@@ -7,6 +7,9 @@ repository = { workspace = true }
 
 [dependencies]
 figment = { workspace = true, features = ["env", "json", "yaml"] }
+http = { workspace = true }
+reqwest = { workspace = true , features = ["default-tls", "json"]}
+saluki-io = { workspace = true }
 saluki-error = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
--- a/lib/saluki-config/src/refresher.rs
+++ b/lib/saluki-config/src/refresher.rs
@@ -1,0 +1,89 @@
+use std::sync::Arc;
+
+use crate::GenericConfiguration;
+use reqwest;
+use saluki_error::GenericError;
+use serde::Deserialize;
+use tokio::time::{sleep, Duration};
+
+#[allow(dead_code)]
+#[derive(Debug, Deserialize)]
+struct ConfigResponse {
+    api_key: String,
+}
+
+/// Configuration for setting up `ConfigRefresher`.
+#[derive(Default, Deserialize)]
+pub struct ConfigRefresherConfiguration {
+    api_key: String,
+    auth_token_file_path: String,
+}
+
+/// The most recent configuration retrieved from the datadog-agent.
+#[derive(Default)]
+pub struct ConfigRefresher {
+    token: String,
+    api_key: String,
+}
+
+#[allow(unused)]
+impl ConfigRefresherConfiguration {
+    /// Creates a new `ConfigRefresherConfiguration` from the given configuration.
+    pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
+        Ok(config.as_typed()?)
+    }
+    /// Create `ConfigRefresher` from `ConfigRefresherConfiguration`
+    pub fn build(&self) -> Result<ConfigRefresher, GenericError> {
+        let raw_bearer_token = std::fs::read_to_string(&self.auth_token_file_path)?;
+        Ok(ConfigRefresher {
+            token: raw_bearer_token,
+            api_key: self.api_key.clone(),
+        })
+    }
+}
+impl ConfigRefresher {
+    /// Start a task that queries the datadog-agent config endpoint every 15 seconds.
+    pub fn spawn_refresh_task(self: Arc<Self>) {
+        tokio::spawn(async move {
+            loop {
+                match self.query_agent().await {
+                    Ok(_) => {
+                        // todo()!
+                    }
+                    Err(_e) => {
+                        // todo()!
+                    }
+                }
+                sleep(Duration::from_secs(15)).await; // Wait for 15 seconds
+            }
+        });
+    }
+    /// Query the datadog-agent config endpoint for the latest config
+    #[allow(unused)]
+    pub async fn query_agent(&self) -> Result<(), GenericError> {
+        let url = format!("https://localhost:5004/config/v1/");
+        let client = reqwest::ClientBuilder::new()
+            .danger_accept_invalid_certs(true) // Allow invalid certificates
+            .build()?;
+
+        let response = client
+            .get(&url) // Use the URL for the GET request
+            .header("Content-Type", "application/json")
+            .header("Authorization", format!("Bearer {}", self.token))
+            .header("DD-Agent-Version", "0.1.0")
+            .header("User-Agent", "agent-data-plane/0.1.0")
+            .send() // Send the request
+            .await?;
+
+        let config_response: ConfigResponse = response.json().await?;
+
+        // self.api_key = config_response.api_key;
+
+        Ok(())
+    }
+
+    /// Most recent `api_key` retrieved from the datadog-agent.
+    pub fn api_key(&self) -> &str {
+        &self.api_key
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1082,6 +1082,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
+name = "foreign-types"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f6f339eb8adc052cd2ca78910fda869aefa38d22d5cb648e6485e4d3fc06f3b1"
+dependencies = [
+ "foreign-types-shared",
+]
+
+[[package]]
+name = "foreign-types-shared"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
+
+[[package]]
 name = "form_urlencoded"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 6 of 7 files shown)


lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,89 @@
...
+            .header("Content-Type", "application/json")
+            .header("Authorization", format!("Bearer {}", self.token))
+            .header("DD-Agent-Version", "0.1.0")
+            .header("User-Agent", "agent-data-plane/0.1.0")
+            .send() // Send the request
+            .await?;
+
+        let config_response: ConfigResponse = response.json().await?;
+
+        // self.api_key = config_response.api_key;
```
rayz, line 80: WIP, not actually updating the field yet, stuck on figuring out how to make `ConfigRefresher` mutable with the Arc

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,89 @@
...
+                }
+                sleep(Duration::from_secs(15)).await; // Wait for 15 seconds
+            }
+        });
+    }
+    /// Query the datadog-agent config endpoint for the latest config
+    #[allow(unused)]
+    pub async fn query_agent(&self) -> Result<(), GenericError> {
+        let url = format!("https://localhost:5004/config/v1/");
+        let client = reqwest::ClientBuilder::new()
```
rayz, line 65: Using `reqwest` here since I needed `.danger_accept_invalid_certs` for the request to work.

The curl command i was trying to imitate was 
` curl  -H "authorization: Bearer $(cat /etc/datadog-agent/auth/auth_token)" -k https://localhost:5004/config/v1/`

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,90 @@
...
+
+#[allow(dead_code)]
+#[derive(Debug, Deserialize)]
+struct ConfigResponse {
+    api_key: String,
+}
+
+/// Configuration for setting up `ConfigRefresher`.
+#[derive(Default, Deserialize)]
+pub struct ConfigRefresherConfiguration {
```
tobz, line 19: I would just call this `RefresherConfiguration`.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        // self.api_key = config_response.api_key;
+
+        Ok(())
+    }
+
+    /// Most recent `api_key` retrieved from the datadog-agent.
+    pub fn api_key(&self) -> &str {
+        &self.api_key
+    }
```
tobz, line 89: Like we talked about DM, I'd say this should be removed and we simply expose a generic lookup method, similar to what `GenericConfiguration` exposes.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,90 @@
...
+pub struct ConfigRefresherConfiguration {
+    api_key: String,
+    auth_token_file_path: String,
+}
+
+/// The most recent configuration retrieved from the datadog-agent.
+#[derive(Default)]
+pub struct ConfigRefresher {
+    token: String,
+    api_key: String,
```
tobz, line 28: Instead of storing the API key directly, probably just hold something like `serde_json::Value` so that we can query it directly, as well as deserialize the config API response to it.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,90 @@
...
+/// Configuration for setting up `ConfigRefresher`.
+#[derive(Default, Deserialize)]
+pub struct ConfigRefresherConfiguration {
+    api_key: String,
+    auth_token_file_path: String,
+}
+
+/// The most recent configuration retrieved from the datadog-agent.
+#[derive(Default)]
+pub struct ConfigRefresher {
```
tobz, line 26: Probably just call this `RefreshableConfiguration`, since it's more about being the thing code uses to query these refreshed values than just the thing that does the refreshing, if that makes sense.

lib/saluki-config/src/refresher.rs
tobz: You'll likely want to use [`arc-swap`](https://docs.rs/arc-swap/latest/arc_swap/) here, which is designed for this sort of use case.

bin/agent-data-plane/Cargo.toml
```diff
@@ -10,6 +10,7 @@ default = []
 fips = ["saluki-app/tls-fips"]
 
 [dependencies]
+arc-swap = { workspace = true }
```
tobz, line 13: Shouldn't need to add this here since you're not actually using it in the binary crate.

lib/saluki-components/Cargo.toml
```diff
@@ -7,6 +7,7 @@ repository = { workspace = true }
 
 [dependencies]
 ahash = { workspace = true }
+arc-swap = { workspace = true }
```
tobz, line 10: Shouldn't need to add this here since you're not actually using it in the crate.

bin/agent-data-plane/src/main.rs
```diff
@@ -174,8 +175,14 @@ fn create_topology(
         .with_transform_builder(host_enrichment_config)
         .with_transform_builder(origin_enrichment_config);
     let internal_metrics_remap_config = AgentTelemetryRemapperConfiguration::new();
-    let dd_metrics_config = DatadogMetricsConfiguration::from_configuration(configuration)
+
+    let config_refresher_configuration = RefresherConfiguration::from_configuration(&configuration)?;
+    let config_refresher: Arc<RefreshableConfiguration> = Arc::new(config_refresher_configuration.build()?);
+    config_refresher.clone().spawn_refresh_task();
```
tobz: I think we could make this less clunky by having `RefresherConfiguration::build` do the spawning for us, taking care of cloning `RefreshableConfiguration`.

This will necessitate needing to change `values` in `RefreshableConfiguration` from `ArcSwap<Value>` to `Arc<ArcSwap<Value>>` (yes, I realize the irony of having to wrap `ArcSwap` in an `Arc` to make it shareable 😄) and adding a derive for `Clone`, but it'll mean we just have callers deal with `RefreshableConfiguration`.

lib/saluki-components/src/destinations/datadog_metrics/mod.rs
```diff
@@ -223,6 +226,11 @@ impl DatadogMetricsConfiguration {
     fn api_base(&self) -> Result<Uri, GenericError> {
         calculate_api_endpoint(self.dd_url.as_deref(), &self.site)
     }
+
+    /// Add option to retrieve config values from a `RefreshableConfiguration`
```
tobz, line 230: Spelling/grammar: `config` -> `configuration`, and add a period.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,107 @@
...
+
+const DATADOG_AGENT_CONFIG_ENDPOINT: &str = "https://localhost:5004/config/v1/";
+
+/// Configuration for setting up `RefreshableConfiguration`.
+#[derive(Default, Deserialize)]
+pub struct RefresherConfiguration {
+    auth_token_file_path: String,
+}
+
+/// The most recent configuration retrieved from the datadog-agent.
```
tobz, line 19: I realize this is indeed very specific to how we're going to depend on the Datadog Agent, but let's go with a more generic description, like this:

```suggestion
/// A configuration whose values are refreshed from a remote source at runtime.
```

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,107 @@
...
+        let raw_bearer_token = std::fs::read_to_string(&self.auth_token_file_path)?;
+        Ok(RefreshableConfiguration {
+            token: raw_bearer_token,
+            values: ArcSwap::from_pointee(serde_json::Value::Null),
+        })
+    }
+}
+impl RefreshableConfiguration {
+    /// Start a task that queries the datadog-agent config endpoint every 15 seconds.
+    pub fn spawn_refresh_task(self: Arc<Self>) {
```
tobz, line 42: I would hide this from `pub` visibility as part of refactoring to make `RefresherConfiguration::build` handle the spawning itself.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,107 @@
...
+            loop {
+                match self.query_agent().await {
+                    Ok(_) => {
+                        // todo()!
+                    }
+                    Err(_e) => {
+                        // todo()!
+                    }
+                }
+                sleep(Duration::from_secs(15)).await; // Wait for 15 seconds
```
tobz, line 53: I would add a field to `RefresherConfiguration` for the refresh interval in seconds (`refresh_interval_secs`), stick this 15 second default up at the top as a const, and then try and pull it out of the configuration in `RefresherConfiguration::from_configuration`, using the const as the default... and then pass it through to `RefreshableConfiguration` and reference it here.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,107 @@
...
+    token: String,
+    values: ArcSwap<Value>,
+}
+
+impl RefresherConfiguration {
+    /// Creates a new `RefresherConfiguration` from the given configuration.
+    pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
+        Ok(config.as_typed()?)
+    }
+    /// Create `RefreshableConfiguration` from `RefresherConfiguration`
```
tobz, line 31: Spelling/grammar nit: trailing period.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub fn spawn_refresh_task(self: Arc<Self>) {
+        tokio::spawn(async move {
+            loop {
+                match self.query_agent().await {
+                    Ok(_) => {
+                        // todo()!
+                    }
+                    Err(_e) => {
+                        // todo()!
+                    }
```
tobz, line 51: I would just have a debug-level log in the `Ok` case, and an error-level log in the `Err` case.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,107 @@
...
+                }
+                sleep(Duration::from_secs(15)).await; // Wait for 15 seconds
+            }
+        });
+    }
+    /// Query the datadog-agent config endpoint for the latest config
+    pub async fn query_agent(&self) -> Result<(), GenericError> {
```
tobz, line 58: I would actually move this all into the refresh task, so that you're constructing the client outside of the loop, and then doing the request inside the loop.

tobz (changes requested):
Left a bunch of comments.

bin/agent-data-plane/src/main.rs
rayz: > This will necessitate needing to change values in RefreshableConfiguration from ArcSwap<Value> to Arc<ArcSwap<Value>>

Can you explain more on why it needs to be wrapped in an `Arc` if the overall struct is in an `Arc`?

lib/saluki-components/src/destinations/datadog_metrics/mod.rs
```diff
@@ -552,14 +569,21 @@ async fn run_endpoint_io_loop<S, B>(
     task_barrier.wait().await;
 }
 
-fn for_resolved_endpoint<B>(endpoint: ResolvedEndpoint) -> impl Fn(Request<B>) -> Request<B> + Clone {
+#[allow(unused)]
```
tobz, line 572: Doesn't seem like this should be here... at least not anymore?

lib/saluki-components/src/destinations/datadog_metrics/request_builder.rs
```diff
@@ -105,6 +105,7 @@ impl MetricsEndpoint {
     }
 }
 
+#[allow(unused)]
```
tobz, line 108: Doesn't seem like this should be here... at least not anymore?

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,131 @@
...
+use crate::{ConfigurationError, GenericConfiguration};
+
+const DATADOG_AGENT_CONFIG_ENDPOINT: &str = "https://localhost:5004/config/v1/";
+const DEFAULT_AUTH_TOKEN_FILE_PATH: &str = "/etc/datadog-agent/auth_token";
+const DEFAULT_REFRESH_INTERVAL_SECONDS: u64 = 15;
+
+/// Configuration for setting up `RefreshableConfiguration`.
+#[derive(Default, Deserialize)]
+pub struct RefresherConfiguration {
+    /// The location of the auth token used by the datadog agent.
```
tobz, line 20:
```suggestion
    /// The location of the auth token used by the Datadog Agent.
```

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,131 @@
...
+
+impl RefresherConfiguration {
+    /// Creates a new `RefresherConfiguration` from the given configuration.
+    pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
+        Ok(config.as_typed()?)
+    }
+    /// Create `RefreshableConfiguration` from `RefresherConfiguration`.
+    pub fn build(&self) -> Result<Arc<RefreshableConfiguration>, GenericError> {
+        let raw_bearer_token = std::fs::read_to_string(&self.auth_token_file_path)?;
+        let refreshable_configuration = Arc::new(RefreshableConfiguration {
```
tobz, line 57: We don't need to wrap `RefreshableConfiguration` in `Arc`, since the internal values state is already wrapped in `Arc`.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,131 @@
...
+                    .send()
+                    .await;
+                match response {
+                    Ok(response) => {
+                        let config_response: Value = response
+                            .json()
+                            .await
+                            .expect("failed to deserialize configuration into json");
+                        self.values.store(Arc::new(config_response));
+                        debug!("Retrieved configuration from datadog-agent.");
```
tobz, line 90: I would change this just to be slightly more generic:

```suggestion
                        debug!(remote_endpoint = DATADOG_AGENT_CONFIG_ENDPOINT, "Retrieved configuration from remote source.");
```

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,131 @@
...
+                    }
+                    Err(e) => {
+                        error!("Error retrieving configuration from datadog-agent: {}", e);
```
tobz, line 93: I would change this just to be slightly more generic:

```suggestion
                        error!(remote_endpoint = DATADOG_AGENT_CONFIG_ENDPOINT, "Failed to retrieve configuration from remote source: {}", e);
```

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,131 @@
...
+                serde_json::from_value(value.clone()).map_err(|_| ConfigurationError::MissingField {
+                    help_text: format!(
+                        "RefreshableConfiguration could not convert key {} value into the proper type.",
+                        key
+                    ),
+                    field: key.to_string().into(),
+                })
+            }
+            None => Err(ConfigurationError::MissingField {
+                help_text: format!("RefreshableConfiguration missing key {}", key),
```
tobz, line 126: The error message already conveys a human-readable string about it missing, so the help text is really just if there's anything the caller should attempt to do to fix the issue, or some hint on how to potentially solve it.

There's really nothing to _do_ in this case, so maybe just something like:

> Try validating remote source provides this field.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// variant will be returned.
+    pub fn get_typed<T>(&self, key: &str) -> Result<T, ConfigurationError>
+    where
+        T: DeserializeOwned,
+    {
+        let values = self.values.load();
+        match values.get(key) {
+            Some(value) => {
+                // Attempt to deserialize the value to type T
+                serde_json::from_value(value.clone()).map_err(|_| ConfigurationError::MissingField {
```
tobz, line 117: The field isn't really missing, it's just not the type we expected. We should be using `ConfigurationError::InvalidFieldType`.

That error variant does want to show the _actual_ type of the field, though, and there's the issue of getting the name of `T`, so you could do something like this:

```rust
ConfigurationError::InvalidFieldType {
  field: key.to_string(),
  expected_ty: std::any::type_name::<T>().to_string(),
  actual_ty: serde_json_value_type_name(value),
}
```

That requires a helper because `serde_json::Value` doesn't have an inherent method to get a stringified name of the variant, so you could add a free function further down like:

```rust
fn serde_json_value_type_name(value: &Value) -> &'static str {
    match value {
        Value::Null => "null",
        Value::Bool(_) => "bool",
        Value::Number(_) => "number",
        Value::String(_) => "string",
        Value::Array(_) => "array",
        Value::Object(_) => "object",
    }
}
```

It won't be perfect because it's not going to help differentiate between arrays of different types, or objects with the wrong structure, and so on... but it should be a good first pass.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,131 @@
...
+    token: String,
+    values: Arc<ArcSwap<Value>>,
+    refresh_interval_seconds: u64,
+}
+
+impl RefresherConfiguration {
+    /// Creates a new `RefresherConfiguration` from the given configuration.
+    pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
+        Ok(config.as_typed()?)
+    }
```
tobz, line 53: Space after this line.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,131 @@
...
+        let raw_bearer_token = std::fs::read_to_string(&self.auth_token_file_path)?;
+        let refreshable_configuration = Arc::new(RefreshableConfiguration {
+            token: raw_bearer_token,
+            values: Arc::new(ArcSwap::from_pointee(serde_json::Value::Null)),
+            refresh_interval_seconds: self.refresh_interval_seconds,
+        });
+        refreshable_configuration.clone().spawn_refresh_task();
+        Ok(refreshable_configuration)
+    }
+}
```
tobz, line 65: Space after this line.

lib/saluki-components/src/destinations/datadog_metrics/mod.rs
```diff
@@ -552,14 +569,21 @@ async fn run_endpoint_io_loop<S, B>(
...
         Scheme::try_from(endpoint.endpoint().scheme()).expect("should not fail to construct new endpoint scheme");
-    let api_key_value =
-        HeaderValue::from_str(endpoint.api_key()).expect("should not fail to construct API key header value");
-
+    let mut api_key: String;
+    if let Ok(refresher_api_key) = refresher.get_typed::<String>("api_key") {
+        api_key = refresher_api_key.clone();
+    } else {
+        api_key = endpoint.api_key().to_string();
+    }
```
tobz, line 585: This is going to apply the API key from the refreshable configuration to _all_ resolved endpoints, which is not what we want to do.

We might need to consider a design where we shove `RefreshableConfiguration` into `ResolvedEndpoint` and move some stuff around, so that we can amortize the cost using the refresher and only apply it where desired.

tobz (changes requested):

lib/saluki-config/Cargo.toml
```diff
@@ -6,8 +6,12 @@ license = { workspace = true }
 repository = { workspace = true }
 
 [dependencies]
+arc-swap = { workspace = true }
 figment = { workspace = true, features = ["env", "json", "yaml"] }
+http = { workspace = true }
+reqwest = { workspace = true, features = ["rustls-tls", "json"] }
```
tobz, line 12: Can you change `rustls-tls` to `rustls-tls-native-roots-no-provider`?

This will avoid pulling in some extra dependencies that we don't need.

bin/agent-data-plane/src/main.rs
```diff
@@ -176,8 +176,13 @@ fn create_topology(
...
         .with_transform_builder(origin_enrichment_config);
     let internal_metrics_remap_config = AgentTelemetryRemapperConfiguration::new();
-    let dd_metrics_config = DatadogMetricsConfiguration::from_configuration(configuration)
+
+    let refresher_configuration = RefresherConfiguration::from_configuration(configuration)?;
+    let refreshable_configuration: RefreshableConfiguration = refresher_configuration.build()?;
+
+    let mut dd_metrics_config = DatadogMetricsConfiguration::from_configuration(configuration)
         .error_context("Failed to configure Datadog Metrics destination.")?;
+    dd_metrics_config.add_refreshable_configuration(refreshable_configuration.clone());
```
tobz, line 185: Don't need to clone this, as far as I can see?

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -113,16 +116,42 @@ impl ResolvedEndpoint {
...
     }
 
     /// Returns the API key associated with the endpoint.
-    pub fn api_key(&self) -> &str {
+    pub fn api_key(&mut self) -> &str {
+        if let Some(config) = &self.config {
+            match config.try_get_typed::<String>("api_key") {
+                Ok(Some(api_key)) => {
+                    if !api_key.is_empty() && self.api_key != api_key {
+                        debug!("Refreshing api key.");
```
tobz, line 143: Just some capitalization stuff and adding the endpoint so we know who got the refreshed API key.

```suggestion
                        debug!(endpoint = %self.endpoint, "Refreshed API key.");
```

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -113,16 +116,42 @@ impl ResolvedEndpoint {
...
+                        self.api_key = api_key;
+                    }
+                }
+                Ok(None) => {
+                    debug!("Failed to retrieve api key from remote source. Falling back to last known api key.");
```
tobz, line 148:
```suggestion
                    debug!("Failed to retrieve API key from remote source (missing or wrong type). Continuing with last known valid API key.");
```

lib/saluki-components/src/destinations/datadog_metrics/endpoint/endpoints.rs
```diff
@@ -113,16 +116,42 @@ impl ResolvedEndpoint {
...
+            config,
+        }
+    }
+
     /// Returns the endpoint of the resolver.
     pub fn endpoint(&self) -> &Url {
         &self.endpoint
     }
 
     /// Returns the API key associated with the endpoint.
```
tobz, line 137:
```suggestion
    /// Returns the API key associated with the endpoint.
    ///
    /// If a refreshable configuration has been configured, the API key will be queried from the
    /// configuration and stored if it has been updated since the last time `api_key` was called.
```

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,197 @@
...
+    refresh_interval_seconds: u64,
+}
+
+impl RefresherConfiguration {
+    /// Creates a new `RefresherConfiguration` from the given configuration.
+    pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
+        Ok(config.as_typed()?)
+    }
+
+    /// Create `RefreshableConfiguration` from `RefresherConfiguration`.
```
tobz, line 76:
```suggestion
    /// Builds a `RefreshableConfiguration`, spawning a background task to periodically pull
    /// configuration data and update the configuration.
    ///
    /// # Errors
    ///
    /// If the authentication token be read from the configured authentication token file
    /// path, an error will be returned.
```

tobz (changes requested):

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,197 @@
...
+    ///
+    /// Defaults to `localhost`.
+    #[serde(default = "default_agent_ipc_host")]
+    agent_ipc_host: String,
+
+    /// The IPC port used by the Datadog Agent.
+    ///
+    /// Defaults to `0`.
+    #[serde(default = "default_agent_ipc_port")]
+    agent_ipc_port: u64,
```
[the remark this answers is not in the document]
rayz: Commenting to document slack discussion: using the endpoint with `cmd_host` and `cmd_port` only provides scrubbed values.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,206 @@
...
+fn default_refresh_interval_seconds() -> u64 {
+    DEFAULT_REFRESH_INTERVAL_SECONDS
+}
+
+fn default_agent_ipc_host() -> String {
+    DEFAULT_AGENT_IPC_HOST.to_owned()
+}
+
+fn default_agent_ipc_port() -> u64 {
+    0
```
tobz, line 61: Zero still doesn't make sense as a value for something we're connecting _to_, so we shouldn't specify a default port at all. This would let us bubble up an error when calling `RefresherConfiguration::from_configuration` if `agent_ipc_port` isn't set.

Probably have the code building the refreshable config in `bin/agent-data-plane/src/main.rs` catch the error and report an INFO-level log that dynamic configuration refreshing will be unavailable due to not being configured, etc etc.

lib/saluki-config/src/refresher.rs
```diff
@@ -0,0 +1,206 @@
...
+    ///
+    /// Defaults to `/etc/datadog-agent/auth_token`.
+    #[serde(default = "default_auth_token_file_path")]
+    auth_token_file_path: String,
+
+    /// The amount of time betweeen each request in seconds.
+    ///
+    /// Defaults to 15 seconds.
+    #[serde(
+        rename = "agent_config_refresh_internal_seconds",
```
tobz, line 30: Sorry for the flip-flopping here, but now that I see how we're aligning more with the `agent_ipc` stuff in the Agent, I think we would actually want this to be `agent_ipc_config_refresh_interval`, which would align with how it's used by the `comp/core/configsync` component.

tobz (changes requested):

PR #352, opened 2024-11-25
[APR-248] enhancement: send requests concurrently in Datadog Metrics destination

tobz:
## Context

When configuring additional endpoints in the Datadog Metrics destination, the destination will send to each endpoint sequentially, waiting for a successful response from an endpoint before sending the request to the next endpoint. This means that slow, or perpetually broken endpoints, cause performance issues with sending to other endpoints that are operating correctly.

## Solution

This PR introduces the ability for the Datadog Metrics destination to send requests to additional endpoints in a _concurrent_ way.

The main chunk of work is in the destination itself, where we've now subdivided the I/O task into individual endpoint I/O tasks, and the main I/O task is responsible for cloning the request and sending it to each endpoint I/O task, as well as some light shutdown coordination.

The endpoint I/O task handles the bulk of the work, in terms of sending the request. As part of doing this, we've moved a large chunk of the request building logic from `RequestBuilder` to the endpoint I/O task, by way of layering in `tower` middleware helpers. Since a large part of building the actual `Request` object was setting the relevant headers and endpoint URI... it makes more sense to not do that at all in `RequestBuilder`, and instead do that in the endpoint I/O task, to avoid duplication of effort and code.

As well, we've switched to using the new `FrozenChunkedBytesBuffer` type which replaces `ReplayBody` as `ReplayBody` is not suitable for concurrent usage, which was a miss on my part when I originally pushed @rayz to use it for adding additional endpoint support in the first place.

Some of the other work in the PR is just optimizing the header stuff as much as we reasonable can to avoid the cost of small allocations for header strings, and what not, that are actually static/fixed in nature.

## Testing

We've added a new SMP experiment -- `dsd_uds_1mb_3k_contexts_dualship` -- for both DSD and ADP, which both exercises the behavior and will help ensure that our work to minimize the cost of cloning the request doesn't regress over time, given the number of potential allocations involved with building and adjusting the request for a specific endpoint.

## Notes

We currently utilize a fixed-size channel between the main I/O and individual endpoint I/O tasks, with a fallible send, which means if the endpoint I/O task is behind, we eventually will fill the channel and start dropping new requests. This channel is currently sized to avoid drops for the experiments we have at the moment, but is not sized, in particular, to stay this way forever: the current size is just a stopgap.

In a follow-up PR, we'll better address how we want to handle slow endpoints, ranging from a disk-backed retry queue to temporarily disabling the target endpoint (intentionally dropping requests while unhealthy), following the current behavior of the Datadog Agent.


tobz:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/saluki/352?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#352** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/352?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/saluki/352?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#349** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/349?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

rayz (dismissed):

PR #353, opened 2024-11-26
experiment: try speeding up `BorrowedTag` splitting via `memchr`

tobz:
## Context

Was examining some profiler output related to DSD metrics parsing, and `BorrowedTag::name_and_value` showed up as taking a reasonably large chunk of time. I used a naive splitting approach (`str::split_once`) when writing `BorrowedTag`, and so it's perhaps not surprising that it's less optimal than it could be.

## Solution

This PR tries out a `memchr`-based solution where we simply search for the first instance of `:` in the raw tag string when creating `BorrowedTag`, and then reuse that in `name`, `value`, and `name_and_value` when deciding whether or not there's anything to split.

In local testing, this has a fairly positive effective on ingest throughput, but we'll see what SMP has to say.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3357,6 +3357,7 @@ dependencies = [
  "ahash",
  "crossbeam-queue",
  "hashbrown 0.15.1",
+ "memchr",
  "metrics",
  "metrics-util",
  "quick_cache",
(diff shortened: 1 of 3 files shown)


lukesteensen:
🎉

PR #367, opened 2024-12-04
chore: add remoteagent proto file

rayz:
files copied from [here](https://github.com/DataDog/datadog-agent/tree/main/pkg/proto/datadog)


lib/datadog-protos/build.rs
```diff
@@ -37,5 +37,12 @@ fn main() {
...
             &["proto"],
         )
+        .expect("failed to build gRPC service definitions for DCA");
+
+    // Handle code generation for gRPC service definitions.
+    tonic_build::configure()
+        .build_server(true)
+        .include_file("remoteagent.mod.rs")
+        .compile_protos(&["proto/datadog/remoteagent/remoteagent.proto"], &["proto"])
         .expect("failed to build gRPC service definitions for DCA")
```
tobz, line 47: We should update the comment here to change `DCA` to `RemoteAgent`.

lib/datadog-protos/build.rs
tobz, line 1: It looks like I missed doing this myself originally, but we should add some additional `cargo:rerun-if-changed` print statements: one for each specific file we're generating gRPC service definitions for, so `api.proto`, `workloadmeta.proto`, and `remoteagent.proto`.

(Just gotta match the relative path syntax.)

tobz (changes requested):

lib/datadog-protos/build.rs
tobz: Sorry, I gave bad instructions here.

What I was trying to say is don't _just_ put `api.proto`, but use the same path as the one we're using to build the protos (`proto/datadog/api/v1/api.proto`) themselves.

tobz (changes requested):

tobz (approved):

PR #370, opened 2024-12-06
chore: normalize metric names in Prometheus destination

tobz:
## Context

Currently, the Prometheus destination will emit its scrape endpoint payload with metric names that are drawn directly from whichever `Metric`s are fed to the destination. For internal telemetry in particular, but really any potential metric, the names make use of periods as a separate, such as `adp.group_allocated_objects_live` or `adp.runtime.worker.local_queue_depth`.

This is problematic because periods, and many characters, are not a valid character in Prometheus metric names.

## Solution

This PR introduces metric name normalization to all metrics coming into the Prometheus destination. We ensure metric names are fully valid with the Prometheus metric name requirements, and take the additional step of converting periods into double underscores, which is an unofficial way of currying separators in the normalized metric name such that it can be used downstream to potentially reconstitute the semantic structure (for instance, realizing that `adp.runtime.worker.local_queue_depth` and `adp__runtime__worker__local_queue_depth` both represent a nested path of `[adp, runtime, worker, local_queue_depth]`).

We don't perform an equivalent sanitization of metric labels, but that will be a follow-up PR, including tests, as we've verified this PR gets us past problems we encountered when trying to scrape ADP's internal telemetry.

### DIFF
--- a/lib/saluki-components/src/destinations/prometheus/mod.rs
+++ b/lib/saluki-components/src/destinations/prometheus/mod.rs
@@ -1,4 +1,4 @@
-use std::{convert::Infallible, fmt::Write as _, sync::Arc};
+use std::{convert::Infallible, fmt::Write as _, num::NonZeroUsize, sync::Arc};
 
 use async_trait::async_trait;
 use ddsketch_agent::DDSketch;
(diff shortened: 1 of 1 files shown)


rayz (approved):
lgtm

PR #377, opened 2024-12-09
Add ability to register ADP as a remote agent to the Datadog Agent

rayz:
# Context

We would like to be able to register ADP as a remote agent to the Core Agent so that status and flares can be requested.

# Solution

This pr creates a new destination component that is intended for both status and flare information. 

# Notes

This pr registers ADP as a remote agent but a following pr is needed for actually sending data over.

# Testing

Run ADP with an agent that has the remote agent registry enabled. 
Run the `status` command on the agent. You should see ADP listed in the `Remote Agents` section.


lib/saluki-components/src/destinations/datadog_status_flare/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+const DEFAULT_ID: &str = "agent-data-plane";
+const DEFAULT_DISPLAY_NAME: &str = "Agent Data Plane";
+const DEFAULT_API_ENDPOINT: &str = "localhost:13580";
+
+/// TODO
+/// wrapped client inside an `Option` to fix Deserialization problems
+#[derive(Deserialize)]
+pub struct DatadogStatusFlareConfiguration {
+    #[serde(default = "default_id", rename = "remote_agent_id")]
+    id: String,
```
tobz, line 30: We don't really want this to be configurable.

The remote agent ID should be derived from the data plane full name (normalized, though) and a randomly generated UUID. Something like:

```rust
let app_details = saluki_metadata::get_app_details();
let formatted_full_name = app_details
    .full_name()
    .replace(" ", "-")
    .replace("_", "-")
    .to_lowercase();

# You'll need to pull in the `uuid` crate for this.
let agent_id = format!("{}-{}", formatted_full_name, uuid::Uuid::now_v7());
```

We would make the field skipped for deserialization, and then generate this in `DatadogStatusFlareConfiguration::from_configuration` one time. The goal is that is should be generated with a consistent format but uniquely per each process... such that we know to expect the format to be `<name>-<uuid>` [^1] for ADP, but that the UUID changes when we restart ADP, so that we can always tie one UUID to one particular process instance.

[^1]: It _could_ be another format, but it should be consistent every time you restart ADP, basically.

lib/saluki-components/src/destinations/datadog_status_flare/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    #[serde(default = "default_display_name", rename = "remote_agent_display_name")]
+    display_name: String,
+
+    #[serde(default = "default_api_endpoint", rename = "remote_agent_api_endpoint")]
+    api_endpoint: String,
+
+    #[serde(default = "default_auth_token", rename = "remote_agent_auth_token")]
+    auth_token: String,
```
tobz, line 39: This should not be configurable.

It is meant to be generated randomly every time a remote agent initializes itself. We wouldn't even want to cache it, really, in the configuration struct... just generate it in `DatadogStatusFlare::run`.

lib/saluki-components/src/destinations/datadog_status_flare/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+
+/// TODO
+/// wrapped client inside an `Option` to fix Deserialization problems
+#[derive(Deserialize)]
+pub struct DatadogStatusFlareConfiguration {
+    #[serde(default = "default_id", rename = "remote_agent_id")]
+    id: String,
+
+    #[serde(default = "default_display_name", rename = "remote_agent_display_name")]
+    display_name: String,
```
tobz, line 33: Similar to the above, we should always derive the display name from `saluki_metadata::get_app_details()` using the "full name" in its unadulterated form.

lib/saluki-components/src/destinations/datadog_status_flare/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    #[serde(default = "default_api_endpoint", rename = "remote_agent_api_endpoint")]
+    api_endpoint: String,
```
tobz, line 36: We should make this be `api_port`, since we don't ever want to expose a port on anything other than `localhost`.

The default port value should also be 5102, which puts it sequentially with the other two ADP ports, 5100 (admin API) and 5101 (telemetry API).

tobz (changes requested):
Overall, I think the component-based approach is a reasonable one.

I started leaving some feedback to better shape the design even though I realize it's still very much a rough draft.

lib/saluki-components/src/destinations/datadog_status_flare/mod.rs
GustavoCaso: Question: 
We would use `localhost` in the case that the Agent and ADP runs in the same machine. Would there be instances in which ADP runs on another machine? Maybe containers? 

Mostly, asking to understand the possible deployment topologies 😄

lib/saluki-components/src/destinations/datadog/status_flare/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+                    }
+                }
+            }
+        });
+
+        /// TODO
+        loop {
+            tokio::time::sleep(Duration::from_secs(30)).await;
+        }
+        Ok(())
```
tobz, line 141: You need to restructure this main loop to work within the component model.

I would go and copy what `Blackhole` has, and go from there. You can basically drop all the events you get on the floor for now, and then you would want to add another match arm to the `select!` macro that watches an `Interval` (more below) to know when it's time to try and (re)register with the Core Agent.

For the interval, you'd use [`interval`](https://docs.rs/tokio/latest/tokio/time/fn.interval.html) from `tokio`. We would just have a default interval of 10 seconds, but then the magic comes later. You can arbitrarily reset an `Interval` to control when it fires next, so we'd do something like:

```rust
let mut register_agent = interval(Duration::from_secs(10));
register_agent.set_missed_tick_behavior(MissedTickBehavior::Delay);

loop {
    select! {
        // We need to respond to health checker requests.
        _ = health.live() => continue,

        // We also need to drain events being sent to us to prevent backpressure,
        // as welli as know when it's time to shutdown.
        result = context.events().next() => if result.is_none() {
            break
        },

        // Time to (re)register with the Core Agent.
        _ = register_agent.tick() => {
            match self.client.register_remote_agent_request(...).await {
                Ok(resp) => {
                    let new_refresh_interval = ...;
                    register_agent.reset_after(new_refresh_interval);
                },
                Err(...) => { .... },
            }
        }
    }
}
```

This ensures that we're always handling health check requests and events being sent to us, and when it's time to (re)register, we do so... and then when registration is successful, we update our registration interval to match. Crucially, the default behavior of `Interval` is that it will immediately tick, so this code will still issue an initial registration call (because `register_agent` will tick immediately) which will let us synchronize with the recommended registration interval.

bin/agent-data-plane/src/main.rs
```diff
@@ -158,6 +163,7 @@ async fn run(started: Instant) -> Result<(), GenericError> {
 
 fn create_topology(
     configuration: &GenericConfiguration, env_provider: ADPEnvironmentProvider, component_registry: &ComponentRegistry,
+    _status_configuration: DatadogStatusFlareConfiguration,
```
tobz, line 166: Drop the underscore on this.

bin/agent-data-plane/src/main.rs
```diff
@@ -200,10 +206,15 @@ fn create_topology(
...
         .add_destination("dd_events_sc_out", events_service_checks_config)?
+        .add_destination("dd_status_flare_out", _status_configuration)?
         .connect_component("dsd_agg", ["dsd_in.metrics"])?
         .connect_component("enrich", ["dsd_agg"])?
         .connect_component("dd_metrics_out", ["enrich"])?
-        .connect_component("dd_events_sc_out", ["dsd_in.events", "dsd_in.service_checks"])?;
+        .connect_component("dd_events_sc_out", ["dsd_in.events", "dsd_in.service_checks"])?
+        .connect_component(
+            "dd_status_flare_out",
+            ["enrich", "dsd_in.events", "dsd_in.service_checks"],
```
tobz, line 216: What we really want to do is actually feed this with internal metrics.

We should probably copy all the non-Prometheus bits out of the if block below, related to internal telemetry, add them here, and then change the if block to just create the Prometheus destination and wire it up to `internal_metrics_remap`.

tobz (changes requested):

lib/saluki-components/src/destinations/datadog_status_flare/mod.rs
tobz: For containerized environments, everything is currently on the same machine: standalone Docker, Kubernetes, etc.

If ADP ever became a mechanism for doing remote aggregation (pool of ADP workers running on a different node than the application send metrics/logs/traces/etc) _then_ this might matter... but I can't really think of a situation where ADP will ever really be decoupled from a local Agent.

tobz (commented):
Changes are looking good.

The thing I remembered, though, is that we need a way to disable this for when we aren't actually running with the Agent, such as in CI. I think we'd basically want an if block in `create_topology` or something... but that gets tricky because we depend on this component now for being ale to always add the internal metrics source. Hmm....

bin/agent-data-plane/src/main.rs
```diff
@@ -93,9 +96,11 @@ async fn run(started: Instant) -> Result<(), GenericError> {
     let env_provider =
         ADPEnvironmentProvider::from_configuration(&configuration, &component_registry, &health_registry).await?;
 
+    let status_configuration = DatadogStatusFlareConfiguration::from_configuration(&configuration).await?;
```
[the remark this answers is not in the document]
rayz: I have it in `run` because `create_topology` isn't async. And the reason `DatadogStatusFlareConfiguration::from_configuration` is async is because of `RemoteAgentClient::from_configuration`

bin/agent-data-plane/src/main.rs
```diff
@@ -95,17 +103,33 @@ async fn run(started: Instant) -> Result<(), GenericError> {
...
+    let config = ServerConfig::builder()
+        .with_no_client_auth()
+        .with_single_cert(cert_chain, rustls::pki_types::PrivateKeyDer::Pkcs8(keys.remove(0)))
+        .unwrap();
+
     let primary_api = APIBuilder::new()
         .with_handler(health_registry.api_handler())
-        .with_handler(component_registry.api_handler());
+        .with_handler(component_registry.api_handler())
+        .with_tls_config(config);
```
tobz, line 132: For now, I would actually add a new method on `APIBuilder` called `with_self_signed_tls` that basically does all of the above code of building the self-signed certificate and the `ServerConfig`, and then have that new method call `APIBuilder::with_tls_config` so we keep it open for future changes where we may want to pass a full `ServerConfig` in.

lib/saluki-components/src/destinations/datadog/status_flare/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+                result = context.events().next() => match result {
+                    Some(_events) => {
+                    },
+                    None => break,
+                },
+
+                // Time to (re)register with the Core Agent.
+                //
+                // TODO: Consider spawning the registration as a task so that the component can keep polling and not slow down the accepting of events and responding of health checks.
+                _ = register_agent.tick() => {
```
rayz, line 142: This component is only built if `adp_standalone_mode == false`, but this call will keep failing if 

```
 remote_agent_registry:
   enabled: true
```
is not set on the `datadog-agent` side.

tobz: 🤷🏻 

I think that's just something we'll have to deal with until we make it the default to enable the remote agent registry. We don't get any visible errors, right, since it's just debug logging if the call fails? Or does it have any other user-visible impact in ADP when the remote agent registry is disabled?

lib/saluki-components/src/destinations/datadog/status_flare/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+
+const DEFAULT_API_LISTEN_PORT: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(5100) };
+
+/// Datadog Status and Flare Destination
+///
+/// Registers ADP as a remote agent to the Core Agent.
+///
+/// ## Missing
+///
+/// - grpc server to respond to Core Agent
```
tobz, line 36: I would say this isn't missing anymore. :)

lib/saluki-components/src/destinations/datadog/status_flare/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+#[derive(Default)]
+struct RemoteAgentImpl {
+    started: DateTime<Utc>,
+}
+
+#[async_trait]
+impl RemoteAgent for RemoteAgentImpl {
+    async fn get_status_details(
+        &self, _request: tonic::Request<GetStatusDetailsRequest>,
+    ) -> std::result::Result<tonic::Response<GetStatusDetailsResponse>, tonic::Status> {
```
tobz, line 171: This should just be `Result` instead of the fully-qualified path.

lib/saluki-components/src/destinations/datadog/status_flare/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+        let response = GetStatusDetailsResponse {
+            main_section: Some(StatusSection { fields: status_fields }),
+            named_sections: HashMap::new(),
+        };
+        Ok(tonic::Response::new(response))
+    }
+
+    async fn get_flare_files(
+        &self, _request: tonic::Request<GetFlareFilesRequest>,
+    ) -> std::result::Result<tonic::Response<GetFlareFilesResponse>, tonic::Status> {
```
tobz, line 183: This should just be `Result` instead of the fully-qualified path.

bin/agent-data-plane/src/main.rs
```diff
@@ -106,17 +109,21 @@ async fn run(started: Instant, logging_api_handler: LoggingAPIHandler) -> Result
...
-    let blueprint = create_topology(&configuration, env_provider, &component_registry)?;
+    let blueprint = create_topology(&configuration, env_provider, &component_registry).await?;
 
     // Build our administrative API server.
     let primary_api_listen_address = configuration
         .try_get_typed("api_listen_address")
         .error_context("Failed to get API listen address.")?
         .unwrap_or_else(|| ListenAddress::Tcp(([0, 0, 0, 0], 5100).into()));
 
+    let remote_agent_server = new_remote_agent_server()?;
```
tobz, line 120: I don't love this, but it also doesn't matter _right now_ because it's basically a no-op anyways.

This will matter more, though, when we actually need to wire up the gRPC server implementation to the data collected by the component, at which point we'll need to do something more like grab this from the component configuration instead, so that we can build it pre-wired to the component.

Again, not a concern for _this_ PR, but we'll have to do something else for sure when actually adding the logic.

lib/saluki-app/src/api.rs
```diff
@@ -48,6 +61,40 @@ impl APIBuilder {
         self
     }
 
+    /// Adds the given tls configuration to this builder.
+    pub fn with_tls_config(mut self, config: ServerConfig) -> Self {
+        self.tls_config = Some(config);
+        self
+    }
+
+    /// Adds a self signed certificate to this builder.
```
tobz, line 70:
```suggestion
    /// Configures this builder to use TLS with a dynamically generated self-signed certificate.
```

lib/saluki-app/src/api.rs
```diff
@@ -48,6 +61,40 @@ impl APIBuilder {
...
+
+        let config = ServerConfig::builder()
+            .with_no_client_auth()
+            .with_single_cert(cert_chain, rustls::pki_types::PrivateKeyDer::Pkcs8(keys.remove(0)))
+            .unwrap();
+
+        self.with_tls_config(config)
+    }
+
+    /// Adds a Tonic router to this builder.
```
tobz, line 90:
```suggestion
    /// Sets the gRPC router to use for this builder.
```

(since this overrides rather than appends)

lib/saluki-app/src/api.rs
```diff
@@ -64,10 +112,18 @@ impl APIBuilder {
 
         // We have to convert this Tower-based `Service` to a Hyper-based `Service` to use it with `HttpServer`, since
         // the two traits are different from a semver perspective.
-        let service = TowerToHyperService::new(self.router);
+        let http_service = self.router;
+
+        let multiplexed_service = TowerToHyperService::new(MultiplexService::new(
+            http_service,
+            self.grpc_router.unwrap().into_router(),
```
tobz, line 119: We shouldn't unwrap here. Maybe just `.unwrap_or_else(|| Server::builder())`?

tobz (changes requested):

lib/saluki-app/src/api.rs
rayz: Hmm, I believe `Server::builder()` needs a `.add_service(svc)` call on it to actually make it return a `Router` which we would then use with the multiplexed service. Otherwise it just returns a `Server`

lib/saluki-components/src/destinations/datadog/status_flare/mod.rs
rayz: Correct, it will just spam the debug logs if it fails.

lib/saluki-app/src/api.rs
tobz: Hmm, doh, that's true.

Looking at the docs, I think we might want to change `grpc_router` to be [`RoutesBuilder`](https://docs.rs/tonic/latest/tonic/service/struct.RoutesBuilder.html): this would let us initialize the field with `RoutesBuilder::default()`, add multiple gRPC services via `RoutesBuilder::add_service`, and then in `serve`, we could consume it and get a `Routes` by calling `RoutesBuilder::routes`, which we would then convert into the proper router type by calling `Routes::into_axum_router`.

tobz (approved):

PR #411, opened 2025-01-07
enhancement: add support for forwarder concurrency to Datadog destinations

tobz:
## Context

This PR adds support for forwarder concurrency in the Datadog destinations. While the premise is simple, there's a lot of work that's gone into this PR and some notes for explaining what we did and _why_ we did it the way we did.

This PR is best reviewed (IMO) commit-by-commit, as there was a number of refactorings done before getting to the core change of adding concurrency.

### Generalize forwarder configuration

The first two commits were simply a little cleanup work to generalize the forwarder configuration -- retries, endpoints, etc -- under a single catch-all type (`DatadogAgentForwarderConfiguration`) and moved it under `saluki_components::destinations::datadog::common`.

We also moved what used to be `run_io_loop`/`run_endpoint_io_loop` in both the DD Metrics and DD Events/Service Checks destinations into the common module, and put them behind a new facade called `TransactionForwarder<B>`. This type facilitates both creating the underlying HTTP client _and_ spawning the I/O loop tasks that actually drive requests to completion.

This commit is large but represents a significant chunk of deduplication.

### Mirroring the "high priority" request channel

The third commit starts to mirror the Datadog Agent forwarder structure, where we add configurability around how many outstanding (waiting to be sent) requests can be queued up for a single endpoint before backpressure is applied.

The most crucial portion of this is that we're no longer fallibly trying to send and dropping things on the floor if the channel is full. This is necessary for future enhancements around adding support for the retry queue.

### Actually adding concurrency support

Finally, the fourth commit adds support for concurrency.

We've approached this in a pretty straightforward way: set a concurrency limit middleware layer on the HTTP client, and then try to keep that many outstanding requests in flight at the same time.

This is materially different from the Datadog Agent where the Agent literally spawns N worker goroutines all pulling from a single input queue, each running a single request at time. I don't think there's actually a meaningful difference between the two approaches, but we'll see what benchmarks say.


lib/saluki-components/src/destinations/datadog/common/io.rs
```diff
@@ -0,0 +1,274 @@
...
+            resolved_endpoint,
+        ));
+
+        endpoint_txs.push((endpoint_url, endpoint_tx));
+    }
+
+    // Listen for requests to forward, and send a copy of each one to each endpoint I/O task.
+    while let Some((metrics_count, request)) = requests_rx.recv().await {
+        for (endpoint_url, endpoint_tx) in &endpoint_txs {
+            if endpoint_tx.send((metrics_count, request.clone())).await.is_err() {
```
lukesteensen, line 151: Mostly just curious, I know this code was moved, but I assume we use some request body type with a cheap clone? Part of me wants to lose that as a generic arg if we aren't using it, but probably not worthwhile.

lukesteensen (approved):
Makes sense to me.

lib/saluki-components/src/destinations/datadog/common/io.rs
tobz: We could/probably should lose it, since, yeah, we use a fixed type, `FrozenChunkedBytesBuffer`... which is indeed cheaply cloneable: the spiritual equivalent of `bytes::Bytes`.

PR #414, opened 2025-01-08
[APR-281] Add support for blocking specific metrics ingested through DogStatsD

rayz:


lib/saluki-components/src/transforms/name_filter/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+use tokio::select;
+use tracing::{debug, error};
+
+#[derive(Deserialize)]
+/// Name Transform.
+pub struct NameFilterConfiguration {
+    #[serde(default, rename = "statsd_metric_namespace")]
+    metric_prefix: String,
+
+    #[serde(
```
tobz, line 25: I think you meant for this to be `serde`?

lib/saluki-components/src/transforms/name_filter/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+                    data[i] = data[j].clone(); // Move item to next position
+                }
+            }
+            data.truncate(i + 1); // Only keep valid elements
+        }
+
+        Blocklist { data, match_prefix }
+    }
+
+    fn test(&self, name: String) -> bool {
```
tobz, line 123: We should call this `contains`, since `test` is insanely generic/abstract... and this should also just take `&str`.

lib/saluki-components/src/transforms/name_filter/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+            debug!("Metric {} excluded due to blocklist.", new_metric_name);
+            return None;
+        }
+
+        // Update metric with new name.
+        let new_context = metric.context().with_name(new_metric_name);
+        let existing_context = metric.context_mut();
+        *existing_context = new_context;
+
+        Some(metric)
```
tobz, line 180: What we really want to avoid doing here is allocating or recreating anything if we didn't actually modify anything. So we know that we're only going to modify the metric name if we have a prefix configured, so we should probably have that be our top-level conditional.

```rust
if self.metric_prefix.is_empty() {
  // just check if this metric is in the blocklist, return Some/None based on that
} else {
  // existing logic goes here
}
```

That gets us away from having to do gymnastics to generate `metric_name`/`new_metric_name` prior to passing it to `self.blocklist.test(..)`.

lib/saluki-components/src/transforms/name_filter/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+
+                        for event in events {
+                            if let Some(metric) = event.try_into_metric() {
+                                if let Some(metric) = self.enrich_metric(metric) {
+
+                                let new_event = Metric::from_parts(
+                                    metric.context().clone(),
+                                    metric.values().clone(),
+                                    metric.metadata().clone(),
+                                );
```
tobz, line 215: Why are we cloning/rebuilding the metric here? We just got it back from `enrich_metric`; it's already good to go.

lib/saluki-components/src/transforms/name_filter/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+            if let Ok(index) = i {
+                if index > 0 && name.starts_with(&self.data[index - 1]) {
+                    return true;
+                }
+            } else if i.is_err() {
+                let index = i.unwrap_err();
+                if index > 0 && name.starts_with(&self.data[index - 1]) {
+                    return true;
+                }
+            }
```
tobz, line 141: Just use a `match` block to extract `index`, and then you can deduplicate this and keep it closer to the original:

```rust
if self.match_prefix {
    let index = i.map_or_else(|idx| idx, |idx| idx);
    if index > 0 && name.starts_with(&self.data[index - 1]) {
        return true;
    }
}
```

(The `map_or_else` bit is just a condensed form of `match` where we return the contained value for either side of `Result<T, E>` just they're both `usize`.)

lib/saluki-components/src/transforms/name_filter/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+    match_prefix: bool,
+}
+
+impl Blocklist {
+    fn new(data: &[String], match_prefix: bool) -> Self {
+        let mut data = data.to_owned();
+        data.sort();
+
+        if match_prefix && !data.is_empty() {
+            let mut i = 0;
```
tobz, line 110: I think a more important code comment to have here is that by doing this, we're only keeping unique prefixes: that's the crux of the filtering we're doing here and how it's then used in `test`.

lib/saluki-components/src/transforms/name_filter/mod.rs
tobz: You might need to tweak `is_excluded` a little bit.

tobz (changes requested):

lib/saluki-components/src/transforms/name_filter/mod.rs
rayz: Changed it but it seems like the `self.data.binary_search` takes in  `&String` so I have to convert it for that.

lib/saluki-components/src/transforms/name_filter/mod.rs
rayz: Sorry, bit confused by this 😅

tobz: Weird, when I reviewed this, it was literally `#[ core (` instead of `#[serde(`. 😮

lib/saluki-components/src/transforms/name_filter/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+            if self.blocklist.contains(&new_metric_name) {
+                debug!("Metric {} excluded due to blocklist.", new_metric_name);
+                return None;
+            }
+
+            // Update metric with new name.
+            let new_context = metric.context().with_name(new_metric_name);
+            let existing_context = metric.context_mut();
+            *existing_context = new_context;
+        }
```
tobz, line 182: Sorry, now that things are a little cleaner, I can see where the Datadog Agent code was going. We only want to prefix the metric if `self.metric_prefix` is actually set, _and_ if the original metric doesn't match any prefix configured in `self.metric_prefix_blacklist`.

As such, we want something like:

```rust
let new_metric_name = if self.is_excluded(...) {
    metric.context().name().clone()
} else {
    MetaString::from(format!(...))
}
```

That would end up generally being pretty efficient since the metric name should be inlined/interned so we're not allocating in the case where the metric has an excluded prefix.

lib/saluki-components/src/transforms/name_filter/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+        "kafka".to_string(),
+        "presto".to_string(),
+        "sidekiq".to_string(),
+        "solr".to_string(),
+        "tomcat".to_string(),
+        "runtime".to_string(),
+    ]
+}
+
+impl NameFilterConfiguration {
```
tobz, line 64: We should probably call this like... `DogStatsDPrefixFilterConfiguration` (and `DogStatsDPrefixFilter` for the component itself) since `NameFilter` doesn't _really_ explain what it's doing (certainly not that it might be prefixing the metrics, at least.)

lib/saluki-components/src/transforms/name_filter/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+
+impl MemoryBounds for NameFilterConfiguration {
+    fn specify_bounds(&self, builder: &mut MemoryBoundsBuilder) {
+        // Capture the size of the heap allocation when the component is built.
+        builder.minimum().with_single_value::<NameFilter>();
+    }
+}
+
+#[derive(Debug)]
+struct Blocklist {
```
tobz, line 99: This should definitely have some unit tests to make sure we're doing the right thing.

tobz (changes requested):

lib/saluki-components/src/transforms/dogstatsd_prefix_filter/mod.rs
```diff
@@ -290,6 +290,29 @@ mod tests {
...
+            metric_prefix: "foo".to_string(),
+            metric_prefix_blacklist: vec![],
+            blocklist: Blocklist::new(&["foobar".to_string(), "test".to_string()], false),
+        };
+        let context = Context::from_static_parts("bar", &[]);
+        let metric = Metric::gauge(context, 1.0);
+        let new_metric = filter.enrich_metric(metric);
+        assert!(new_metric.is_none());
+
+        let filter = DogstatsDPrefixFilter {
```
rayz, line 305: While testing/looking at the agent code, I saw that a metric won't be blocked if the name is the same as a blocked metric prefix but not in the block list 🤔 . (`foo` prefix is not added to make `foofoo` since `foo` in `metric_prefix_blacklist` but `foo` is not in `blocklist`)

lib/saluki-components/src/transforms/dogstatsd_prefix_filter/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+        "presto".to_string(),
+        "sidekiq".to_string(),
+        "solr".to_string(),
+        "tomcat".to_string(),
+        "runtime".to_string(),
+    ]
+}
+
+impl DogstatsDPrefixFilterConfiguration {
+    /// Creates a new `NameFilterConfiguration` from the given configuration.
```
tobz, line 70: Missed spot with the rename.

bin/agent-data-plane/src/main.rs
```diff
@@ -176,6 +177,7 @@ fn create_topology(
     let origin_enrichment_config = OriginEnrichmentConfiguration::from_configuration(configuration)
         .error_context("Failed to configure origin enrichment transform.")?
         .with_environment_provider(env_provider);
+    let name_filter_configuration = DogstatsDPrefixFilterConfiguration::from_configuration(configuration)?;
```
tobz, line 180: Probably change all the instance of `name_filter` (both this variable name and the component ID below) to something like `dsd_prefix_filter`?

lib/saluki-components/src/transforms/dogstatsd_prefix_filter/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+                if s == metric_name || self.blocklist.match_prefix && metric_name.starts_with(s) {
+                    return None;
+                }
+            }
+        } else {
+            // Enrich metric with prefix if prefix is allowed.
+            let new_metric_name = if self.is_excluded(metric_name) {
+                metric.context().name().clone()
+            } else {
+                MetaString::from(format!("{}{}", self.metric_prefix, metric_name))
```
tobz, line 176: Do we need to add a separator (like a period) between the prefix and metric name? Or does the Agent expect the user to add the trailing period to the prefix directly in their configuration?

lib/saluki-components/src/transforms/dogstatsd_prefix_filter/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+}
+
+impl DogstatsDPrefixFilter {
+    fn enrich_metric(&self, mut metric: Metric) -> Option<Metric> {
+        let metric_name = metric.context().name().deref();
+
+        if self.metric_prefix.is_empty() {
+            for s in &self.blocklist.data {
+                if s == metric_name || self.blocklist.match_prefix && metric_name.starts_with(s) {
+                    return None;
```
tobz, line 168: Probably also add the debug message from below (about excluding the metric) to this block as well?

lib/saluki-components/src/transforms/dogstatsd_prefix_filter/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+    }
+}
+
+#[async_trait]
+impl Transform for DogstatsDPrefixFilter {
+    async fn run(mut self: Box<Self>, mut context: TransformContext) -> Result<(), GenericError> {
+        let mut health = context.take_health_handle();
+        health.mark_ready();
+
+        debug!("Agent metric name filter transform started.");
```
tobz, line 207: I would change this from `Agent metric name filter` to `DogStatsD Prefix Filter`, just to make sure we're consistent with how other components log their human-friendly name.

tobz: (And then also update down below as well.)

lib/saluki-components/src/transforms/dogstatsd_prefix_filter/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+        let filter = DogstatsDPrefixFilter {
+            metric_prefix: "foo".to_string(),
+            metric_prefix_blacklist: vec![],
+            blocklist: Blocklist::default(),
+        };
+        let context = Context::from_static_parts("bar", &[]);
+        let metric = Metric::gauge(context, 1.0);
+        let new_metric = filter.enrich_metric(metric).unwrap();
+
+        assert_eq!(new_metric.context().name().clone().into_owned(), "foobar".to_string());
```
tobz, line 259: FWIW, there's equality comparisons for `MetaString`/`&str`, so I _think_ you could just do `assert_eq!(new_metric.context().name(), "foobar");` and have it work.

lib/saluki-components/src/transforms/dogstatsd_prefix_filter/mod.rs
tobz: That makes sense to me, but admittedly is why I think the Agent code (and this code as a result) is a little confusing... because it's coupling the logic for prefixing and the logic for the blocklist stuff into the same area.

lib/saluki-components/src/transforms/dogstatsd_prefix_filter/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    #[test]
+    fn test_metric_prefix_add() {
+        let filter = DogstatsDPrefixFilter {
+            metric_prefix: "foo".to_string(),
+            metric_prefix_blacklist: vec![],
+            blocklist: Blocklist::default(),
+        };
+        let context = Context::from_static_parts("bar", &[]);
+        let metric = Metric::gauge(context, 1.0);
```
tobz, line 256: FWIW, you can just do `Metric::gauge("bar", 1.0);` since there's an implementation to create a `Context` purely from a single static string, which implies no tags.

tobz (changes requested):
Just some very minor changes and a few questions.

lib/saluki-components/src/transforms/name_filter/mod.rs
tobz: Ah, yeah. You can use `binary_search_by` in order to customize the comparison so that we can use a borrowed string reference as the needle:

```rust
    let i = self.data.binary_search_by(|k| k.as_str().cmp(name));
```

lib/saluki-components/src/transforms/dogstatsd_prefix_filter/mod.rs
rayz: Oops good catch. The agent adds the trailling `.` if the user doesn't already have it as that as the suffix in the prefix. Updated the code to mimic this behavior

lib/saluki-components/src/transforms/dogstatsd_prefix_filter/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+    topology::OutputDefinition,
+};
+use saluki_error::GenericError;
+use saluki_event::{metric::Metric, DataType};
+use serde::Deserialize;
+use stringtheory::MetaString;
+use tokio::select;
+use tracing::{debug, error};
+
+#[derive(Deserialize)]
```
tobz, line 20: Nit: attributes should go below doc comments.

```rust
/// Foobar.
/// ...
/// ...
#[derive(Whatever)]
pub struct Blah {
    ...
}
```

tobz (changes requested):
One small nit and then this should be good to go.

tobz (approved):

PR #422, opened 2025-01-13
[APR-253] enhancement: track bounds expressions and generate basic sizing guide

lukesteensen:
The larger change here is to the way that we track memory bounds expressions. Rather than simply adding everything up immediately, we now keep track of the structure of the expressions themselves, noting which are config values, struct sizes, etc. This allows us to export those calculations to build tools related to the memory bounds.

The second part is less invasive, and simply adds a flag that when set will have ADP write out an html file with the memory bounds calculations embedded as JSON data, and some functionality to evaluate those expressions with varying config values, etc. It also includes some very basic heuristics to generate config changes from a workload spec (currently just metric cardinality).

### DIFF
--- a/lib/memory-accounting/src/api.rs
+++ b/lib/memory-accounting/src/api.rs
@@ -69,8 +69,9 @@ impl MemoryRegistryState {
             component_usage.insert(
                 component_name.to_string(),
                 ComponentUsage {
-                    minimum_required_bytes: bounds.self_minimum_required_bytes,
-                    firm_limit_bytes: bounds.self_firm_limit_bytes,
+                    // TODO: figure out what's actually going on here, this might not be a valid change
+                    minimum_required_bytes: bounds.total_minimum_required_bytes(),
+                    firm_limit_bytes: bounds.total_firm_limit_bytes(),
                     actual_live_bytes: stats_snapshot.live_bytes(),
                 },
             );
--- a/lib/saluki-components/src/sources/dogstatsd/mod.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/mod.rs
@@ -4,7 +4,7 @@ use std::{num::NonZeroUsize, time::Duration};
 use async_trait::async_trait;
 use bytes::{Buf, BufMut};
 use bytesize::ByteSize;
-use memory_accounting::{MemoryBounds, MemoryBoundsBuilder};
+use memory_accounting::{MemoryBounds, MemoryBoundsBuilder, UsageExpr};
 use metrics::{Counter, Gauge, Histogram};
 use saluki_config::GenericConfiguration;
 use saluki_context::{
@@ -331,12 +331,19 @@ where
         builder
             .minimum()
             // Capture the size of the heap allocation when the component is built.
-            .with_single_value::<DogStatsD>()
+            .with_single_value::<DogStatsD>("source struct")
             // We allocate our I/O buffers entirely up front.
-            .with_fixed_amount(self.buffer_count * get_adjusted_buffer_size(self.buffer_size))
+            .with_expr(UsageExpr::product(
+                "buffers",
+                UsageExpr::config("dogstatsd_buffer_count", self.buffer_count),
+                UsageExpr::config("dogstatsd_buffer_size", get_adjusted_buffer_size(self.buffer_size)),
+            ))
             // We also allocate the backing storage for the string interner up front, which is used by our context
             // resolver.
-            .with_fixed_amount(self.context_string_interner_bytes.as_u64() as usize);
+            .with_expr(UsageExpr::config(
+                "dogstatsd_string_interner_size",
+                self.context_string_interner_bytes.as_u64() as usize,
+            ));
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,6 +37,7 @@ dependencies = [
  "saluki-io",
  "saluki-metadata",
  "serde",
+ "serde_json",
  "stringtheory",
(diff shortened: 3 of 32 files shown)


lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -300,12 +300,19 @@ impl MemoryBounds for DogStatsDConfiguration {
...
             // Capture the size of the heap allocation when the component is built.
-            .with_single_value::<DogStatsD>()
+            .with_single_value::<DogStatsD>("source struct")
             // We allocate our I/O buffers entirely up front.
-            .with_fixed_amount(self.buffer_count * get_adjusted_buffer_size(self.buffer_size))
+            .with_expr(UsageExpr::product(
+                "buffers",
+                UsageExpr::config("dogstatsd_buffer_count", self.buffer_count),
+                UsageExpr::config("dogstatsd_buffer_size", get_adjusted_buffer_size(self.buffer_size)),
+            ))
```
tobz, line 309: Not a criticism of what you have, but this sort of example usage really drives home that it _feels_ like we may eventually want to do something to spruce this up and make it more foolproof.

Like I don't want to just proc macro-ify everything for the sake of writing proc macros, but this is definitely the sort of thing that could lend itself to using macros to help drive automatically, perhaps.

tobz: Random thought just to get it down on paper: maybe a proc macro that just generates methods based on a configuration type, where then instead of what is in place here, it would let you do something like:

```rust

#[derive(Computed)]
pub struct DogStatsDConfiguration {
...
}

# Calculating the bounds further down:
builder.minimum().with_expr(computed! { "buffers" => config(buffer_count) * config(buffer_size) });
```

which would desugar to something like:

```rust
impl DogStatsDConfigurationi {
    fn _computed_buffer_count(config: &DogStatsDConfiguration) -> UsageExpr {
        UsageExpr::config("dogstatsd_buffer_count", config.buffer_count)
    }
    
    fn _computed_buffer_size(config: &DogStatsDConfiguration) -> UsageExpr {
        UsageExpr::config("dogstatsd_buffer_size", config.buffer_size)
    }
}

...

builder.minimum().with_computed("buffers", UsageExpr::product(
    "buffers",
    DogStatsDConfiguration::_computed_buffer_count(self),i
    DogStatsDConfiguration::_computed_buffer_size(self),
));
```

Like obviously it's not perfect, and a little kludgy under the hood, but if it could help developers maybe... then I dunno, maybe worth some pain? Anyways, just scribbling out the idea. :P

lib/memory-accounting/src/api.rs
```diff
@@ -69,8 +69,9 @@ impl MemoryRegistryState {
             component_usage.insert(
                 component_name.to_string(),
                 ComponentUsage {
-                    minimum_required_bytes: bounds.self_minimum_required_bytes,
-                    firm_limit_bytes: bounds.self_firm_limit_bytes,
+                    // TODO: figure out what's actually going on here, this might not be a valid change
```
tobz, line 72: Regarding the TODO... what behavior did you see that led you to make this change?

tobz (commented):
Overall, I like this approach a lot, and I think it lays a lot of the groundwork for further investment/improvements in UX down the road. 👍🏻

lib/memory-accounting/src/api.rs
lukesteensen: No behavior that made me suspicious, just that I was moving quickly through these changes and didn't verify that there wasn't some semantic difference in this conversion that I'd neglected.

lib/saluki-components/src/sources/dogstatsd/mod.rs
lukesteensen: Yeah, I definitely agree it could be nice to make this more seamless and that's clashing with my natural inclination to avoid proc macros 😅 

I do think configuration values are the lowest hanging fruit for something like this, since we very much want to match exactly the actual configuration key and that's already part of things like `serde(rename = "...")`, etc. I'm hesitant to go full `configurable_component` with it, but something in that direction would very likely make sense.

lib/memory-accounting/src/api.rs
tobz: Ahh, gotcha. 👍🏻

lib/saluki-components/src/sources/dogstatsd/mod.rs
tobz: Yeah, I agree. It would have to be a delicate balance of extracting the information that's already there (like the user-visible configuration setting name via `#[serde(rename = "...")]`) vs requiring additional input beyond the initial `#[derive(...)]` to properly drive the calculations.

lib/saluki-components/src/transforms/origin_enrichment/mod.rs
```diff
@@ -0,0 +1,252 @@
+#![allow(dead_code)]
```
tobz, line 1: This file should be gonezo at this point.

Looks like you're up-to-date in terms of your parent... so maybe this just somehow snuck in during a rebase?

tobz (changes requested):

lib/saluki-components/src/transforms/origin_enrichment/mod.rs
lukesteensen: Yep, rebase gone wrong.

tobz (approved):

tobz (approved):

PR #426, opened 2025-01-15
enhancement: batch timestamped (passthrough) metrics for a short period of time before forwarding

tobz:
## Summary

This PR adds the ability for the `aggregate` transform to batch "passthrough" (pre-aggregated) metrics for short periods of time, in larger-than-normal event buffers, with the express goal of improving the efficiency of handling pre-aggregated metrics.

We've updated the logic of the transform to follow an equivalent behavior in the Datadog Agent's "no aggregation pipeline", which looks something like:

- batch pre-aggregated metrics into their own buffer (defaults to 2048 points)
- if the batch is filled to capacity, flush it to the "forwarder" (Datadog Metrics destination, in our case)
- every two seconds, check if we have received any pre-aggregated metrics within the last second: if not, flush the batch to the "forwarder"

This aims to improve how many pre-aggregated metrics are packed into an individual series/sketch request, improving efficiency and reducing the number of requests that have to be sent. There's still a difference in number of series/sketch requests sent between Core Agent and ADP even with this batching behavior in place, which I'm still currently investigating in staging.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Tested in staging.

(more detail to be added here as I test further)

## References

N/A

### DIFF
--- a/lib/saluki-components/src/transforms/aggregate/mod.rs
+++ b/lib/saluki-components/src/transforms/aggregate/mod.rs
@@ -479,6 +479,8 @@ impl PassthroughBatcher {
 
             if self.active_buffer.try_push(event).is_some() {
                 error!("Event buffer is full even after forwarding events. Dropping event.");
+                self.telemetry.increment_events_dropped();
+                return;
             }
         }
 
--- a/lib/saluki-components/src/transforms/aggregate/mod.rs
+++ b/lib/saluki-components/src/transforms/aggregate/mod.rs
@@ -1,4 +1,7 @@
-use std::{num::NonZeroU64, time::Duration};
+use std::{
+    num::NonZeroU64,
+    time::{Duration, Instant},
+};
 
 use async_trait::async_trait;
 use hashbrown::{hash_map::Entry, HashMap};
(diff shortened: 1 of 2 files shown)


lib/saluki-components/src/transforms/aggregate/mod.rs
```diff
@@ -223,159 +287,244 @@ impl MemoryBounds for AggregateConfiguration {
...
+    async fn push_metric(&mut self, metric: Metric, forwarder: &Forwarder) {
+        // Try pushing the metric into our active buffer.
+        //
+        // If our active buffer is full, then we'll flush the buffer, grab a new one, and push the metric into it.
+        if let Some(event) = self.active_buffer.try_push(Event::Metric(metric)) {
+            debug!("Passthrough event buffer was full. Flushing...");
+            self.forward_events(forwarder).await;
+
+            if self.active_buffer.try_push(event).is_some() {
+                error!("Event buffer is full even after forwarding events. Dropping event.");
```
rayz, line 481: Do we need to add a return here to prevent line 490 from hitting?

rayz (approved):

rayz (approved):

PR #428, opened 2025-01-17
enhancement: incorporate origin key into context key hash to shrink size of `ContextKey`

tobz:
## Context

In #415, we made a number of changes related to origin enrichment, one of which involved bundling the "origin" key -- a spiritual equivalent to `ContextKey`, but for origin data -- into `ContextKey` itself.

This had the side effect of changing the size of `ContextKey` from 8 bytes (`u64`) to 24 bytes (`u64` + `Option<OriginKey>`, where `OriginKey` was just a `u64`). As `ContextKey` is the key type for our actual context cache in `ContextResolver`, this means we're spending more memory for every cached context.

## Solution

We've slightly reworked how we generate, and pass around, the `OriginKey` during context resolving. We've also folded in the hash of the origin key to the context's hash itself, rather than carrying the origin key as a separate field. This brings `ContextKey` back down to 8 bytes.

### DIFF
--- a/lib/saluki-context/src/context.rs
+++ b/lib/saluki-context/src/context.rs
@@ -243,7 +243,18 @@ where
     }
 }
 
-impl<'a> Tagged for &'a [&'static str] {
+impl<'a> Tagged for &'a [&'a str] {
+    fn visit_tags<F>(&self, mut visitor: F)
+    where
+        F: FnMut(&str),
+    {
+        for tag in self.iter() {
+            visitor(tag);
+        }
+    }
+}
+
+impl<'a, const N: usize> Tagged for [&'a str; N] {
     fn visit_tags<F>(&self, mut visitor: F)
     where
         F: FnMut(&str),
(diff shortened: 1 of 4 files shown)


rayz (approved):

lib/saluki-context/src/context.rs
```diff
@@ -243,7 +243,18 @@ where
     }
 }
 
-impl<'a> Tagged for &'a [&'static str] {
+impl<'a> Tagged for &'a [&'a str] {
```
lukesteensen, line 246: nit: It feels like a lot of these could be covered with something like 

```
impl<T, S> Tagged for T
where
    T: IntoIterator<Item=S>,
    S: AsRef<str>,
{
    fn visit_tags<F>(self, mut visitor: F)
    where
        F: FnMut(&str),
    {
        for tag in self {
            visitor(tag.as_ref());
        }
    }
}
```

lukesteensen (approved):

tobz:
A note: this is on hold while getting the origin enrichment refactoring approved/merged, since it will likely need to change a little bit to fit into that.

tobz:
Superseded by #452.

PR #440, opened 2025-01-23
chore: slightly revamp External Data support to expose both pod and container entity ID

tobz:
## Context

We weren't entirely using External Data in the correct way prior to this PR, as the Datadog Agent actually enriches for the pod UID contained in External Data, and _then_ enriches based on the container ID that is resolved by _using_ the External Data. We only ever turned External Data identifiers into the container ID, so we were potentially missing things in doing it that way.

This PR adds a new type, `ResolvedExternalData`, which holds the entity IDs of the pod and container, respectively, and can be used to store the result of resolved a raw External Data identifier. This will be necessary for follow-up PRs that need to access the resolved entities from External Data when doing origin detection/enrichment.

We've reworked existing usages of External Data, although these usages will in fact also go away in the aforementioned follow-up PRs... but I opted to make the PRs atomic and so it is what it is. :)

PR #439, opened 2025-01-23
chore: switch metric-specific destinations to use new `Tagged` trait for metric tags

tobz:
## Context

This PR just updates the metrics-related destinations to actually take advantage of `Tagged` by using it on the `Context` itself, so they're not having to reach into specific getters for tags.

This sets us up later to incorporate visiting origin tags in the same call, without the callers needing to know or care deeply about the difference between instrumented and origin tags.

PR #438, opened 2025-01-23
chore: make `Tagged` relevant to materialized tags, not raw tags

tobz:
## Context

Prior to this PR, the `Tagged` trait was something we added to abstract over the numerous possible ways to supply the borrowed tags that get fed in when resolving a context.

This ends up being overkill and unnecessary without the `Resolvable` machinery, and would be better served as a way to abstract getting the materialized tags from different implementations of tag storage.

## Solution

This PR reworks the `Tagged` trait slightly by moving it under `saluki_context::tags`, having it give back `&Tag` instead of `&str`, and adding implementations for the major tag storage types, `TagSet`, `SharedTagSet`, and `Context`. This gives us a unified approach to visiting tags, regardless of where they're stored.

We've also reverted back, essentially, to the previous method signatures for hashing a context and creating context keys, by simply taking a value that can be turned into an iterator with items that dereference to a string. This more than satisfies the needs of the aforementioned methods, and lets us avoid a second trait related to tags, which would be otherwise confusing.

PR #436, opened 2025-01-23
chore: update to latest Datadog Agent RC for image builds/benchmarks

tobz:
## Context

Just bumping the Agent RC version so our converged images track closer to where we want to be.

PR #441, opened 2025-01-23
chore: introduce first-class origin tags abstraction

tobz:
## Context

Currently, we've frontloaded origin detection/enrichment in the ingestion phase, where we strive to both resolve the origin of a metric and then also resolve the tags for that origin, cache them, and return a stable reference to those tags.... all before we resolve the context.

This technically works, but has one major issue: origin tags arrive incrementally from the Core Agent, especially during startup. Taking this approach, which necessitates caching to maintain reasonable performance, leads to issues with missing tags and effectively breaks origin enrichment.

## Solution

We've taken the tack of breaking apart the resolving of an origin -- taking the raw identifiers and computing the unique origin identifier -- and the resolving of an origins tags.

Our new approach maintains much of the same machinery -- turning a raw `OriginInfo<'a>` into an `OriginKey` that is later used to look up the origin tags -- but we've added the aspect of deferring the actual resolution of the origin tags until they're queried.

To support this, we've created a new abstraction for origin tags called `OriginTags`. This type is the facade to visiting the tags attached to a given origin. It is powered by `OriginKey`, and a new trait, `OriginTagsResolver`, that replaces the old `OriginEnricher`. This allows us to resolve a stable origin identifier (`OriginKey`), bundle that identifier (and the resolver used to get it) into `OriginTags` itself, and then at the point where the origin tags are querier, the resolver is used again to look up the tags associated with the `OriginKey`.

Instead of merging and allocating storage for these tags, we've also added a new abstraction -- `OriginTagsVisitor` -- that is used to visit the tags attached to an origin. This lets us maintain a single copy of the tags in `TagStore` and simply iterate over them on-demand when requested.

A new type, `OriginTagsQuerier`, manages the link between the tag store, External Data resolver, and the logic to determine how to "resolve" an origin to its constituent parts -- pod, container, etc -- and sits within `saluki-env`. This type is wired into the ADP-specific workload provider, and bridged back to `saluki-context` by implementing the new `OriginTagsResolver`, such that configuring the DogStatsD source to use it is as simple as referencing the existing workload provider used in ADP.

Additionally, the `Tagged` implementation for `Context` now transparently visits the origin tags as well, which means both instrumented and origin tags are transparently chained together when visiting the tags for a context.


tobz:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/saluki/441?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#442** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/442?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#441** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/441?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/saluki/441?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#440** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/440?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#439** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/439?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#438** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/438?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#437** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/437?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#436** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/436?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -97,12 +96,10 @@ const fn default_dogstatsd_permissive_decoding() -> bool {
 ///
 /// Accepts metrics over TCP, UDP, or Unix Domain Sockets in the StatsD/DogStatsD format.
 #[derive(Deserialize)]
-pub struct DogStatsDConfiguration {
-    /// Origin enrichment configuration.
-    ///
-    /// See [`OriginEnrichmentConfiguration`] for more details.
-    #[serde(default)]
-    origin_enrichment: OriginEnrichmentConfiguration,
+pub struct DogStatsDConfiguration<R = ()> {
```
lukesteensen, line 99: It'd be nice not to add this generic, since I don't think it's really serving any meaningful purpose. You can't build the source with the default unit type, and there's really only one implementation of the trait.

More generally, abstracting with the trait leads to a bigger stack of nouns to hold in your head when reviewing this stuff (i.e. `WorkloadProvider` -> `OriginTagsResolver` -> `OriginTagsQuerier` -> `TagStoreQuerier` + `DogStatsDOriginTagResolver`), which is especially confusing because at least `TagStoreQuerier` is used at multiple levels. That's not to say the design is wrong, it's just complex. If there's a workable way to simplify it, great, and if not then it might be useful have some kind of ascii diagram in a comment that illustrates how all the pieces fit together 😅

lukesteensen (approved):
Just a note on something it'd be nice to simplify and/or document a bit more, which can certainly be a followup.

PR #442, opened 2025-01-23
chore: add resolver test for origin uniqueness + small cleanup

tobz:
## Context

Just adds a simple test for ensuring that differing origins lead to different contexts, and reworks a helper on `ContextResolverBuilder` to clean things up somewhat.

PR #444, opened 2025-01-24
fix: merge gauges properly and avoid dropping existing points

tobz:
## Context

When metrics are merged in the aggregate transform, we call `MetricValues::merge`, which is meant to uniformly handle the merge of different points for all metric types. This method handles both the merging of points based on their timestamp, as well as the specific merge behavior for the point values themselves.

An issue was pointed out with how ADP was handling gauges, and after some debugging, it became apparent that when merging gauges, we were actually ignoring the _timestamp_ but, which meant that if you had an existing point for a gauge, say with timestamp X, and you merged in another point with a timestamp Y, the first point would be entirely dropped.

This manifested itself as a fun bug where, due to the bucket width and flush interval defaults, a gauge being reported every 10 seconds was only ever emitted every 30 seconds.

## Solution

While gauges nominally have "last write wins" behavior, that logic only applies for a given timestamp. This PR fixes the logic of merging gauges such that we don't drop the previous points, and only apply the LWW behavior when merging two points with an identical timestamp.

Further, we've added a number of unit tests to ensure the merging behavior is correct for each metric type. We've additionally added a number of `From` and `Display` impls on the related types... some of which we only needed for initial debugging (but might be valuable in the future) and some of which are used by the unit tests.


lib/saluki-event/src/metric/value/mod.rs
```diff
@@ -440,16 +432,16 @@ impl MetricValues {
...
                 if *a_interval != b_interval {
                     *a_points = b_points;
                     *a_interval = b_interval;
                 } else {
-                    a_points.merge(b_points);
+                    a_points.merge(b_points, merge_scalar_sum);
                 }
             }
-            (Self::Gauge(a), Self::Gauge(b)) => *a = b,
+            (Self::Gauge(a), Self::Gauge(b)) => a.merge(b, merge_scalar_latest),
```
tobz, line 444: This is the fix, basically. Right here. This one lil guy.

tobz:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/saluki/444?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#444** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/444?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/saluki/444?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#442** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/442?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#443](https://github.com/DataDog/saluki/pull/443) <a href="https://app.graphite.dev/github/pr/DataDog/saluki/443?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#441** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/441?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#440** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/440?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#439** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/439?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#438** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/438?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#437** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/437?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#436** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/436?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lukesteensen (dismissed):

lukesteensen (approved):

lukesteensen (approved):

PR #453, opened 2025-01-27
[APR-286] Add filter for disabling metric payloads from being sent

rayz:
# Context

We would like to be able to filter metrics at the source component level before they materialize into `Events` in order to avoid doing any extra work if we know that the metrics will not be sent.

This pr adds the filter mechanism for `enable_payloads.series` and `enable_payloads.sketches`.

# Implementation

We create a new trait `Filter` with a `allow_metric` method.


lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -52,6 +52,9 @@ use self::framer::{get_framer, DsdFramer};
 mod origin;
 use self::origin::{origin_info_from_metric_packet, OriginEnrichmentConfiguration};
 
+mod filters;
+use filters::{EnablePayloadsFilter, Filter};
```
tobz, line 56: Small nit: I/we prefer to qualify child modules via `self` just to avoid _any_ confusion about them being an external crate or not.

```suggestion
use self::filters::{EnablePayloadsFilter, Filter};
```

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -298,6 +321,10 @@ impl SourceBuilder for DogStatsDConfiguration {
             }),
             codec,
             context_resolver,
+            pre_filters: vec![Arc::new(EnablePayloadsFilter::new(
```
tobz, line 324: Flip this around, so it's `Arc<Vec<dyn Filter + Send + Sync>>`, so that we're only cloning the the `Arc<...>` when we create a new listener/handler/stream, instead of actually cloning the `Vec<...>`.

lib/saluki-event/src/metric/value/mod.rs
```diff
@@ -470,6 +470,19 @@ impl MetricValues {
             Self::Distribution(_) => "distribution",
         }
     }
+
+    /// Returns whether the metric is a serie.
```
tobz, line 474:
```suggestion
    /// Returns `true` if this metric is a serie.
```

lib/saluki-event/src/metric/value/mod.rs
```diff
@@ -470,6 +470,19 @@ impl MetricValues {
...
+    pub fn is_serie(&self) -> bool {
+        matches!(
+            self,
+            Self::Counter(_) | Self::Rate(_, _) | Self::Gauge(_) | Self::Set(_)
+        )
+    }
+
+    /// Returns whether the metric is a sketch.
```
tobz, line 482:
```suggestion
    /// Returns `true` if this metric is a sketch.
```

lib/saluki-event/src/metric/value/mod.rs
```diff
@@ -470,6 +470,19 @@ impl MetricValues {
...
+    pub fn is_sketch(&self) -> bool {
+        matches!(self, Self::Histogram(_) | Self::Distribution(_))
```
tobz, line 484: Does this match the Agent behavior? I'm trying to think through this one because by default, we would materialize aggregates over the histogram and send them as gauges... but if we set histograms to be copied to distributions, then we would turn them into sketches, and so on... 🤔

rayz: Oops, looks like its a [serie](https://github.com/DataDog/datadog-agent/blob/main/pkg/metrics/histogram.go#L156). Updated

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -791,6 +828,12 @@ fn handle_frame(
     let event = match parsed {
         ParsedPacket::Metric(metric_packet) => {
             let events_len = metric_packet.num_points;
+            for filter in filters.iter() {
+                if !filter.allow_metric(&metric_packet) {
+                    debug!(%metric_packet.metric_name, "Metric filtered out.");
+                    return Ok(None);
```
tobz, line 834: Had a thought: what does the Agent do here in terms of telemetry? Should we tracking that we dropped this metric/packet?

tobz (approved):
LGTM overall, but just had one question.

lib/saluki-components/src/sources/dogstatsd/mod.rs
rayz: Doesn't look like there's any tracking https://github.com/DataDog/datadog-agent/blob/main/pkg/serializer/serializer.go#L336-L339

tobz (approved):

PR #463, opened 2025-01-30
chore: clean up origin enrichment logic / type names / etc

tobz:
## Summary

This PR is a follow-up to #441, in particular around cleaning up the layout and names of the various pieces related to origin detection and enrichment.

Specifically:

- we've removed the generic parameter from `DogStatsDConfiguration`, opting instead to simply allow passing a `WorkloadProvider` which we box internally (since it will only be called when resolving tags, instead of every time we process a metric)
- we've moved DogStatsD-specific origin detection/enrichment logic _back_ to the DogStatsD source (we originally moved it to `saluki-env` because of the complicated interdependence of types, but now I believe it is sufficiently decoupled without having had to add additional abstractions, so it should live close to the DSD source as it is DSD-specific)
- we've moved all things External Data to live in `saluki-context` as it is intimately related to contexts/origins
- `WorkloadProvider` now exposes two new methods related to generating an `OriginKey` from an origin, and then visiting the tags related to a key
- we've refactored `OriginTagsVisitor` to just be `TagVisitor`, which now acts as a mirror to `Tagged`
- we've renamed the various origin/External Data types to follow a common pattern of `X` for the owned variants, `XRef<'a>` for borrowed variants, and `ResolvedX` for versions where we've resolved the actual entity IDs involved

With this PR, we've normalized (IMO) the data flow in a single direction, with unified types:

- we start at the context resolver, which requires an implementation of `OriginTagsResolver`: this resolver handles creating the `OriginKey` from an `OriginRef<'a>` and then visiting the tags associated with that key
- the DSD source provides an implementation of `OriginTagsResolver` based on the workload provider, and acts as a thin shim over the workload provider that can apply the necessary logic of which entity IDs to pull tags for based on how origin detection is configured
- the workload provider given to the DSD source focuses exclusively on taking entity IDs and getting their tags, which should allow us to optimize how (and for how long) we store tag data in the future

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

TODO.

## References

N/A


tobz:
* **#464** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/464?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#463** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/463?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/saluki/463?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/saluki-context/src/origin.rs
```diff
@@ -67,7 +70,7 @@ impl fmt::Display for OriginTagCardinality {
     }
 }
 
-/// Information about the entity where a metric originated from.
+/// A borrowed representation of an origin.
```
lukesteensen, line 73: Is this really a borrowed representation of an origin? It feels more like "potentially incomplete collection of borrowed things that are associated with an origin and can be used to look one up", which obviously won't work as a name. But when I see `Ref` I think of something more akin to a pointer/reference, i.e. that's very direct and specific. Anyway, this might be too nitpicky but figured it was worth mentioning just for the sake of making names as clear as possible.

lukesteensen (approved):

lib/saluki-context/src/origin.rs
tobz: Hmm, no, that's a very fair point.

Let me noodle on other naming schemes.

lib/saluki-components/src/destinations/prometheus/mod.rs
```diff
@@ -329,7 +329,7 @@ fn format_tags(tags_buffer: &mut String, context: &Context) -> bool {
     let mut has_tags = false;
     let mut exceeded = false;
 
-    context.visit_tags(|tag| {
+    context.visit_tags(&mut |tag: &saluki_context::tags::Tag| {
```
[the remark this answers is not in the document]
tobz: Cleaned up in db64e11.

lib/saluki-context/src/origin.rs
tobz: I adjusted `OriginRef<'a>` and `ExternalDataRef<'a>` to be `RawOrigin<'a>` and `ExternalDataRef<'a>`, respectively. While `ExternalDataRef<'a>` _is_ much closer to a pointer/reference to `ExternalData`, you can't actually get one from `ExternalData`, so it's never used as reference, only a zero-copy lookup key, same as `RawOrigin<'a>`.

Let me know what you think.

lukesteensen (approved):
Nice! I like that a lot better.

PR #477, opened 2025-02-05
[WIP] DogstatsD Mapper

rayz:
## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

### DIFF
--- a/lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
+++ b/lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
@@ -1,0 +1,245 @@
+use std::collections::HashMap;
+use std::num::NonZeroUsize;
+use std::str::FromStr;
+use std::sync::LazyLock;
+
+use async_trait::async_trait;
+use bytesize::ByteSize;
+use memory_accounting::{MemoryBounds, MemoryBoundsBuilder};
+use regex::Regex;
+use saluki_config::GenericConfiguration;
+use saluki_context::{ContextResolver, ContextResolverBuilder};
+use saluki_core::{
+    components::transforms::{SynchronousTransform, SynchronousTransformBuilder},
+    topology::interconnect::FixedSizeEventBuffer,
+};
+use saluki_error::{generic_error, GenericError};
+use serde::{Deserialize, Serialize};
+use serde_with::{serde_as, DisplayFromStr, PickFirst};
+
+pub const MATCH_TYPE_WILDCARD: &str = "wildcard";
+pub const MATCH_TYPE_REGEX: &str = "regex";
+
+static ALLOWED_WILDCARD_MATCH_PATTERN: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9\-_*.]+$").expect("Invalid regex in ALLOWED_WILDCARD_MATCH_PATTERN"));
+
+/// DogstatsD mapper transform.
+#[serde_as]
+#[derive(Deserialize)]
+#[allow(dead_code)]
+pub struct DogstatsDMapperConfiguration {
+    #[serde(skip)]
+    context_string_interner_bytes: ByteSize,
+
+    #[serde_as(as = "PickFirst<(DisplayFromStr, _)>")]
+    #[serde(default)]
+    dogstatsd_mapper_profiles: MapperProfileConfigs,
+}
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[allow(dead_code)]
+struct MapperProfileConfigs(pub Vec<MappingProfileConfig>);
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[allow(dead_code)]
+struct MappingProfileConfig {
+    pub name: String,
+    pub prefix: String,
+    pub mappings: Vec<MetricMappingConfig>,
+}
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[allow(dead_code)]
+struct MetricMappingConfig {
+    #[serde(rename = "match")]
+    metric_match: String,
+    #[serde(rename = "match_type")]
+    match_type: String,
+    name: String,
+    tags: HashMap<String, String>,
+}
+
+#[allow(dead_code)]
+struct MetricMapper {
+    profiles: Vec<MappingProfile>,
+    context_resolver: ContextResolver,
+}
+
+#[allow(dead_code)]
+struct MappingProfile {
+    name: String,
+    prefix: String,
+    mappings: Vec<MetricMapping>,
+}
+
+#[allow(dead_code)]
+struct MetricMapping {
+    name: String,
+    tags: HashMap<String, String>,
+    regex: Regex,
+}
+
+impl FromStr for MapperProfileConfigs {
+    type Err = serde_json::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let profiles: Vec<MappingProfileConfig> = serde_json::from_str(s)?;
+        Ok(MapperProfileConfigs(profiles))
+    }
+}
+
+fn build_regex(match_re: &str, match_type: &str) -> Result<Regex, GenericError> {
+    let mut pattern = match_re.to_owned();
+    if match_type == MATCH_TYPE_WILDCARD {
+        // Check it against the allowed wildcard pattern
+        if !ALLOWED_WILDCARD_MATCH_PATTERN.is_match(&pattern) {
+            return Err(generic_error!(
+                "invalid wildcard match pattern `{}`, it does not match allowed match regex `{}`",
+                pattern,
+                ALLOWED_WILDCARD_MATCH_PATTERN.as_str()
+            ));
+        }
+        // Disallow "**"
+        if pattern.contains("**") {
+            return Err(generic_error!(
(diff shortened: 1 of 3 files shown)


lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+
+        Ok(MetricMapper { profiles })
+    }
+}
+
+impl DogstatsDMapperConfiguration {
+    /// Creates a new `DogstatsDMapperConfiguration` from the given configuration.
+    pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
+        let mut config: Self = config.as_typed()?;
+        config.context_string_interner_bytes = ByteSize::kib(512);
```
rayz, line 201: I copied the value from the `AgentTelemetryRemapper` as I'm unsure how to determine what value to set this as.

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+
+static ALLOWED_WILDCARD_MATCH_PATTERN: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9\-_*.]+$").expect("Invalid regex in ALLOWED_WILDCARD_MATCH_PATTERN"));
+
+/// DogstatsD mapper transform.
+#[serde_as]
+#[derive(Deserialize)]
+pub struct DogstatsDMapperConfiguration {
+    #[serde(skip)]
+    context_string_interner_bytes: ByteSize,
```
tobz, line 31: Related to your comment below about not knowing the right value to pick here....

I'd say you should make this configurable, like how it is in the DogStatsD source, since there's a potential for the default value to be insufficient. We can hardcode the size for the remapper because we know the remapped metrics have a fixed relationship to internal telemetry, which is also fixed.

However, we're deal with user-submitted metrics here, which means we might map 10, 100, or 10,000 metrics... we're not sure, so we need to allow for that possibility by ensuring our context interner can be sized up or down.

I'd say that the default should probably just be 64KB, though. I would guess that mapper usage is low enough that 95% of customers will never come close to that number, and if they do, it'll just heap allocate if necessary.

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+
+impl SynchronousTransform for DogstatsDMapper {
+    fn transform_buffer(&self, event_buffer: &mut FixedSizeEventBuffer) {
+        let context_string_interner_size = NonZeroUsize::new(self.context_string_interner_bytes.as_u64() as usize)
+            .ok_or_else(|| generic_error!("context_string_interner_size must be greater than 0"))
+            .unwrap();
+        let mut context_resolver = ContextResolverBuilder::from_name("dogstatsd_mapper")
+            .expect("resolver name is not empty")
+            .with_interner_capacity_bytes(context_string_interner_size)
+            .build();
```
tobz, line 244: You need to build the resolver when we build the component itself, and then stick it in `DogstatsDMapper` and use it from there... otherwise we're never really interning. :P

tobz: Might even be easiest just to pass it directly into `MetricMapper` and use it internally so that the higher-level transform code doesn't need to futz around with it.

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+                    let mut name = String::new();
+                    captures.expand(&mapping.name, &mut name);
+
+                    let mut tags = Vec::with_capacity(mapping.tags.len());
+                    for (tag_key, tag_value_expr) in &mapping.tags {
+                        let mut expanded_value = String::new();
+                        captures.expand(tag_value_expr, &mut expanded_value);
+                        tags.push(format!("{}:{}", tag_key, expanded_value));
+                    }
+                    return context_resolver.resolve(&name, tags, None);
```
tobz, line 80: So two things jump out at me here:

- we're ignoring the existing tags on the metric and just keeping whatever new tags we generate from the metric name
- we're going to lose origin tags since we're not doing anything with keeping the origin info

For the first one, and perhaps the second one: does the Agent actually throw away the existing tags on the metric in the process of mapping it?

If so, then this code would be fine as-is, I think? If not, then we'd obviously need to clone the existing tags, but we'd also need to copy over the origin information... and we don't really have a way to do that currently, so I'm going to think on that a little bit.

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+struct MapperProfileConfigs(pub Vec<MappingProfileConfig>);
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+struct MappingProfileConfig {
+    pub name: String,
+    pub prefix: String,
+    pub mappings: Vec<MetricMappingConfig>,
```
tobz, line 45: I don't think these _need_ to be `pub`?

tobz (changes requested):
Just some initial feedback from things I immediately saw. I'm still going through the mapping logic itself.

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
rayz: Oops thanks for catching this. The new tags should be added to the existing tags. Will update this

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
rayz: I had that originally but I think I ran into the issue of `transform_buffer` useing `&self` and to call `context_resolver.resolve(...)` I needed a `&mut self` on the `map` method.

tobz: Hmm, yes, you're right.

I don't know why I made `SynchronousTransform::transform_buffer` take `&self` instead of `&mut self`.

I'm gonna go fix that right now in a small PR.

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+        let metric = Metric::counter(context, 1.0);
+        let existing_tags = metric.context().tags();
+        let context = mapper
+            .map(
+                &input_metric_name,
+                existing_tags.clone(),
+                metric.context().origin_tags().clone(),
+            )
+            .unwrap();
+        let expected_metric_name = MetaString::from_static("custom_metric.process");
```
tobz, line 298: FWIW, I don't _think_ you should need to make this `MetaString` explicitly: `MetaString` dereferences to `&str`, which should mean it's fine to just do `assert_eq!(context.name(), "custom_metric.process");`

Did you try that and it didn't happen to work? Genuinely curious because that's how I would _want_ it to work, and if it doesn't, I want to tweak `MetaString` so that it does work.

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    fn mapper(json_data: &str) -> MetricMapper {
+        let mpc: MapperProfileConfigs = serde_json::from_str(json_data).unwrap();
+        let context_string_interner_bytes = ByteSize::kib(64);
+        mpc.build(context_string_interner_bytes).unwrap()
+    }
+
+    #[test]
+    fn test_mapper_wildcard() {
+        let json_data = r#"[{"name":"my_custom_metric_profile","prefix":"custom_metric.","mappings":[{"match":"custom_metric.process.*.*","match_type":"wildcard","name":"custom_metric.process","tags":{"tag_key_1":"$1","tag_key_2":"$2"}}]}]"#;
```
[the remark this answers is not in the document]
rayz: Gotcha, had it as one line to stop the text from taking as much space in the test. Will keep this in mind for the future

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
```diff
@@ -0,0 +1,373 @@
...
+                if let Some(new_context) = self.metric_mapper.try_map(metric.context()) {
+                    *metric.context_mut() = new_context;
+                }
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
tobz, line 271: I think we're missing some simple tests to exercise the error cases:

- unknown match type
- invalid match string when using `wildcard` match type
- consecutive asterisks when using `wildcard` match type

tobz (changes requested):

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
rayz: Done. Included all the same tests from the [datadog-agent mapper tests](https://github.com/DataDog/datadog-agent/blob/main/comp/dogstatsd/mapper/mapper_test.go) so the expected behavior *should* be the same

tobz (approved):

PR #483, opened 2025-02-10
chore: refactor Datadog Status/Flare component in preparation for shared telemetry via reflector

tobz:
## Summary

This PR does a bunch of refactoring around how we create/spawn the component previously known as the "Datadog Status/Flare Destination".

Primarily, it moves the code into the ADP binary crate itself, and de-componentizes it: now we just have standalone types, we spawn a standalone task, and so on. This is in preparation of having it draw its telemetry data from a new "reflector"-based pattern. We've also done some other work around `ListenAddress` to make it possible to cleanly pull the "what is our secure API listen address?" information necessary when registering ourselves as a remote agent.

This PR also includes the basic code for the reflector, which is included to intend to show roughly what we'll be moving _to_. This PR doesn't include all of the work to actually utilize the reflector, though, which will be part of a follow-up PR.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Built and ran ADP locally in non-standalone mode, and ensured that the Remote Agent task that replaces the component still runs and registers ADP as a remote agent.

## References

N/A


rayz (approved):
LGTM, just curious: are there any existing code we could migrate to use this new reflector pattern besides the shared telemetry?

PR #496, opened 2025-02-12
chore: wire up internal metrics reflector to status/flare component

tobz:
## Summary

This PR is a follow-up to #483 where we moved the Datadog Status/Flare component directly into the ADP binary, and structured it more as a direct bit of logic/async task. This work was in preparation of driving it with a new mechanism, also added in the PR, that we call a "reflector": a small, generic mechanism for processing a stream of data and maintaining some bit of state based on it. It follows the behavior of the identically-named mechanism used in Kubernetes controller implementations.

In this PR, we've actually added the first reflector implementation to be used: internal metrics. This is a basic implementation that captures all counters and gauges, and keeps an aggregated view of them: total sum of all observed values for counters, and most recent value for gauges. This means that we can essentially query the reflector and get the total sum of a given counter since the process started, or the most recent value for a gauge... and we don't have to deal with "points" and timestamps, or care too deeply about the specifics of `Context`, `Metric`, `MetricValues`, and so on. The exposed state is meant to provide a more ergonomic view over the data, and is designed specifically to allow for quickly and easily querying for a specific metric.

We've also wired up the Remote Agent API (neé Datadog Status/Flare component) to consume internal metrics through the reflector, and ADP will now respond to status requests with relevant DogStatsD-specific information, as seen in the screenshot below:

image

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

All testing was done locally by manually querying the Remote Agent gRPC service that is exposed on ADP's privileged API endpoint, while generating DSD load to ensure DSD telemetry was being updated.

## References

N/A


rayz (approved):
👌

PR #533, opened 2025-02-25
chore: update to nom v8

tobz:
## Summary

This PR updates `nom` to `8.0.0` which had some breaking changes due to how combinators have been restructured. We don't _need_ to upgrade, but I'm curious if we happen to notice anything in the benchmarks due to the changes they've made.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

N/A

## References

N/A


tobz:
Looks like the top end got _slightly_ faster, and nothing else regressed... so why not update it, I guess?

rayz (approved):

PR #541, opened 2025-03-03
enhancement: emit telemetry about registered components in health registry

### LINKED ISSUE #538: Emit telemetry about component healthy.
## Context

Currently, we emit no telemetry about the health of registered components beyond what is available in the health API endpoints themselves. This can make it difficult to debug after the fact when components have a change in readiness/liveness that leads to eventually failing health checks and being restarted.

We should consider emitting some additional internal telemetry:

- `health.component_ready` (value of `1` when component is marked as ready, `0` otherwise)
- `healthy.component_alive` (value of `1` when component is marked as live, `0` otherwise)

Both of these metrics would tagged with the component ID used to register the component.

This would let us trivially graph the readiness/liveness for each component, and understand when it changes, how often it changes, and so on.

tobz:
## Summary

This PR adds new internal telemetry to the health registry (`saluki_health::HealthRegistry`) about the status of registered components in order to be able to track changes to component readiness/liveness over time.

We've added two new gauges -- one for readiness, and one for liveness -- as well as a histogram -- for tracking liveness probe response latency -- which are all tagged with `component_id`, which corresponds to the name given when registering the component in the registry.

These metrics correspond to the output given when querying the readiness/liveness API endpoints themselves, and are updated in lockstep.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Built and ran ADP locally, and observed the new metrics as part of the telemetry scrape endpoint:

```text
toby@consigliera:~/src/saluki$ curl -s localhost:5102/metrics | grep health
# TYPE adp__health__component__alive gauge
adp__health__component__alive{component_id="topology.transforms.enrich"} 1
adp__health__component__alive{component_id="topology.destinations.dd_metrics_out"} 1
...
# TYPE adp__health__component__ready gauge
adp__health__component__ready{component_id="topology.transforms.internal_metrics_remap"} 1
adp__health__component__ready{component_id="topology.sources.internal_metrics_in"} 1
...
# TYPE adp__health__component__liveness_latency_secs histogram
adp__health__component__liveness_latency_secs_bucket{component_id="topology.sources.dsd_in",le="0.000029192926025390623"} 1
adp__health__component__liveness_latency_secs_bucket{component_id="topology.sources.dsd_in",le="0.0000656840835571289"} 4
...
```

## References

Closes #538.

### DIFF
--- a/lib/saluki-health/src/lib.rs
+++ b/lib/saluki-health/src/lib.rs
@@ -8,6 +8,7 @@ use std::{
 };
 
 use futures::StreamExt as _;
+use metrics::{gauge, histogram, Gauge, Histogram};
 use saluki_error::{generic_error, GenericError};
 use serde::Serialize;
 use stringtheory::MetaString;
@@ -34,7 +35,7 @@ struct ComponentHealth {
 
 /// A handle for updating the health of a component.
 pub struct Health {
-    ready: Arc<AtomicBool>,
+    shared: Arc<SharedComponentState>,
     request_rx: mpsc::Receiver<LivenessRequest>,
     response_tx: mpsc::Sender<LivenessResponse>,
 }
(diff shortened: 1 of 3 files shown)


rayz (approved):

lib/saluki-health/src/lib.rs
```diff
@@ -64,38 +70,80 @@ impl Health {
     }
 }
 
-#[derive(Eq, PartialEq)]
+#[derive(Clone, Copy, Eq, PartialEq)]
```
rayz, line 73: Is this utilized anywhere?

tobz: In `ComponentState::mark_live`, `ComponentState::mark_not_live`, and `ComponentState::mark_dead`, we pass `self.health` to `Telemetry::update_liveness`, which required making `HealthState` either `Copy` or passing it by reference... I just chose to make it `Copy` since it's not more efficient to pass it by reference.

Mostly just stylistic.

PR #547, opened 2025-03-05
enhancement: wire up support for retry queue w/ high vs low priority semantics for DD destinations

tobz:
## Summary

This PR adds full retry support for Datadog destinations, matching the behavior of the Datadog Agent.

### Background

When we talk about "retries" in the Datadog Agent, we're talking about the retrying of HTTP transactions sent by the "forwarder", which is used for sending metric, event, and service check payloads. Despite what might be commonly assumed to be "retry support", the Datadog Agent has a somewhat different spin on retries:

- all transactions (requests) sent to a forwarder first enter the "high priority" queue which is drained by a configurable number of "workers" (tasks that only care about driving the HTTP requests themselves)
- workers also drain transactions from a "low priority" queue, while giving preference to draining all high priority transactions first
- when a transaction fails in a retriable way, two things happen:
  - the endpoint to which the transaction was sent is marked as having failed, putting it into a "blocked" state (for a configurable amount of time)
  - the transaction is "requeued" and sent to the retry queue
- while an endpoint is blocked, all transactions pulled off a queue for processing that target that endpoint will be immediately diverted to the retry queue
- periodically, the retry queue is drained and reinjected into the low priority queue
- for any action where a transaction would be added to the high or low priority queues, if the given queue is currently full, the transaction is instead sent back to the retry queue
- the retry queue has a finite amount of capacity it will use for enqueued transactions, and will drop older transactions to make room for incoming transactions

As you can see, the logic is somewhat intertwined and complex, but at the core:

- we keep a main queue of transactions to process (high-priority queue), and fall back to enqueuing pending transactions (whether new or to be retried) to a lower-priority queue, both of which are bounded in space
- we always prefer to drain high-priority transactions first, to prioritize fresh data being processed
- when a request fails, we don't sequentially retry it, but instead defer it to be retried later

### Approach

This PR has a number of parts to try and satisfy these requirements, which I'll enumerate below.

#### `Transaction<B>` and `TransactionBody<B>`

As transactions that are to be retried are sent to `RetryQueue<T>`, this means that they will potentially be serialized to disk, which means that we can't simply take our existing `Request<B>` values and serialize them. We have to consider if the body type is serializable, and we also need to carry other metadata about the request, such as the event count.

In order to handle this, we've introduced `Transaction<B>` and `TransactionBody<B>`. These types provide a wrapper over an otherwise generic `http::Request<B>` and the relevant metadata (event count, anything else we need to add in the future, etc) as well as a (de)serialization implementation to allow using the same type for an "original" (not rehydrated from disk) transaction as a disk-backed one. This mostly centers around abstracting the body types over two possible implementations to handle either scenario.

It's a lot of boilerplate, but it's boilerplate we'll almost never need to touch going forward.

#### `FixedBody`

This is just a simple wrapper type to replace cases where our request body was `String`, in order to make it compatible with `Transaction<B>`/`TransactionBody<B>`.

#### `PendingTransactions`

`PendingTransactions` a simple abstraction that wraps over the high-priority queue (a simple `VecDeque<T>`) and the low-priority queue (`RetryQueue<T>`). We've simplified things compared to the Datadog Agent because otherwise we would need a separate task/chunk of logic to periodically drain the retry queue and then feed it into yet another channel. We don't _need_ that abstraction in order to meet any concurrency-related behavior, and we can avoid arbitrarily delaying the retry of transactions (i.e., waiting five seconds when the backoff is actually less than that.)

#### `TransactionForwarder` 

The vast majority of the changes are in the `TransactionForwarder` code itself, in terms of wiring up `RetryCircuitBreaker` to provide the non-sequential retry behavior, and `PendingTransactions` to manage the selection of transactions to process from the high and low priority queues.

This is where we specifically handle:

- ensuring we get transactions into the "pending" transaction queue (which handles the high- and low-priority queues, and their respective ordering/behavior) as fast as possible to avoid backpressure in the main endpoint I/O task
- waiting for our service to be "ready", where service readiness is tied directly to any indicated backoff (due to an endpoint becoming "blocked"), before processing the next highest-priority transaction
- driving in-flight transactions to completion

#### Datadog Metrics destination

We had to make some changes here related to memory bounds and buffer pool sizing. As the introduction of some of the aforementioned code means that we can now have many more outstanding transactions sitting in memory, our previous buffer pool sizing is now insufficient: we can't simply build one request and wait for it to be sent, otherwise we'd immediately block ourselves from processing in-flight metrics as soon as a single request triggered a retry.

We've wired up both the memory bounds and buffer pool creation to the relevant forwarder configuration, such that we size our buffer pool to ensure that we have enough chunks to build as many transactions as we're configured to be able to keep around in-memory. We now use an elastic object pool for this as we would otherwise be allocating a _lot_ of memory upfront to support the default queue sizes used by the Datadog Agent.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

A combination of unit testing and local testing. Local testing involves simulating an endpoint failure by simply terminating the dummy HTTP blackhole server that ADP was configured to send to, allowing transactions to build up, and then starting the server again to let them drain.

## References

N/A

### DIFF
--- a/lib/saluki-components/src/destinations/datadog/common/io.rs
+++ b/lib/saluki-components/src/destinations/datadog/common/io.rs
@@ -4,7 +4,7 @@ use futures::FutureExt as _;
 use http::{Request, Uri};
 use http_body::Body;
 use http_body_util::BodyExt as _;
-use hyper::body::Incoming;
+use hyper::{body::Incoming, Response};
 use saluki_config::RefreshableConfiguration;
 use saluki_core::task::spawn_traced;
 use saluki_error::{generic_error, GenericError};
@@ -28,11 +28,12 @@ use tracing::{debug, error};
 use super::{
     config::ForwarderConfiguration,
     endpoints::ResolvedEndpoint,
-    telemetry::ComponentTelemetry,
+    middleware::{for_resolved_endpoint, with_version_info},
+    telemetry::{ComponentTelemetry, TransactionQueueTelemetry},
     transaction::{Transaction, TransactionBody},
 };
-use crate::destinations::datadog::common::middleware::{for_resolved_endpoint, with_version_info};
 
+/// A handle to the transaction forwarder.
 pub struct Handle<B>
 where
     B: ReadIoBuffer + Clone,
@@ -45,6 +46,11 @@ impl<B> Handle<B>
 where
     B: ReadIoBuffer + Clone,
 {
+    /// Sends a transaction to the forwarder.
+    ///
+    /// # Errors
+    ///
+    /// If the endpoint I/O task has unexpectedly stopped and can no longer accept transactions, an error will be returned.
     pub async fn send_transaction(&self, transaction: Transaction<B>) -> Result<(), GenericError> {
         match self.transactions_tx.send(transaction).await {
             Ok(()) => Ok(()),
@@ -52,6 +58,7 @@ where
         }
     }
 
+    /// Triggers the forwarder to shutdown and waits to shutdown to complete.
     pub async fn shutdown(self) {
         let Self {
             transactions_tx,
@@ -69,6 +76,7 @@ where
 pub struct TransactionForwarder<B> {
     config: ForwarderConfiguration,
     telemetry: ComponentTelemetry,
+    metrics_builder: MetricsBuilder,
     client: HttpClient<TransactionBody<B>>,
     endpoints: Vec<ResolvedEndpoint>,
 }
@@ -79,6 +87,7 @@ where
     B::Data: Send,
     B::Error: std::error::Error + Send + Sync,
 {
+    /// Creates a new `TransactionForwarder` instance from the given configuration.
     pub fn from_config<F>(
         config: ForwarderConfiguration, maybe_refreshable_config: Option<RefreshableConfiguration>, endpoint_name: F,
         telemetry: ComponentTelemetry, metrics_builder: MetricsBuilder,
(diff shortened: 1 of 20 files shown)


lib/saluki-components/src/destinations/datadog/common/io.rs
```diff
@@ -193,33 +210,48 @@ async fn run_endpoint_io_loop<S, B>(
...
         .service(service);
 
+    let retry_queue = RetryQueue::new(endpoint_url.clone(), config.retry().retry_queue_max_size_bytes);
+    let mut pending_transactions = PendingTransactions::new(config.endpoint_buffer_size(), retry_queue);
+
     let mut in_flight = JoinSet::new();
     let mut done = false;
 
     loop {
         select! {
```
rayz, line 230: Rust question: `select` just chooses the branch that completes first right? Is there a way to prioritize which one runs? (Let's say maybe we want to finish all requests to completion before starting new requests.

tobz: Yes, and yes.

By default, the `select!` will generate code to randomize the order it checks the branches in. You can add a little extra directive -- `biased;` -- which turns that off and always checks them in the order they're defined in:

```rust
select! {
  biased;

  _ = my_future => { ... }
}
```

Generally you want the randomization for fairness, and to potentially avoid deadlocking yourself if you're doing something complex... but there are situations where biased polling makes sense, like if one of your branches is checking if you should shutdown, since otherwise you might take another branch (perhaps one that causes you to do a lot of work or go to sleep) a few times before you take the shutdown one, and so on.

lib/saluki-components/src/destinations/datadog/common/retry.rs
```diff
@@ -65,9 +69,19 @@ pub struct RetryConfiguration {
 
     /// Whether or not a successful request should completely the error count.
```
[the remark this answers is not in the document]
tobz: Good eye! Fixed.

rayz (approved):

tobz:
Note to self: we still want to explore augmenting the circuit breaker to have support for "soft start" behavior, where it also does a reverse of the exponential backoff as it starts to let traffic back through.

This will likely be useful to help avoid slamming endpoints with the transactions sitting in the internal queues, exhibiting some mechanical sympathy on both ADP and the downstream endpoints.

PR #550, opened 2025-03-07
chore: update deps and fix cargo-deny config

tobz:
## Summary

Two parts:

- update dependencies (simple)
- fix our `cargo-deny` configuration to compensate for two recent RUSTSEC advisories that we would dispute as being not relevant to us
  - `paste` is now "unmaintained", but it's been stable basically since it was created, so who cares?
  - `protobuf` has an issue related to uncontrolled recursion when decoding payloads, but we only use it for _encoding_ payloads, and never for decoding

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

N/A

## References

N/A

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -3,7 +3,10 @@ version = 2
 db-path = "~/.cargo/advisory-db"
 db-urls = ["https://github.com/rustsec/advisory-db"]
 yanked = "warn"
-ignore = []
+ignore = [
+  { id = "RUSTSEC-2024-0436", reason = "paste is a stable crate and we do not consider it being unmaintained as a security risk" },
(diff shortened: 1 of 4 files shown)


deny.toml
```diff
@@ -3,7 +3,10 @@ version = 2
 db-path = "~/.cargo/advisory-db"
 db-urls = ["https://github.com/rustsec/advisory-db"]
 yanked = "warn"
-ignore = []
+ignore = [
+  { id = "RUSTSEC-2024-0436", reason = "paste is a stable crate and we do not consider it being unmaintained as a security risk" },
```
jszwedko, line 7: TIL there is a `reason` field we can use in the `ignore` block.

jszwedko (approved):

PR #551, opened 2025-03-07
enhancement: add `RetryCircuitBreaker` middleware

tobz:
## Summary

This PR implements a new Tower middleware, `RetryCircuitBreaker`. It is designed to help mimic the retry behavior used in the Datadog Agent which involves a blend of both traditional retry behavior (try again after a period of time) and circuit breaker behavior (all requests made while the circuit is open will fail).

Currently, we use the generic retry middleware provided in the `tower` crate, but this comes with a particular downside: requests have to proceed sequentially, and one request being stuck blocks all other requests behind it. The Datadog Agent will intentionally divert additional incoming requests to a "retry queue" when a worker is fully saturated, and when a given endpoint is currently being backed off due to a previous error(s). This is hard to accomplish as-is with the generic retry middleware because we can't differentiate between the request still executing on its first attempt or having gone into trying further attempts, which means once we make the service call, we're stuck waiting for it.

`RetryCircuitBreaker` is a middleware implementation that allows using an existing retry `Policy` from `tower` -- so we can take advantage of what we've already got -- and uses it to determine if a response constitutes a success or failure. When a request fails, the middleware itself modulates its readiness -- if we're instructed to back off for N seconds, the service will then not report as ready for N seconds -- but the response is immediately returned, indicating to the caller that the call failed and the circuit is now open.

What this allows us to do is await service readiness to match the behavior of not starting a request _at all_ while we're in backoff, which in turn will allow us to choose to divert incoming requests to a retry queue until the service becomes ready. Additionally, the service can return the original request when the circuit breaker is open or when the request fails in a retryable way, which means the code using this middleware doesn't have to hold on to its own copy of the request to potentially requeue it later on if it fails.

## Notes

I'm leaving any changes to actually _use_ this middleware for follow-up PRs.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests only.

## References

N/A

### DIFF
--- a/lib/saluki-io/src/net/util/middleware/retry_circuit_breaker.rs
+++ b/lib/saluki-io/src/net/util/middleware/retry_circuit_breaker.rs
@@ -1,0 +1,438 @@
+use std::{
+    future::Future,
+    pin::Pin,
+    sync::{Arc, Mutex},
+    task::{ready, Context, Poll},
+};
+
+use futures::FutureExt as _;
+use pin_project_lite::pin_project;
+use tower::{retry::Policy, Service};
+
+/// An error from [`RetryCircuitBreaker`].
+#[derive(Debug)]
+pub enum Error<E, R> {
+    /// The inner service responded with an error.
+    Service(E),
+
+    /// The circuit breaker is open and requests are being rejected.
+    Open(R),
+}
+
+impl<E, R> PartialEq for Error<E, R>
+where
+    E: PartialEq,
+    R: PartialEq,
+{
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Service(a), Self::Service(b)) => a == b,
+            (Self::Open(a), Self::Open(b)) => a == b,
+            _ => false,
+        }
+    }
+}
+
+pin_project! {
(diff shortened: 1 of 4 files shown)


lib/saluki-io/src/net/util/middleware/retry_circuit_breaker.rs
```diff
@@ -0,0 +1,438 @@
...
+        let result = fut.await;
+        assert_eq!(result, good_req.as_service_response());
+
+        // Second request should fail and should be a service error, because without being able to clone the request, it
+        // can't be retried anyways.
+        let mut circuit_breaker = RetryCircuitBreaker::new(LoopbackService, NonCloneableTestRetryPolicy);
+        let svc = circuit_breaker.ready().await.expect("should never fail to be ready");
+        let fut = svc.call(bad_req.clone());
+        let result = fut.await;
+        assert_eq!(result, bad_req.as_open_response());
```
rayz, line 373:
```suggestion
        assert_eq!(result, bad_req.as_service_response());
```

rayz (approved):
TIL: circuit breaker design pattern

LGTM

PR #554, opened 2025-03-10
enhancement: use zstd by default for metric payloads

lukesteensen:
## Summary

Add support for zstd compression of metric payloads and make it the default behavior to match the Agent.

This required a change to how our chunked buffers accept writes, as we were seeing some deadlock-like behavior with the previous implementation that waited to collect the full capacity for each write. This PR changes that to write as much as it can to the available buffers, even if that is not the full amount, and allow the rest to be written on subsequent calls.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## References

AGTMETRICS-56

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ snafu = { version = "0.8", default-features = false, features = ["std"] }
 tokio = { version = "1", default-features = false }
 tracing = { version = "0.1", default-features = false, features = ["std"] }
 ahash = { version = "0.8", default-features = false, features = ["std", "runtime-rng"] }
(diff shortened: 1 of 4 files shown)


rayz (approved):
LGTM.  Are we also going to support the `zstd_compression_level` config thats in the datadog-agent as well?

lukesteensen:
@rayz Yeah, I think we'll want to add more knobs around compression generally. Given that there might be some discussion around that, I'm inclined to go ahead and merge this as-is and then write a ticket for making it configurable.

PR #564, opened 2025-03-20
enhancement: add new zstd compressor that can obey a maximum size limit

tobz:
## Summary

This PR introduces a new compression type, `ZstdLimitedCompressor`, which is meant as a deeply-integrated wrapper over Zstandard which integrates with our object pool regime as well as directly supports the ability to inform the caller when a compressed payload would exceed a configured limit, allowing the caller to finalize the payload and start a new one.

This is intended to be used in Datadog destinations, where there are limits on the size of compressed payloads, as avoiding the generation of compressed payloads that exceed the limit is more efficient, as it allows us to avoid having to "split" the payload, which then forces us to send additional HTTP requests, wasting more network bandwidth and CPU.

I've tried very hard to document the code to explain the "how" and "why" of this implementation, so I would highly suggest reading the documentation of `ZstdLimitedCompressor` first before going through any actual code review.

Notably, this PR does _not_ switch anything to actually use it, as I wanted to have this code be reviewed standalone after getting it to a point of being well documented and well tested.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests, and property testing.

## References

N/A


lib/saluki-io/src/compression/zstd.rs
```diff
@@ -0,0 +1,757 @@
...
+            Ok(WriteStatus::Ok) => panic!("should not be able to fully finalize to the output buffer"),
+            Ok(WriteStatus::OutputBufferFull) => {
+                // We have a full output buffer, so let's check the status.
+                assert_eq!(output.as_slice().len(), output.capacity());
+            }
+            Err(e) => panic!("finalize failed: {}", e),
+        }
+    }
+
+    #[test_strategy::proptest(async = "tokio")]
```
tobz, line 681: We use this non-traditional way of creating a `proptest`-based test as `proptest`'s normal `proptest!` declarative macro _cannot_ be used when wrapped around an async test function that uses `#[tokio::test]`.

PR #575, opened 2025-03-26
chore(ci): unify image builds into build stage + debug vs release cleanup

tobz:
## Summary

This PR involves a bunch of cleanup around our CI pipeline for building ADP images and publishing ADP releases.

There's a lot going on here, so I'm just going to try to summarize:

1. We now build all ADP images (debug, release, etc) in the `build` stage so that we can consolidate _how_ we build them, varying only the necessary parameters, etc, to generate the different images.
2. We still build ADP images for SMP runs in the `benchmark` stage, mostly because of the fact it works and it depends on spelunking through Git history to find the merge base to get the baseline commit (and it lets us avoid the default multi-platform image builds, so things can start faster)
3. We've removed the `optimized-debug-release` Cargo profile and just use `optimized-release` now in its place: we always want to have debug symbols present for built images, since we currently have no need to ship stripped binaries (we did this for published releases, but they're beta, so it's better to have debug symbols at the moment).

Other than that, this is mostly cleanup: DRYing up variable usage, naming things a little more succinctly or consistently, and so on.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

N/A

## References

N/A


rayz (approved):

.gitlab/benchmark.yml
spencergilbert, line 1: Should these use the `BUILD_PROFILE` variable from `.gitlab-ci.yml`?

spencergilbert (approved):

.gitlab/benchmark.yml
tobz: Yeah, we should make sure our benchmarks are always running the same level of optimization as released binaries.

Good callout. 👍🏻

tobz: Fixed in a26f2d38.

PR #595, opened 2025-04-08
experiment(dsd): dedicated non-caching context resolver for no-agg DSD metrics

tobz:
## Summary

This PR introduces the ability to disable caching of contexts in `ContextResolver` and updates the DogStatsD source to use a dedicated context resolver for pre-aggregated metrics.

Normally, all incoming metrics would have their contexts resolved through `ContextResolver`, whether the metrics need to be aggregated or not. This means that we always pay the cost of caching the context of pre-aggregated metrics, and holding the context until it hits idle expiration (or not), when the value of caching such contexts is drastically less than non-aggregated metrics, due to their far more infrequent emissions.

Overall, this can lead to a large amount of memory usage (and the resulting permanent growth of the cache data structures) that we should be able to more aggressively reclaim.

This PR adds the ability to disable caching of contexts during resolving -- we skip adding them to the cache, and tracking their expiration, while all else stays the same -- and we now use that in the DogStatsD source.

Ultimately, the experiment here is to see: for workloads that are heavily biased towards pre-aggregated metrics, can we improve ADP's memory usage by a meaningful amount by always allocating contexts which will allow them to be dropped as _soon_ as they are no longer referenced?

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Added a single unit test for ensuring that disabling caching in `ContextResolver` actually skips caching. Remainder of testing, namely performance, will happen in staging.

## References

N/A

### DIFF
--- a/lib/saluki-components/src/sources/dogstatsd/mod.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/mod.rs
@@ -1,5 +1,5 @@
 use std::sync::{Arc, LazyLock};
-use std::{num::NonZeroUsize, time::Duration};
+use std::time::Duration;
 
 use async_trait::async_trait;
 use bytes::{Buf, BufMut};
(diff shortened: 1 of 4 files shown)


rayz (approved):
LGTM

Issue #596, opened 2025-04-08
Improve efficiency and memory usage of `host_tags` synchronous transform.


tobz:
## Context

The `host_tags` synchronous transform is used to apply host tags to metrics for some configurable duration after initial startup, when the Datadog backend may not yet be associating host tags to metrics automatically when the host itself is new.

We do this by simply querying the host tags from the perspective of the Datadog Agent, and then extending the metric's instrumented tags until we've passed the configurable duration, and then the transform becomes a no-op from that point.

However, we have to do normal context resolving to create the updated context with the aforementioned host tags, which is where we're currently doing things inefficiently.

## Solution

We should do three main things:

- build a list of tags via `&str` before resolving, rather than allocating owned `String`s
- set an expiration on the resolver to actively clean up entries after they're idle for long enough
- hold the resolver (and host tags) in `Option<T>`, and drop them once the configurable duration has been exceeded

The first item addresses unnecessary allocations as the resolver is going to intern the tag strings regardless. We're still likely to have to allocate when interning, but we don't need to do additional temporary allocations first if we can help it.

The second and third items let the transform, and resolver, remove infrequently used contexts and avoid just holding on to all resolved contexts for all time. The third item, in particular, _fully_ removes all allocations related to the host tags-updated contexts as it will drop the interner, the resolver, and the host tags we've collected.

PR #597, opened 2025-04-08
chore: set expiration on host tags transform resolver

rayz:
## Summary

This pr introduces a couple of changes to make the host tags transform more efficient.
- Using `&str` instead of `String` for tags
- Using `with_idle_context_expiration` with the `self.expected_tags_duration`
- Dropping the resolver and host tags when `self.expected_tags_duration` has elapsed.
## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Performance

## References

Closes: #596

### DIFF
--- a/lib/saluki-components/src/transforms/host_tags/mod.rs
+++ b/lib/saluki-components/src/transforms/host_tags/mod.rs
@@ -60,12 +60,14 @@ impl SynchronousTransformBuilder for HostTagsConfiguration {
         let context_resolver = ContextResolverBuilder::from_name("host_tags")
             .expect("resolver name is not empty")
             .with_interner_capacity_bytes(context_string_interner_size)
+            .with_idle_context_expiration(Duration::from_secs(self.expected_tags_duration))
+            .with_expiration_interval(Duration::from_secs(1))
             .build();
         Ok(Box::new(HostTagsEnrichment {
             start: Instant::now(),
-            context_resolver,
+            context_resolver: Some(context_resolver),
             expected_tags_duration: Duration::from_secs(self.expected_tags_duration),
-            host_tags,
+            host_tags: Some(host_tags),
         }))
     }
 }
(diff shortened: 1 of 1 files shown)


lib/saluki-components/src/transforms/host_tags/mod.rs
```diff
@@ -60,12 +60,14 @@ impl SynchronousTransformBuilder for HostTagsConfiguration {
         let context_resolver = ContextResolverBuilder::from_name("host_tags")
             .expect("resolver name is not empty")
             .with_interner_capacity_bytes(context_string_interner_size)
+            .with_idle_context_expiration(Duration::from_secs(self.expected_tags_duration))
```
tobz, line 63: We should ideally set this _lower_, that way infrequent contexts can be cleared out. As is, we would essentially never expire any context up until the point we dropped the entire resolver, so it would just keep growing and growing.

If we set it lower (say thirty seconds? I'm not sure we care deeply about making this configurable), then we can gradually recover some memory while it's active.

tobz (changes requested):
Overall, this is looking good.

One thing I do think we should have is just a simple unit test that ensures we enrich correctly _and_ that we stop enriching after the duration has elapsed.

lib/saluki-components/src/transforms/host_tags/mod.rs
rayz: The part where it checks if the duration elapsed happens in `transform_buffer` which I wasn't sure how to test so I just set `context_resolver` and `host_tags` to `None` for that test. Happy to update if there is a better way to test the 'stop enrichment' part though 🤔 .

lib/saluki-components/src/transforms/host_tags/mod.rs
```diff
@@ -124,3 +129,44 @@ impl SynchronousTransform for HostTagsEnrichment {
...
+    use saluki_event::metric::Metric;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_host_tags_enrichment() {
+        let context_resolver = ContextResolverBuilder::from_name("host_tags")
+            .expect("resolver name is not empty")
+            .with_interner_capacity_bytes(NonZeroUsize::new(ByteSize::kib(64).as_u64() as usize).unwrap())
+            .build();
```
tobz, line 152: This can just be `let context_resolver = ContextResolverBuilder::for_tests().build();` since we don't care about the resolver name of the interner capacity, as the default capacity is more than enough for this test.

lib/saluki-components/src/transforms/host_tags/mod.rs
```diff
@@ -124,3 +129,44 @@ impl SynchronousTransform for HostTagsEnrichment {
...
+    };
+
+    use bytesize::ByteSize;
+    use saluki_context::{Context, ContextResolverBuilder};
+    use saluki_event::metric::Metric;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_host_tags_enrichment() {
```
tobz, line 148: We try to avoid stuttering, so no `test_` prefix in the function name... and maybe just call it something like `basic`?

lib/saluki-components/src/transforms/host_tags/mod.rs
```diff
@@ -124,3 +129,44 @@ impl SynchronousTransform for HostTagsEnrichment {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+
```
tobz, line 135: Stray empty line?

lib/saluki-components/src/transforms/host_tags/mod.rs
```diff
@@ -124,3 +129,44 @@ impl SynchronousTransform for HostTagsEnrichment {
...
+        let context_resolver = ContextResolverBuilder::from_name("host_tags")
+            .expect("resolver name is not empty")
+            .with_interner_capacity_bytes(NonZeroUsize::new(ByteSize::kib(64).as_u64() as usize).unwrap())
+            .build();
+
+        let mut host_tags_enrichment = HostTagsEnrichment {
+            start: Instant::now(),
+            context_resolver: Some(context_resolver),
+            expected_tags_duration: Duration::from_secs(30),
+            host_tags: Some(vec!["hosttag1".to_string(), "hosttag2".to_string()]),
```
tobz, line 158: It would be nice to lift these out from their inline definition, so we can then actually check that the enriched metric has these as the tags.

tobz (changes requested):
Some small nits.

lib/saluki-components/src/transforms/host_tags/mod.rs
tobz: I think that's fine overall.

The main two way to test this are: set the resolver/tags to `None` and call `enrich_metric` _or_ pause/advance time manually and call `transform_buffer`.

The "if elapsed > expected duration { ... }" bit is straightforward enough that I hesitate to worry too much about _directly_ testing it.

tobz (approved):

PR #598, opened 2025-04-09
enhancement(contexts): improve context resolver telemetry and slightly improve non-cached overhead

### LINKED ISSUE #571: Context resolver increments "resolved new context total" metric twice.
## Context

Currently, `ContextResolver` will increment the `context_resolver_resolved_new_context_total` metric twice when resolving a new context: once in `create_context`, and another time in `resolve_inner`.

This means that, simply, we're double counting new context creation. I think we should move all context-specific metric operations into `create_context` -- including incrementing `active_contexts` -- to avoid any confusion about who owns the responsibility of those metrics.

tobz:
## Summary

This PR primarily improves the telemetry of the context resolver, in terms of getting telemetry updates out of the hot path for resolving, as well as resolving some overreporting.

## Change Type

- [x] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Utilized existing unit tests and ran correctness test harness.

## References

Closes #571.

### DIFF
--- a/lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
+++ b/lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
@@ -305,8 +305,8 @@ mod tests {
         assert_eq!(context.tags().len(), expected_tags.len(), "unexpected number of tags");
     }
 
-    #[test]
-    fn test_mapper_wildcard_simple() {
+    #[tokio::test]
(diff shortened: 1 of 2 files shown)


tobz:
Looking at the SMP data, this doesn't actually appear to be an improvement as I thought it might be.

Going to take a closer look at the profiling data but this might be a bust.

rayz (approved):

Issue #395, opened 2024-12-16
Add support for blanket enriching metrics ingested through DogStatsD with additional tags.


tobz:
## Context

The Datadog Agent supports adding a blanket set of tags metrics ingested through DogStatsD via the `dogstatsd_tags` configuration setting.

PR #605, opened 2025-04-10
chore: add support for enriching metrics with tags from the `dogstatsd_tags` configuration setting

rayz:
## Summary

This pr introduces a new wrapper `TagsWrapper` with an iterator that iterates over `RawTagsIter` and also the additional tags provided by the `dogstatsd_tags` setting.
## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Existing unit tests and also a new one.

## References

- Closes: #395


lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -394,6 +400,7 @@ pub struct DogStatsD {
     codec: DogstatsdCodec,
     context_resolvers: ContextResolvers,
     pre_filters: Arc<Vec<Box<dyn Filter + Send + Sync>>>,
+    additional_tags: Vec<String>,
```
tobz, line 403: We should make this, and the subsequent fields for the listener/handler context, use `Arc<[String]>` instead, and then we can save on the allocations when cloning for each handler.

`Arc<[T]> implements `From<Vec<T>>` so it should be as simple as `additional_tags: self.additional_tags.clone().into()` up on line 359.

lib/saluki-context/src/tags/raw.rs
```diff
@@ -130,3 +131,64 @@ fn split_at_delimiter(input: &str, delimiter: u8) -> Option<(&str, &str)> {
         }
     }
 }
+
+/// A wrapper over a `RawTags` iterator that allows for additional tags to be appended to the end of the iterator.
+#[derive(Clone)]
+pub struct TagsWrapper<'a> {
```
tobz, line 137: Can we call this `ChainedRawTags`?

The doc comment is fine, but I'd just like to see the type name to slightly more descriptive.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -837,7 +850,7 @@ async fn drive_stream(
 
 fn handle_frame(
     frame: &[u8], codec: &DogstatsdCodec, context_resolvers: &mut ContextResolvers, source_metrics: &Metrics,
-    peer_addr: &ConnectionAddress, enabled_filter: EnablePayloadsFilter,
+    peer_addr: &ConnectionAddress, enabled_filter: EnablePayloadsFilter, additional_tags: &Arc<[String]>,
```
tobz, line 853: I'd make this one be `&[String]` as well, since we don't need to know, at this point, that it's actually an `Arc`.

tobz (changes requested):

PR #610, opened 2025-04-11
chore(chained): require name for each subtransform to improve internal telemetry

### LINKED ISSUE #602: Update Chained transform to require short name for each synchronous transform that is configured.
## Context

Currently, the Chained transform allows specifying multiple synchronous "subtransforms" that will be applied in the order that they're configured. We indicate this in the resulting internal telemetry by using the form of `subtransforms_<id>`, where `id` is the zero-based index of the order of the subtransforms... so `subtransform_0` is the first subtransform, and so on.

However, when looking at the metrics, this becomes very opaque, as you need to already know what subtransform is configured in a certain "slot"... and that value could change between releases.

We should update the Chained transform to actually require a short name for each subtransform, such that we can concatenate in the identifier we use. Instead of `subtransform_0`, we might end up with `0_host_enrichment`, and so on. We want to keep the _order_ they're defined in, since ordering could be relevant to understanding how the behavior of one subtransform influences the following subtransform(s)... but having the name alone immediately allows operators to understand what subtransform the metrics are referring to.

tobz:
## Summary

This PR updates the Chained transform to use a new string identifier format for subtransforms -- `<index>_<name>` -- to improve the resulting internal telemetry that is generated for each subtransform.

Callers must now provide a name when adding a subtransform, and this name is sanitized to lowercase ASCII alphanumeric (and underscores), and then used to form the string identifier. This means that instead of seeing metrics with a component ID tag like `topology.components.transforms.enrich.subtransform_0`, we now get something like `topology.components.transforms.enrich.0_host_enrichment`. The `host_enrichment` portion can be trivially searched for in the codebase, and in most cases, it will likely be indicative enough on its own without the need to search.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Built and run ADP locally, and observed the change as part of internal telemetry:

```
adp__group_allocated_bytes_total{group_id="topology.components.transforms.enrich.1_dogstatsd_mapper"} 0
adp__group_allocated_bytes_total{group_id="topology.components.transforms.enrich.0_host_enrichment"} 0
...
```

## References

Closes #602.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3120,6 +3120,7 @@ dependencies = [
  "pin-project",
  "protobuf",
  "regex",
+ "saluki-common",
  "saluki-config",
  "saluki-context",
  "saluki-core",
--- a/bin/agent-data-plane/src/main.rs
+++ b/bin/agent-data-plane/src/main.rs
@@ -198,13 +198,13 @@ async fn create_topology(
     let host_enrichment_config = HostEnrichmentConfiguration::from_environment_provider(env_provider.clone());
     let dsd_mapper_config = DogstatsDMapperConfiguration::from_configuration(configuration)?;
     let mut enrich_config = ChainedConfiguration::default()
-        .with_transform_builder(host_enrichment_config)
-        .with_transform_builder(dsd_mapper_config);
+        .with_transform_builder("host_enrichment", host_enrichment_config)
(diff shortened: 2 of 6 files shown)


rayz (approved):
lgtm

PR #611, opened 2025-04-14
Create PoC checks agent based on sopell/checks-cont PR

GustavoCaso:
## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance


lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -0,0 +1,712 @@
...
+        });
+
+        tokio::pin!(check_shutdown);
+
+        loop {
+            select! {
+                Some(check_metric) = check_metrics_rx.recv() => {
+                    info!("Received check metric: {:?}", check_metric);
+                    let mut event_buffer = context.event_buffer_pool().acquire().await;
+                    let event: Event = check_metric.try_into().expect("can't convert");
```
scottopell, line 494: I believe this is a starting point for the breadcrumb of "what metric types get translated to what". This `try_into` `Event` is doing a lot of work.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -0,0 +1,712 @@
...
+        // Convert Vec<String> to Vec<Tag>
+        let tags: Vec<Tag> = self.tags.into_iter().map(Tag::from).collect();
+
+        // Convert Vec<Tag> to TagSet
+        let tagset: TagSet = TagSet::from_iter(tags);
+
+        let context = Context::from_parts(self.name, tagset);
+        let metadata = MetricMetadata::default();
+
+        match self.metric_type {
```
scottopell, line 686: This is where the "magic" from the `try_into` happens, we just ignore any metric type that isn't gauge/counter.

lib/saluki-components/src/sources/checks/python_exposed_modules.rs
```diff
@@ -0,0 +1,160 @@
...
+        name,
+        metric_type: mtype.into(),
+        value,
+        tags,
+    };
+
+    match module.getattr("SUBMISSION_QUEUE") {
+        Ok(py_item) => match py_item.extract::<Py<python_scheduler::PythonSenderHolder>>() {
+            Ok(q) => {
+                let res = pyo3::Python::with_gil(|py| q.bind_borrowed(py).borrow_mut().sender.clone());
```
scottopell, line 35: I think this is a bit convoluted, if you understood the pyo3 borrow system better you could end up with some cleaner code. IIRC @GeorgeHahn actually had a suggestion here at one point

lib/saluki-components/src/sources/checks/python_exposed_modules.rs
```diff
@@ -0,0 +1,160 @@
...
+        20 => tracing::event!(tracing::Level::INFO, "Python Log: {}", message),
+        10 => tracing::event!(tracing::Level::DEBUG, "Python Log: {}", message),
+        7 => tracing::event!(tracing::Level::TRACE, "Python Log: {}", message),
+        _ => tracing::event!(tracing::Level::TRACE, "Python Log: {}", message),
+    };
+}
+
+#[pyfunction]
+fn set_check_metadata(check_id: String, name: String, value: String) {
+    debug!("Called set_check_metadata({}, {}, {})", check_id, name, value);
```
scottopell, line 138: I never dug into this to see what this check_metadata is actually used for/by, this could be a tricky one with race conditions around checks being scheduled/descheduled.

lib/saluki-components/src/sources/checks/python_scheduler.rs
```diff
@@ -0,0 +1,486 @@
...
+    }
+
+    fn possible_import_paths_for_module(&self, module_name: &str) -> Vec<String> {
+        vec![module_name.to_string(), format!("datadog_checks.{}", module_name)]
+    }
+
+    fn register_check_impl(&mut self, check: &RunnableCheckRequest) -> Result<CheckHandle, GenericError> {
+        let check_module_name = &check.check_request.name;
+        // if there is a specific source, then this will populate into locals and can be found
+        if let Some(py_source) = &check.check_source_code {
```
scottopell, line 143: as noted elsewhere, this only exists for tests at this point, "production" usage of checks do not ship python source code, its always a module loaded

lib/saluki-components/src/sources/checks/python_scheduler.rs
```diff
@@ -0,0 +1,486 @@
...
+            num_instances = check.check_request.instances.len()
+        );
+        for (idx, instance) in check.check_request.instances.iter().enumerate() {
+            let instance = instance.clone();
+            let init_config = check.check_request.init_config.clone();
+
+            let check_handle = check_handle.clone();
+            trace!("Spawning task for check instance {idx}");
+            self.tlm.check_instances_started.increment(1);
+            let handle = tokio::task::spawn(async move {
```
scottopell, line 320: This is where we _actually_ schedule a check to run. We simply use tokio to schedule a lightweight "task" (similar to goroutine) and avoid any interaction with OS threads.

lib/saluki-components/src/sources/checks/python_scheduler.rs
```diff
@@ -0,0 +1,486 @@
...
+                let mut interval =
+                    tokio::time::interval(Duration::from_millis(instance.min_collection_interval_ms().into()));
+                loop {
+                    interval.tick().await;
+                    // run check
+                    info!("Running check instance {idx}");
+                    pyo3::Python::with_gil(|py| {
```
scottopell, line 327: This will block until gil access is acquired, so this again abstracts any low-level concurrency/scheduling concerns and relies on tokio's scheduler.

lib/saluki-components/src/sources/checks/python_scheduler.rs
```diff
@@ -0,0 +1,486 @@
...
+                            .expect("Could not set name");
+                        kwargs
+                            .set_item("init_config", init_config.to_pydict(&py))
+                            .expect("could not set init_config");
+                        kwargs
+                            .set_item("instances", instance_list)
+                            .expect("could not set instance list");
+
+                        // TODO should this be moved up outside of the loop?
+                        // I kind of think so, this makes a new check instance every tick
```
scottopell, line 343: Worth highlighting this 😅 , this is the kind of thing I mentioned that @ofek would have a much better grasp on

scottopell (commented):
Just leaving some comments for readers

lib/saluki-components/src/sources/checks/python_exposed_modules.rs
GeorgeHahn: My previous suggestion is lost to the sands of time, but looking at this with fresh eyes makes me wonder if the runtime GIL acquisition can be replaced with a function parameter. The docs [here](https://docs.rs/pyo3/latest/pyo3/marker/struct.Python.html#obtaining-a-python-token) make it sound like that's possible and faster.

PR #616, opened 2025-04-14
experiment: track per-task poll count/duration to hunt for slow tasks

tobz:
## Summary

In many cases, understanding the polling behavior of a task -- the top-level future that is spawned on the asynchronous runtime -- is crucial to understanding where scheduling tail latency originates from. In some cases, when a single task -- despite having multiple worker threads -- takes a long time to complete being polled, it can introduce scheduling latency for other waiting tasks, increasing how long it takes before they are polled and can continue to be driven to completion.

This PR adds support for tracking the poll behavior of tasks by adding support to wrap top-level futures with instrumentation around how many times they are polled, as well as tracking the duration of each poll, and emitting it as internal telemetry.

We've done a bunch of work in this PR to support this:

- added `InstrumentedTask<F>`, which is a newtype wrapper over an arbitrary `Future`, which tracks its poll count and poll durations tied to a user-supplied task name
- added `TaskInstrumented`, a helper trait for exposing a way to wrap `Future`s with `InstrumentedTask<F>` in a way similar to how we instrument futures for `tracing` or `memory_accounting`
- moved all custom spawn functions from `saluki_core::task` to `saluki_common::task`, and added new variants for specifying a name for a given task, as well as augmented the original versions to specify a name that maps to the location where the future was spawned, similar to the logic used by `tokio-console`
- updated nearly all relevant usages of the existing traced spawn functions, or bare `tokio::spawn` usages, to use the new `spawn_traced_named` functions

For every task future we spawn this way, we emit a new histogram metric, `runtime_task.poll_duration_seconds`, which tracks the poll duration of the future in seconds. This metric has a `task_name` tag, which is either the user-specified task name, or a default name of `file-<file path>@<line number>-<column>`, which corresponds to the location that the relevant `spawn_traced_named` function was called at.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

N/A

## References

N/A


tobz:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/saluki/616?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#616** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/616?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/saluki/616?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#619** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/619?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/saluki-components/src/destinations/prometheus/mod.rs
```diff
@@ -334,13 +334,11 @@ fn write_metrics(
...
-                        write!(payload_buffer, "{}_bucket{{{}", &prom_context.metric_name, tags_buffer).unwrap();
-                        if !tags_buffer.is_empty() {
-                            payload_buffer.push(',');
-                        }
-                        writeln!(payload_buffer, "le=\"{}\"}} {}", upper_bound_str, count).unwrap();
+                    write!(payload_buffer, "{}_bucket{{{}", &prom_context.metric_name, tags_buffer).unwrap();
+                    if !tags_buffer.is_empty() {
+                        payload_buffer.push(',');
                     }
+                    writeln!(payload_buffer, "le=\"{}\"}} {}", upper_bound_str, count).unwrap();
```
tobz, line 341: I had originally done the change to not show buckets with a zero count to clean up the output, but realized it just sort of hides the necessary data to show which buckets things actually fall between.

lib/saluki-core/src/observability/metrics.rs
```diff
@@ -154,6 +155,8 @@ async fn flush_metrics(flush_interval: Duration) {
 
     let state = RECEIVER_STATE.get().expect("metrics receiver should be set");
 
+    let mut histogram_samples = Vec::<f64>::new();
```
tobz, line 158: Just a small optimization to reuse the buffer for holding histogram samples instead of throwing it away each time.

lib/saluki-components/src/destinations/datadog/metrics/mod.rs
rayz, line 1: https://github.com/DataDog/saluki/blob/50391297a4fc1a43339f8d5f70da1a89bdf3e4d2/lib/saluki-components/src/destinations/datadog/metrics/mod.rs#L254

Should this be `spawn_traced_named` too? When would we want to use `spawn_traced` over `spawn_traced_name`?

rayz (approved):

lib/saluki-components/src/destinations/datadog/metrics/mod.rs
tobz: Generally speaking, we always give each spawned task a name, whether `spawn_traced` or `spawn_traced_named` is used. For `spawn_traced`, we use a default name based on where `spawn_traced` is called from. In this case, it would be something like... `task_name:file-lib/saluki-components/src/destinations/datadog/metrics/mod.rs@254`.

So `spawn_traced_named` should be used when we want to give a human-friendly name... which, to be fair, should be "always".

tobz:
### Merge activity

* **Apr 21, 8:58 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/DataDog/saluki/616).

PR #618, opened 2025-04-15
[AGENTRUN-221] chore(checks-agent): create initial `checks-agent` binary

GustavoCaso:
## Summary

This PR introduces the new agent binary `checks-agent`. During the last quarter, the Agent Runtimes worked on extracting the checks outside of the main Agent. This effort was part of an exploration of finding the next Agent architecture that allows us to scale without increasing our customers' operational cost. The end result was that the overhead of extracting separate parts of the Agent into Go processes was not feasible due to the overhead. More importantly, the lack of fine-grained control within Go (garbage-collected language). You can read more about it [here](https://docs.google.com/document/d/1L9Sftu_cPDDYWazBNpR87zqsqhLY3zKp32byjPjDMGg/edit?pli=1&tab=t.3miccnxavpy2). 

We are going to explore the possibility of using Rust and the Saluki framework to extract functionality from the main Agent. 

This PR creates a simple agent `checks-agent` that can parse configuration, and has a simple topology. From this point, we would start benchmarking against the Go Agent (version) and continue to add check features. By the end of the quarter, we should have a clear understanding of whether using Rust and Saluki is the right choice for the next Agent architecture. 

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

### DIFF
--- a/.ci/templates/generate-result-links-markdown.sh
+++ b/.ci/templates/generate-result-links-markdown.sh
@@ -83,7 +83,7 @@ dsd_only_experiments=$(comm -13 adp-experiments dsd-experiments)
 common_experiments=$(comm -12 adp-experiments dsd-experiments)
 
 # Write out our table of links, doing common experiments first, then ADP-only, then DSD-only.
-echo "## Experiment Result Links"
+echo "## ADP Experiment Result Links"
 echo ""
 echo "| experiment | link(s) |"
 echo "|------------|---------|"
@@ -109,6 +109,11 @@ for experiment in $dsd_only_experiments; do
 
     echo "| $experiment (DSD only) | \\[[Profiling (DSD)]($dsd_continuous_profiler_url)\\] \\[[SMP Dashboard]($adp_smp_dashboard_url)\\] |"
 done
+
+echo "## Checks Agent Experiment Result Links"
+echo ""
+echo "| experiment | link(s) |"
+echo "|------------|---------|"
 
 checks_continuous_profiler_url=$(get_continuous_profiler_url "$checks_run_id" "$checks_start_time" "$checks_end_time" "quality_gates_idle_rss")
 checks_smp_dashboard_url=$(get_adp_smp_dashboard_url "$checks_run_id" "$dsd_run_id" "$checks_start_time" "$checks_end_time" "quality_gates_idle_rss")
--- a/.ci/templates/generate-result-links-markdown.sh
+++ b/.ci/templates/generate-result-links-markdown.sh
@@ -42,6 +42,9 @@ ensure_file_exists "adp_job_end_time"
 ensure_file_exists "dsd_run_id"
 ensure_file_exists "dsd_job_start_time"
 ensure_file_exists "dsd_job_end_time"
+ensure_file_exists "checks_run_id"
+ensure_file_exists "checks_job_start_time"
+ensure_file_exists "checks_job_end_time"
 
 adp_run_id=$(cat adp_run_id)
 adp_start_time=$(cat adp_job_start_time)
@@ -49,6 +52,9 @@ adp_end_time=$(cat adp_job_end_time)
 dsd_run_id=$(cat dsd_run_id)
 dsd_start_time=$(cat dsd_job_start_time)
 dsd_end_time=$(cat dsd_job_end_time)
+checks_run_id=$(cat checks_run_id)
+checks_start_time=$(cat checks_job_start_time)
+checks_end_time=$(cat checks_job_end_time)
 
 # Load the job start/end times and figure out which job started first and which job ended last, which we'll use as the
 # start/end time for our dashboard, which shows both sides -- ADP and DSD -- in the same pane of glass.
@@ -103,3 +109,8 @@ for experiment in $dsd_only_experiments; do
 
     echo "| $experiment (DSD only) | \\[[Profiling (DSD)]($dsd_continuous_profiler_url)\\] \\[[SMP Dashboard]($adp_smp_dashboard_url)\\] |"
 done
+
+checks_continuous_profiler_url=$(get_continuous_profiler_url "$checks_run_id" "$checks_start_time" "$checks_end_time" "quality_gates_idle_rss")
+checks_smp_dashboard_url=$(get_adp_smp_dashboard_url "$checks_run_id" "$dsd_run_id" "$checks_start_time" "$checks_end_time" "quality_gates_idle_rss")
+
+echo "| quality_gates_idle_rss | \\[[Profiling (Checks Agent)]($checks_continuous_profiler_url)\\] \\[[SMP Dashboard]($checks_smp_dashboard_url)\\] |"
--- a/.gitlab/benchmark.yml
+++ b/.gitlab/benchmark.yml
@@ -401,6 +401,7 @@ comment-smp-result-links:
   needs:
     - run-benchmarks-adp
     - run-benchmarks-dsd
+    - run-benchmarks-checks-agent
   image: "${SALUKI_SMP_CI_IMAGE}"
   before_script:
     - *setup-smp-env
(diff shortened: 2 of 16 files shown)


lib/saluki-components/src/sources/heartbeat/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+use memory_accounting::{MemoryBounds, MemoryBoundsBuilder};
+use saluki_context::Context;
+use saluki_core::components::{sources::*, ComponentContext};
+use saluki_core::topology::OutputDefinition;
+use saluki_error::GenericError;
+use saluki_event::{metric::Metric, DataType, Event};
+use tokio::{select, time::interval};
+use tracing::{debug, error};
+
+/// Configuration for a source that doesn't produce any events or optionally emits heartbeat metrics.
```
tobz, line 13:
```suggestion
/// Heartbeat source.
///
/// Emits a "heartbeat" metric on a configurable interval.
```

Since users generally only ever interact with the component configuration type (when building the topology), we strive to put the human-friendly name/description of a component in the doc comments of that configuration type itself.

lib/saluki-components/src/sources/heartbeat/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+        loop {
+            select! {
+                _ = &mut global_shutdown => {
+                    debug!("Received shutdown signal.");
+                    break;
+                },
+                _ = health.live() => continue,
+                _ = tick_interval.tick() => {
+                    // Create a simple heartbeat metric
+                    let metric_context = Context::from_static_name("none_source.heartbeat");
```
tobz, line 61:
```suggestion
                    let metric_context = Context::from_static_name("heartbeat");
```

tobz: Kind of an anti-pattern to put the name of the source in the metric.

test/smp/regression/checks-agent/cases/quality_gates_idle_rss/experiment.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+    # Enable internal telemetry endpoint.
+    DD_TELEMETRY_ENABLED: "true"
+    DD_PROMETHEUS_LISTEN_ADDR: tcp://127.0.0.1:5102
+
+    DD_PRINT_JEMALLOC_STATS: "true"
+
+checks:
+  - name: memory_usage
+    description: "Acceptable upper bound on the memory used by ADP when idle."
```
tobz, line 34:
```suggestion
    description: "Acceptable upper bound on the memory used by Checks Agent when idle."
```

tobz (changes requested):
Bunch of little nits/suggestions.

.gitlab/benchmark.yml
```diff
@@ -5,9 +5,13 @@
   - export SMP_ECR_HOST="${SMP_ACCOUNT_ID}.dkr.ecr.us-west-2.amazonaws.com"
   - git fetch --verbose --deepen=999 origin ${CI_COMMIT_BRANCH} main
   - export BASELINE_SALUKI_SHA=$(git merge-base origin/main origin/${CI_COMMIT_BRANCH})
+  - export BASELINE_CHECKS_AGENT_SHA=$(git merge-base origin/poc-checks-agent-part-1 origin/${CI_COMMIT_BRANCH})
   - export COMPARISON_SALUKI_SHA="${CI_COMMIT_SHA}"
+  - export COMPARISON_CHECKS_AGENT_SHA="${CI_COMMIT_SHA}"
```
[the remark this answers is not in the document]
GustavoCaso: Since I will use stack PRs, and the base is `rigin/poc-checks-agent-part-1,` I thought it would be good to have separate ENV vars. If we merge the work in the main, I agree that they will become redundant.

Would you be okay with leaving for now until we decide whether to merge this PR?

docker/Dockerfile.checks-agent
```diff
@@ -0,0 +1,76 @@
+ARG BUILD_IMAGE
```
tobz, line 1: Not something to address in this PR, but mostly so I don't forget: it strikes me that we should collapse this and `Dockerfile.agent-data-plane` into a single Dockerfile that takes an argument that we pass to `cargo build` for which binary to build, since everything else is identical.

lib/saluki-components/src/sources/heartbeat/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+impl Default for HeartbeatConfiguration {
+    fn default() -> Self {
+        Self {
+            heartbeat_interval_secs: 10,
+        }
+    }
+}
+
+/// Source that doesn't produce any events or optionally emits heartbeat metrics.
+struct HeartbeatSource {
```
tobz, line 31:
```suggestion
struct Heartbeat {
```

(Just making this match how other components define their primary type.)

lib/saluki-components/src/sources/heartbeat/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+                _ = tick_interval.tick() => {
+                    // Create a simple heartbeat metric
+                    let metric_context = Context::from_static_name("none_source.heartbeat");
+                    let metric = Metric::gauge(metric_context, 1.0);
+
+                    let events = vec![Event::Metric(metric)];
+                    if let Err(e) = context.forwarder().forward(events).await {
+                        error!(error = %e, "Failed to forward heartbeat event.");
+                    } else {
+                        debug!("Emitted heartbeat metric.");
```
tobz, line 61:
```suggestion
                    let mut buffered_forwarder = context.forwarder().buffered().expect("default output must always exist");
                    if let Err(e) = buffered_forwarder.push(Event::Metric(metric)).await {
                        error!(error = %e, "Failed to forward event.");
                    } else {
                        if let Err(e) = buffered_forwarder.flush().await {
                            error!(error = %e, "Failed to forward events.");
                        } else {
                            debug!("Emitted heartbeat metric.");
                        }
                    }
```

This will be a little more efficient since we skip allocating the `Vec<Event>` and just use `BufferedForwarder<'a, _>` to directly write the event into an event buffer.

(The code is a little verbose, so I opened https://github.com/DataDog/saluki/issues/621 to provide an easier approach.)

tobz (changes requested):

docker/Dockerfile.checks-agent
GustavoCaso: I can take care of that, once we decide to merge this to main

lib/saluki-components/src/sources/heartbeat/mod.rs
GustavoCaso: Awesome. Thanks for the tip. Still getting familiar with Rust and any feedback is very welcome

.ci/templates/generate-result-links-markdown.sh
```diff
@@ -103,3 +109,8 @@ for experiment in $dsd_only_experiments; do
 
     echo "| $experiment (DSD only) | \\[[Profiling (DSD)]($dsd_continuous_profiler_url)\\] \\[[SMP Dashboard]($adp_smp_dashboard_url)\\] |"
 done
+
+checks_continuous_profiler_url=$(get_continuous_profiler_url "$checks_run_id" "$checks_start_time" "$checks_end_time" "quality_gates_idle_rss")
+checks_smp_dashboard_url=$(get_adp_smp_dashboard_url "$checks_run_id" "$dsd_run_id" "$checks_start_time" "$checks_end_time" "quality_gates_idle_rss")
+
+echo "| quality_gates_idle_rss | \\[[Profiling (Checks Agent)]($checks_continuous_profiler_url)\\] \\[[SMP Dashboard]($checks_smp_dashboard_url)\\] |"
```
tobz, line 116: Maybe we change the table header (on line 86) to say `ADP Experiment Result Links` and then down here, we `echo` a similar table header that says `Checks Agent Experiment Result Links` since right now we'll have two rows with an identical `experiment` value of `quality_gates_idle_rss`, only differing in terms of the link text shown.

GustavoCaso: https://github.com/DataDog/saluki/pull/618/commits/6a26a3b69e216ade872c018cc373be7b4b6ab2c2

tobz (approved):

PR #628, opened 2025-04-18
enhancement: move internal telemetry/APIs/etc to dedicated runtimes for isolation

tobz:
## Summary

This PR reorganizes some of the "internal", non-data pipeline-oriented aspects of the ADP binary to better isolate their behavior from one another. This is meant to ensure that issues/degradation in the primary topology don't negatively impact the control plane, such as by preventing internal telemetry from being scraped due to topology stalls, as well as isolate the primary topology from control plane computation, such as high duration task polls which can drive up tail latency for sensitive tasks (like DogStatsD I/O handlers).

Specifically, we now isolate internal processes -- (un)privileged APIs, health registry, internal telemetry pipeline, etc -- within dedicated current-thread async runtimes. This lets us keep our existing code and utilize existing paradigms, but simply isolates the related work/tasks to their own runtime executor threads.

Notable changes:

- `bin/agent-data-plane/src/internal` now holds all internal-y things, and `bin/agent-data-plane/src/api` has been folded into it
- topologies are now named as we now run two distinct topologies (primary and internal telemetry) by default (so `topology.primary.components.sources.dsd_in` instead of `topology.components.sources.dsd_in`)
- the global event buffer pool sizing can now be configured on `TopologyBlueprint` (necessary for being able to build the internal telemetry topology while not having a firm upper bound of like 80MB)
- Tokio runtime metrics are now tagged by a `runtime_id` to allow for differentiating between the primary runtime and internal control plane runtimes
- we print the component bounds whether the bounds are satisfied or not, allowing an operator to actually understand _why_ bounds cannot be met

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Tested locally to ensure that APIs are still accessible and that internal telemetry can still be accessed/reported.

## References

N/A

### DIFF
--- a/bin/agent-data-plane/src/internal/control_plane.rs
+++ b/bin/agent-data-plane/src/internal/control_plane.rs
@@ -1,23 +1,61 @@
 use std::future::pending;
 
-use saluki_app::api::APIBuilder;
+use memory_accounting::ComponentRegistry;
+use saluki_app::{api::APIBuilder, memory::MemoryProfilingAPIHandler, prelude::acquire_logging_api_handler};
 use saluki_common::task::spawn_traced_named;
 use saluki_config::GenericConfiguration;
-use saluki_core::state::reflector::Reflector;
 use saluki_error::{generic_error, ErrorContext as _, GenericError};
+use saluki_health::HealthRegistry;
 use saluki_io::net::ListenAddress;
 use tracing::{error, info};
 
-mod remote_agent;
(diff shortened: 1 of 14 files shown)


rayz (approved):

tobz:
### Merge activity

* **Apr 21, 8:58 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/DataDog/saluki/628).

PR #629, opened 2025-04-21
experiment: Erlang/OTP-style supervisor system

tobz:
## Summary

Work in progress.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance


tobz:
* **#629** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/629?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/saluki/629?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/saluki-core/src/runtime/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+//! executing on a thread pool. We refer to these as **processes.** In other systems, these might be called _actors_,
+//! _tasks_, _fibers_, _virtual threads_, _goroutines_, or something else. Processes are lightweight and able to be
+//! (generally) created and destroyed cheaply.
+//!
+//! Processes have a few key attributes and invariants:
+//!
+//! - every process is a future that runs as an independent asynchronous task on a Tokio runtime
+//! - every process has a unique numerical identifier and a semi-unique name
+//!
+//! Processes cannot run by themselves, however. They must be _supervised_.
```
blt, line 25: Confirming, unlike Erlang processes saluki-core processes do not have a mailbox / input queue associated for message passing?

lib/saluki-core/src/runtime/process.rs
```diff
@@ -0,0 +1,276 @@
...
+/// Process name.
+///
+/// A human-readable name for a process that only contains alphanumeric characters, underscores, and periods.
+///
+/// Process names are scoped, such that the resulting process name is nested. For example, if a supervisor has a name of
+/// `topology_sup`, and a child process is added to that supervisor with a name of `worker`, the resulting process name
+/// for the child process will be `topology_sup.worker`. Process names can be arbitrarily nested in this way.
+///
+/// Process names will be sanitized if they contain invalid characters, such as hyphens or spaces. Invalid characters
+/// will be replaced with underscores.
```
blt, line 45: I'm not opposed, just curious why we don't reject invalid names outright? Haven't confirmed yet, it's possible the intention here is for process names to arrive via user config, in which case this does make some sense.

lib/saluki-core/src/runtime/mod.rs
blt: Confirmed later, no mailbox. Probably worth adding this to the documentation in case someone -- like myself -- comes along with an Erlang mental model inappropriately.

lib/saluki-core/src/runtime/supervisor.rs
```diff
@@ -0,0 +1,555 @@
...
+                _ = &mut shutdown => {
+                    debug!(supervisor_id = %self.supervisor_id, "Shutdown triggered, shutting down all child processes.");
+                    worker_state.shutdown_workers().await;
+                    break;
+                },
+                worker_task_result = worker_state.wait_for_next_worker() => match worker_task_result {
+                    // TODO: Erlang/OTP defaults to always trying to restart a process, even if it doesn't terminate due to a
+                    // legitimate failure. It does allow configuring this behavior on a per-process basis, however. We don't
+                    // support dynamically adding child processes, which is the only real use case I can think of for having
+                    // non-long-lived child processes... so I think for now, we're OK just always try to restart.
```
blt, line 301: +1

lib/saluki-core/examples/basic_supervisor.rs
```diff
@@ -0,0 +1,175 @@
+use std::time::Duration;
```
blt, line 1: I like how the API works. Probably worth developing a lint or something to confirm that processes actually check their shutdown signal properly. Shame we don't have the ability to preempt.

blt (approved):

lib/saluki-core/src/runtime/process.rs
tobz: Yeah, more or less. Process names will come, in part, from user-defined identifiers that might be provided at a layer above (or higher) the point where processes themselves are constructed, so we want to be more liberal in what we accept so that code doesn't have to bubble up that sort of constraint.

PR #633, opened 2025-04-21
fix(ci): pass proper env var for ADP binary 'full name' value

tobz:
## Summary

In CI, we use the `emit-build-metadata` Make target to generate the "build metadata" that gets bundled into the ADP binary: build data, short/full name of the binary, and so on. We were incorrectly passing the "full name" to the relevant Docker build step (`APP_FULLNAME` instead of `APP_FULL_NAME`), leading to the default value of `unknown` being used.

This manifests as showing a name of `unknown` when ADP connects to the Core Agent's "Remote Agent Registry", as well as a User Agent of `unknown/0.1.0` when sending outbound HTTP requests, and so on.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Will require waiting for CI to build an image that we can test locally.

## References

N/A

### DIFF
--- a/.gitlab/build.yml
+++ b/.gitlab/build.yml
@@ -20,7 +20,7 @@
       --build-arg APP_IMAGE=${GBI_BASE_IMAGE}
       --build-arg BUILD_PROFILE=${BUILD_PROFILE}
       --build-arg BUILD_FEATURES=${BUILD_FEATURES}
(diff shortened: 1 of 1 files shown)


rayz (approved):
Tested with the custom image and it now shows `agent-data-plane` as expected 👍

Issue #626, opened 2025-04-18
Implement support for restricting retry disk buffer usage when disk usage exceeds a configurable threshold.


tobz:
## Context

The Datadog Agent's disk buffering capabilities include support for not only limiting the usage by a fixed amount, but also in terms of avoiding causing disk usage overall to exceed a configurable threshold. This means that, for example, the Agent can be configured to not write additional files to disk if the underlying volume/mountpoint is, for example, 95% full.

This is a safety feature to ensure that the Agent is a model citizen on a given system: even if it's still obeying its own capacity limits, we don't want to be the process that causes a disk to become entirely full, as this can have negative consequences both for the Agent and other processes.

PR #642, opened 2025-04-23
enhancement: implement support for restricting retry disk buffer usage when disk usage exceeds a configurable threshold

rayz:
## Summary

This pr adds a configuration option for `forwarder_storage_max_disk_ratio`. 

This is used to compute the maximum bytes that are allowed to be stored on disk which follows the[ algorithm used in the datadog-agent here](https://github.com/DataDog/datadog-agent/blob/main/comp/forwarder/defaultforwarder/internal/retry/disk_usage_limit.go#L41-L50)

The configuration options for `forwarder_storage_path` and `forwarder_storage_max_size_in_bytes` were missing so we added those as well.

This pr also enable disk persistence if `forwarder_storage_max_size_in_bytes > 0` 
## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

New unit test
## References

- Closes: #626

### DIFF
--- a/lib/saluki-components/src/destinations/datadog/common/io.rs
+++ b/lib/saluki-components/src/destinations/datadog/common/io.rs
@@ -268,7 +268,10 @@ async fn run_endpoint_io_loop<S, B>(
                 ))),
             )
             .await
-            .expect("Failed to initialize disk persistence for retry queue.");
+            .unwrap_or_else(|e| {
+                error!(endpoint_url, error = %e, "Failed to initialize disk persistence for retry queue. Transactions will not be persisted.");
+                RetryQueue::new(endpoint_url.clone(), config.retry().queue_max_size_bytes())
+            });
     }
     let mut pending_txns = PendingTransactions::new(config.endpoint_buffer_size(), retry_queue, txnq_telemetry);
 
--- a/lib/saluki-components/src/destinations/datadog/common/io.rs
+++ b/lib/saluki-components/src/destinations/datadog/common/io.rs
@@ -1,4 +1,4 @@
-use std::{collections::VecDeque, error::Error as _, sync::Arc, time::Duration};
+use std::{collections::VecDeque, error::Error as _, path::PathBuf, sync::Arc, time::Duration};
 
 use futures::FutureExt as _;
 use http::{Request, Uri};
@@ -14,7 +14,7 @@ use saluki_io::{
         client::http::HttpClient,
         util::{
             middleware::{RetryCircuitBreakerError, RetryCircuitBreakerLayer},
-            retry::{PushResult, RetryQueue, Retryable},
+            retry::{DiskUsageRetrieverImpl, PushResult, RetryQueue, Retryable},
         },
     },
 };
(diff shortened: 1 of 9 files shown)


lib/saluki-components/src/destinations/datadog/common/retry.rs
```diff
@@ -90,6 +130,24 @@ impl RetryConfiguration {
...
+    }
+
+    /// Returns the path to the directory where the retry queue will be stored on disk.
+    #[allow(unused)]
+    pub fn storage_path(&self) -> &str {
+        &self.storage_path
+    }
+
+    /// Returns the maximum disk usage ratio for storing transactions on disk.
+    #[allow(unused)]
```
rayz, line 146: Looks like we don't actually have `with_disk_persistence` in use except for in tests (?) These unused functions would be used in combination with `with_disk_persistence` [here](https://github.com/DataDog/saluki/blob/9d02818eee138de7feb1f4a84b74a33cb9b3742f/lib/saluki-components/src/destinations/datadog/common/io.rs#L257)

lib/saluki-io/src/net/util/retry/queue/persisted.rs
```diff
@@ -92,6 +98,18 @@ where
         self.entries.len()
     }
 
+    fn on_disk_bytes_limit(&self) -> Result<u64, GenericError> {
+        let total_space = total_space(&self.root_path)
```
rayz, line 102: `total_space` and `available_space` are still sync operations in `fs4` 🤔

lib/saluki-io/src/net/util/retry/queue/persisted.rs
```diff
@@ -92,6 +134,14 @@ where
         self.entries.len()
     }
 
+    fn on_disk_bytes_limit(&self) -> Result<u64, GenericError> {
+        let total_space = self.disk_usage_retriever.inner.total_space()? as f64;
+        let available_space = self.disk_usage_retriever.inner.available_space()? as f64;
```
tobz, line 139: Damn, I just realized after looking at `fs4` more closely that it only has natively async functions as an extension to types like `tokio::fs::File`. 😣

We'll want to change this function to be `async` and then wrap the entire body of the function with `tokio::task::spawn_blocking(async move { ... })`, and then await on the `JoinHandle<T>`.

This will also require us to make `DiskUsageRetrieverWrapper` cloneable (since we'll have to clone to send it to the blocking pool) and the contents it's wrapping.

Sorry about the initial confusion on `fs4` being natively async-friendly.

lib/saluki-io/src/net/util/retry/queue/persisted.rs
```diff
@@ -92,6 +134,14 @@ where
         self.entries.len()
     }
 
+    fn on_disk_bytes_limit(&self) -> Result<u64, GenericError> {
```
tobz, line 137: It would be nice to just give this a doc comment explaining that we use this function to determine if writing to disk would cause the underlying volume to end up with more than `self.storage_max_disk_ratio` in terms of used space, even if we haven't yet hit `self.max_on_disk_bytes`.

tobz (changes requested):

lib/saluki-io/src/net/util/retry/queue/persisted.rs
tobz: Ah, sorry, yeah... I was being a doofus when I said `async move`.

lib/saluki-components/src/destinations/datadog/common/io.rs
```diff
@@ -254,7 +254,22 @@ async fn run_endpoint_io_loop<S, B>(
...
+            .with_disk_persistence(
+                PathBuf::from(config.retry().storage_path()),
+                config.retry().storage_max_size_bytes(),
+                config.retry().storage_max_disk_ratio(),
+                Arc::new(DiskUsageRetrieverImpl::new(PathBuf::from(
+                    config.retry().storage_path(),
+                ))),
+            )
+            .await
+            .expect("Failed to initialize disk persistence for retry queue.");
```
tobz, line 271: Hmmm, the not so great part is that we would just be panicking and killing the endpoint I/O task here.

What does the Agent do in terms of error handling? Is there error handling during the _construction_ of the on-disk retry queue?

tobz: For context, what I'm trying to suss out is if we need to be pushing this error up the stack to end up killing the process in the same way we would when encountering any other sort of start-up error.

rayz: Looks like it continues but just without disk persistence. https://github.com/DataDog/datadog-agent/blob/main/comp/forwarder/defaultforwarder/internal/retry/transaction_retry_queue.go#L61-L67

Updated the code to remove the expect

tobz: Alright, neat, that makes things simpler. 👍🏻

tobz (approved):
LGTM.

PR #651, opened 2025-04-25
[AGENTRUN-271] chore(check-agent) Autodiscovery Provider

GustavoCaso:
## Summary

Added an Autodiscovery provider. 

Provided to separate implementations:
- Local 
- Remote 

And a boxed implementation that can use either the local or remote. That way, based on the setup we can use the core agent or the local file system to receive events. 

The API expose a single function `async fn subscribe(&mut self, sender: mpsc::Sender<AutodiscoveryEvent>) -> Result<(), Self::Error>;`

The caller can provide their queue and consume the event once the provider pushes information. 

### Update 29/04/2025

Refactor the AutodiscoveryProvider to avoid having to use `mut &self` receiver.
We used the interior immutability pattern. We use OnceCell to ensure the internal state is initialized just once.
Used `tokio::sync::broadcast` channel to instead of accepting a channel in the subscribe function we return a receiver.
That way, we no longer have to mutate the internal state to keep track of the different subscribers.
Also, we no longer need to implement our broadcast logic.

The final API is `async fn subscribe(&self) -> Receiver<AutodiscoveryEvent>;`

I planned on adding the `autodiscovery` as part of the `EnvProvider` trait, but decided against it, since ADP currently has no use for it. In the future, if we have any need for it, we can add it 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

I provided unit tests.


lib/saluki-env/src/autodiscovery/providers/remote_agent.rs
```diff
@@ -0,0 +1,128 @@
...
+            listener_started: false,
+        }
+    }
+
+    async fn start_background_listener(&mut self) -> Result<(), GenericError> {
+        if self.event_senders.lock().unwrap().is_empty() {
+            return Err(GenericError::msg(
+                "Cannot start background listener without subscribers",
+            ));
+        }
```
GustavoCaso, line 34: Since the call this function after pushing a sender, we can remove this check and avoid acquire the lock twice on the first `subscribe` call

lib/saluki-env/src/autodiscovery/providers/remote_agent.rs
```diff
@@ -0,0 +1,128 @@
...
+    {
+        let senders_guard = senders.lock().unwrap();
+        for (idx, sender) in senders_guard.iter().enumerate() {
+            match sender.try_send(event.clone()) {
+                Ok(_) => {}
+                Err(TrySendError::Closed(_)) => {
+                    to_remove.push(idx);
+                }
+                Err(TrySendError::Full(_)) => {
+                    warn!("Channel full - subscriber cannot keep up with autodiscovery events");
```
GustavoCaso, line 87: Can we do anything in this case?

lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -0,0 +1,401 @@
...
+        service_id: String::new(),
+        tagger_entity: String::new(),
+        cluster_check: false,
+        node_name: String::new(),
+        source: path.to_string_lossy().to_string(),
+        ignore_autodiscovery_tags: false,
+        metrics_excluded: false,
+        logs_excluded: false,
+        advanced_ad_identifiers: Vec::new(),
+    };
```
GustavoCaso, line 111: We could parse using a YAML parser if we think it would be useful :)

lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -0,0 +1,401 @@
...
+
+        // Scan for initial configurations
+        let mut local_known = self.known_configs.clone();
+        scan_and_emit_events(&self.search_paths, &mut local_known, &self.event_senders).await?;
+        self.known_configs = local_known;
+
+        // Start background monitoring if not already active
+        if !self.monitoring_active {
+            self.start_background_monitor(30)?; // Check every 30 seconds
+        }
```
GustavoCaso, line 260: The local implementation is not ideal in the sense that every subscription will trigger a broadcast to all the senders. We can probably change the implementation so `scan_and_emit_events` can accept a list of senders that way in the subscribe function we could pass a list with the newly added sender, then inside the `start_background_monitor` we can pass the full list of senders.

lib/saluki-env/src/autodiscovery/providers/remote_agent.rs
```diff
@@ -0,0 +1,128 @@
...
+                        for proto_config in response.configs {
+                            let config = Config::from(proto_config.clone());
+                            let event = AutodiscoveryEvent { config };
+
+                            broadcast_event(&event_senders, event).await;
+                        }
+                    }
+                    Err(status) => {
+                        error!("Error from autodiscovery stream: {}", status);
+                        break;
```
tobz, line 62: Using `break` here means the first error we get, even if transient/recoverable, will kill the task and halt any further AD events.

You should wrap the all of the contents of this async block in a `loop`, and add a small sleep before breaking, like so:

```rust
loop {
    let events = client.get_autodiscovery_stream();
  
    while let Some(result) = events.next().await {
        match result {
            Ok(response) => { /* response was good, process it */ },
            Err(status) => {
                warn!(?status, "Encountered error while listening for autodiscovery events.  Retrying in 1 second...");

                tokio::time::sleep(Duration::from_secs(1)).await;
                break;
            }
        }
    }
}
```

lib/saluki-env/src/autodiscovery/providers/remote_agent.rs
```diff
@@ -0,0 +1,128 @@
...
+
+            warn!("Autodiscovery stream closed");
+        });
+
+        self.listener_started = true;
+        Ok(())
+    }
+}
+
+async fn broadcast_event(senders: &Arc<Mutex<Vec<mpsc::Sender<AutodiscoveryEvent>>>>, event: AutodiscoveryEvent) {
```
tobz, line 75: Related to the questions below: we should just use `tokio::sync::broadcast`, which already solves this in a cleaner way.

lib/saluki-env/src/autodiscovery/mod.rs
```diff
@@ -0,0 +1,253 @@
...
+pub struct AutodiscoveryEvent {
+    /// Configuration payload
+    pub config: Config,
+}
+
+/// Provides autodiscovery functionality.
+///
+/// This trait is used to discover and monitor configuration files for checks.
+#[async_trait]
+pub trait AutoDiscoveryProvider {
```
tobz, line 156: `AutoDiscoveryProvider` -> `AutodiscoveryProvider`

tobz (changes requested):
Just doing a first pass here.

lib/saluki-env/src/autodiscovery/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+}
+
+/// Provides autodiscovery functionality.
+///
+/// This trait is used to discover and monitor configuration files for checks.
+#[async_trait]
+pub trait AutodiscoveryProvider {
+    /// Subscribe for autodiscovery events.
+    ///
+    /// This method returns a channel where autodiscovery events will be sent.
```
tobz, line 159: In general, we strive to avoid code comments about the arguments or return type of a method/function unless there's conditional logic or something that isn't obvious from the return type.

For example, if we were instead returning `Result<Receiver<AutodiscoveryEvent>, GenericError>`, then we might want to explain that in the normal case, the caller will get `Ok(Receiver<AutodiscoveryEvent>)`, and that if there's an err, they get `Err(e)` containing the error, and the common reason an error might be returned.

In this case, we _only_ ever return `Receiver<AutodiscoveryEvent>`, so the caller knows they're getting a receiver of some sort, with autodiscovery events. Very clear by itself already.

lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -0,0 +1,298 @@
...
+/// A local auto-discovery provider that uses the file system.
+pub struct LocalAutoDiscoveryProvider {
+    search_paths: Vec<PathBuf>,
+    sender: Sender<AutodiscoveryEvent>,
+    listener_init: OnceCell<()>,
+}
+
+impl LocalAutoDiscoveryProvider {
+    /// Creates a new `LocalAutoDiscoveryProvider` that will monitor the specified paths.
+    pub fn new<P: AsRef<std::path::Path>>(paths: Vec<P>) -> Self {
```
tobz, line 23: Import `Path` here so we aren't using the fully-qualified path to it.

lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -0,0 +1,298 @@
...
+                }
+                if !p.as_ref().is_dir() {
+                    warn!("Skipping path '{}', it is not a directory.", p.as_ref().display());
+                    return None;
+                }
+                Some(p.as_ref().to_path_buf())
+            })
+            .collect();
+
+        let (sender, _) = broadcast::channel::<AutodiscoveryEvent>(100);
```
tobz, line 39: We should lift up this `100` to a constant (for example, `const AD_STREAM_CAPACITY: usize = 100;`) just so that it's a little more discoverable when reading the file.

lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -0,0 +1,298 @@
...
+                    warn!("Error scanning for configurations: {}", e);
+                }
+            }
+        });
+    }
+}
+
+/// Parse a YAML file into a Config object
+async fn parse_config_file(path: &PathBuf) -> Result<(String, Config), GenericError> {
+    let content = fs::read_to_string(path).await?;
```
tobz, line 71: One nice part about `GenericError` is that we can bundle context in the errors, similar to the common Go pattern of building nested errors (e.g. `fmt.Errorf("failed to read configuration file: %v", err)`).

You do have to import a trait for it (`use saluki_error::ErrorContext as _;`), but then for any time you're dealing with a `Result<T, E>`, you can do something like this:

```rust
let content = fs::read_to_string(path).await
    .error_context("Failed to read local autodiscovery configuration file.")?;
```

We get a proper error chain _and_ a nice, usable error message to better inform the operator what went wrong.

lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -0,0 +1,298 @@
...
+        .to_string_lossy()
+        .to_string()
+        .replace(['/', '\\'], "_");
+
+    // Create a Config
+    let config = Config {
+        name: path.file_name().unwrap().to_string_lossy().to_string(),
+        event_type: EventType::Schedule,
+        init_config: content.as_bytes().to_vec(),
+        instances: Vec::new(), // Would parse from YAML in a real implementation
```
tobz, line 85: Is there any reason we shouldn't make it actually read from YAML as part of this PR?

lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -0,0 +1,298 @@
...
+        }
+    }
+}
+
+#[async_trait]
+impl AutodiscoveryProvider for LocalAutoDiscoveryProvider {
+    async fn subscribe(&self) -> Receiver<AutodiscoveryEvent> {
+        self.listener_init
+            .get_or_init(|| async {
+                self.start_background_monitor(30).await;
```
tobz, line 200: Similar comment here about lifting up this interval constant to a `const` for discoverability.

lib/saluki-env/src/autodiscovery/providers/remote_agent.rs
```diff
@@ -0,0 +1,84 @@
...
+
+        Self {
+            client,
+            sender,
+            listener_init: OnceCell::new(),
+        }
+    }
+
+    async fn start_background_listener(&self) {
+        debug!("Starting autodiscovery background listener");
```
tobz, line 32: General comment: we should always strive to use proper punctuation in log messages, so just adding a period to the end of this.

tobz (changes requested):
A bunch of little things here and there, and some questions about the structure of the data and how we're modeling it.

lib/saluki-env/src/autodiscovery/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+            EventType::Schedule
+        }
+    }
+}
+
+/// Kubernetes namespaced name
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct KubeNamespacedName {
+    /// Kubernetes resource name
+    pub name: String,
```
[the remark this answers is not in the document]
GustavoCaso: Changed all usage of `String` with `MetaString` 😄

lib/saluki-env/src/autodiscovery/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+            ignore_autodiscovery_tags: proto.ignore_autodiscovery_tags,
+            metrics_excluded: proto.metrics_excluded,
+            logs_excluded: proto.logs_excluded,
+        }
+    }
+}
+
+/// An autodiscovery event
+#[derive(Debug, Clone)]
+pub struct AutodiscoveryEvent {
```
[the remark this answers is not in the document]
GustavoCaso: Use an enum to differentiate between events

lib/saluki-env/src/autodiscovery/providers/local.rs
GustavoCaso: I parsed the ones that we care for checks `init_config` and `instances`. I think we can leave for now like this and the more we use other fields we can add the parsig bit in the local implmentation

lib/saluki-env/src/autodiscovery/mod.rs
GustavoCaso: Thanks for the context. Updated the comment 😄

lib/saluki-env/src/autodiscovery/mod.rs
```diff
@@ -96,21 +98,19 @@ pub struct Config {
...
-
         // Convert advanced AD identifiers from proto
         let advanced_ad_identifiers = proto
             .advanced_ad_identifiers
             .into_iter()
             .map(|adv_id| {
                 let kube_service = adv_id.kube_service.map(|svc| KubeNamespacedName {
-                    name: svc.name,
-                    namespace: svc.namespace,
+                    name: MetaString::from(svc.name),
```
tobz, line 107: Generally speaking, if a type implements `From<T>` -- so in this case, `impl From<String> for MetaString -- you can just do `value.into()` instead of the full `T::from(value)`.

lib/saluki-env/src/autodiscovery/mod.rs
```diff
@@ -236,4 +296,56 @@ mod tests {
...
+        };
+
+        proto_config.advanced_ad_identifiers = vec![adv_id];
+
+        let event = AutodiscoveryEvent::from(proto_config.clone());
+
+        match event {
+            AutodiscoveryEvent::Schedule { config: _config } => {}
+            _ => panic!("Expected an Schedule event"),
+        }
```
tobz, line 340: There's a handy macro, `matches!`, which returns a boolean value based on if the pattern match is a hit or not:

```suggestion
        assert!(matches!(event, AutodiscoveryEvent::Schedule { .. }), "expected schedule event");
```

We're also taking advantage of the ability to ignore unbound fields of the enum variant by using `..`. This is the equivalent of doing `foo: _` for every field in the variant.

lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -66,32 +76,76 @@ impl LocalAutoDiscoveryProvider {
...
+        name: MetaString::from(path.file_name().unwrap().to_string_lossy().to_string()),
+        init_config,
+        instances,
+        metric_config: HashMap::new(),
+        logs_config: HashMap::new(),
         ad_identifiers: Vec::new(),
-        provider: "local".to_string(),
-        service_id: String::new(),
-        tagger_entity: String::new(),
+        provider: MetaString::from_static(""),
```
tobz, line 143: Change these to `MetaString::empty()`.

Same thing.

lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -66,32 +76,76 @@ impl LocalAutoDiscoveryProvider {
...
+}
+
 /// Parse a YAML file into a Config object
 async fn parse_config_file(path: &PathBuf) -> Result<(String, Config), GenericError> {
     let content = fs::read_to_string(path).await?;
 
+    let check_config: CheckConfig = match serde_yaml::from_str(&content) {
+        Ok(read) => read,
+        Err(e) => {
+            warn!("Can't decode yaml as check configuration: {}", content);
```
tobz, line 92: You're already emitting a warning log when you call `parse_config_file` and it returns `Err(e)`, so no need to also emit a warning log here too.

tobz (changes requested):

lib/saluki-env/src/autodiscovery/providers/local.rs
GustavoCaso: removed 🔥

lib/saluki-env/src/autodiscovery/providers/local.rs
GustavoCaso: Done

lib/saluki-env/src/autodiscovery/mod.rs
GustavoCaso: Thanks for the tip

lib/saluki-env/src/autodiscovery/mod.rs
GustavoCaso: Done ✅

lib/saluki-env/src/autodiscovery/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+    let map: HashMap<String, serde_yaml::Value> = serde_yaml::from_str(&parse_bytes)?;
+
+    let mut result = HashMap::<MetaString, serde_yaml::Value>::new();
+
+    for (key, value) in map {
+        result.insert(key.into(), value);
+    }
+
+    Ok(result)
+}
```
tobz, line 163: For the sake of this method, we can just have it spit out `HashMap<String, serde_yaml::Value>` to avoid to complete rebuilding of the map just to convert the `String` to `MetaString`.

lib/saluki-env/src/autodiscovery/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+            node_name: proto.node_name.into(),
+            source: proto.source.into(),
+            ignore_autodiscovery_tags: proto.ignore_autodiscovery_tags,
+            metrics_excluded: proto.metrics_excluded,
+            logs_excluded: proto.logs_excluded,
+        }
+    }
+}
+
+fn bytes_to_hasmap(bytes: Vec<u8>) -> Result<HashMap<MetaString, serde_yaml::Value>, GenericError> {
```
tobz, line 151: `bytes_to_hasmap` -> `bytes_to_hashmap`

lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -0,0 +1,332 @@
...
+            return Err(GenericError::from(e).context("Failed to decode yaml as check configuration."));
+        }
+    };
+
+    // Build config ID from the file path
+    let config_id = path
+        .canonicalize()?
+        .to_string_lossy()
+        .to_string()
+        .replace(['/', '\\'], "_");
```
tobz, line 101: Right now, this will do synchronous file I/O to canonicalize the path, so this little change just ensures we do it asynchronous (using the `tokio` variants of functions in `tokio::fs`), and additionally, we've skipped the `to_string()` call since `replace` operates on `&str`, so we're allocating the intermediate `String` needlessly since `replace` is going to allocate one anyways.

```suggestion
    let canonicalized_path = fs::canonicalize(&path).await?;
    let config_id = canonicalized_path
        .to_string_lossy()
        .replace(['/', '\\'], "_");
```

lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -0,0 +1,332 @@
...
+                }
+            }
+        });
+    }
+}
+
+#[derive(Debug, Deserialize)]
+struct CheckConfig {
+    init_config: serde_yaml::Mapping,
+    instances: Vec<serde_yaml::Mapping>,
```
tobz, line 82: I think this should just be `Vec<HashMap<String, serde_yaml::Value>>`, and then you can skip all the additional logic down in `parse_config_file` that's manually doing this at the moment.

lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -0,0 +1,332 @@
...
+use tokio::sync::OnceCell;
+use tokio::time::{interval, Duration};
+use tracing::{debug, info, warn};
+
+use crate::autodiscovery::{AutodiscoveryEvent, AutodiscoveryProvider, Config};
+
+const BG_MONITOR_INTERVAL: u64 = 30;
+
+/// A local auto-discovery provider that uses the file system.
+pub struct LocalAutoDiscoveryProvider {
```
tobz, line 20: `LocalAutoDiscoveryProvider` -> `LocalAutodiscoveryProvider`

lib/saluki-env/src/autodiscovery/providers/remote_agent.rs
```diff
@@ -0,0 +1,82 @@
...
+use futures::StreamExt;
+use tokio::sync::broadcast::{self, Receiver, Sender};
+use tokio::sync::OnceCell;
+use tracing::{debug, info, warn};
+
+use crate::autodiscovery::{AutodiscoveryEvent, AutodiscoveryProvider};
+use crate::helpers::remote_agent::RemoteAgentClient;
+
+/// An autodiscovery provider that uses the Datadog Agent's internal gRPC API to receive autodiscovery updates.
+pub struct RemoteAgentAutoDiscoveryProvider {
```
tobz, line 13: `RemoteAgentAutoDiscoveryProvider` -> `RemoteAgentAutodiscoveryProvider`

lib/saluki-env/src/autodiscovery/providers/remote_agent.rs
```diff
@@ -0,0 +1,82 @@
...
+                let mut autodiscovery_stream = client.get_autodiscovery_stream();
+
+                while let Some(result) = autodiscovery_stream.next().await {
+                    match result {
+                        Ok(response) => {
+                            for proto_config in response.configs {
+                                let event = AutodiscoveryEvent::from(proto_config);
+                                match sender.send(event) {
+                                    Ok(_) => (),
+                                    Err(e) => {
```
tobz, line 50: An error during send  on a broadcast channel implies there are no active receivers, so you can effectively ignore them.

lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -0,0 +1,332 @@
...
+        emit_event(sender, event);
+    }
+
+    Ok(())
+}
+
+fn emit_event(sender: &Sender<AutodiscoveryEvent>, event: AutodiscoveryEvent) {
+    match sender.send(event) {
+        Ok(_) => (),
+        Err(e) => {
```
tobz, line 213: An error during send on a broadcast channel implies there are no active receivers, so you can effectively ignore them.

lib/saluki-env/src/autodiscovery/providers/remote_agent.rs
```diff
@@ -0,0 +1,82 @@
...
+        let mut client = self.client.clone();
+        let sender = self.sender.clone();
+
+        tokio::spawn(async move {
+            info!("Listening to autodiscovery events from remote agent.");
+
+            loop {
+                let mut autodiscovery_stream = client.get_autodiscovery_stream();
+
+                while let Some(result) = autodiscovery_stream.next().await {
```
tobz, line 43: We might want to add a debug log right above this line to the effect of:

```rust
debug!("Polling autodiscovery event stream.");
```

While the info log we have above tells us that we've started listening _overall_, we don't have a way to know if the stream is ending gracefully and being restarted often, which adding this debug log would give us.

tobz (changes requested):

lib/saluki-env/src/autodiscovery/mod.rs
```diff
@@ -0,0 +1,348 @@
+//! Autodiscovery provider.
+//!
+//! This module provides the `Autodiscovery` trait, which deals with providing information about autodiscovery.
+
+pub mod providers;
```
tobz, line 5: Move this line down below all imports (`use`).

(This is something we, unfortunately, cannot yet lint/control with the formatting tool.)

tobz (changes requested):
One small nit and then we'll be good to go.

tobz (approved):

PR #654, opened 2025-04-29
fix: adjust preaggr subdomain to always be api

lukesteensen:
## Summary

The new pipeline endpoint is only available at `api.`, so we need to adjust the subdomain of the additional forwarder.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Performance

## References

TPOT-19


lukesteensen:
Really need to get cursor set up to fmt reliably...

PR #655, opened 2025-04-29
experiment(env): normalize tag store to reduce overhead

tobz:
## Summary

This PR refactors `TagStore`, and related components, to switch to a normalized data model that doesn't express the same dataflow/event-based semantics of the previous design. This is in service to reducing the amount of duplicated data, and overhead in building unified tag sets, for entities in scenarios where the present workloads involve a high number of entities and entity tags.

### Current design

Currently, `TagStore` is a hierarchical design, where each entity (container ID, pod UID, process ID, etc) has its own specific tags, and entities are linked together in a hierarchy mapping: process ID X belongs to container Y, container Y belongs to pod Z, and so on. Each of these "entities" (pod UID, container ID, process ID, and more) can have their own individual tags, but when querying for the tags of a given entity, we want the aggregated set of tags for the entity itself and all of its ancestors. For example, if we query the tags of a container entity, we should get its tags and also any tags of the pod it's linked to, if it is in fact linked to a pod. This allows us to query the highest precedence entity ID available to us while getting all of the relevant tags.

### Issues with current design

However, this design has a few problems in its current implementation, and based on how we actually _use_ `TagStore`:

- this design means we have to manage separate, distinct sets of tags for each entity, which means we're constantly cloning the tags of linked ancestors when building the tags for a specific entity (inefficient)
- high overhead when updating the tags of an entity with descendants (all tags for descendants now have to be rebuilt)
- use of a "tag snapshot" which means constant cloning of all unified tagsets (the map of entity ID -> tags, for each cardinality level) on each update
- we actually try to get the tags for every available entity ID we scraped for a given metric, which means we might be evaluating the same tags many times (e.g., we might evaluate pod-specific tags multiple times if we have the pod UID entity, container ID entity, _and_ process ID entity), as this behavior matches how the Datadog Agent performs origin-based enrichment p

### Updated design

This PR updates the design of `TagStore` to be normalized, such that we don't build unified tagsets for each entity, as we no longer track entity hierarchy mappings. We only track the tags set directly for a given entity. Off the bat, this means no unnecessary duplication of tags since we're not merging/unifying them together, and so we only store the minimum amount necessary. This is OK because we know origin detection will query the tags for each entity ID it has: if we have a pod UID, we'll query the tags for that, and if we have a container ID, we'll query the tags for _that_, and so on.

Secondly, we've updated the design to add a concept of "aliases." While the hierarchical design had inefficiencies, one thing it did allow for is trivially associating something like a process ID with the container ID it belongs to. In cases where _all_ we have is a process ID, we can still manage to sometimes link that back to the container ID. This is relevant because process IDs don't have their own tags: along with container inodes, these are simply a way to "alias" to a container ID, where the actual set of tags we care about will be present. The addition of aliases means we can preserve that behavior of getting the "tags" for process IDs and container inodes, aka the container ID they're linked to.

### Unrelated changes

We've removed a bunch of ancillary stuff that wasn't in use, such as different ways to add tags to an entity (i.e., we supported both extending the existing set of tags _and_ overriding them, but only ever wanted the override behavior, and thus only ever used that approach) and more.

As well, we've switched to pushing the usage of `TagVisitor` deeper into `saluki-env`, which now lets us avoid having to build merged tagsets of the different cardinality groupings and we can simply query each necessary cardinality and visit the respective tagsets in order.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Unit tests. Still need to deploy to staging to verify any performance improvements.

## References

AGTMETRICS-184


lib/saluki-env/src/workload/providers/remote_agent/api.rs
```diff
@@ -41,44 +49,34 @@ impl RemoteAgentWorkloadState {
...
-            while let Some(parent) = entity_mappings.get(current_entity) {
-                ancestors.insert(0, parent);
-                current_entity = parent;
-            }
-
+            let alias = entity_aliases.get(entity_id);
             let low_cardinality_tags = self
                 .tag_querier
-                .get_entity_tags(entity_id, OriginTagCardinality::Low)
+                .get_unaliased_entity_tags(entity_id, OriginTagCardinality::Low)
```
rayz, line 62: Why `.get_unaliased_entity_tags` here? Doesn't the old implementation using `TagSnapshot` have the unified tags so the dump would show those tags for the `entity_id` but here we are leaving it empty if we have something that has an alias.

rayz (approved):

Issue #646, opened 2025-04-24
Add support for copying histograms to distributions in DogStatsD source.


tobz:
## Context

The DogStatsD implementation in Datadog Agent supports copying histograms to distributions by setting a configuration field, `histogram_copy_to_distribution`, to `true`. This does say it says: it will _copy_ (not convert) each histogram to a distribution, and send them both downstream to the aggregator. The Agent allows setting a specific prefix to add to the distribution's name (`histogram_copy_to_distribution_prefix`) to help differentiate them, but it's not required to be populated.

We should add support for this by exposing similar configuration fields, and logic, in the DogStatsD source, to allow copying histograms to distributions before forwarding.

## Open Questions

If we naively allowed the prefix to be empty, as the Datadog Agent does, we'd end up with an identical context which would be problematic from an aggregation standpoint. If we tried to wait to post-aggregation to do the copying, it would be too late since the histogram would be decomposed into its summary statistics by then.

Does this mean we need to incorporate metric type into contexts as well? How is the Datadog Agent handling this? Should we just _force_ a non-empty prefix?

tobz:
I investigated the question of how the Datadog Agent is handling things when the prefix is empty.

Looking at the [relevant code](https://github.com/DataDog/datadog-agent/blob/main/pkg/aggregator/time_sampler.go#L90-L104), we see that the time samplers handle distributions differently from all other metric types. This means that even though the context key is identical for both a histogram and its distribution-based copy, the two don't intersect as they're stored in separate places.

This means that we _do_ have to support a scenario where no prefix is given to the distribution copy. However, I think it also means we can approach this as an aggregator problem. If we handle this in the aggregator, we can add it as an additional stage of histogram processing:
- process the various aggregate statistics of the histogram, each emitted as their own metric
- if copying is enabled, take the original histogram metric and convert it to a distribution

PR #657, opened 2025-05-02
enhancement: add support for `histogram_copy_to_distribution` feature

rayz:
## Summary

This pr adds support for the `histogram_copy_to_distribution_feature`. The copy is done in the `aggregate` transform.
## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Set `histogram_copy_to_distribution` to `true` and set `histogram_copy_to_distribution_prefix`. Emit a histogram metric and see that the histogram statistics show up and as well as a new distribution metrics on the Datadog UI

## References

- Closes: #646


lib/saluki-components/src/transforms/aggregate/mod.rs
```diff
@@ -776,6 +776,17 @@ where
                 forwarder.push(Event::Metric(new_metric)).await?;
             }
 
+            if hist_config.copy_to_distribution() {
+                let new_context = context.with_name(format!(
+                    "{}{}",
```
rayz, line 781: They just append the prefix and name in the agent 
https://github.com/DataDog/datadog-agent/blob/main/comp/dogstatsd/server/server.go#L681

lib/saluki-components/src/transforms/aggregate/mod.rs
```diff
@@ -776,6 +776,17 @@ where
...
+                    hist_config.copy_to_distribution_prefix(),
+                    context.name()
+                ));
+
+                let new_metric = Metric::from_parts(new_context, values, metadata.clone());
```
rayz, line 786: We convert the histograms to sketches later [here](https://github.com/DataDog/saluki/blob/4169391f3a0f639320eedeadd7249a5e86a39392/lib/saluki-components/src/destinations/datadog/metrics/request_builder.rs#L614-L622), so I don't think we would need to do any conversion here (?).

lib/saluki-components/src/transforms/aggregate/config.rs
```diff
@@ -89,13 +89,21 @@ struct RawHistogramConfiguration {
     /// percentile of 0.95 would be represented as `95percentile`, while a percentile of 0.05 would be represented as
     /// `5percentile`.
     histogram_percentiles: Vec<String>,
+
+    /// Whether to copy the histogram to the distribution metric.
```
tobz, line 93:
```suggestion
    /// Whether to copy histograms to distributions.
    ///
    /// Emits a copy of each histogram as a distribution, potentially with
    /// using either the original histogram's name, or an updated name
    /// based on the value of `copy_histogram_to_distribution_prefix`.
    ///
    /// Defaults to `false`.
```

lib/saluki-components/src/transforms/aggregate/config.rs
```diff
@@ -89,13 +89,21 @@ struct RawHistogramConfiguration {
...
+    histogram_copy_to_distribution: bool,
+
+    /// Prefix to append to the distribution metric name.
```
tobz, line 96:
```suggestion
    /// Prefix to append to the name of distributions copied from histograms.
    ///
    /// Defaults to an empty string (no prefixing).
```

lib/saluki-components/src/transforms/aggregate/config.rs
```diff
@@ -104,20 +112,36 @@ impl Default for RawHistogramConfiguration {
...
+            copy_to_distribution_prefix,
         }
     }
 
     /// Returns the configured aggregate statistics to calculate.
     pub fn statistics(&self) -> &[HistogramStatistic] {
         &self.statistics
     }
+
+    /// Returns `true` if the histogram should be copied to the distribution metric.
```
tobz, line 136:
```suggestion
    /// Returns `true` if histograms should be copied to distributions.
```

lib/saluki-components/src/transforms/aggregate/config.rs
```diff
@@ -104,20 +112,36 @@ impl Default for RawHistogramConfiguration {
...
+    pub fn copy_to_distribution(&self) -> bool {
+        self.copy_to_distribution
+    }
+
+    /// Returns the prefix to append to the distribution metric name.
```
tobz, line 141:
```suggestion
    /// Returns the prefix to append to the name of distributions copied
    /// from histograms.
```

lib/saluki-components/src/transforms/aggregate/mod.rs
tobz: We shouldn't be creating a new context if we don't have to. That's effectively what we're doing here, even if the prefix is the same... because `Context::with_name` will blindly clone the context's inner data before altering the name. It doesn't check if the name is actually change it not.

lib/saluki-components/src/transforms/aggregate/mod.rs
tobz: You're right that we do treat histograms like distributions when sending them out of the Datadog Metrics destination.

I think I'd still want to see us convert to a proper `Distribution` here because the expectation is that we're actively copying a `Histogram` to a `Distribution` by the time it leaves the aggregate transform.

If some other component expected this to happen, but we made this optimization assuming that we only ever fed into the Datadog Metrics destination... well, then we'd have a mismatch in expected behavior.

lib/saluki-components/src/transforms/aggregate/mod.rs
```diff
@@ -1264,6 +1279,44 @@ mod tests {
...
+        let count_metric = Metric::rate("metric1.count", 0.0, Duration::from_secs(BUCKET_WIDTH_SECS));
+        let sum_metric = Metric::gauge("metric1.sum", 0.0);
+        let p50_metric = Metric::gauge("metric1.p50", 0.0);
+
+        // We use a less strict error ratio (how much the expected vs actual) for the percentile check, as we generally
+        // expect the value to be somewhat off the exact value due to the lossy nature of `DDSketch`.
+        assert_flushed_scalar_metric!(count_metric, &flushed_metrics[1], [bucket_ts(1) => 5.0]);
+        assert_flushed_scalar_metric!(p50_metric, &flushed_metrics[2], [bucket_ts(1) => 3.0], error_ratio => 0.0025);
+        assert_flushed_scalar_metric!(sum_metric, &flushed_metrics[3], [bucket_ts(1) => 15.0]);
+    }
```
tobz, line 1318: Seems like we should have a `assert_flushed_distribution_metric!` call here to assert the distribution metric we're copying the histogram to is accurate: expected prefixed name, the same points, etc.

tobz (changes requested):

lib/saluki-components/src/transforms/aggregate/mod.rs
```diff
@@ -750,6 +751,32 @@ where
         // If we're dealing with a histogram, we calculate a configured set of aggregates/percentiles from it, and emit
         // them as individual metrics.
         MetricValues::Histogram(ref mut points) => {
+            // Convert histogram to distribution
+            if hist_config.copy_to_distribution() {
+                for (timestamp, histogram) in points.clone() {
+                    if timestamp.is_some() {
```
rayz, line 757: Skipping if `timestamp` is `None` (when would this be?) since `SketchPoints` cannot use a default timestamp of `0`

lib/saluki-event/src/metric/value/sketch.rs
```diff
@@ -188,6 +188,19 @@ impl<'a> IntoIterator for &'a SketchPoints {
...
 }
 
+impl FromIterator<(Option<NonZeroU64>, DDSketch)> for SketchPoints {
+    fn from_iter<T: IntoIterator<Item = (Option<NonZeroU64>, DDSketch)>>(iter: T) -> Self {
+        let mut sketch_points = SketchPoints(TimestampedValues::default());
+        for (ts, sketch) in iter {
+            sketch_points
+                .0
+                .values
+                .push(TimestampedValue::from((ts.unwrap().get(), sketch)));
```
tobz, line 198: We shouldn't be unwrapping if we're knowingly accepting `Option<T>` here.. so we should add a new `From` impl -- `impl<T> From<(Option<NonZeroU64>, T)> for TimestampedValue<T>` -- to let us convert properly.

tobz (changes requested):

tobz (approved):

PR #662, opened 2025-05-05
[AGENTRUN-287] feature(checks-agent): add  checks source and scheduler

GustavoCaso:
## Summary

This adds the checks source and scheduler. 

The check's source reads events from the autodiscovery provider, and based on the type of event, it would schedule or unschedule events into the scheduler. 

The check collector maintains a list of all checks, stored in a HashSet to ensure that duplicated events, likely to occur from a reconnect, do not schedule duplicate checks. 

The scheduler is not 100% feature parity with the one in Go, but it is a good start to continue adding more features to it. We introduce the trait `CheckBuilder`, which is similar to the loader functionality in the Go version. We could add different implementations: Python, Rust, dynamic loading, etc. We also introduce the `Check` trait. 

The scheduler uses the same logic as the one in Go to store checks in buckets based on their interval. 
There are a few features missing from the implementation:
- Schedule checks without an interval 
- Stats 

I wanted to first get an initial round of feedback to ensure we are going in the right direction 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance


Cargo.toml
```diff
@@ -177,6 +177,8 @@ axum-extra = { version = "0.9", default-features = false }
 papaya = { version = "0.2", default-features = false }
 tempfile = { version = "3", default-features = false }
 fs4 = { version = "0.13.1", default-features = false }
+fnv = { version = "1.0.7", default-features = false }
```
tobz, line 180: Super small note: we generally prefer to scope this to the minimum version necessary, kind of similar to what Go does by default.

Cargo will assume here, for example, that we _need_ `fnv` to be at least `1.0.7`, even if `1.0.0` would be fine. Even if we only put `1`, Cargo will use the latest version it reasonably can, will adhering to semver.

So for `fnv`, we can say `1`.

Cargo.toml
```diff
@@ -177,6 +177,8 @@ axum-extra = { version = "0.9", default-features = false }
...
+twox-hash = { version = "1.6.1", default-features = false }
```
tobz, line 181: Looks like `twox-hash` is already at `2.x`, so we should use that.

bin/checks-agent/src/env_provider.rs
```diff
@@ -25,7 +25,7 @@ use tracing::{debug, warn};
 #[allow(dead_code)]
 pub struct ChecksAgentEnvProvider {
     host_provider: BoxedHostProvider,
-    autodiscovery_provider: BoxedAutodiscoveryProvider,
+    pub autodiscovery_provider: BoxedAutodiscoveryProvider,
```
tobz, line 28: We should expose this by adding a method on `ChecksAgentEnvProvider` instead, like so:

```rust
pub fn autodiscovery(&self) -> &BoxedAutodiscoveryProvider {
    &self.autodiscovery_provider
}
```

This will get the code a little closer to what it will eventually look like when we actually expose the autodiscovery provider through the `EnvironmentProvider` trait.

bin/checks-agent/src/env_provider.rs
```diff
@@ -54,7 +54,11 @@ impl ChecksAgentEnvProvider {
 
         let autodiscovery_provider = if in_standalone_mode {
             debug!("Using local autodiscovery provider due to standalone mode.");
-            BoxedAutodiscoveryProvider::from_provider(LocalAutodiscoveryProvider::new(vec!["dist"]))
+            let mut config_dir = config.get_typed_or_default::<String>("checks_config_dir");
+            if config_dir.is_empty() {
+                config_dir = "./dist/conf.d".to_string();
```
tobz, line 59: This should be whatever the default Datadog Agent checks configuration directory is, like `/etc/datadog-agent/conf.d` or whatever.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+                                        let check_id = config.id(digest, instance);
+                                        if self.checks.contains(&check_id) {
+                                            continue;
+                                        }
+
+                                        for builder in self.check_builders.iter_mut() {
+                                            if let Some(check) = builder.build_check(&check_id, &config) {
+                                                checks.push(check);
+                                                self.checks.insert(check_id.clone());
+                                                continue;
```
tobz, line 164: This `continue` seems unnecessary since there's no more logic in the loop body after this point.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+                    break
+                },
+
+                _ = self.health.live() => continue,
+
+                event = self.event_rx.recv() => match event {
+                        Ok(event) => {
+                            match event {
+                                AutodiscoveryEvent::Schedule { config } => {
+                                    let digest = &config.digest();
```
tobz, line 151: I would drop taking the reference here (so make it just `config.digest()`) and then take the reference at the point of usage down when you're calling `config.id(...)`.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+                                            }
+                                        }
+                                    }
+
+                                    for check in checks {
+                                        self.scheduler.schedule(check);
+                                    }
+                                }
+                                AutodiscoveryEvent::Unscheduled { config } => {
+                                    let digest = &config.digest();
```
tobz, line 174: Same comment as above about referencing at the point of usage instead.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+                                    for instance in &config.instances {
+                                        let check_id = config.id(digest, instance);
+                                        if self.checks.contains(&check_id) {
+                                            continue;
+                                        }
+
+                                        for builder in self.check_builders.iter_mut() {
+                                            if let Some(check) = builder.build_check(&check_id, &config) {
+                                                checks.push(check);
+                                                self.checks.insert(check_id.clone());
```
tobz, line 163: This seems like it would allow creating multiple versions of the same check since we're not checking if the check ID has already been registered.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+impl Source for ChecksSource {
+    async fn run(self: Box<Self>, mut context: SourceContext) -> Result<(), GenericError> {
+        let mut global_shutdown = context.take_shutdown_handle();
+        let mut health = context.take_health_handle();
+
+        let mut listener_shutdown_coordinator = DynamicShutdownCoordinator::default();
+
+        let check_collector_health = context
+            .health_registry()
+            .register_component("checks_collectorr")
```
tobz, line 211: `checks_collectorr` -> `checks_collector`

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+            shutdown_handle,
+            health,
+            event_rx,
+            check_builders,
+            checks: HashSet::new(),
+            scheduler: Scheduler::new(check_runners),
+        }
+    }
+
+    pub async fn run(mut self) {
```
tobz, line 134: It seems like it would end up being cleaner just to have this logic be in the body of `Source::run` down below since you're not creating multiple instances of `ChecksCollector` or anything.

lib/saluki-core/src/topology/blueprint.rs
```diff
@@ -143,7 +143,7 @@ impl TopologyBlueprint {
         let component_id = self
             .graph
             .add_source(component_id, &builder)
-            .error_context("Failed to build/validate topology graph.")?;
+            .error_context("Failed to add_source to topology graph.")?;
```
tobz, line 146: `add_source` -> `add source`

lib/saluki-core/src/topology/blueprint.rs
```diff
@@ -172,7 +172,7 @@ impl TopologyBlueprint {
         let component_id = self
             .graph
             .add_transform(component_id, &builder)
-            .error_context("Failed to build/validate topology graph.")?;
+            .error_context("Failed to add_transform to topology graph.")?;
```
tobz, line 175: `add_transform` -> `add transform`

lib/saluki-core/src/topology/blueprint.rs
```diff
@@ -203,7 +203,7 @@ impl TopologyBlueprint {
         let component_id = self
             .graph
             .add_destination(component_id, &builder)
-            .error_context("Failed to build/validate topology graph.")?;
+            .error_context("Failed to add_destination to topology graph.")?;
```
tobz, line 206: `add_destination` -> `add destination`

lib/saluki-core/src/topology/blueprint.rs
```diff
@@ -238,7 +238,7 @@ impl TopologyBlueprint {
         for source_output_component_id in source_output_component_ids.into_iter() {
             self.graph
                 .add_edge(source_output_component_id, destination_component_id.as_ref())
-                .error_context("Failed to build/validate topology graph.")?;
+                .error_context("Failed to connect_component to topology graph.")?;
```
tobz, line 241:
```suggestion
                .error_context("Failed to add component connection to topology graph.")?;
```

lib/saluki-env/src/autodiscovery/mod.rs
```diff
@@ -96,6 +99,49 @@ pub struct Config {
     pub logs_excluded: bool,
 }
 
+impl Config {
+    /// Get the ID for an instance
+    pub fn id(&self, digest: &u64, instance: &HashMap<MetaString, serde_yaml::Value>) -> String {
```
tobz, line 104: You don't need to pass primitives, like integers, by reference unless you need to actually provide a mutable reference so the callee is able to mutate it.

Otherwise, it's not any cheaper: `u64` is 8 bytes, and on 64-bit platforms, so is the pointer backing the reference to it.

tobz, line 104: Separately: what's the _point_ of this function?

My naive thought is that the config instance should be wrapped up in some sort of helper type -- let's call it `ConfigInstance` -- where it holds both the raw instance config _and_ the ID for that instance.

Doing that would allow you to handle all of the error checking, which is being skipped here, as part of initially creating the `Config`, instead of having to do it all on-demand.

lib/saluki-env/src/autodiscovery/mod.rs
```diff
@@ -162,16 +208,35 @@ fn bytes_to_hashmap(bytes: Vec<u8>) -> Result<HashMap<MetaString, serde_yaml::Va
...
+            return value.to_string();
+        }
+    }
+    if let Some(namespace) = instance.get("namespace") {
+        if let Some(value) = namespace.as_str() {
+            return value.to_string();
+        }
+    }
+    "".to_string()
+}
```
tobz, line 228: Functions like this are another good reason why a check config instance could stand to be wrapped up in a custom struct, to hide this logic and simply expose it as a basic method.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+/// A no-op implementation of CheckBuilder
+struct NoOpCheckBuilder;
+
+impl CheckBuilder for NoOpCheckBuilder {
+    fn build_check(&self, _check_id: &str, _check_request: &Config) -> Option<Arc<dyn Check + Send + Sync>> {
+        None
+    }
+}
```
tobz, line 108: We should create a new module -- `lib/saluki-components/src/sources/checks/builder` -- and put this stuff under there.

tobz (changes requested):

lib/saluki-components/src/sources/checks/mod.rs
GustavoCaso: On line 156

```rs
if self.checks.contains(&check_id) {
  continue;
}
```

So the logic is:

1. Check if the check ID is present, if so skip
2. Check which check builder support that check
3. If any check builder, push the check 

Maybe I'm missing something

lib/saluki-env/src/autodiscovery/mod.rs
GustavoCaso: That could work. 

The idea of having digest and id functions separated is to avoid computing the digest multiple times for checks with multiple instances. 

> Doing that would allow you to handle all of the error checking, which is being skipped here, as part of initially creating the Config, instead of having to do it all on-demand.

I guess we could do it, but doing it on demand means that if we receive non-check events from the autodiscovery provider, we can skip them without doing any extra computation. BTW, that event skipping is not implemented yet

lib/saluki-components/src/sources/checks/mod.rs
GustavoCaso: The continue is meant to signal that the first check builder that returns the check is the one that we would execute. Now that I have said it out loud, I realized that I meant to use `break`

lib/saluki-components/src/sources/checks/scheduler.rs
```diff
@@ -0,0 +1,506 @@
...
+    fn stop_interval_ticker(&self, interval_secs: u64) {
+        let mut tickers = self.interval_handles.lock().unwrap();
+        if let Some(handle) = tickers.remove(&interval_secs) {
+            handle.abort();
+            debug!("Stopped ticker for interval {}s", interval_secs);
+        }
+    }
+
+    /// Adjust the number of workers
+    fn adjust_worker_count(&self, desired_count: usize) {
```
scottopell, line 231: I'm a bit skeptical that "number of workers" is a useful construct for the python check runners.

Since python execution is blocked on the GIL anyway, there is no "concurrency" here.
And even if we look ahead to a `nogil` future, its not clear to me why we would manage our own runner abstraction level when we can defer this to `tokio` for scheduling.

GustavoCaso: @scottopell that is a very valid point. 

Still, I want to understand this last part for your comment:
> And even if we look ahead to a nogil future, its not clear to me why we would manage our own runner abstraction level when we can defer this to tokio for scheduling.

Each scheduler worker is a tokio thread, that which would be manage by the tokio scheduler https://github.com/DataDog/saluki/blob/883e35684c91f2361fbf78cbdbb5770d87af2e64/lib/saluki-components/src/sources/checks/scheduler.rs#L254

Are you suggesting to not even worry about having workers at all and running each check on a tokio thread, and let tokio handle everything?

lib/saluki-env/src/autodiscovery/mod.rs
tobz: If we're receiving non-check events, though... then they wouldn't have anything in `instances`, no?

That's mostly what I'm getting at: if we've decoded an event far enough that we know it has check config instances in `instances`, then we could go one small step further during the creation of `Config` and wrap that up in some hypothetical type like `ConfigInstance`, and we'd only be computing the digest/ID when creating `ConfigInstance`, because we know we always need to query those values at least once.

lib/saluki-components/src/sources/checks/scheduler.rs
```diff
@@ -0,0 +1,506 @@
...
+
+/// Message type for worker communication
+enum WorkerMessage {
+    RunCheck(Arc<dyn Check + Send + Sync>),
+    Shutdown,
+}
+
+/// A scheduler that manages the execution of checks.
+/// It maintains a dynamic pool of workers and organizes checks by their intervals.
+pub struct Scheduler {
```
tobz, line 20: I think this could be restructured a little differently to end up with a more structured approach to the concurrency involved, but I had a few behavioral questions first:

- is the scheduler meant to actually drop the scheduling of checks if the check runners are busy?
- is running the checks _exactly_ on their interval a requirement or do we only care that we've waited for `interval` time between each run of a check?

lib/saluki-env/src/autodiscovery/mod.rs
GustavoCaso: I've decoded the config further and split between checks config and generic configuration events. I added two specific legs to the `AutodiscoveryEvent` enum to differentiate between generic and check events as well. 

Here is the code https://github.com/DataDog/saluki/pull/662/commits/16d66ab224e510dd5236cb852399ee300f7f1c03

Please let me know if that's what you think.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+            .register_component("checks_collector")
+            .expect("Failed to register checks collector health");
+        let mut event_rx = self.autodiscovery.subscribe().await;
+
+        let mut check_builders: Vec<Arc<dyn CheckBuilder + Send + Sync>> = vec![Arc::new(NoopCheckBuilder)];
+        let mut shutdown_handle = listener_shutdown_coordinator.register();
+        let mut check_ids = HashSet::new();
+        let scheduler = Scheduler::new(self.check_runners);
+
+        spawn_traced_named("checks-collector", async move {
```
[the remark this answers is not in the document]
GustavoCaso: Done https://github.com/DataDog/saluki/pull/662/commits/6099c960c180c268bd464238a91f03e388af2339

lib/saluki-env/src/autodiscovery/mod.rs
```diff
@@ -59,19 +62,65 @@ pub struct AdvancedADIdentifier {
...
 
-/// Configuration data
+trait RawData {
+    fn get_value(&self) -> &HashMap<MetaString, serde_yaml::Value>;
+
+    fn get(&self, key: &str) -> Option<&serde_yaml::Value> {
+        self.get_value().get(key)
+    }
+
+    fn to_bytes(&self) -> Result<Vec<u8>, GenericError> {
```
tobz, line 72: I might suggest calling this `to_yaml` just to be explicit, but that could happen in a follow-up PR.

tobz (approved):
Definitely some cleanup to do in follow-up PRs but this is reasonable enough for now. 👍🏻

PR #676, opened 2025-05-09
enhancement(core): add `remove_if` method to `FixedSizeEventBuffer` for efficient removal

tobz:
## Summary

This PR adds a new method, `remove_if`, to `FixedSizeEventBuffer`, that allows for removing specific events when they aren't needed any longer. The previous best way to accomplish this -- `extract` -- was inefficient for that usage as it involved intermediate allocations to do so.

We've updated the Preaggregation Filter transform to take advantage of this new method, and we also fixed up both methods to properly update the "seen" data types for `FixedSizeEventBuffer`, and updated the unit tests to account for that as well.

## Change Type

- [x] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests.

## References

AGTMETRICS-233


lukesteensen (approved):
Looks good! This does make me wonder if it'd be simpler to have buffers hold a single type.

PR #678, opened 2025-05-09
chore(core): add new payload model + skeleton of forwarder component type

tobz:
## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Non-functional (chore, refactoring, docs)
- [ ] Performance

PR #677, opened 2025-05-09
chore(core): move `saluki-event` crate into `saluki_core::data_model::event`

tobz:
## Summary

This PR moves the entirety of the `saluki-event` crate into a new section in `saluki-core`: `saluki_core::data_model::event`. This is in preparation of expanding the data model of Saluki to include data types other than "events".

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Non-functional (chore, refactoring, docs)
- [ ] Performance

## How did you test this PR?

Normal tests, and the fact it compiles.

## References

AGTMETRICS-233

PR #679, opened 2025-05-09
chore(core): add new encoder component type

tobz:
## Summary

This PR adds a new core component type: encoder.

Encoders are meant to act as the presentation layer, transposing telemetry events into output payloads before handing off to forwarders. Encoders emit payloads of varying shapes and encoding/serialization formats.

A simple way to think of encoders is that they are almost always used in conjunction with a specific forwarder to achieve the equivalent of a destination: there might be a Datadog Metrics encoder to encode metrics into an HTTP payload using the proper Protocol Buffers-encoded format, and then an HTTP forwarder that specifically handles HTTP payloads and does so with a certain configuration (retry behavior, etc).

This split allows what might traditionally be multiple destinations instead be multiple encoders and a single forwarder, allowing for more efficient use of existing resources, such as network connections.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

N/A

## References

AGTMETRICS-233


tobz:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/saluki/679?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#679** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/679?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/saluki/679?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#678** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/678?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#677** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/677?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/saluki-core/src/components/renderers/builder.rs
```diff
@@ -0,0 +1,30 @@
...
+
+    /// Data types emitted as output payloads by this renderer.
+    fn output_payload_type(&self) -> PayloadType;
+
+    /// Builds an instance of the renderer.
+    ///
+    /// ## Errors
+    ///
+    /// If the renderer cannot be built for any reason, an error is returned.
+    async fn build(&self, context: ComponentContext) -> Result<Box<dyn Renderer + Send>, GenericError>;
```
lukesteensen, line 29: It'd be nice to make this not async, at least to start. Feels like it shouldn't need to be.

lib/saluki-core/src/components/renderers/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// The renderer context provides access primarily to the event stream, used to receive events sent to the renderer,
+    /// and the forwarder, used to send payloads to the downstream forwarder in the topology, as well as other
+    /// information such as the component context.
+    ///
+    /// Renderers are expected to run indefinitely until their event stream is terminated, or an error occurs.
+    ///
+    /// # Errors
+    ///
+    /// If an unrecoverable error occurs while running, an error is returned.
+    async fn run(self: Box<Self>, context: RendererContext) -> Result<(), GenericError>;
```
lukesteensen, line 35: Related, this would be an interesting one to try to scope down the API a lot more. Essentially have individual implementations focus only on the actual logic, and wrap it with some kind of driver that is the spawned task. So this trait could be more like `add_event`, `flush`, `tick`, etc, all synchronous if possible.

Could also be something to explore later, once the concept is more established.

lukesteensen (commented):
Honestly don't love the name 😅, makes me think of graphics or HTML. It's basic, but I feel like `PayloadBuilder` or something like that would capture the idea better.

lib/saluki-core/src/components/renderers/builder.rs
tobz: This is one that probably doesn't _need_ to be `async`. I suspect we may need to make it `async` in short order, though, given that renderers/payload builders will touch object pools and such.

I'll make it synchronous for now.

lib/saluki-core/src/components/renderers/mod.rs
tobz: Hmm, maybe.

It would definitely be possible, although I wonder if we can properly encode all necessary behaviors with just the individual operations (`add_event`, `flush`, etc) and a generic driver on top.

One thought is that we do that with a default implementation of `run`. Basically just have it process all ready events (via `add_event`) and then `flush` at the end before looping all over again. If code needed a custom implementation of `run` -- specialized flushing behavior, etc -- they could just override `run` in their implementation, or create a wrapper `Renderer` impl or something.

tobz: Or, alternatively, maybe we leave the exact trait as-is but add something equivalent to `SynchronousTransform`, and _that's_ where we implement the model you mention.. and perhaps we end up with like two actual implementations of `Renderer`, like one for "add and flush as soon as possible" and one for "configurable batching/buffering/dwell time"... but all of the interesting implementations are all of `SynchronousRenderer` or whatever.

lib/saluki-core/src/components/renderers/builder.rs
```diff
@@ -0,0 +1,30 @@
...
+pub trait RendererBuilder: MemoryBounds {
+    /// Data types allowed as input payloads to this renderer.
+    fn input_event_type(&self) -> EventType;
+
+    /// Data types emitted as output payloads by this renderer.
+    fn output_payload_type(&self) -> PayloadType;
+
+    /// Builds an instance of the renderer.
+    ///
+    /// ## Errors
```
GustavoCaso, line 26:
```suggestion
    /// # Errors
```

lib/saluki-core/src/components/renderers/context.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub fn new(
+        component_context: ComponentContext, events: EventStream, memory_limiter: MemoryLimiter,
+        component_registry: ComponentRegistry, health_handle: Health, health_registry: HealthRegistry,
+        thread_pool: Handle,
+    ) -> Self {
+        Self {
+            component_context,
+            events,
+            memory_limiter,
+            health_handle: Some(health_handle),
```
GustavoCaso, line 29: Do we need to warp it into an Option? It seems that we then unwrap it or panic on line 42.

tobz: We do want it to be `Option<T>` in this case, because it should only be taken once... and we're just consolidating that logic/behavior by doing it all (including the `expect` call) in `take_health_handle` instead of making every component do it themselves. [^1]

[^1]: All component context types have a `take_health_handle` method.

lib/saluki-core/src/components/renderers/mod.rs
tobz: I ended up switching to having both `Renderer` and adding a new `IncrementalRenderer` trait, which defines a method for processing a single event, and a method for flushing. Processing the event gives back a result that indicates if processing can proceed (or if a flush must be triggered) which I think could be a good middle ground for the moment.

This will let us implement the actual renderers in terms of how to process each event and flush, while allowing us to decouple all of the other thorny parts like buffering, shutdown, etc.

rayz (approved):
lgtm

tobz:
### Merge activity

* **May 19, 1:43 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/DataDog/saluki/679).

PR #691, opened 2025-05-16
Gustavo.caso/python check builder

GustavoCaso:
## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

### DIFF
--- a/lib/saluki-components/src/sources/checks/builder/python.rs
+++ b/lib/saluki-components/src/sources/checks/builder/python.rs
@@ -1,38 +1,302 @@
 //! A Python implementation of CheckBuilder
 
+use std::collections::HashMap;
+use std::path::Path;
 use std::sync::Arc;
+use std::time::Duration;
 
 use async_trait::async_trait;
-use saluki_env::autodiscovery::CheckConfig;
-use tokio::sync::OnceCell;
+use pyo3::types::{PyDict, PyList, PyNone, PyTuple, PyType};
+use pyo3::PyObject;
+use pyo3::{prelude::*, IntoPyObjectExt};
+use saluki_env::autodiscovery::{Data, Instance, RawData};
+use saluki_error::{generic_error, GenericError};
+use stringtheory::MetaString;
+use tracing::{debug, error, info, warn};
 
 use crate::sources::checks::builder::CheckBuilder;
 use crate::sources::checks::check::Check;
 
+#[allow(dead_code)]
+struct PythonCheck {
+    check_class: PyObject,
+    version: String,
+    interval: Duration,
+    instance: PyObject,
+    source: String,
+    id: String,
+}
+
+impl Check for PythonCheck {
+    fn run(&self) -> Result<(), GenericError> {
+        let result = pyo3::Python::with_gil(|py| self.instance.call_method(py, "run", (), None));
+        match result {
+            Ok(_) => Ok(()),
+            Err(e) => Err(generic_error!(e)),
+        }
+    }
+    /// Get the interval of the check
+    fn interval(&self) -> &Duration {
+        &self.interval
+    }
+    /// Get the id of the check
+    fn id(&self) -> &str {
+        &self.id
+    }
+}
+
 pub struct PythonCheckBuilder {
-    able_to_run_checks: OnceCell<()>,
+    valid: bool,
+    agent_check_class: Option<PyObject>,
 }
 
 impl PythonCheckBuilder {
     pub fn new() -> Self {
-        Self {
-            able_to_run_checks: OnceCell::new(),
-        }
-    }
+        pyo3::prepare_freethreaded_python();
+        let result = Python::with_gil(|py| -> Option<PyObject> {
+            if let Ok(sys) = py.import("sys") {
+                if let Ok(path_attr) = sys.getattr("path") {
+                    match path_attr.downcast::<PyList>() {
+                        Ok(p) => {
+                            // Get the project root to use as a base for relative paths
+                            let project_root = match std::env::current_exe() {
+                                Ok(path) => {
+                                    // Get the project root directory (two levels up from executable)
+                                    let mut project_root = path.clone();
+                                    // Go up from executable to debug dir
+                                    if let Some(parent) = project_root.parent() {
+                                        project_root = parent.to_path_buf();
+                                        // Go up from debug dir to target dir
+                                        if let Some(parent) = project_root.parent() {
+                                            project_root = parent.to_path_buf();
+                                            // Go up from target dir to project root
+                                            if let Some(parent) = project_root.parent() {
+                                                project_root = parent.to_path_buf();
+                                            }
+                                        }
+                                    }
+                                    project_root
+                                }
+                                Err(e) => {
+                                    error!("Failed to get current executable path: {}", e);
+                                    Path::new(".").to_path_buf()
+                                }
+                            };
 
-    async fn ensure_able_to_run_checks(&self) {
-        self.able_to_run_checks
-            .get_or_init(|| async {
-                // TODO: Implement this
-            })
-            .await;
+                            // Add paths relative to executable location
+                            let dist_path = project_root.join("dist");
+                            let checks_d_path = dist_path.join("checks.d");
+
+                            p.insert(0, &dist_path).unwrap(); // path.GetDistPath()
+                            p.insert(0, &checks_d_path).unwrap(); // custom checks in checks.d subdir
+                            p.insert(0, Path::new("/etc/datadog-agent/checks.d/")).unwrap(); // custom checks in checks.d subdir
+
+                            info!("Python sys.path is: {:?}", p);
+                        }
+                        Err(e) => {
+                            error!(%e, "Could not get sys.path");
+                            return None;
+                        }
+                    };
+                }
+            } else {
+                error!("Could not import sys module");
+                return None;
+            }
+
+            // Validate that python env is correctly configured
+            let modd = match py.import("datadog_checks.checks") {
+                Ok(m) => m,
+                Err(e) => {
+                    if let Some(traceback) = e.traceback(py) {
+                        error!("Traceback: {}", traceback.format().expect("Could format traceback"));
+                    }
+                    error!(%e, "Could not import datadog_checks module");
+                    return None;
+                }
+            };
+            match modd.getattr("AgentCheck") {
+                Ok(c) => {
+                    info!("All pre-requisites for running python checks.");
+                    Some(c.unbind())
+                }
+                Err(e) => {
+                    if let Some(traceback) = e.traceback(py) {
+                        error!("Traceback: {}", traceback.format().expect("Could format traceback"));
+                    }
+                    error!(%e, "Could not get AgentCheck class.");
+                    None
+                }
+            }
+        });
+
+        match result {
+            Some(agent_check_class) => Self {
+                valid: true,
+                agent_check_class: Some(agent_check_class),
+            },
+            None => Self {
+                valid: false,
+                agent_check_class: None,
+            },
+        }
     }
 }
 
 #[async_trait]
 impl CheckBuilder for PythonCheckBuilder {
-    async fn build_check(&self, _check_id: &str, _check_request: &CheckConfig) -> Option<Arc<dyn Check + Send + Sync>> {
-        self.ensure_able_to_run_checks().await;
-        None
+    async fn build_check(
+        &self, name: &str, instance: &Instance, init_config: &Data, source: &MetaString,
+    ) -> Option<Arc<dyn Check + Send + Sync>> {
+        if !self.valid {
+            warn!("Python check builder is not configured. Cannot build check.");
+            return None;
+        } else {
+            let mut load_errors = vec![];
+            for import_str in [name.to_string(), format!("datadog_checks.{}", name)].iter() {
+                match self.register_check_from_imports(name, import_str, instance, init_config, source) {
+                    Ok(handle) => return Some(handle),
+                    Err(e) => {
+                        load_errors.push(e.root_cause().to_string());
+                    }
+                }
+            }
+            error!("Could not load check {} errors: {:?}", name, load_errors);
+            return None;
+        }
+    }
+}
+
+impl PythonCheckBuilder {
+    fn register_check_from_imports(
+        &self, name: &str, import_path: &str, instance: &Instance, init_config: &Data, source: &MetaString,
+    ) -> Result<Arc<dyn Check + Send + Sync>, GenericError> {
+        pyo3::Python::with_gil(|py| {
+            let module = py.import(import_path)?;
+
+            let base_class = self
+                .agent_check_class
+                .as_ref()
+                .ok_or_else(|| generic_error!("PythonCheckBuilder is not configured. Cannot build check."))?;
+
+            let check_class = base_class.bind(py);
+
+            let checks = module
+                .dict()
+                .iter()
+                .filter(|(_name, value)| {
+                    let class_bound: &Bound<PyType> = match value.downcast() {
+                        Ok(c) => c,
+                        Err(_) => return false,
+                    };
+                    if class_bound.is(check_class) {
+                        // skip the base class
+                        return false;
+                    }
+                    if let Ok(true) = class_bound.is_subclass(check_class) {
+                        return true;
+                    }
+                    false
+                })
+                .collect::<Vec<_>>();
+            if checks.is_empty() {
+                return Err(generic_error!("No checks class found in source"));
+            }
+            if checks.len() >= 2 {
+                return Err(generic_error!("Multiple checks classes found in source"));
+            }
+            let (check_key, check_value) = &checks[0];
+            debug!("Found base class {check_key} for check {import_path} {check_value:?}");
+            let mut version = "unversioned".to_string();
+            if let Ok(check_version) = check_value.getattr("__version__") {
+                if let Ok(v) = check_version.extract::<String>() {
+                    version = v;
+                }
+            }
+
+            let kwargs = PyDict::new(py);
+            let parsed_config =
+                map_to_pydict(init_config.get_value(), &py).expect("Could not convert init_config to pydict");
+
+            let parse_instance =
+                map_to_pydict(instance.get_value(), &py).expect("Could not convert instance to pydict");
+
+            let py_vec = PyTuple::new(py, vec![parse_instance]).expect("Could not create tuple.");
+
+            kwargs.set_item("name", name).expect("Could not set name");
+            kwargs
+                .set_item("init_config", parsed_config)
+                .expect("could not set init_config");
+            kwargs
+                .set_item("instances", py_vec)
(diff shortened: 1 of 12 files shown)


lib/saluki-components/src/sources/checks/builder/python.rs
```diff
@@ -0,0 +1,312 @@
...
+                valid: false,
+                agent_check_class: None,
+            },
+        }
+    }
+}
+
+#[async_trait]
+impl CheckBuilder for PythonCheckBuilder {
+    async fn build_check(
```
tobz, line 154: Why did we make `CheckBuilder::build_check` async? I don't see us doing anything asynchronous here.

lib/saluki-components/src/sources/checks/builder/python.rs
```diff
@@ -0,0 +1,312 @@
...
+
+pub struct PythonCheckBuilder {
+    valid: bool,
+    agent_check_class: Option<PyObject>,
+}
+
+impl PythonCheckBuilder {
+    pub fn new() -> Self {
+        pyo3::prepare_freethreaded_python();
+        let result = Python::with_gil(|py| -> Option<PyObject> {
```
tobz, line 62: A lot of this code here feels like it ends up doing the error logging, etc, but then continuing on, eventually leading us to returning `Self` with `valid` either set to `true` or `false`.

Why aren't we just make `PythonCheckBuilder::new` return `Result<Self, GenericError>` instead? If checks are configured, and we can't initialize Python, it's hard to imagine why we'd want to continue with startup unless we think we can somehow resolve the problem by retrying, etcv... but there's no indication of that here.

lib/saluki-components/src/sources/checks/check.rs
```diff
@@ -5,11 +5,16 @@ use saluki_error::GenericError;
 /// Check trait
 ///
 /// This trait allow us to have different check implementations.
+#[allow(dead_code)]
 pub trait Check {
     /// Run the check
     fn run(&self) -> Result<(), GenericError>;
     /// Get the interval of the check
-    fn interval(&self) -> Duration;
+    fn interval(&self) -> &Duration;
```
tobz, line 13: `std::time::Duration` is `Copy`: there's almost never a good reason to pass a reference to it.

lib/saluki-components/src/sources/checks/check.rs
```diff
@@ -5,11 +5,16 @@ use saluki_error::GenericError;
 /// Check trait
 ///
 /// This trait allow us to have different check implementations.
+#[allow(dead_code)]
 pub trait Check {
     /// Run the check
```
tobz, line 10:
```suggestion
    /// Run the check.
    ///
    /// # Errors
    ///
    /// If a problem occurs while running the check, an error is returned.
```

lib/saluki-components/src/sources/checks/check.rs
```diff
@@ -5,11 +5,16 @@ use saluki_error::GenericError;
...
     fn run(&self) -> Result<(), GenericError>;
     /// Get the interval of the check
-    fn interval(&self) -> Duration;
+    fn interval(&self) -> &Duration;
     /// Get the id of the check
     fn id(&self) -> &str;
+    /// Get the version of the check
```
tobz, line 16: Add a trailing period to the doc comment.

lib/saluki-components/src/sources/checks/check.rs
```diff
@@ -5,11 +5,16 @@ use saluki_error::GenericError;
...
+    fn version(&self) -> &str;
+    /// Get the source of the check
```
tobz, line 18: Add a trailing period to the doc comment.

lib/saluki-components/src/sources/checks/scheduler.rs
```diff
@@ -72,7 +89,7 @@ impl Scheduler {
             self.start_interval_ticker(interval_secs);
         }
 
-        debug!("Scheduled check '{}' with interval {}s", check_id, interval_secs);
+        info!("Scheduled check '{}' with interval {}s", check_id, interval_secs);
```
tobz, line 92:
```suggestion
        info!(check_id, check_interval_secs = interval_secs, "Scheduled periodic check.", check_id);
```

We generally try to prefer using structured key/value pairs, because it lets us get structured logs for free.

lib/saluki-components/src/sources/checks/scheduler.rs
```diff
@@ -54,6 +55,22 @@ impl Scheduler {
...
+            if channels_guard.is_empty() {
+                warn!("Failed to schedule one-time check '{}': No workers available", check_id);
+                return;
+            }
+
+            let channel_idx = rand::thread_rng().gen_range(0..channels_guard.len());
+            if let Err(e) = channels_guard[channel_idx].try_send(WorkerMessage::RunCheck(Arc::clone(&check))) {
+                error!("Failed to enqueue a one-time check '{}': {}", check_id, e);
+            }
+            info!("Scheduled one-time check '{}' to worker {}", check_id, channel_idx);
```
tobz, line 70:
```suggestion
            info!(check_id, "Scheduled one-time check.");
```

Similar comment about structured key/value pairs as made below.

As far as not mentioning the worker index, I would say that we ought to just have each worker emit a log message about each check as they execute it. As is, if it's not relevant to mention the worker index for a scheduled check, doesn't seem worth it to mention it for a one-time check either? (This isn't a strongly held opinion but I think it ends up a little cleaner to just have workers log when they're running a check then the schedule logic itself emitting that.)

lib/saluki-components/src/sources/checks/scheduler.rs
```diff
@@ -238,6 +254,7 @@ impl Scheduler {
             for _ in 0..(desired_count - current_count) {
                 self.add_worker();
             }
+
             info!("Increased worker count to {}", desired_count);
```
tobz, line 258:
```suggestion
            info!("Check worker count updated to {}.", desired_count);
```

lib/saluki-components/src/sources/checks/builder/python.rs
```diff
@@ -0,0 +1,303 @@
...
+            if let Ok(check_version) = check_value.getattr("__version__") {
+                if let Ok(v) = check_version.extract::<String>() {
+                    version = v;
+                }
+            }
+
+            let min_interval = if let Some(value) = instance.get("min_collection_interval") {
+                value.as_u64().unwrap_or_default()
+            } else {
+                0
```
scottopell, line 226: default should likely be 15 seconds, 0 seconds will infinitely loop

lib/saluki-components/src/sources/checks/check.rs
```diff
@@ -5,11 +5,16 @@ use saluki_error::GenericError;
 /// Check trait
 ///
 /// This trait allow us to have different check implementations.
```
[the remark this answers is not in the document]
scottopell: and logs

lib/saluki-components/src/sources/checks/check.rs
```diff
@@ -5,11 +5,16 @@ use saluki_error::GenericError;
...
+#[allow(dead_code)]
 pub trait Check {
     /// Run the check
     fn run(&self) -> Result<(), GenericError>;
     /// Get the interval of the check
-    fn interval(&self) -> Duration;
+    fn interval(&self) -> &Duration;
     /// Get the id of the check
```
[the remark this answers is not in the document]
scottopell: I'd actually leave out the word "internal", the `check_id` is a concept that was ported directly from agent code and must match the logic in the Agent

lib/saluki-components/src/sources/checks/scheduler.rs
```diff
@@ -54,6 +55,22 @@ impl Scheduler {
         let check_id = check.id();
         let interval_secs = check.interval().as_secs();
 
+        if interval_secs == 0 {
+            // Check with no interval are push to a random worker immediately
+            let channels_guard = self.worker_channels.lock().unwrap();
```
scottopell, line 60: unwrap could be handled in the below check in an `if let` for better safety

docker/Dockerfile.agent-data-plane
```diff
@@ -1,6 +1,12 @@
+ARG DD_AGENT_VERSION=latest-jmx
```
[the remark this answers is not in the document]
GustavoCaso: Absolutely. Great call 
Just for the curious. To share some of why I did it at first is that we needed to pass the CI test, but then later learn about the internal images, and forgot to remove them.

lib/saluki-components/src/sources/checks/builder/python.rs
GustavoCaso: > 0 seconds will infinitely loop

Will schedule immediately, and not create an interval entry. 

Do we set up a default in the Agent (Go) 

https://github.com/DataDog/datadog-agent/blob/cbc65dc8c8e403204a371689c27148af95a4c25a/pkg/collector/python/check.go#L269-L271


```go
// See if a collection interval was specified
if commonOptions.MinCollectionInterval > 0 {
    c.interval = time.Duration(commonOptions.MinCollectionInterval) * time.Second
}
```

lib/saluki-components/src/sources/checks/builder/python.rs
GustavoCaso: It is a leftover from when I was trying to lazy-load the Python interpreter. I can remove it

GustavoCaso: I changed the function not to be async, and made the Python environment check lazy loaded using `OnceLock`
https://github.com/DataDog/saluki/pull/691/commits/df816042672a7d7b9c0a8d4e899ba5bafef74375

docker/Dockerfile.agent-data-plane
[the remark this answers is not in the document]
GustavoCaso: Removed 🔥

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -102,7 +102,8 @@ impl Source for ChecksSource {
         info!("Checks source started.");
 
         let mut event_rx = self.autodiscovery.subscribe().await;
-        let mut check_builders: Vec<Arc<dyn CheckBuilder + Send + Sync>> = vec![Arc::new(NoopCheckBuilder)];
+        let mut check_builders: Vec<Arc<dyn CheckBuilder + Send + Sync>> =
+            vec![Arc::new(PythonCheckBuilder::new()), Arc::new(NoopCheckBuilder)];
```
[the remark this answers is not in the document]
GustavoCaso: Removed `NoopCheckBuilder` 🔥

lib/saluki-components/src/sources/checks/check.rs
GustavoCaso: Thanks for the info. TIL

lib/saluki-components/src/sources/checks/scheduler.rs
```diff
@@ -54,6 +55,22 @@ impl Scheduler {
...
+            if channels_guard.is_empty() {
+                warn!("Failed to schedule one-time check '{}': No workers available", check_id);
+                return;
+            }
+
+            let channel_idx = rand::thread_rng().gen_range(0..channels_guard.len());
+            if let Err(e) = channels_guard[channel_idx].try_send(WorkerMessage::RunCheck(Arc::clone(&check))) {
+                error!("Failed to enqueue a one-time check '{}': {}", check_id, e);
```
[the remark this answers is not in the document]
GustavoCaso: Completely agree with this. I would like this PR to focus solely on adding the Python bindings, and work on a separate PR to modify the scheduler. Can we change that on a follow-up PR?

docker/Dockerfile.agent-data-plane
```diff
@@ -78,7 +82,7 @@ USER root
 RUN test -d /usr/share/ca-certificates || (apt-get update && \
     apt-get install -y --no-install-recommends ca-certificates=20240203~22.04.1 && \
     apt-get clean && rm -rf /var/lib/apt/lists)
-COPY --from=builder /adp/target/${BUILD_PROFILE}/agent-data-plane /usr/local/bin/agent-data-plane
+COPY --from=builder /adp/target/bin/agent-data-plane /usr/local/bin/agent-data-plane
```
tobz, line 85: Why are we still modifying the ADP Dockerfile? 😅

lib/saluki-components/src/sources/checks/builder/python.rs
```diff
@@ -0,0 +1,312 @@
...
+                agent_check_class: Some(agent_check_class),
+            },
+            Err(e) => {
+                warn!("Could not load Python runtime. Cannot build Python check. {:?}", e);
+                Self {
+                    valid: false,
+                    agent_check_class: None,
+                }
+            }
+        }
```
tobz, line 153: I'm still not convinced that this is the right pattern.

Why are we allowing execution to continue if the Python interpreter/environment fails to initialize properly?

docker/Dockerfile.agent-data-plane
GustavoCaso: Ohh, that was done previously to allow us not to rely on the BUILD_PROFILE on the last build step. I can remove it if you prefer. The final build is not affected nor includes the Python interpreter

lib/saluki-components/src/sources/checks/builder/python.rs
GustavoCaso: > Why are we allowing execution to continue if the Python interpreter/environment fails to initialize properly?

That is how the Go Agent works; if it fails to load the Python runtime, it does not crash, it simply continues. In this case, if we were to have other check builders we would like to try to build the check with those builders.

docker/Dockerfile.agent-data-plane
tobz: Yeah, I'd rather just leave the file as-is if we don't _need_ to change it.

lib/saluki-components/src/sources/checks/builder/python.rs
tobz: Fair enough.

lib/saluki-components/src/sources/checks/builder/python.rs
```diff
@@ -0,0 +1,312 @@
...
+    }
+}
```
tobz, line 155:
```suggestion
struct PythonEnvBuilder {
    agent_check_class: Option<PyObject>,
}

impl PythonEnvBuilder {
    fn get_instance() -> &'static PythonEnvBuilder {
        PYTHON_CHECK_BUILDER.get_or_init(PythonEnvBuilder::initialize)
    }

    fn initialized(&self) -> bool {
        self.agent_check_class.is_some()
    }

    fn initialize() -> Self {
        pyo3::prepare_freethreaded_python();
        let result = Python::with_gil(|py| -> Result<PyObject, GenericError> {
            let sys_path_attr = py.import("sys")
                .error_context("Could not import sys module.")
                .and_then(|sys| sys.getattr("path").error_context("Could not get 'sys.path' attribute."))?;

            let sys_path = sys_path_attr.downcast::<PyList>()
                .map_err(|_| GenericError::msg("Could not downcast 'sys.path' to list."))?;

            // Add additional paths to sys.path to support loading checks.
            let dist_path = Path::new("./dist");
            let checks_d_path = dist_path.join("checks.d");

            sys_path.insert(0, dist_path.to_string_lossy().as_ref()).unwrap(); // common modules are shipped in the dist path directly or under the "checks/" sub-dir
            sys_path.insert(0, checks_d_path.to_string_lossy().as_ref()).unwrap(); // integrations-core legacy checks
            sys_path.insert(0, "/etc/datadog-agent/checks.d/").unwrap(); // Agent checks folder

            debug!("Updated Python system path (sys.path) to {:?}.", sys_path);

            // Import the Datadog Checks module, ensuring it loads correctly, and grab a reference to the base AgentCheck class.
            let dd_checks_module = match py.import("datadog_checks.checks") {
                Ok(m) => m,
                Err(e) => {
                    if let Some(traceback) = e.traceback(py) {
                        error!("Traceback: {}", traceback.format().expect("Could format traceback"));
                    }
                    return Err(generic_error!("Could not import datadog_checks module {:?}", e));
                }
            };

            dd_checks_module.getattr("AgentCheck")
                .map(|c| c.unbind())
                .error_context("Could not get AgentCheck class from datadog_checks module.")
        });

        match result {
            Ok(agent_check_class) => {
                info!("Python runtime loaded successfully and initialized for checks.");
                Self {
                    agent_check_class: Some(agent_check_class),
                }
            },
            Err(e) => {
                warn!(error = %e, "Failed to load/initialize Python runtime for checks. Python checks will be unavailable.");
                Self {
                    agent_check_class: None,
                }
            }
        }
    }
}
```

GustavoCaso: Beautiful 🎉

lib/saluki-components/src/sources/checks/builder/python.rs
```diff
@@ -0,0 +1,276 @@
...
+
+pub struct PythonCheckBuilder;
+
+impl CheckBuilder for PythonCheckBuilder {
+    fn build_check(
+        &self, name: &str, instance: &Instance, init_config: &Data, source: &MetaString,
+    ) -> Option<Arc<dyn Check + Send + Sync>> {
+        let env = PythonEnvBuilder::get_instance();
+        if env.initialized() {
```
tobz, line 124: We generally prefer to have preconditions early on to let us return early. For example, if we know the Python environment isn't initialized, we should return `None` early:

```rust
let env = PythonEnvBuilder::get_instance();
if !env.initialized() {
    return None;
}

let agent_check_class = ...
```

lib/saluki-components/src/sources/checks/scheduler.rs
```diff
@@ -238,7 +257,8 @@ impl Scheduler {
             for _ in 0..(desired_count - current_count) {
                 self.add_worker();
             }
-            info!("Increased worker count to {}", desired_count);
+
+            info!(worker_count = desired_count, "Check worker count updated to.");
```
tobz, line 261: Ahh, looks like I left an extra word in my original suggestion:

```suggestion
           info!(worker_count = desired_count, "Check worker count updated.");
```

lib/saluki-components/src/sources/checks/check.rs
```diff
@@ -2,14 +2,25 @@ use std::time::Duration;
 
 use saluki_error::GenericError;
 
-/// Check trait
+/// A check.
 ///
-/// This trait allow us to have different check implementations.
+/// Checks run some arbitrary chunk of logic on a configured interval, potentially producing outputs
+/// such as service checks, events, metrics, and logs.
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
GustavoCaso: We are not referencing `version` nor `source`. 😢 
On a follow up PR we would reference them 😄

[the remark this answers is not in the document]
tobz: Ah, gotcha. Makes sense.

tobz (approved):

PR #696, opened 2025-05-19
enhancement(context): cheaply clone incoming context name/tags where possible

tobz:
## Summary

This PR updates `ContextResolver` to cheaply clone the name and/or tags of a context, when possible, based on changes to `MetaString` and a new trait, `CheapMetaString`, which exposes the ability to determine when a string value can be cheaply cloned to `MetaString`.

In many cases, components need to re-resolve a metric context as they change the tags or remap the name of a metric. For metric names/tags that can inlined, this is effectively free. However, when the strings cannot be inlined, the context resolver must fallback to either interning or heap allocating the values. This means that even in cases where the original name/tag was static or already interned, we will re-intern it (or worse, heap allocate it), wasting additional capacity for a `MetaString` that was already able to be cheaply cloned.

In this PR, we've done three main things:

- expose a new method on `MetaString`, `is_cheaply_cloneable`, to allow detecting when a clone will avoid a heap allocation
- add a new trait, `CheapMetaString`, to abstract over various string/string-y types and allow for returning cheap clones (based on `MetaString`) of those types if supported
- use that trait in `ContextResolver` to cheaply clone the context name and/or tags when possible

After this PR, when the name/tags being fed to `ContextResolver::resolve` (or `ContextResolver::resolve_with_origin_tags`) can be determined to be cheaply cloneable, we do so. This means that, for example, if we're handling a bunch of previously-interned tags, and adds some new tags that we've interned ourselves (or they're static, or inlined, etc), then re-resolving a context with those tags should result in no duplicate interner entries or heap allocations.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Added a new unit test to verify that the context resolver's interner is not touched when the input name/tags are cheaply cloneable.

## References

AGTMETRICS-233

### DIFF
--- a/lib/saluki-context/src/resolver.rs
+++ b/lib/saluki-context/src/resolver.rs
@@ -871,4 +871,43 @@ mod tests {
         assert_eq!(context1, context2);
         assert!(!context1.ptr_eq(&context2));
     }
+
+    #[test]
+    fn cheaply_cloneable_tags() {
+        const BIG_TAG_ONE: &str = "long-tag-that-cannot-be-inlined-just-to-be-doubly-sure-on-top-of-being-static";
+        const BIG_TAG_TWO: &str = "another-long-boye-that-we-are-also-sure-wont-be-inlined-and-we-stand-on-that";
+
+        // Create a context resolver with a proper string interner configured:
+        let mut resolver = ContextResolverBuilder::for_tests()
+            .with_interner_capacity_bytes(NonZeroUsize::new(1024).expect("not zero"))
+            .build();
+
+        // Create our context with cheaply cloneable tags, aka static strings:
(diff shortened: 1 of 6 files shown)


lib/saluki-context/src/resolver.rs
```diff
@@ -564,6 +567,81 @@ async fn drive_telemetry(context_cache: Arc<ContextCache>, interner: GenericMapI
     }
 }
 
+mod private {
+    pub trait Sealed {}
```
rayz, line 571: TIL: Sealed Traits (though searching the codebase I see we use this concept elsewhere too :hmmnod: )

rayz (approved):

PR #698, opened 2025-05-20
enhancement(stringtheory): add new `MetaString` variant for `Arc<str>`-backed strings

### LINKED ISSUE #697: Update `MetaString` to support "shared" variant based on `Arc<str>`.
## Context

Currently, `MetaString` only supports sharing of semi-owned strings when they're interned, which is effectively a specialized form of `Arc<str>`.

It would be useful, however, to allow for literally using `Arc<str>` values to back `MetaString`, as in some cases it would be otherwise toilsome/noisy to whip up a string interner just to intern a few strings that you want to cheaply share.

This should be possible as our tagging scheme supports a large number of niches (~100) and we only use two of them at the moment for static strings and interned strings.

tobz:
## Summary

This PR adds support to `MetaString` to abstract over `Arc<str>` by adding a new variant in our tag byte niche. This is a simpler approach when intern-like behavior is required but the number of strings is small, and the use of an interner would be overkill compared to simply creating a few `Arc<str>` instances by hand.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests.

## References

AGTMETRICS-233

Closes #697

### DIFF
--- a/lib/stringtheory/src/lib.rs
+++ b/lib/stringtheory/src/lib.rs
@@ -209,7 +209,7 @@ impl DiscriminantUnion {
         // - Static strings and interned strings only occupy the first two fields, which means their capacity should not
         //   be used.
         //
-        // As such, we encode the five possible string types as follows:
+        // As such, we encode the possible string types as follows:
         //
         // - when all fields are zero, we have an empty string
         // - when the last byte does have the top-bit set, we have an owned string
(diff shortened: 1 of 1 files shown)


rayz (approved):
https://github.com/DataDog/saluki/blob/3f316e652886aa44558813994da344d2fc0ba7cd/lib/stringtheory/src/lib.rs#L297

https://github.com/DataDog/saluki/blob/3f316e652886aa44558813994da344d2fc0ba7cd/lib/stringtheory/src/lib.rs#L212

nit: six now?

tobz:
> nit: six now?

Good call, you right. Will fix.

PR #711, opened 2025-05-22
[AGENTRUN-330] feature(checks-agent): allow python checks to send metrics

GustavoCaso:
## Summary

We create the necessary Python modules for our dummy Python to work. Those modules include the `datadog_agent` and `aggregator`

We are not implementing all the functions of the modules; we are only adding the necessary code for sending metrics (We  are not going to deploy this in production; we are gathering numbers for running Python in Rust)

This works show how we could add the different Python modules using the pyo3 crate

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance


lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -158,6 +166,19 @@ impl Source for ChecksSource {
...
+                    let mut event_buffer = context.event_buffer_pool().acquire().await;
+                    let event: Event = check_metric.try_into().expect("can't convert");
+                    if let Some(_unsent) = event_buffer.try_push(event) {
+                        error!("Event buffer full, dropping event");
+                        continue;
+                    }
+                    if let Err(e) = context.dispatcher().dispatch_buffer(event_buffer).await {
+                        error!(error = %e, "Failed to forward check metrics.");
+                    }
                 }
```
tobz, line 182: I would move this to a background task that only focuses on draining check metrics and forwarding them. Something like:

```rust
async fn drain_and_dispatch_check_metrics(context: SourceContext, check_metrics_rx: mpsc::Receiver<CheckMetric>) {
    let mut flush_interval = tokio::time::interval(Duration::from_secs(1));
    let mut buffered_dispatcher = context.dispatcher().buffered().expect("checks source should always have default output");

    loop {
        select! {
            result = check_metrics_rx().recv() => match result {
                None => break,
                Some(check_metric) => {
                    trace!("Received check metric: {:?}", check_metric);
                    let event: Event = check_metric.try_into().expect("can't convert");
                    if let Err(e) = buffered_dispatcher.push(event).await {
                        error!(error = %e, "Failed to forward check metrics.");
                    }
                }
            },

            _ = &mut flush_interval => {
                match buffered_dispatcher.flush().await {
                    Ok(()) => debug!("Flushed pending check metrics."),
                    Err(e) => error!(error = %e, "Failed to flush check metrics."),
                }
            }
        }
    }

    // Final dispatcher flush after all checks have closed.
    if let Err(e) = buffered_dispatcher.flush().await {
        error!(error = %e, "Failed to flush check metrics.");
    }
}
```

This approach gives us a few benefits, some objective and some subjective:

- we decouple the handling of check metrics from the source's main loop, so less risk of backpressure due to dispatching which means less chance of responding slowly to autodiscovery events or liveness probes
- we batch check metrics by using `BufferedDispatcher` and flushing on a short interval, which saves on event buffer pool acquisitions and helps downstream processing efficiency
- `BufferedDispatcher` also handles all the logic of acquiring event buffers and flushing them if full, so we only need to care about pushing in new events, and triggering a flush if we want to ensure metrics aren't buffered for too long

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -101,8 +106,11 @@ impl Source for ChecksSource {
 
         info!("Checks source started.");
 
+        let (check_metrics_tx, mut check_metrics_rx) = mpsc::channel(10_000_000);
```
tobz, line 109: Bounded channels: good.
Bounded at 10 million items: no no no no no no. :P

You can almost always start with something closer to like 64 - 128 for a bounded channel.

lib/saluki-components/src/sources/checks/mod.rs
GustavoCaso: Amazing. Thanks for the suggestion.

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub sender: Sender<CheckMetric>,
+}
+
+/// submit_metric is called from the AgentCheck implementation when a check submits a metric.
+/// Python signature:
+///     aggregator.submit_metric(self, self.check_id, mtype, name, value, tags, hostname, flush_first_value)
+#[allow(clippy::too_many_arguments)]
+#[pyfunction]
+#[pyo3(pass_module)]
+pub(crate) fn submit_metric(
```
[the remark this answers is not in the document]
GustavoCaso: I've used the `#[pymodule]` macro to create the aggregator module, which is way more declarative than the previous way. 

As for the internal state eg, the metrcis sender queue. I added some functions to store the state for the Python checks, and changed the `submit_metric` function to check that global state instead of looking at the module attributes, reducing the jumps between Rust and Python to just once 😄 

Let me know what you think

lib/saluki-components/src/sources/checks/builder/python/builder.rs
```diff
@@ -54,19 +58,32 @@ impl Check for PythonCheck {
...
+        Self {
+            check_events_tx,
+            custom_checks_folders,
+            agent_check_class: None,
+        }
+    }
+
+    fn get_instance(&self) -> &'static PythonEnvBuilder {
+        PYTHON_ENV_BUILDER.get_or_init(|| self.initialize())
     }
```
tobz, line 76: This is a little weird, having `PythonEnvBuilder::new` and `PythonEnvBuilder::initialize` both return `Self`.

I would just do something like:

- change `PYTHON_ENV_BUILDER` to `PYTHON_ENV_INITIALIZE: OnceLock<()>`
- change `PythonEnvBuilder::new` to do all of the initialization logic and return `Result<Self, GenericError>`
- in `PythonEnvBuilder::new`, have it try to claim the right to initialize the Python interpreter by calling `if PYTHON_ENV_INITIALIZE.set(()).is_err() { ... }` and returning an error if that call fails
- otherwise, continue with initialization and eventually return `Self` at the end

This will ensure we're guarding against multiple initializations _and_ make the code a bit clearer, I think.

tobz (changes requested):

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+fn get_or_init_sender_storage() -> &'static Arc<Mutex<Option<Sender<CheckMetric>>>> {
+    METRIC_SENDER.get_or_init(|| Arc::new(Mutex::new(None)))
+}
+
+fn get_metric_sender() -> Option<Sender<CheckMetric>> {
+    let storage = get_or_init_sender_storage();
+    storage.lock().ok().and_then(|guard| guard.clone())
+}
```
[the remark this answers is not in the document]
GustavoCaso: I've used a mutex because multiple checks could be running simultaneously, as the check scheduler have various workers. 

Since we have a single `Sender` for multiple `tokio::spawn` tasks, I thought we would need to use a Mutex. 

Was that a wrong assumption?

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -0,0 +1,158 @@
...
+
+use pyo3::prelude::*;
+use saluki_error::{generic_error, GenericError};
+use tokio::sync::mpsc::Sender;
+use tracing::{debug, error, trace};
+
+use crate::sources::checks::check_metric::{CheckMetric, MetricType};
+
+// Global state to store the sender
+static METRIC_SENDER: OnceLock<Arc<Mutex<Sender<CheckMetric>>>> = OnceLock::new();
```
tobz, line 11: You still don't need a `Mutex` for this.

All the sender methods on `Sender` are already safe to use from multiple callers (`&self`) and the initialization of `METRIC_SENDER` itself is already synchronized by `OnceLock`.

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -0,0 +1,158 @@
...
+        );
+
+        let check_metric = CheckMetric::new(name.clone(), mtype.into(), value, tags.clone());
+
+        match try_send_metric(check_metric) {
+            Ok(()) => {}
+            Err(e) => {
+                error!("Failed to send metric: {}", e);
+            }
+        };
```
tobz, line 70:
```suggestion
        if let Err(e) = try_send_metric(check_metric) {
            error!("Failed to send metric: {}", e);
        };
```

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -0,0 +1,158 @@
...
+    fn set_hostname(hostname: String) {
+        trace!("Called set_hostname({})", hostname);
+        // In a function context without a struct, we cannot actually "set" the hostname persistently.
+    }
+
+    #[pyfunction]
+    fn reset_hostname() {
+        trace!("Called reset_hostname()");
+        // Similar to `set_hostname`, we cannot reset without a persistent structure.
+    }
```
tobz, line 114: We let checks set and reset the hostname.... what the hell? 😂 

How does this checks-specific hostname actually get utilized?

lib/saluki-components/src/sources/checks/check_metric.rs
```diff
@@ -0,0 +1,113 @@
...
+
+        let context = Context::from_parts(check_metric.name, tagset);
+        let metadata = MetricMetadata::default();
+
+        match check_metric.metric_type {
+            MetricType::Gauge => Event::Metric(Metric::from_parts(
+                context,
+                MetricValues::gauge(check_metric.value),
+                metadata,
+            )),
```
tobz, line 76: Probably a little cleaner to do something like:

```rust
let values = match check_metric.metric_type {
    MetricType::Gauge => MetricValues::gauge(check_metric.value),
    ...
}:

Event::Metric(Metric::from_parts(context, values, metadata))
```

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -173,8 +192,55 @@ impl Source for ChecksSource {
...
+
+    let buffered_dispatcher = context
+        .dispatcher()
+        .buffered()
+        .expect("checks source should always have default output");
+
+    // Final dispatcher flush after all checks have closed.
+    if let Err(e) = buffered_dispatcher.flush().await {
+        error!(error = %e, "Failed to flush check metrics.");
+    }
```
tobz, line 245: No need for this: you're constructing it fresh so it's only ever going to be empty.

lib/saluki-components/src/sources/checks/check_metric.rs
```diff
@@ -0,0 +1,113 @@
...
+    tags::{Tag, TagSet},
+    Context,
+};
+use saluki_core::data_model::event::metric::{Metric, MetricMetadata, MetricValues};
+use saluki_core::data_model::event::Event;
+use tracing::warn;
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum MetricType {
+    Gauge = 0,
```
tobz, line 13: No need for the common Go pattern of setting the value of the first variant... since it will already be zero if you cast the variant to `i32` (e.g., `MetricType::Gauge as i32`).

Only relevant if you want the offset to be something other than 0.

tobz (changes requested):

lib/saluki-components/src/sources/checks/check_metric.rs
GustavoCaso: Noted! Thanks for the info 😄

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
GustavoCaso: I copied these methods from @scottopell work during innovation week. (Shame on me for not double checking)

I believe these methods are not part of the building we offer to the Python checks. Doing a bit of search, I see we have them as part of the `DatadiogAegntStub ' in the integrations-core repo, probably for testing. 

https://github.com/DataDog/integrations-core/blob/98186cfdf5231a4efab3fbd05cae8f105da5430c/datadog_checks_base/datadog_checks/base/stubs/datadog_agent.py#L94-L98

I will remove them. 

> We let checks set and reset the hostname.... what the hell? 😂
That would be crazy if we allow Python to mess up with the hostname

lib/saluki-components/src/sources/checks/check_metric.rs
GustavoCaso: Done https://github.com/DataDog/saluki/pull/711/commits/c5233ba99425a31001a200bc781b9cd36ea9e22a

lib/saluki-components/src/sources/checks/check_metric.rs
GustavoCaso: Done

tobz (approved):

PR #735, opened 2025-06-04
[AGENTRUN-350] chore(checks-agent): remove blackhole destination and use a real destination

GustavoCaso:
## Summary

On the previous PR we added support for sending metrics from the Python checks https://github.com/DataDog/saluki/pull/711

This PR modifies the `checks-agent` binary to use the transformers (Aggregation, and Host Enrich) and send the data to the DD metrics intake. 

The aggregation transformer might require some changes to make sure the check metrics are emitted on the right interval. Nevertheless, for our benchmarking work, accuracy is not as important 😄 

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

### DIFF
--- a/bin/checks-agent/src/env_provider.rs
+++ b/bin/checks-agent/src/env_provider.rs
@@ -1,10 +1,14 @@
 use memory_accounting::ComponentRegistry;
 use saluki_config::GenericConfiguration;
-use saluki_env::autodiscovery::providers::{
-    BoxedAutodiscoveryProvider, LocalAutodiscoveryProvider, RemoteAgentAutodiscoveryProvider,
+use saluki_env::{
+    autodiscovery::providers::{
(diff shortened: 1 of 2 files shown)


bin/checks-agent/src/env_provider.rs
```diff
@@ -69,10 +74,24 @@ impl ChecksAgentEnvProvider {
...
+impl EnvironmentProvider for ChecksAgentEnvProvider {
+    type Host = BoxedHostProvider;
+    type Workload = Option<RemoteAgentWorkloadProvider>;
+
+    fn host(&self) -> &Self::Host {
+        &self.host_provider
+    }
+
+    fn workload(&self) -> &Self::Workload {
+        &self.workload_provider
```
tobz, line 95: Just have this return `&None`, since we never actually set `self.workload_provider` to anything other than `None`... and then you can just skip having that field entirely. :D

tobz (approved):
Looks good to me, with one very small nit.

PR #744, opened 2025-06-06
chore(common): move time-related functions from `saluki-env` + add coarse time support

tobz:
## Summary

This PR does two main things: moves time functions out of `saluki-env` into `saluki-common`, and adds support for "coarse" time.

Coarse time is simply a way to get the current time from a cached source with an acceptable loss in accuracy in order to avoid the overhead of getting the current time. In this case, we use a standard approach of spawning a background thread which updates an atomic, every 250ms, with the current time. Since we only have second granularity for timestamps, this means we may be behind by up to 250ms, but most usages don't necessary care about that.

For example, cache expiration (which this was added with in mind) is already somewhat lossy, in that values aren't _immediately_ removed when expiration occurs, but some short while after... now we may just remove values 1-2 seconds later, when using coarse time, due to added accuracy error... but recording those timestamps via coarse time will be vastly cheaper, CPU-wise.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

N/A

## References

AGTMETRICS-233

### DIFF
--- a/lib/saluki-context/src/expiry.rs
+++ b/lib/saluki-context/src/expiry.rs
@@ -1,12 +1,12 @@
 use std::{
     marker::PhantomData,
     sync::{Arc, Mutex},
-    time::{Duration, SystemTime},
+    time::Duration,
 };
 
 use crossbeam_queue::ArrayQueue;
 use quick_cache::Lifecycle;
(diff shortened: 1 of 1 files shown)


rayz (approved):
lgtm

PR #748, opened 2025-06-09
enhancement(common): add simple concurrent cache implementation

tobz:
## Summary

This PR adds a simple, concurrent cache implementation to `saluki-common`. This implementation is based on the caching setup in `saluki-context`, which uses `quick_cache` under the hood. It maintains a number of things from the original implementation, and adds/simplifies others:

- dedicated builder type (`CacheBuilder`) for configuring a cache
- dedicated telemetry for each cache instance (current size, limit, hits/misses, expirations, etc)
- support for time-to-idle expiration (based on a `BP-Wrapper`-inspired approach for reduced lock contention)
- support for customized item "weights" to support limiting the cache by total number of items, or by total memory used by items, or any other key/value-derivable indicators

We've left the cache implementation used in `saluki-context` as-is for the moment, and will switch to the `saluki-common`-based implementation in a follow-up PR.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Limited number of unit tests.

Expiration and telemetry are still untested as they depend on mockable time, and aren't using time from `tokio` which is currently the only approach we have for mocking time.

This will be resolved in a follow-up PR.

## References

AGTMETRICS-233


lib/saluki-common/src/cache/expiry.rs
```diff
@@ -0,0 +1,251 @@
...
+            // Calculate the cutoff time for entries to be considered expired.
+            //
+            // Our math here shouldn't ever fail, but if it does, we just return and wait for the next attempt.
+            let now = get_unix_timestamp();
+            let last_seen_cutoff = match now.checked_sub(inner.time_to_idle.as_secs()) {
+                Some(cutoff) => cutoff,
+                None => return,
+            };
+
+            let expired_entries = inner
```
rayz, line 209: Should we call `process_pending_ops` before draining? Since wouldn't there be some `ExpirationOp::Accessed` that might be in the queue which would affect the access time and so potentially some entries would be removed even if they were accessed recently?

tobz: Good catch!

This would indeed tighten up how precise we are with expirations. I'll fix this up.

rayz (approved):

PR #749, opened 2025-06-09
experiment(context): support structural sharing via chained pointers in `SharedTagSet`

tobz:
## Summary

When dealing with the tags of a metric, the most common scenario is that a metric is ingested with its own set of tags, and then an additional set (or sets) of tags are added based on the metric's "origin": the container and/or Kubernetes pod it originated from. Origin tags are generally stored in a map, keyed by an "entity ID": container ID, pod UID, and so on. This means that when resolving all of the origin tags for a given metric, we might be grabbing anywhere from 1-5 different sets of tags with the aim to combine them all together when rendering a metric, and so on.

In the last major refactor to origin tags, we provided an efficient interface for querying over this dynamic number of underlying tags through a visitor pattern. However, this pattern is limiting because if we want to store all of the tags together, in a single concrete object... we would have to allocate a new `TagSet` to do so. All of the origin tags are already cheaply shareable via `SharedTagSet`, but what we lack is a way to cheaply hold a _set_ of those shared tag sets to satisfy our "single concrete object" constraint.

This PR refactors `SharedTagSet` to support "structural sharing", where we hold a vector of _pointers_ to the underlying tagsets, and then provide convenience methods over the top of that. Before, combining two `SharedTagSet`s together would mean allocating a new `TagSet` and copying the tags from each shared tagset into it. Now, extending `SharedTagSet` from another `SharedTagSet` is possible by allocating a much smaller underlying vector, one that only grows based on the number of chained tagsets rather than the overall number of tags.

Our design utilizes `smallvec` instead of `Vec<T>`. This lets us hold at least two tagset pointers in `SharedTagSet` _without_ any underlying allocation as `smallvec` is able to inline them. After two pointers, the vector will have to be expanded, which will cause an allocation... but again, this does not scale with the number of tags, but the number of chained tagset pointers.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Existing unit tests.

## References

AGTMETRICS-233


rayz (approved):
lgtm

PR #758, opened 2025-06-11
enhancement: add subcommand support to ADP binary

brianna-wang-dd:
## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

### DIFF
--- a/bin/agent-data-plane/src/config.rs
+++ b/bin/agent-data-plane/src/config.rs
@@ -39,4 +39,4 @@ pub struct RunConfig {
     /// Path to the configuration file
     #[arg(long = "config", default_value = "/etc/datadog-agent/datadog.yaml")]
     pub config: PathBuf,
-} 
+}
--- a/bin/agent-data-plane/src/config.rs
+++ b/bin/agent-data-plane/src/config.rs
@@ -1,0 +1,42 @@
+use std::path::PathBuf;
+
+use clap::{ArgAction, Args, Parser, Subcommand};
+use tracing::level_filters::LevelFilter;
+
+#[derive(Parser)]
+#[command(about)] // Q: Is a description needed here?
+pub struct Cli {
+    /// Enable verbose output. (Specify twice for more verbosity.)
+    #[arg(global = true, short = 'v', long, action = ArgAction::Count, default_value_t = 0)]
+    verbose: u8,
+
+    /// Subcommand to run.    
+    #[command(subcommand)]
+    pub action: Action,
+}
+
+impl Cli {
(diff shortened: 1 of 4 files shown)


bin/agent-data-plane/src/config.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// Run subcommand for the ADP binary
+    #[command(name = "run")]
+    Run(RunConfig),
+}
+
+/// Run subcommand configuration
+#[derive(Args, Debug)]
+pub struct RunConfig {
+    /// Path to the configuration file
+    #[arg(long = "config", default_value = "/etc/datadog-agent/datadog.yaml")]
```
rayz, line 28: Can we add `short` too? Agent docs show `run -c bin/agent/dist/datadog.yaml` so it would be nice to have that as well.

bin/agent-data-plane/src/main.rs
```diff
@@ -65,16 +70,31 @@ async fn main() {
         fatal_and_exit(format!("failed to initialize TLS: {}", e));
     }
 
-    match run(started).await {
-        Ok(()) => info!("Agent Data Plane stopped."),
-        Err(e) => {
-            error!("{:?}", e);
-            std::process::exit(1);
+    match cli.action {
```
rayz, line 73: I think we can reduce some of this duplicated code by defining `let run_config: RunConfig` outside the `match` and then just assigning to it.

bin/agent-data-plane/src/config.rs
```diff
@@ -0,0 +1,30 @@
...
+    verbose: u8,
+
+    /// Subcommand to run.    
+    #[command(subcommand)]
+    pub action: Option<Action>, // Makes subcommands optional
+}
+
+#[derive(Subcommand)]
+pub enum Action {
+    /// Run subcommand for the ADP binary
```
tobz, line 19:
```suggestion
    /// Runs the data plane.
```

bin/agent-data-plane/src/config.rs
```diff
@@ -0,0 +1,30 @@
...
+    #[command(name = "run")]
+    Run(RunConfig),
+}
+
+/// Run subcommand configuration
+#[derive(Args, Debug)]
+pub struct RunConfig {
+    /// Path to the configuration file
```
tobz, line 27: Style note: For user-facing doc comments, which is anything that starts with `///`, we strive to have full/complete sentences, including proper punctuation.

So this needs a period at the end.

tobz (changes requested):

rayz (approved):

tobz (approved):

PR #767, opened 2025-06-16
enhancement: add log level override subcommand to ADP binary

### LINKED ISSUE #731: Add log level override subcommand to ADP binary.
## Context

We should add two debug subcommands for dynamically configuring the log level of the ADP process: `debug set-log-level` and `debug reset-log-level`.

For setting the log level, this should be `agent-data-plane debug set-log-level <filter directives> <duration in seconds>`. The subcommand would call out to `/logging/override`. We would want to make sure we take care to report back errors to the caller, since the API endpoint enforces a limit on how long the duration can be.

For resetting the log level, this should be `agent-data-plane debug reset-log-level`. The subcommand would call out to `/logging/reset`.

(Depends on #730.)

brianna-wang-dd:
## Summary
This PR adds two debug subcommands for dynamically configuring the log level of the ADP process: `debug set-log-level` and `debug reset-log-level`.

For setting the log level, this should be `agent-data-plane debug set-log-level`. The subcommand would call out to /logging/override. For resetting the log level, this should be `agent-data-plane debug reset-log-level`. The subcommand would call out to /logging/reset.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Manually tested by running the run-adp-standalone and testing various set and reset commands

## References

Closes #731.


bin/agent-data-plane/src/config.rs
```diff
@@ -15,6 +15,10 @@ pub enum Action {
     /// Runs the data plane.
     #[command(name = "run")]
     Run(RunConfig),
+
+    /// Configures the log level of the data plane.
```
tobz, line 19: Since this will end up eventually having more sub-subcommands related to debugging, better to say something generic, along the lines of `Various debugging commands.`, etc.

bin/agent-data-plane/src/config.rs
```diff
@@ -24,3 +28,27 @@ pub struct RunConfig {
...
 }
+
+/// Debug subcommand configuration.
+#[derive(Subcommand, Debug)]
+pub enum DebugConfig {
+    /// Resets log level.
+    #[command(name = "reset-log-level")]
+    ResetLogLevel,
+
+    /// Overrides the default log filtering directives.
```
tobz, line 39:
```suggestion
    /// Overrides the current log level.
```

bin/agent-data-plane/src/config.rs
```diff
@@ -24,3 +28,27 @@ pub struct RunConfig {
...
+}
+
+/// Set log level configuration.
+#[derive(Args, Debug)]
+pub struct SetLogLevelConfig {
+    /// Filter directives to apply.
+    #[arg(required = true)]
+    pub filter_directives: String,
+
+    /// Duration in seconds for which the override will be applied.
```
tobz, line 51:
```suggestion
    /// Amount of time to apply the log level override, in seconds.
```

tobz (approved):
LGTM.

We'll want to follow this up with a PR to improve the subcommand documentation (e.g., explain what sort of filtering directives are possible) as well as split out the logic for each subcommand into dedicated modules/files so it's not all clumped together in `main.rs`... but this is fine for now. 👍🏻

Issue #499, opened 2025-02-13
Failure to visit cgroups hierarchy due to missing file.


tobz:
Saw this error log in staging complaining about a missing file when trying to visit the cgroups hierarchy:

```
yyyy-MM-dd HH:mm:ss UTC | DATAPLANE | ERROR | (lib/saluki-env/src/workload/helpers/* | error:No such file or directory (os error 2) | Failed to visit cgroups hierarchy.
```

This error log can crop up for any underlying error we encounter when traversing the cgroup hierarchy, which potentially means it's similar to #498 in that we found a cgroup controller file for a container, but the container exited before we actually read the file? Something along those lines.

It might just be that we want to use more customized error messages in `visit_subdirectories` to differentiate where in the traversal it failed, and on what path the failure occurred.

jszwedko:
Closed by https://github.com/DataDog/saluki/pull/2460

PR #769, opened 2025-06-16
fix(env): allow extracting container ID from relative cgroup controller path

tobz:
## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance


tobz:
> Nice, this is #499 right?

This is actually subtly different.

When the codepath that this PR fixes was failing, it was only failing logically: we would end up swallowing any I/O errors in `CgroupsReader::try_cgroup_from_path` and just return `None` instead.

PR #771, opened 2025-06-17
enhancement: include ADP log file to flare response

rayz:
## Summary

This pr updates the logger to log to both stdout and to a log file specified by a new env var `DD_ADP_LOG_FILE` using `BasicRollingFileAppender` from the `tracing-appender` crate. 

The `DD_LOG_FILE_MAX_SIZE` and `DD_LOG_FILE_MAX_ROLLS` env vars are also added to mimic the datadog agent's logging configuration.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Ran adp and the datadog agent. Request a flare and check that the ADP log file is included.

image


bin/agent-data-plane/src/internal/remote_agent.rs
```diff
@@ -204,9 +205,25 @@ impl RemoteAgent for RemoteAgentImpl {
...
         &self, _request: tonic::Request<GetFlareFilesRequest>,
     ) -> Result<tonic::Response<GetFlareFilesResponse>, tonic::Status> {
-        let response = GetFlareFilesResponse {
-            files: HashMap::default(),
-        };
+        let mut files = HashMap::new();
+
+        let log_file_path = PathBuf::from(
+            std::env::var("DD_ADP_LOG_FILE").unwrap_or(saluki_app::logging::DEFAULT_ADP_LOG_FILE.to_string()),
+        );
```
tobz, line 212: We should be pulling this from the configuration itself, not just an environment variable.

lib/saluki-app/src/logging.rs
```diff
@@ -61,7 +77,9 @@ pub fn fatal_and_exit(message: String) {
 /// # Errors
 ///
 /// If the logging subsystem was already initialized, an error will be returned.
-pub fn initialize_logging(default_level: Option<LevelFilter>) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+pub fn initialize_logging(
+    default_level: Option<LevelFilter>,
+) -> Result<WorkerGuard, Box<dyn std::error::Error + Send + Sync>> {
```
tobz, line 82: I think there's enough things to be tweaked now (some of which I bring up in comments further down in the file) that we should probably expose a configuration object for all of the logging settings: default log level, whether or not to enable JSON logging, log file stuff, etc.

... and to complete the thought, take such a configuration as the one and only argument to this function.

lib/saluki-app/src/logging.rs
```diff
@@ -115,19 +133,47 @@ fn initialize_logging_inner(
...
+    let file_appender = BasicRollingFileAppender::new(
+        adp_log_file,
+        RollingConditionBasic::new().max_size(log_file_max_size),
+        log_file_max_rolls,
+    )?;
+    let (file_nb, guard) = tracing_appender::non_blocking(file_appender);
+    let file_level_filter = EnvFilter::builder()
+        .with_default_directive(default_level.unwrap_or(LevelFilter::INFO).into())
+        .with_env_var("DD_LOG_LEVEL")
+        .from_env_lossy();
```
tobz, line 152: This is going to cause us to log to a file by default, for anything using this... which we don't want to do.

We should -- in conjunction with my comment above about a single configuration type for logging stuff -- make this optional based on whether or not a log file is configured. `tracing_subscriber` supports optional layers through `impl<L> Layer for Option<L>` so we can just have something like `let maybe_file_layer = if enabled { Some(<actual layer here>) } else { None };` and then pop it in with `.with(maybe_file_layer)`.

lib/saluki-app/src/logging.rs
```diff
@@ -115,19 +133,47 @@ fn initialize_logging_inner(
...
+
     if is_json {
         let json_layer = initialize_tracing_json();
         tracing_subscriber::registry()
             .with(json_layer.with_filter(filter_layer))
+            .with(
+                tracing_subscriber::fmt::Layer::new()
+                    .with_writer(file_nb)
+                    .with_filter(file_level_filter),
```
tobz, line 161: This is going to use the default logging format of `tracing-subscriber` which isn't going to match either format we use here for stdout.

It's not clear to me if setting the logging to JSON mode in the Core Agent (or subagents) would also affect their log files... but I know that we would at least want the non-JSON mode to have the same formatting for the log file.

Might be able to just clone the JSON/pretty layers, or might need to initialize 'em twice, not sure.

lib/saluki-app/src/logging.rs
```diff
@@ -115,19 +133,47 @@ fn initialize_logging_inner(
...
+    let log_file_max_rolls = std::env::var("DD_LOG_FILE_MAX_ROLLS")
+        .map(|s| s.parse::<usize>().unwrap_or(DEFAULT_LOG_FILE_MAX_ROLLS))
+        .unwrap_or(DEFAULT_LOG_FILE_MAX_ROLLS);
+    let file_appender = BasicRollingFileAppender::new(
+        adp_log_file,
+        RollingConditionBasic::new().max_size(log_file_max_size),
+        log_file_max_rolls,
+    )?;
+    let (file_nb, guard) = tracing_appender::non_blocking(file_appender);
+    let file_level_filter = EnvFilter::builder()
```
tobz, line 149: Cleaner to just take advantage of `shared_level_filter`, I think, since it already does this much for us, minus any ability to be dynamically updated.

tobz (changes requested):

lib/saluki-app/src/logging.rs
rayz: Had type issues substituting `shared_level_filter` for `file_level_filter` here as well.

PR #779, opened 2025-06-20
chore: refactoring subcommands and improving documentation

brianna-wang-dd:
## Summary
This PR includes no functional changes. It refactors the subcommands into different modules and includes more documentation on the different existing subcommands.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance


bin/agent-data-plane/src/run_subcommand.rs
```diff
@@ -0,0 +1,241 @@
+use std::time::{Duration, Instant};
```
tobz, line 1: I would move this -- and the other subcommand-specific files -- to a new directory, like `cli` or something.

So this would become `bin/agent-data-plane/src/cli/run.rs`, the debug stuff would be `bin/agent-data-plane/src/cli/debug.rs`, and so on.

tobz (changes requested):

bin/agent-data-plane/src/cli/debug.rs
```diff
@@ -0,0 +1,63 @@
...
+    let response = match client.post("https://localhost:5101/logging/reset").send().await {
+        Ok(resp) => resp,
+        Err(e) => {
+            error!("Failed to send request: {}", e);
+            std::process::exit(1);
+        }
+    };
+
+    if response.status().is_success() {
+        println!("Log level reset successful");
```
tobz, line 33: We should always be using the `tracing` macros for logging (i.e. `info!`, `error!`, etc), and never `println!`/`eprintln!` (except in test code, since initializing `tracing` there would be cumbersome.)

tobz: Also, we always have proper punctuation, so this log message (and the equivalent one in `set_log_level` should have a trailing period.

tobz (changes requested):
Two small things.

Issue #774, opened 2025-06-18
Track length of interner separate from offset into backing buffer.


tobz:
## Context

Currently, `InternerStorage` manages a direct allocation and a list of reclaimed entries, which allows it to intern new strings either by reusing a reclaimed entry _or_ operating a bump allocator fashion, simply writing into the backing buffer from the current buffer offset.

We expose the "length" of the interner storage, which is based on the aforementioned offset, and act as if that's the current usage of the interner, in terms of _active_ entries. This is incorrect, because the offset (which is confusingly called `len`, my bad) is never decremented, so interner usage only ever goes up... but even then, it's still the wrong value. We want to show the current usage of _active_ entries.

## Solution

We should rename the `len` field to `offset` to better indicate the actual purpose of the field, and additionally add a new field called `len` which contains the size of all _active_ entries.

This new `len` field would be incremented when adding a new entry (whether reusing a reclaimed entry or not) and decremented when an entry is reclaimed. Essentially, it should equal `offset - <sum of length of current reclaimed entries>`.

We should also consider updating and merging #345, which deals with trying to avoid adding a reclaimed entry when the entry is adjacent to the available capacity by simply decrementing the offset.

PR #781, opened 2025-06-25
track length of interner separate from offset into backing buffer

brianna-wang-dd:
## Summary
This PR renames the `len` field in `InternerStorage` to `offset` to better indicate the actual purpose of the field and adds a new field called `len` which contains the size of all active entries.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?
Added a unit test to check that `len` reflects the number of bytes occupied by active entries.

## References
Closes: Track length of interner separate from offset into backing buffer. <https://github.com/DataDog/saluki/issues/774>


lib/stringtheory/src/interning/map.rs
```diff
@@ -587,7 +593,7 @@ impl GenericMapInterner {
 
     /// Returns the total number of bytes in the interner.
     pub fn len_bytes(&self) -> usize {
-        self.state.lock().unwrap().storage.len
+        self.state.lock().unwrap().storage.offset
```
rayz, line 596: Here is where `len_bytes` is used. 
https://github.com/DataDog/saluki/blob/b7f0294306ac737cfda562efd0864b9759008771/lib/saluki-context/src/resolver.rs#L531

I think we actually want this to return the new `len` field instead of `offset`,

rayz (approved):

lib/stringtheory/src/interning/map.rs
```diff
@@ -253,15 +254,16 @@ impl InternerStorage {
...
             capacity,
+            len: 0,
             reclaimed: ReclaimedEntries::new(),
         }
     }
 
     /// Returns the total number of unused bytes that are available for interning.
     fn available(&self) -> usize {
-        self.capacity.get() - self.len
+        self.capacity.get() - self.offset
```
tobz, line 266: So this is actually where we _would_ want to use `self.len`, because we're trying to make `self.len` indicate how many bytes are _actually_ in use, not just how far into the buffer we've managed to go.

tobz (changes requested):

PR #787, opened 2025-07-02
[AGENTRUN-612] (feature): add support for `service_check` and `events` for Checks source

GustavoCaso:
## Summary

Continue adding support for ADP to execute Python checks. 
This PR adds support for: 
- `aggregator.submit_service_check`
- `aggregator.submit_event`
- `datadog_agent.hostname`
- `datadog_agent.get_config`

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Added unit tests for the `python_modules.rs`. We execute Python code and load the custom modules and ensure the functionality works as expected.


bin/checks-agent/src/env_provider.rs
```diff
@@ -79,6 +79,10 @@ impl ChecksAgentEnvProvider {
     pub fn autodiscovery_provider(&self) -> &BoxedAutodiscoveryProvider {
         &self.autodiscovery_provider
     }
+
+    pub fn host(&self) -> &BoxedHostProvider {
+        &self.host_provider
+    }
```
tobz, line 85: Shouldn't need to expose this here, since you get it through the impl of the `EnvironmentProvider` trait already.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -75,27 +83,48 @@ impl ChecksConfiguration {
...
+                Some(host) => Ok(Box::new(ChecksSource {
+                    autodiscovery: Arc::clone(autodiscovery),
+                    check_runners: self.check_runners,
+                    custom_checks_dirs: if !self.additional_checksd.is_empty() {
+                        Some(self.additional_checksd.split(",").map(|s| s.to_string()).collect())
+                    } else {
+                        None
+                    },
+                    configuration: self.full_configuration.clone(),
+                    host: Arc::clone(host),
```
tobz, line 111: You could just query the hostname here and pass it directly instead of having to pass through a `HostProvider` impl all the way to `ChecksSource`.

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -65,23 +107,51 @@ pub mod aggregator {
...
-    fn metrics(name: String) -> Vec<String> {
-        println!("metrics called for: {}", name);
-        vec![] // Dummy return
+        if let Ok(service_check_status) = status.try_into() {
+            let mut tag_set = TagSet::default();
+            for tag in tags {
+                tag_set.insert_tag(tag);
+            }
+
+            let service_check = ServiceCheck::new(name.clone(), service_check_status)
```
tobz, line 129: No need to clone `name` here, since we already own it.

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -65,23 +107,51 @@ pub mod aggregator {
...
+                .with_tags(tag_set.into_shared())
+                .with_hostname(if hostname.is_empty() {
+                    None
+                } else {
+                    Some(hostname.into())
+                })
+                .with_message(Some(message.unwrap_or_default().into()));
```
tobz, line 136: This should probably be like... `message.map(Into::into)` to do the conversion from `Option<String>` to `Option<MetaString>`.

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -1,29 +1,71 @@
...
+        None => "".to_string(),
+    }
+}
+
+fn fetch_hostname() -> String {
+    match GLOBAL_HOST.get() {
+        Some(host) => host.clone(),
+        None => "".to_string(),
+    }
+}
```
tobz, line 67: This _could_ return `&'static str` since it's effectively static now.

```suggestion
fn fetch_hostname() -> &'static str {
    match GLOBAL_HOST.get() {
        Some(host) => host.as_str(),
        None => "",
    }
}
```

tobz (changes requested):

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -1,29 +1,71 @@
...
     }
 }
 
+fn try_send_service_check(service_check: ServiceCheck) -> Result<(), GenericError> {
+    match METRIC_SENDER.get() {
+        Some(sender) => sender
+            .try_send(Event::ServiceCheck(service_check))
+            .map_err(|e| generic_error!("Failed to send metric: {}", e)),
+        None => Err(generic_error!("Metric sender not initialized.")),
+    }
```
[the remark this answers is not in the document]
GustavoCaso: Maybe is something else, but I get this error when trying to use `blocking_send`: 

> thread 'tokio-runtime-worker' panicked at lib/saluki-components/src/sources/checks/builder/python/python_modules.rs:39:14:
Cannot block the current thread from within a runtime. This happens because a function attempted to block the current thread while the thread is being used to drive asynchronous tasks.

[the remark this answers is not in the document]
GustavoCaso: We could create a separate tokio spawn for each time we send metric and inside the spawned task us `blocking_send`. What do you think?

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -65,23 +120,122 @@ pub mod aggregator {
...
+                    Some(hostname.into())
+                })
+                .with_message(message.map(Into::into));
+
+            if let Err(e) = try_send_service_check(service_check) {
+                error!("Failed to send service check: {}", e);
+            }
+        } else {
+            error!("Invalid service check status: {}", status);
+        }
```
tobz, line 156: We prefer to unnest conditional structures like this, so that it looks more like:

```rust
let check_status = match status.try_into() {
    Ok(check_status) => check_status,
    _ => {
        error!(check_id, check_status = status, "Check returned invalid/unknown status value.");
        return;
    }
};

// Construct the tags, `ServiceCheck`, etc.
```

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -65,23 +120,122 @@ pub mod aggregator {
...
     }
 
     #[pyfunction]
-    fn reset() {
-        println!("reset called");
+    fn submit_event(_class: PyObject, _check_id: String, event: PyObject) {
+        let extracted_event = Python::with_gil(|py| -> Result<PythonEvent, PyErr> {
+            let python_event = event.downcast_bound::<PyDict>(py)?;
+            let extracted_event: PythonEvent = python_event.extract()?;
+            Ok(extracted_event)
```
tobz, line 192: Does this still compile if you change these two lines to just be `python_event.extract()`? You're already declaring the return type for the closure so I would expect it to be able to infer the types.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -60,12 +59,21 @@ pub struct ChecksConfiguration {
     /// Autodiscovery provider to use.
     #[serde(skip)]
     autodiscovery_provider: Option<Arc<dyn AutodiscoveryProvider + Send + Sync>>,
+
+    /// Hostname provider to use.
+    #[serde(skip)]
+    hostname_provider: Option<Arc<dyn HostProvider<Error = GenericError> + Send + Sync>>,
```
tobz, line 65: Since you're now holding two distinct providers, it might be a little cleaner to just hold the environment provider instead. You can look at the Host Enrichment transform for an example of how we do that, including where to place the necessary bounds on the `SourceBuilder` implementation for stuff like the error type, etc.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -236,24 +289,57 @@ async fn drain_and_dispatch_check_metrics(
...
-                        .expect("checks source should always have default output");
+                Some(check_event) => {
+                    trace!("Received check event: {:?}", check_event);
+
+                    match check_event.event_type() {
+                       EventType::Metric => {
+                            let mut buffered_dispatcher = context
+                                .dispatcher()
+                                .buffered_named("metrics")
+                                .expect("checks source should always have default output");
```
tobz, line 302: This message doesn't make sense anymore since it's a named output... same with the other equivalent calls for service checks and events.

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -120,3 +274,137 @@ pub mod datadog_agent {
...
+    }
+
+    struct TestResults {
+        hostname: String,
+        config_value: String,
+    }
+
+    #[test]
+    fn test_python_datadog_agent_integration() {
+        std::env::set_var("DD_TEST_FOO", "bar");
```
tobz, line 369: This is very brittle -- setting environment variables at runtime -- but there's no better option at the moment for what you're trying to do... so I'm just saying this to remind myself that we need a answer for mocking out stuff in configuration.

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -1,29 +1,84 @@
...
 }
 
+fn try_send_service_check(service_check: ServiceCheck) -> Result<(), GenericError> {
+    match METRIC_SENDER.get() {
+        Some(sender) => sender
+            .try_send(Event::ServiceCheck(service_check))
+            .map_err(|e| generic_error!("Failed to send service check: {}", e)),
+        None => Err(generic_error!("Metric sender not initialized.")),
+    }
+}
```
tobz, line 56:
```suggestion
    METRIC_SENDER.get()
        .ok_or_else(|| generic_error!("Metric sender not initialized."))
        .and_then(|sender| sender.blocking_send(Event::ServiceCheck(service_check)))
        .error_context("Failed to send service check. Source disconnected.")
```

This will condense things slightly, and it will also ensure we don't arbitrarily lose events when the checks source is processing a big burst of them.

We should do this for all three sender methods.

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -1,29 +1,84 @@
...
+// Global state to store the configuration
+static GLOBAL_HOST: OnceLock<String> = OnceLock::new();
 
 /// Sets the metric sender to be used by the aggregator module.
-pub fn set_metric_sender(check_metrics_tx: Sender<CheckMetric>) -> &'static Sender<CheckMetric> {
+pub fn set_metric_sender(check_metrics_tx: Sender<Event>) -> &'static Sender<Event> {
     METRIC_SENDER.get_or_init(|| check_metrics_tx)
 }
 
+/// Stores the configuratioon for integration to use in the Python module.
```
tobz, line 30: `configuratioon` -> `configuration`

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -1,29 +1,84 @@
...
+pub fn set_configuration(configuration: GenericConfiguration) -> &'static GenericConfiguration {
+    GLOBAL_CONFIGURATION.get_or_init(move || configuration)
+}
+
+/// Stores the configuratioon for integration to use in the Python module.
```
tobz, line 35: Looks like the doccomment for `set_configuration` got copypasted here instead?

tobz (changes requested):

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
GustavoCaso: `blocking_send` causes an issue. Look at my other comment https://github.com/DataDog/saluki/pull/787#discussion_r2245602259

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
GustavoCaso: I tried, and it does not compile :(

lib/saluki-components/src/sources/checks/mod.rs
GustavoCaso: Gave it a go here https://github.com/DataDog/saluki/pull/787/commits/abd50d4facf511d2753a2a51e353cecee3f2180a

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -57,82 +56,122 @@ pub struct ChecksConfiguration {
...
-    /// Sets the autodiscovery provider to use.
-    pub fn with_autodiscovery_provider<A>(mut self, provider: A) -> Self
-    where
-        A: AutodiscoveryProvider + Send + Sync + 'static,
-    {
-        self.autodiscovery_provider = Some(Arc::new(provider));
-        self
+impl<E> ChecksConfiguration<E> {
+    /// Sets the env provider to use.
+    pub fn with_env_provider<T>(self, env_provider: T) -> ChecksConfiguration<T> {
```
tobz, line 77: Two things here:

- we try to prefer full words for _method_ names, and abbreviations/acronyms for variable names (just so the API documentation stays as human-friendly as possible)
- (this one is very very subjective) when following this sort of pattern, where you're replacing a value of type T (that implements a common trait) with another concrete type that implements the same common trait, we try to match the generic type parameter name but suffixed with a `2` (so `T2`), to sort of indicate that we're replacing the existing `T` with something `T`-like.. so in this case, our struct uses `E`, so we want to use `E2`

```suggestion
    /// Sets the environment provider to use.
    pub fn with_environment_provider<E2>(self, env_provider: E2) -> ChecksConfiguration<E2> {
```

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -57,82 +56,122 @@ pub struct ChecksConfiguration {
     #[serde(default)]
     additional_checksd: String,
 
-    /// Autodiscovery provider to use.
     #[serde(skip)]
-    autodiscovery_provider: Option<Arc<dyn AutodiscoveryProvider + Send + Sync>>,
+    env_provider: Option<E>,
```
tobz, line 60: This shouldn't need to be `Option<E>` any longer, I think.

tobz (changes requested):
Just a few small comments.

tobz (approved):

PR #788, opened 2025-07-02
experiment(topology): try using an on-demand object pool for global event buffers pool

tobz:
## Summary

This PR introduces a new "on demand" object pool and switches to using it for the global event buffer pool in order to address the risk of deadlock under high load.

The global event buffer pool is inextricably tied to components dispatching events to the next component in the topology. This means that many components, when there is a steady flow of metrics, are constantly acquiring event buffers and then dispatching them. In some cases, however, the volume of metrics can outpace the rate at which event buffers are processed, leading to scenarios where a deadlock is possible.

One such scenario is the aggregate transform. It processes incoming event buffers, aggregate the metrics within them, and then periodically flushes the aggregated metrics. In order to flush, it has to acquire event buffers, which then get filled and dispatched. If the global event buffer pool is exhausted, perhaps by a high rate of incoming metrics, then the transform can be starved of the event buffers it needs to continue flushing. This is a problem because in order to free up event buffers to use when flushing, the in-flight event buffers have to be consumed/processed... which doesn't happen until the flush completes. Boom, deadlock.

This PR attempts to address this by allowing on-demand event buffer allocation. The thought process here is that our global event buffer pool being fixed size (in terms of maximum event buffers) is a means to an end: establish an upper bound of the memory usage related to dispatching events. We can't simply increase the maximum, because theoretically no number is high enough: we could consume all possible event buffers while the aggregate transform (or any other component) is doing some work, leading to starvation. Our key insight here is that we are ultimately bounded by the capacity of the interconnect channels between components: if the channel is full, you can't send another event buffer, and so code that is acquiring event buffers to fill them and dispatch them will allocate one, fill it, try to dispatch it, and then encounter backpressure when dispatching... ultimately limiting any further event buffer acquisition.

This ensures that all components trying to acquire an event buffer can do so until they hit backpressure, and since we know the number of components and the capacity of the interconnect channels, we can accurately calculate bounds. Win win.

In this PR, we maintain usage of the "object pool" mechanism, even though there is no pooling going on, and we just allocate on demand. We may choose to remove this abstraction in the future but I wanted to leave it for now just to keep that flexibility.

## Change Type

- [x] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Existing unit tests/correctness tests in CI + tested locally and ensured normal DSD pipeline in ADP still forwarded metrics. 

## References

AGTMETRICS-233

### DIFF
--- a/lib/saluki-core/src/topology/blueprint.rs
+++ b/lib/saluki-core/src/topology/blueprint.rs
@@ -17,10 +17,8 @@ use crate::{
     data_model::event::Event,
 };
 
-// SAFETY: These are obviously all non-zero.
+// SAFETY: This is obviously non-zero.
 const DEFAULT_EVENT_BUFFER_SIZE: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(1024) };
-const DEFAULT_EVENT_BUFFER_POOL_SIZE_MIN: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(32) };
-const DEFAULT_EVENT_BUFFER_POOL_SIZE_MAX: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(512) };
 
 /// A topology blueprint error.
 #[derive(Debug, Snafu)]
@@ -53,8 +51,6 @@ pub struct TopologyBlueprint {
     destinations: HashMap<ComponentId, RegisteredComponent<Box<dyn DestinationBuilder + Send>>>,
     component_registry: ComponentRegistry,
     event_buffer_pool_buffer_size: usize,
-    event_buffer_pool_size_min: usize,
-    event_buffer_pool_size_max: usize,
 }
 
 impl TopologyBlueprint {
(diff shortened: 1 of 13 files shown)


lib/saluki-core/src/topology/blueprint.rs
```diff
@@ -63,78 +61,84 @@ impl TopologyBlueprint {
...
+        self.recalculate_bounds();
+        self
+    }
+
+    fn recalculate_bounds(&mut self) {
+        // Adjust the bounds related to interconnects.
+        //
+        // Every non-source component has an interconnect.
+        let total_interconnect_capacity =
+            (self.transforms.len() + self.destinations.len()) * self.interconnect_capacity.get();
```
lukesteensen, line 106: We don't do it now, but technically you could have more edges in the graph than this (e.g. two parallel transforms both sending to two sinks).

lukesteensen (approved):

lib/saluki-core/src/topology/blueprint.rs
tobz: Hmm... I don't think that changes anything?

Each non-source component has a single interconnect channel, regardless of how many upstream components are connected to it.

tobz: 😂 No worries.

PR #796, opened 2025-07-09
chore(core): overhaul interconnect types to be generic

tobz:
## Summary

This PR refactors a number of interconnect types to be generic, allowing us to reuse them for both event-based and payload-based components.

### Notes

The changes in this PR are widespread, but here's the high-level notes:

- `Dispatcher` and `Consumer` (neé `EventStream`) are now generic over the item type.
  - This is the core change, so that we can reuse both of them for `Encoder` and `Forwarder`.
  - We renamed `EventStream` to `Consumer` because now it's not event-centric.
  - We added some helper stuff -- `Dispatchable` and `DispatchBuffer` -- to allow for type-driven API methods on `Dispatcher` to support buffering events within `FixedSizeEventBuffer` while not forcing payload-based components to have to buffer their items, instead being able to simply dispatch a `Payload` directly.
- We codified a fixed "events" buffer type as well as a "payloads" buffer type, and the corresponding dispatcher/consumer types utilizing them: `EventsBuffer`/`EventsDispatcher`/`EventsConsumer` and `PayloadsBuffer`/`PayloadsDispatcher`/`PayloadsConsumer`.
  - Technically, `PayloadsBuffer` is just `Payload` but I wanted the naming to match. 🙈 
  - I was thinking that we'd sort of do this with `TopologyConfiguration`, but it ended up being too ugly and was easier to just hard-code the types in a single place, which leads me to..
- Removed `TopologyConfiguration` since it had no immediate path to cleanly being weaved in.
  - We may want something like it in the future, but it's too infectious of a change at the moment and we don't have a strong need for that level of flexibility.
- `FixedSizeEventBuffer` now has a const generic that defines the capacity, and is no longer poolable.
  - I figured that since we weren't _actually_ pooling it anymore, we should decouple ourselves from that.
  - We still wanted a mechanism to create new buffers but ensure they all had a uniform capacity, which we previously got from the object pool's builder configuration... which we now achieve with the const generic parameter.
  - This means it supports `Default`, and `DispatchBuffer`, and `Dispatcher` can create them on-demand as needed.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Existing unit tests.

## References

AGTMETRICS-233


tobz:
* **#796** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/796?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/saluki/796?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rayz (approved):

PR #797, opened 2025-07-10
[AGTMETRICS-310] add support for streaming configuration updates from the datadog agent

rayz:
## Summary

This PR adds the implementation for streaming configuration events from the core agent. This is gated behind the `DD_ADP_USE_NEW_CONFIG_STREAM_ENDPOINT` environment variable. If present, ADP will block until a snapshot has been retrieved via the `ready` in `GenericConfiguration`.

Implementation:

`RemoteAgentClient` is used to establish a stream with the core agent and a background task handles sending the updates to the `GenericConfiguration` which has its own task `run_dynamic_config_updater` (which is only run when `with_dynamic_configuration` option is used. This task is responsible for maintaining a copy of the dynamic state, and the current state so that a diff can be produced. Each change is then sent to any subscribers of `broadcast::Sender<ConfigChangeEvent>`. Note: This pr doesn't add any subscribers yet, it only provides the functionality.

Providers are merged in based on the order they were added. For our setup:
```
ConfigurationLoader::default()
            .try_from_yaml(&run_config.config)
            .with_dynamic_configuration(receiver)
            .from_environment("DD")?
            .with_default_secrets_resolution()
 ```
 Environment variables will have the highest priority over values retrieved dynamically here.

A new `config` subcommand has been added, along with a scrubber that hides sensitive values.

## Notes
1. The old method of retrieving up-to-date configuration settings via `RefreshableConfiguration` has been removed.
2. This branch served as the base for setting up the configuration stream. Relevant prs that have been since merged into this one:
   https://github.com/DataDog/saluki/pull/882
   https://github.com/DataDog/saluki/pull/842
   https://github.com/DataDog/saluki/pull/810

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

1. Manual test, updating configs and running the new `config` subcommand.
2. Unit tests for config diffs

### CODE REVIEW

File: lib/saluki-common/src/scrubber.rs [hunk 1]
```diff
@@ -0,0 +1,416 @@
...
+                rc_app_key_replacer,
+                uri_password_replacer,
+                password_replacer,
+            ],
+        }
+    }
+}
+
+/// A YAML scrubber that can be configured with different replacers.
+pub struct Scrubber {
```
tobz, line 142: From what I can see, none of these methods have to be `async`.

lib/saluki-components/src/common/datadog/endpoints.rs
```diff
@@ -205,7 +205,7 @@ impl ResolvedEndpoint {
     }
 
     /// Creates a new  `ResolvedEndpoint` instance from an existing `ResolvedEndpoint`, adding an optional `RefreshableConfiguration`.
-    pub fn with_refreshable_configuration(self, config: Option<RefreshableConfiguration>) -> Self {
+    pub fn with_refreshable_configuration(self, config: Option<GenericConfiguration>) -> Self {
```
tobz, line 208: Seems like we'll also want to remove the "refreshable" configuration wording in this area of the code as well.

Cargo.toml
```diff
@@ -66,7 +66,7 @@ serde = { version = "1", default-features = false, features = [
   "std",
 ] }
 snafu = { version = "0.8", default-features = false, features = ["std"] }
-tokio = { version = "1", default-features = false }
+tokio = { version = "1", default-features = false, features = ["rt", "rt-multi-thread", "io-util", "macros"] }
```
tobz, line 69: Hmmm, do we need this here? I see we're adding these features in `lib/saluki-common/Cargo.toml` directly.

(Ignoring that I don't think the scrubber actually needs to be `async` and so we might not need these features at all for `saluki-common`.)

Cargo.toml
```diff
@@ -113,7 +113,7 @@ rustls = { version = "0.23", default-features = false, features = [
 slab = { version = "0.4", default-features = false }
 tokio-util = { version = "0.7.10", default-features = false }
 tower = { version = "0.5", default-features = false }
-tracing-subscriber = { version = "0.3", default-features = false }
+tracing-subscriber = { version = "0.3.20", default-features = false }
```
tobz, line 116: I don't think we explicitly need this since it's fixed in `main`.

tobz (changes requested):

lib/saluki-components/src/common/datadog/endpoints.rs
rayz: I changed the wording for now, but I think we may just remove this all together once we actually use the new config subscriber model on the components.

PR #812, opened 2025-07-16
[AGTMETRICS-315] add support for dogstatsd stats subcommand

brianna-wang-dd:
## Summary

This PR adds support for the `dogstatsd stats` subcommand. It adds a new destination and creates an API handler that contains the `/dogstatsd/stats` endpoint, which is called with the `dogstatsd stats` subcommand. `dogstatsd stats` returns basic statistics about the metrics received by the agent data plane, including the name, count, time last seen, and tags.

Example output:
image

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Manually tested with `echo`ing sample metrics.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -90,7 +90,7 @@ default:
 # Run a job on development branches.
 .on_development_branch:
   rules:
-    if: $CI_COMMIT_BRANCH != "main" && $CI_COMMIT_TAG == ""
+    if: $CI_COMMIT_BRANCH != "main" && $CI_COMMIT_TAG == null
 
 # Common build-specific variables that need to be shared across stages.
 .build-common-variables:


bin/agent-data-plane/src/config.rs
```diff
@@ -19,6 +19,10 @@ pub enum Action {
     /// Various debugging commands.
     #[command(subcommand)]
     Debug(DebugConfig),
+
+    /// Prints the version of the agent data plane.
+    #[command(name = "dogstatsd-stats")]
+    DogstatsdStats,
```
tobz, line 25: We should make this a subcommand under `dogstatsd`.

We'll eventually end up having other DSD-specific things (capturing the raw payloads, etc), so it'll look a little nicer as proper subcommands, and we don't need to _exactly_ match how it's done in the Core Agent since users are going to have to type `agent-data-plane ...` instead of `agent ...` to begin with.

tobz: Sorry, missed your response...

Yes, we want there to be a `dogstatsd` subcommand group, with this PR adding a `statsd` subcommand in that group.

tobz: Sorry: `stats`, not `statsd`. 😅

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -173,6 +185,7 @@ async fn create_topology(
         .add_encoder("dd_events_encode", dd_events_config)?
         .add_encoder("dd_service_checks_encode", dd_service_checks_config)?
         .add_forwarder("dd_out", dd_forwarder_config)?
+        .add_destination("dsd_stats_out", dsd_stats_config.clone())?
```
rayz, line 188: `dogstatsd-stats` isn't enabled by default in `datadog-agent` since there is a performance cost I believe. It is gated behind the `dogstatsd_metrics_stats_enable` env var.

rayz (changes requested):

lib/saluki-components/src/destinations/dsd_stats/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+    fn specify_bounds(&self, builder: &mut MemoryBoundsBuilder) {
+        builder
+            .minimum()
+            .with_single_value::<DogStatsDStatisticsConfiguration>("configuration struct");
+
+        builder.firm().with_expr(UsageExpr::constant("api handler state", 48));
+        builder.firm().with_expr(UsageExpr::constant("destination state", 24));
+        builder
+            .firm()
+            // Capture the size of the "split re-encode" buffers in the request builders, which is where we keep owned
```
rayz, line 211: I think we can remove this (split re-encode buffer) since it doesn't apply here I believe.

lib/saluki-components/src/destinations/dsd_stats/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+        let mut health = context.take_health_handle();
+
+        health.mark_ready();
+
+        loop {
+            select! {
+                _ = health.live() => {
+                    continue
+                },
+                maybe_request = async { (self.rx.lock().await).recv().await } => {
```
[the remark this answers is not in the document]
brianna-wang-dd: I ran into an issue where `rx` was shared between `DogStatsDStats` and `DogStatsDStatisticsConfiguration`, so the lock was used. Let me think a bit more about this to see if there's anything in the design that can be changed to avoid using a lock!

lib/saluki-components/src/destinations/dsd_stats/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+                                let tags_formatted = tags.join(",");
+                                let key = if tags.is_empty() {
+                                    metric_name.clone()
+                                } else {
+                                    format!("{}|{}", metric_name.clone(), tags_formatted)
+                                };
+
+                                let now = SystemTime::now();
+                                let datetime = chrono::DateTime::<chrono::Local>::from(now);
+                                let sample = self.stats.metrics_received.entry(key).or_insert_with(|| MetricSample {
```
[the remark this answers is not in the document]
tobz: I had the thought that this will actually end up considering the origin tags as part of the key, which we don't want to do... 😓 

We can just create our own simplified context type (call it `ContextNoOrigin`, maybe?) that just holds the metric name (`MetaString`) and instrumented tags (`SharedTagSet`), both of which we should be able to clone very, very cheaply.

lib/saluki-components/src/destinations/dsd_stats/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+
+/// DogStatsD destination that collects metrics and processes statistics.
+pub struct DogStatsDStats {
+    rx: Arc<Mutex<mpsc::Receiver<tokio::sync::oneshot::Sender<Stats>>>>,
+    stats: Stats,
+}
+
+#[async_trait::async_trait]
+impl Destination for DogStatsDStats {
+    async fn run(mut self: Box<Self>, mut context: DestinationContext) -> Result<(), GenericError> {
```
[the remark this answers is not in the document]
brianna-wang-dd: Ohh I see, so we want to break the select down into more cases now that we're implementing the collection period and the try again later. Thanks Toby for typing out all the code with the detailed comments!! 🙏

lib/saluki-components/src/destinations/dsd_stats/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    fn generate_routes(&self) -> Router<Self::State> {
+        Router::new().route("/dogstatsd/stats", get(Self::stats_handler))
+    }
+}
+
+impl DogStatsDStatisticsConfiguration {
+    /// Creates a new 'DogStatsDStatisticsConfiguration' from the given configuration.
+    pub fn from_configuration(_: &GenericConfiguration) -> Result<Self, GenericError> {
```
GustavoCaso, line 245: Are we using the generic configuration?

brianna-wang-dd: Oop, we're not using the generic configuration anymore so it can be removed.

tobz (approved):

PR #818, opened 2025-07-21
enhancement: add cardinality field support to DogStatsD

### LINKED ISSUE #799: Add cardinality field support to DogStatsD.
## Context

The latest update to the DogStatsD protocol ([here](https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=metrics#dogstatsd-protocol-v16)) added support for specifying the cardinality of any tags added via origin detection.

We don't not yet support this, and we should! :)

This issue depends on #798 as we need to add it for metrics, events, _and_ service checks... and adding it for events and service checks doesn't make sense until we actually support origin detection for those to begin with.

rayz:
## Summary

This pr adds support for parsing the `card` field for metrics, events and service checks.
## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

unit tests
## References

- Closes #799


lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
```diff
@@ -238,6 +239,18 @@ fn parse_dogstatsd_metric<'a>(
                     let (_, external_data) = all_consuming(preceded(tag("e:"), external_data)).parse(chunk)?;
                     maybe_external_data = Some(external_data);
                 }
+                // Cardinality: client-provided cardinality for the metric.
+                b'c' if chunk.starts_with(message::CARDINALITY_PREFIX) => {
+                    let (_, cardinality) = all_consuming(preceded(
+                        tag(message::CARDINALITY_PREFIX),
+                        alt((tag("none"), tag("low"), tag("orchestrator"), tag("high"))),
```
tobz, line 246: I would bundle this into a standalone function, similar to `external_data`, so that we don't have to concern `saluki-context` with using `simdutf8` or parsing from `&[u8]`, etc.

tobz (changes requested):

PR #820, opened 2025-07-22
Rayz/generic configuration dynamic

rayz:
## Summary

Adds the `with_dynamic_configuration` option to `ConfigurationLoader`.

When enabled, `into_generic` will create `RefreshableConfiguration`. A handle is exposed for the refreshable values, for both the priviledged api and `RemoteAgent` to use.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

### DIFF
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -1,9 +1,5 @@
-use std::{
-    sync::Arc,
-    time::{Duration, Instant},
-};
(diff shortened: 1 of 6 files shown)


rayz:
The `get` / `try_get_typed` methods in `GenericConfiguration` haven't been updated yet to use the `RefreshableConfiguration` yet. One thing to note is, the keys from the values updated by the `RemoteAgent` have `_` instead of `.`

PR #821, opened 2025-07-23
Dynamic Provider

rayz:
## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

### DIFF
--- a/lib/saluki-config/src/lib.rs
+++ b/lib/saluki-config/src/lib.rs
@@ -12,9 +12,11 @@ use serde::Deserialize;
 use snafu::{ResultExt as _, Snafu};
 use tracing::debug;
 
+mod dynamic;
 mod provider;
 mod refresher;
 mod secrets;
+use self::dynamic::Provider;
 use self::provider::ResolvedProvider;
 pub use self::refresher::{RefreshableConfiguration, RefresherConfiguration};
 
@@ -118,11 +120,22 @@ impl LookupSource {
 /// - YAML file
 /// - JSON file
 /// - environment variables (must be prefixed; see [`from_environment`][Self::from_environment])
-#[derive(Default)]
 pub struct ConfigurationLoader {
     inner: Figment,
     lookup_sources: HashSet<LookupSource>,
     dynamic_configuration_enabled: bool,
+    dynamic_values: Option<Arc<ArcSwap<serde_json::Value>>>,
+}
+
+impl Default for ConfigurationLoader {
+    fn default() -> Self {
+        Self {
+            inner: Figment::new(),
+            lookup_sources: HashSet::new(),
+            dynamic_configuration_enabled: false,
+            dynamic_values: None,
+        }
+    }
 }
 
 impl ConfigurationLoader {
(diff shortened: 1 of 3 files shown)


lib/saluki-config/src/lib.rs
```diff
@@ -302,10 +316,15 @@ impl ConfigurationLoader {
 
     /// Enables dynamic configuration.
     ///
-    /// This will enable the dynamic configuration feature, which allows the configuration to be sourced from the core agent.
-    pub fn with_dynamic_configuration(mut self) -> Self {
+    /// This will enable the dynamic configuration feature, which allows the configuration to be sourced from a dynamic provider. The dynamic provider is updated at runtime, and `figment` will re-read from it when configuration values are requested.
+    pub fn with_dynamic_configuration(mut self) -> Result<Self, ConfigurationError> {
         self.dynamic_configuration_enabled = true;
```
tobz, line 321: Wonder if we should just drop this field entirely since we can determine it by whether or not `dynamic_values` is `None`/`Some`?

lib/saluki-config/src/lib.rs
```diff
@@ -322,37 +341,59 @@ impl ConfigurationLoader {
...
-        let inner: Value = self.inner.extract()?;
-
-        let (refreshable_config, refreshable_values) = if self.dynamic_configuration_enabled {
-            let refresher_settings = inner.deserialize::<RefresherConfiguration>()?;
-            debug!("Dynamic configuration enabled.");
-
-            let values = Arc::new(ArcSwap::from_pointee(serde_json::Value::Null));
-            let refreshable = refresher_settings.build(values.clone()).await.context(Generic)?;
-            (Some(refreshable), Some(values))
+        let refreshable_values = if self.dynamic_configuration_enabled {
```
tobz, line 344: Dropping this field, as mentioned above, would let us get rid of checks like these and instead just directly reference `dynamic_values`.

lib/saluki-config/src/lib.rs
```diff
@@ -382,20 +423,6 @@ pub struct GenericConfiguration {
...
-    fn get(&self, key: &str) -> Option<&Value> {
-        match self.inner.value.find_ref(key) {
-            Some(value) => Some(value),
-            None => {
-                // We might have been given a key that uses nested notation -- `foo.bar` -- but is only present in the
-                // environment variables. We specifically don't want to use a different separator in environment
-                // variables to map to nested key separators, so we simply try again here but with all nested key
-                // separators (`.`) replaced with `_`, to match environment variables.
-                let key = key.replace('.', "_");
-                self.inner.value.find_ref(&key)
```
tobz, line 394: Seems like we haven't added this back anywhere else... which would definitely break something things off the top of my head (mostly ADP-specific configuration settings.)

tobz (changes requested):

lib/saluki-config/src/lib.rs
rayz: Added `get` back. It now returns a type `T` but it looks like we were only ever using it in the typed functions anyways.

PR #834, opened 2025-07-29
chore(ci): tighten up timeouts for build jobs + remove unnecessary k8s requests/limits

tobz:
## Summary

We've been hitting a number of timed out build jobs over the past few weeks for our ADP images. I've been chatting with the internal team responsible for the CI infrastructure about the _why_, but in the interim, we should tighten our job timeout to attempt to get away from a stalled job more quickly. This PR sets our timeout to 10m for all ADP/`checks-agent` builds as it's right around the p90 job duration for successful runs of those jobs.

This PR also removes all variables related to setting Kubernetes requests/limits for Docker build-based jobs, as they don't influence the underlying BuildKit worker used for building the image. We've kept them for test jobs, which is where we _do_ execute `cargo` directly on the top-level CI runner.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

N/A

## References

AGTMETRICS-233

### DIFF
--- a/.gitlab/benchmark.yml
+++ b/.gitlab/benchmark.yml
@@ -25,14 +25,10 @@ build-adp-baseline-image:
   image: "${SALUKI_SMP_CI_IMAGE}"
   needs: []
   retry: 2
-  timeout: 20m
+  timeout: 10m
   before_script:
     - *setup-smp-env
   variables:
-    # Compiling Rust is intensive. ¯\_(ツ)_/¯
-    KUBERNETES_CPU_REQUEST: "16"
-    KUBERNETES_MEMORY_REQUEST: "8Gi"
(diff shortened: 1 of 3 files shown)


.gitlab/benchmark.yml
```diff
@@ -47,7 +43,7 @@ build-adp-baseline-image:
     # Actually build.
     - aws ecr get-login-password --region us-west-2 --profile ${AWS_NAMED_PROFILE} | docker login --username AWS --password-stdin ${SMP_ECR_HOST}
     - docker buildx build
-      --platform linux/amd64,linux/arm64
+      --platform linux/amd64
```
tobz, line 46: Changing this because SMP also doesn't run on ARM64, so there's no point in building for it.

PR #841, opened 2025-07-31
chore(ci): skip SMP benchmarks on pushes to main/release tags

tobz:
## Summary

This PR skips the SMP benchmarks for pushes to `main` and release tags. Since there's no reasonable "baseline" to test against, running the benchmarks doesn't make sense... and just ends up always failing which is confusing if you don't know that is expected.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Can't test this without actually merging. 😅 

## References

AGTMETRICS-233

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -90,7 +90,7 @@ default:
 # Run a job on development branches.
 .on_development_branch:
   rules:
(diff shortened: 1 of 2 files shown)


brianna-wang-dd (approved):
lgtm!

PR #853, opened 2025-08-06
enhancement(common): reusable deduplication helper

tobz:
## Summary

This PR adds a reusable deduplication helper, primarily for tag deduplication.

Currently, tags deduplication happens through the `Deduplicated` helper, exposed by `TagsExt::deduplicated`. This creates a wrapper iterator that uses a hash set internally to skip duplicates: simple and effective. However, one issue is that we're allocating a new hash set every time we create one of these helpers. This means we're (re)allocating constantly as we deduplicate tag sets when building request payloads for Datadog payloads, and so on.

This PR introduces a new helper, `ReusableDeduplicator`, that allows reusing the underlying hash set between iterators. We use it to create a wrapper iterator that borrows the hash set for as long as necessary to iterate and deduplicate.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Existing unit and correctness tests.

## References

AGTMETRICS-233

### DIFF
--- a/lib/saluki-core/src/data_model/event/service_check/mod.rs
+++ b/lib/saluki-core/src/data_model/event/service_check/mod.rs
@@ -1,6 +1,7 @@
 //! Service checks.
 
-use saluki_context::tags::{SharedTagSet, TagsExt};
+use saluki_common::iter::ReusableDeduplicator;
+use saluki_context::tags::SharedTagSet;
 use serde::{ser::SerializeMap as _, Serialize, Serializer};
 use stringtheory::MetaString;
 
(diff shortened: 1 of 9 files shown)


lib/saluki-core/src/data_model/event/service_check/mod.rs
```diff
@@ -229,7 +230,10 @@ impl<'a> Serialize for DeduplicatedTagsSerializable<'a> {
     where
         S: Serializer,
     {
-        let deduplicated_tags = self.tags.into_iter().chain(self.origin_tags).deduplicated();
+        let chained_tags = self.tags.into_iter().chain(self.origin_tags);
+
+        let mut tags_deduplicator = ReusableDeduplicator::new();
```
rayz, line 235: We wouldn't get the benefits here right?

rayz (approved):
LGTM, just one comment

lib/saluki-core/src/data_model/event/service_check/mod.rs
tobz: Unfortunately, yeah: no benefits here.

This is something I hope we can, potentially in the future, replace with a protobuf-based encoder component so that we can match events/metrics.

PR #858, opened 2025-08-08
chore: add blocking until initial configuration is received

brianna-wang-dd:
## Summary
This PR reorganizes and blocks the `run` function in order to obtain a snapshot of the configuration from the core agent before the topology is set up.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?
Tested manually with Raymond's datadog-agent branch `rayz/agent-ipc-config-stream` (PR: <https://github.com/DataDog/datadog-agent/pull/39877>)

### DIFF
--- a/lib/saluki-env/src/configstream/stream.rs
+++ b/lib/saluki-env/src/configstream/stream.rs
@@ -16,8 +16,7 @@ use crate::helpers::remote_agent::RemoteAgentClient;
 pub async fn create_config_stream(
     config: &GenericConfiguration, shared_config: Arc<ArcSwap<Value>>, snapshot_received: Arc<AtomicBool>,
 ) -> Result<(), GenericError> {
-    let config = config.clone();
-    let mut client = match RemoteAgentClient::from_configuration(&config).await {
+    let mut client = match RemoteAgentClient::from_configuration(config).await {
         Ok(client) => client,
         Err(e) => {
             error!("Failed to create remote agent client: {}.", e);


lib/saluki-env/src/configstream/stream.rs
```diff
@@ -0,0 +1,99 @@
...
+use serde_json::{Map, Value};
+use tracing::error;
+
+use crate::helpers::remote_agent::RemoteAgentClient;
+
+/// Creates a new `ConfigStreamer` that receives a stream of config events from the remote agent.
+pub async fn create_config_stream(
+    config: &GenericConfiguration, shared_config: Arc<ArcSwap<Value>>, snapshot_received: Arc<AtomicBool>,
+) -> Result<(), GenericError> {
+    let config = config.clone();
```
tobz, line 19: Why are we cloning this? `RemoteAgentClient::from_configuration` takes `&GenericConfiguration`, and we never pass `config` to the async task that we spawn... 🤔

tobz (changes requested):

lib/saluki-env/src/configstream/stream.rs
brianna-wang-dd: OH GOOD CATCH! I think there was an ownership issue when the `from_configuration` code was in the async block, and then I moved it out but didn't remove the cloning. 🙈

PR #866, opened 2025-08-13
chore(dogstatsd): re-organize and clean up DSD codec

tobz:
## Summary

As stated in the PR title.

This is really just isolating metrics, events, and service check handling code to their own code files, and moving the helpers to their own spot. We've added a bunch of doc comments and what not as well, but mostly it's the code reorganization.

This is all in service, for what it's worth, of adding a lot more testing for parsing and not wanting `lib/saluki-io/src/deser/codec/dogstatsd/mod.rs` to end up with a `tests` module that's like 2,000 lines long.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Existing unit tests.

## References

AGTMETRICS-233


rayz (approved):
lgtm

PR #872, opened 2025-08-20
chore: gate config stream creation and initial config blocking behind `adp.use_new_config_stream_endpoint`

brianna-wang-dd:
## Summary
This PR checks the `adp.use_new_config_stream_endpoint ` setting and only creates the config stream and blocks for the initial configuration when this setting is configured true.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?
Manually tested with setting `DD_ADP_USE_NEW_CONFIG_STREAM_ENDPOINT=true` versus no setting and setting as `false`.

### DIFF
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -60,7 +60,7 @@ pub async fn run(started: Instant, run_config: RunConfig) -> Result<(), GenericE
                 return Err(e);
             }
             // If configured, block until a initial configuration is received.
-            if configuration.get_typed_or_default::<bool>("adp.block_until_config_snapshot") {
+            if configuration.get_typed_or_default::<bool>("adp.use_new_config_stream_endpoint") {
                 info!("Waiting for initial configuration from Datadog Agent...");
                 let mut attempts = 0;
                 const CHECK_INTERVAL_MS: u64 = 100;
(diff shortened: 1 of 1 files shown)


bin/agent-data-plane/src/cli/run.rs
```diff
@@ -59,24 +59,25 @@ pub async fn run(started: Instant, run_config: RunConfig) -> Result<(), GenericE
...
-                tokio::time::sleep(Duration::from_millis(CHECK_INTERVAL_MS)).await;
-                attempts += 1;
-
-                if attempts % 100 == 0 {
-                    info!(
-                        "Still waiting for initial configuration... ({}s elapsed)",
-                        attempts / 10
-                    );
+            // If configured, block until a initial configuration is received.
+            if configuration.get_typed_or_default::<bool>("adp.block_until_config_snapshot") {
```
tobz, line 63: I'm realizing now that we should actually include even the config stream creation in this block, and should change the setting to `adp.use_new_config_stream_endpoint`.

Testing locally, with this PR as-is, we still always see at least one error message from the config streamer trying to start the stream but failing to if the Agent isn't new enough to have `StreamConfigEvents`.

I hadn't considered this originally, but really what I was imagining was that we don't use this blocking logic, _or_ the new endpoint, unless explicitly configured to do so.

tobz (changes requested):

PR #874, opened 2025-08-22
chore: add configurability for cached contexts/tagsets limit

rayz:
## Summary

This pr adds configurability in the dogstatsd source for the cached contexts and tagsets limit.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance


lib/saluki-components/src/sources/dogstatsd/resolver.rs
```diff
@@ -54,6 +59,7 @@ impl ContextResolvers {
 
         let no_agg_resolver = ContextResolverBuilder::from_name(format!("{}/dsd/no_agg", context.component_id()))?
             .with_interner_capacity_bytes(context_string_interner_size)
+            .with_cached_contexts_limit(cached_contexts_limit)
```
tobz, line 62: Don't need this here since we don't cache for the no-agg resolver.

tobz (changes requested):

PR #882, opened 2025-08-27
Add broadcast for config updates

rayz:
## Summary

POC for adding a similar mechanism to the `OnUpdate` used in the core-agent. It is not quite the same since we are not using callbacks here.

This pr updates `run.rs` so that a bootstrap configuration is created. If dynamic configuration is enabled, a `mpsc::channel` is created. The sender and bootstrap config is used for `create_config_stream(bootstrap_config, sender)`.

The `receiver` is used in the background task `run_dynamic_config_updater` which is spawned when `into_generic` is called. 

This task is responsible for:
- receiving the initial snapshot and using `Notify` to signal that it has been received. (A new method `ready` has been added to `GenericConfiguration` and it will block until the signal has been received. This is only used if dynamic configuration is enabled.)
- listening for all subsequent updates on a mpsc channel, diffing the changes against the current state, and broadcasting `ConfigChangeEvents` to any subscribers.

Configuration priority is determined by the order of which the providers are added.

Component OnUpdate Mechanism: This system enables components to react to changes. By calling `config.subscribe_for_updates()`, a component can receive a stream of `ConfigChangeEvents`. It can then listen for changes to relevant keys and update its internal state accordingly.
 

Example: https://github.com/DataDog/saluki/pull/883 (edit: outdated)

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

1. Run ADP with the agent
2. Run `config` subcommand in ADP
3. Change some configs
4. Run `config` subcommand in ADP


lib/saluki-config/src/dynamic/event.rs
```diff
@@ -0,0 +1,26 @@
+//! Defines the event type for configuration changes.
+
+use serde_json::Value as JsonValue;
+
+/// An event that occurs when the configuration changes.
+#[derive(Debug, Clone)]
+pub enum ConfigChangeEvent {
+    /// A snapshot of the configuration has been received.
+    Snapshot,
```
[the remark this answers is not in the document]
rayz: Removed 👍

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -40,45 +38,48 @@ pub async fn run(started: Instant, run_config: RunConfig) -> Result<(), GenericE
...
+    let in_standalone_mode = static_config.get_typed_or_default::<bool>("adp.standalone_mode");
+    let use_new_config_stream_endpoint =
+        static_config.get_typed_or_default::<bool>("adp.use_new_config_stream_endpoint");
+
+    let configuration = if !in_standalone_mode && use_new_config_stream_endpoint {
+        // If we're not in standalone mode and the config stream is enabled, we
+        // set up a channel for the config stream to communicate with the updater task.
+        let (sender, receiver) = mpsc::channel(100);
+
+        if let Err(e) = create_config_stream(&static_config, sender).await {
```
tobz, line 60: `create_config_stream` should just create the channel and return the receiver to be used.

lib/saluki-config/src/dynamic/diff.rs
```diff
@@ -0,0 +1,47 @@
+//! Functions for diffing configuration values.
+
+use super::event::ConfigChangeEvent;
+
+/// Diffs two configuration values and returns a list of changes.
+pub fn diff_config(old_config: &figment::value::Value, new_config: &figment::value::Value) -> Vec<ConfigChangeEvent> {
```
tobz, line 6: I think we just want a basic unit test, or two, for this. Primarily to show that we emit the right event type for modifications vs additions, and also to show that for dictionaries, we emit an event _per_ key rather than just saying "yeah, these two dicts are different".

lib/saluki-config/src/dynamic/event.rs
```diff
@@ -0,0 +1,38 @@
...
+        new_value: JsonValue,
+    },
+}
+
+/// An update message for the dynamic configuration state, sent from the config stream to the updater task.
+#[derive(Clone, Debug)]
+pub enum ConfigUpdate {
+    /// A complete snapshot of the configuration. The existing state should be replaced.
+    Snapshot(serde_json::Value),
+    /// A partial update for a single key-value pair. This should be merged into the existing state.
```
tobz, line 31: For multi-sentence doc comments like this, we should split them up like so:

```
/// A complete snapshot of the configuration.
///
/// The existing state should be replaced.
```

lib/saluki-config/src/lib.rs
```diff
@@ -420,6 +594,16 @@ pub struct GenericConfiguration {
...
+    /// Waits for the configuration to be ready, if dynamic configuration is enabled.
+    ///
+    /// This should NOT be called unless dynamic configuration is in use.
+    ///
+    /// If dynamic configuration is in use, this method will asynchronously wait until the first snapshot has been
+    /// received and applied.
+    pub async fn ready(&self) {
+        self.inner.ready_signal.notified().await;
+    }
+
```
tobz, line 606: One problem with this approach is that if called multiple times, it would always have to wait for the _next_ snapshot even... which may never come. Another problem is that for people not using dynamic configuration, calling this represents a deadlock.

I _think_ we could do better here, but it's a little more code to wrangle. We could switch to using `tokio::sync::Mutex<Option<oneshot::Receiver<()>>>` instead. Our logic would instead look something like this:

```
// We need a lock to both ensure that multiple callers can race against this,
// and to allow us mutable access to consume the receiver:
let maybe_ready_rx = self.inner.ready_signal.lock().await;
if let Some(ready_rx) = maybe_ready_rx.take() {
    // We're the first caller to wait for readiness.
    ready_rx.await;
}

// If we got here, then that means the first snapshot has already been received,
// either by us or by a previous caller that actually consumed the ready receiver.
```

This would solve both problems: once the receiver is gone, our "got the initial snapshot" signal becomes latched (problem one) and if we simply set the value of the mutex to `None` when we have no dynamic configuration, then we never await anything here in the case of a static configuration (problem two).

bin/agent-data-plane/src/internal/control_plane.rs
```diff
@@ -40,7 +40,7 @@ pub fn spawn_control_plane(
     let privileged_api = APIBuilder::new()
         .with_self_signed_tls()
         .with_optional_handler(acquire_logging_api_handler())
-        .with_optional_handler(config.get_refreshable_handle().map(ConfigAPIHandler::from_state))
+        .with_optional_handler(Some(ConfigAPIHandler::new(config.clone())))
```
tobz, line 43: I don't think this needs to be optional anymore.

lib/saluki-config/src/dynamic/diff.rs
```diff
@@ -0,0 +1,111 @@
...
+    use figment::{providers::Serialized, value::Value, Figment};
+    use serde_json::json;
+
+    use super::*;
+
+    fn to_figment_value(json: serde_json::Value) -> Value {
+        let serialized = Serialized::defaults(json);
+        let value: Value = Figment::from(serialized).extract().unwrap();
+        value
+    }
```
tobz, line 60: Probably want to add a dummy test for when there's no change?

tobz (approved):

PR #890, opened 2025-09-02
Add some more logic to Python checks

fabbing:
## Summary

This PR adds the `tracemalloc_enabled`, `headers` and implements `get_version` functions for Python checks.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Added a small unit test that checks this functions return value.

## References

---

I'm not totally convinced by a few things:
- I introduced the `ExecutionContext` type to cache information needed for Python checks: it justs cache the hostname from the `EnvironmentProvider`, and provides the default HTTP headers. In the future, it should also cache information retrieved form a running datadog-agent.
I'm not a fan of leaking this Python check specific type to the entire `Checks` infrastructure, but it may be also provide useful information to them. Should I move it around?
- I'm not happy with the growing number of `static` globals in `python_module.rs`. Maybe should we group them under an unique `struct`?
-  I've added a small config file `lib/saluki-components/src/sources/checks/builder/python/tests/test_checks_config.yaml` to be used for the unit test. I'm not sure what's the best practice for where to store such test files?

### DIFF
--- a/lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
+++ b/lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
@@ -280,7 +280,7 @@ pub mod datadog_agent {
     #[pyfunction]
     fn get_version() -> &'static str {
         trace!("Called get_version()");
-        "0.0.0"
+        saluki_metadata::get_app_details().version().raw()
     }
 
     #[pyfunction]
--- a/lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
+++ b/lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
@@ -83,7 +83,7 @@ fn fetch_tracemalloc_enabled() -> bool {
 }
 
 fn fetch_http_headers() -> &'static HashMap<String, String> {
-    static EMPTY: LazyLock<HashMap<String, String>> = LazyLock::new(|| HashMap::new());
+    static EMPTY: LazyLock<HashMap<String, String>> = LazyLock::new(HashMap::new);
 
     match GLOBAL_EXECUTION_CONTEXT.get() {
         Some(ec) => ec.http_headers(),
--- a/lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
+++ b/lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
@@ -266,15 +266,21 @@ pub mod datadog_agent {
     use super::*;
 
     #[pyfunction]
+    fn get_config(config_option: String) -> String {
+        trace!("Called get_config({})", config_option);
+        get_config_key(config_option)
+    }
+
+    #[pyfunction]
     fn get_hostname() -> &'static str {
         trace!("Called get_hostname()");
         fetch_hostname()
     }
 
     #[pyfunction]
-    fn get_config(config_option: String) -> String {
-        trace!("Called get_config({})", config_option);
-        get_config_key(config_option)
+    fn tracemalloc_enabled() -> bool {
+        trace!("Called tracemalloc_enabled()");
+        fetch_tracemalloc_enabled()
     }
 
     #[pyfunction]
@@ -284,21 +290,15 @@ pub mod datadog_agent {
     }
 
     #[pyfunction]
-    fn set_check_metadata(check_id: String, name: String, value: String) {
-        debug!("Called set_check_metadata({}, {}, {})", check_id, name, value);
-        // Again, we can only log this because there's no structure to store it.
-    }
-
-    #[pyfunction]
     fn headers() -> &'static HashMap<String, String> {
         trace!("Called headers()");
         fetch_http_headers()
     }
 
     #[pyfunction]
-    fn tracemalloc_enabled() -> bool {
-        trace!("Called tracemalloc_enabled()");
-        fetch_tracemalloc_enabled()
+    fn set_check_metadata(check_id: String, name: String, value: String) {
+        debug!("Called set_check_metadata({}, {}, {})", check_id, name, value);
+        // Again, we can only log this because there's no structure to store it.
     }
 }
 
--- a/lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
+++ b/lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
@@ -3,7 +3,6 @@ use std::sync::{LazyLock, OnceLock};
 
 use pyo3::prelude::*;
 use pyo3::types::PyDict;
-use saluki_config::GenericConfiguration;
 use saluki_context::tags::TagSet;
 use saluki_core::data_model::event::{
     eventd::{AlertType, EventD, Priority},
@@ -18,21 +17,14 @@ use crate::sources::checks::check_metric::{CheckMetric, MetricType};
 use crate::sources::checks::execution_context::ExecutionContext;
 
 // Global state to store the sender
-static METRIC_SENDER: OnceLock<Sender<Event>> = OnceLock::new();
-// Global state to store the configuration
-static GLOBAL_CONFIGURATION: OnceLock<GenericConfiguration> = OnceLock::new();
+static GLOBAL_METRIC_SENDER: OnceLock<Sender<Event>> = OnceLock::new();
 
 // Global state to store the execution context
 static GLOBAL_EXECUTION_CONTEXT: OnceLock<ExecutionContext> = OnceLock::new();
 
 /// Sets the event sender to be used by the aggregator module.
 pub fn set_event_sender(check_metrics_tx: Sender<Event>) -> &'static Sender<Event> {
-    METRIC_SENDER.get_or_init(|| check_metrics_tx)
-}
-
-/// Sets the configuration to be used by the datadog_agent module.
-pub fn set_configuration(configuration: GenericConfiguration) -> &'static GenericConfiguration {
-    GLOBAL_CONFIGURATION.get_or_init(move || configuration)
+    GLOBAL_METRIC_SENDER.get_or_init(|| check_metrics_tx)
 }
 
 /// Sets the hostname to be used by the datadog_agent module.
@@ -41,7 +33,7 @@ pub fn set_execution_context(execution_context: ExecutionContext) -> &'static Ex
 }
 
 fn try_send_metric(metric: CheckMetric) -> Result<(), GenericError> {
-    match METRIC_SENDER.get() {
+    match GLOBAL_METRIC_SENDER.get() {
         Some(sender) => sender
             .try_send(metric.into())
             .map_err(|e| generic_error!("Failed to send metric: {}", e)),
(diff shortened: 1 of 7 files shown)


lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -67,18 +70,33 @@ fn try_send_event(event: EventD) -> Result<(), GenericError> {
...
+        Some(ExecutionContext { hostname, .. }) => hostname.as_str(),
         None => "",
     }
 }
 
+fn fetch_tracemalloc_enabled() -> bool {
+    match GLOBAL_CONFIGURATION.get() {
+        Some(configuration) => configuration.get_typed_or_default::<bool>("tracemalloc_debug"),
+        None => false, // FIXME: this returns a supposed default. Can we get the default without a configuration?
+    }
```
tobz, line 88: Can we get a default with the configuration? Yeah, just set it to whatever the default is for the setting itself. :P

What I'm more interested with, overall, is how we can get away from these global statics and their associated getters. It _feels_ like it should be possible for us to define a global static on the _Python_ side which references back to a value we own whose lifetime is tied to the source component itself... but I haven't personally investigated whether this is possible and/or doable without changing the Python API that checks depend on.

lib/saluki-components/src/sources/checks/execution_context.rs
```diff
@@ -0,0 +1,50 @@
...
+use saluki_env::{EnvironmentProvider, HostProvider};
+use saluki_metadata;
+use tracing::warn;
+
+// Cache execution information from datadog agent for Python checks
+#[allow(dead_code)] // FIXME temporary
+#[derive(Clone)]
+pub struct ExecutionContext {
+    pub hostname: String,
+    pub http_headers: HashMap<String, String>,
```
tobz, line 12: I would expose getter methods for these so that we don't need to be destructuring `ExecutionContext` every time we want to use it, and updating those destructuring blocks if we want to internally refactor it.

tobz (changes requested):
> I'm not a fan of leaking this Python check specific type to the entire Checks infrastructure, but it may be also provide useful information to them. Should I move it around?

I think it's fine as-is. This _is_ the sort of thing we'd want to expose to Rust-based checks, or WebAssembly-checks, or whatever.

> I'm not happy with the growing number of static globals in python_module.rs. Maybe should we group them under an unique struct?

Me either. I left a small comment with my thoughts on this.

> I've added a small config file `lib/saluki-components/src/sources/checks/builder/python/tests/test_checks_config.yaml` to be used for the unit test. I'm not sure what's the best practice for where to store such test files?

That's fine for the moment, I'd say.

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -35,9 +35,11 @@ use self::check::Check;
 
 mod builder;
 
+mod execution_context;
 #[cfg(feature = "python-checks")]
 use self::builder::python::builder::PythonCheckBuilder;
 use self::builder::CheckBuilder;
+use self::execution_context::ExecutionContext; // FIXME: leaking context for other checks
```
GustavoCaso, line 42: Is it actually leaking for other checks? Probably, this execution context is something we would use for other types of checks. Also, it's too early to know 😄

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
fabbing: I'll try to investigate if this is possible, when time allows!

lib/saluki-components/src/sources/checks/execution_context.rs
fabbing: I drop this particular change in the rebase, since I misunderstood what you meant there 😅

fabbing:
Rebased again to sync with `main` and clean up the history a bit!

lib/saluki-components/src/sources/checks/execution_context.rs
```diff
@@ -0,0 +1,73 @@
+use std::collections::HashMap;
+
+use saluki_config::GenericConfiguration;
+use saluki_env::{EnvironmentProvider, HostProvider};
+use saluki_metadata;
+use tracing::warn;
+
+// Cache execution information from datadog agent for Python checks
```
tobz, line 8:
```suggestion
/// Global/shared configuration for checks.
```

In Rust, "doc" comments are indicated by three forward slashes (instead of the traditional two), which means they get consumed and used for things like IDE hover hints, generated documentation (`docs.rs`, etc), and so on... so any time we're adding documentation on public types/methods/fields, we want to use the triple forward slash format.

I've also just provided a slightly more generic description here, as we try to avoid specifically always referring back to the Datadog Agent in terms of what something is related to.

lib/saluki-components/src/sources/checks/execution_context.rs
```diff
@@ -0,0 +1,73 @@
...
+#[derive(Clone)]
+pub struct ExecutionContext {
+    configuration: GenericConfiguration,
+    hostname: String,
+    http_headers: HashMap<String, String>,
+}
+
+impl ExecutionContext {
```
tobz, line 16: All of the public methods in this `impl` block should have doc comments.

lib/saluki-components/src/sources/checks/execution_context.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub fn configuration(&self) -> &GenericConfiguration {
+        &self.configuration
+    }
+
+    pub fn hostname(&self) -> &str {
+        &self.hostname
+    }
+
+    #[allow(dead_code)]
+    pub fn set_hostname<S: AsRef<str>>(self, hostname: S) -> Self {
```
tobz, line 63: For setter methods, we follow this naming scheme:

- `set_foo`: used when the method receiver takes a mutable reference to `self` (`fn set_foo(&mut self, ...)`)
- `with_foo`: used when the method receiver consumes `self` (`fn with_foo(self, ...)`)

lib/saluki-components/src/sources/checks/execution_context.rs
```diff
@@ -0,0 +1,73 @@
...
+        ExecutionContext {
```
tobz, line 64: You can simply use `Self` here.

tobz (changes requested):

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -378,37 +390,75 @@ mod tests {
...
+            .from_yaml("src/sources/checks/builder/python/tests/test_checks_config.yaml")
+            .expect("configuration should be loaded")
+            .into_generic()
+            .expect("convert to generic configuration");
+
+        let execution_context = ExecutionContext::new(generic_configuration).set_hostname("agent-test-host");
+        set_execution_context(execution_context);
+
+        pyo3::append_to_inittab!(datadog_agent);
+        pyo3::prepare_freethreaded_python();
```
sgnn7, line 430: Drive-by: this seems deprecated: https://docs.rs/pyo3/latest/pyo3/fn.prepare_freethreaded_python.html

lib/saluki-components/src/sources/checks/execution_context.rs
fabbing: Thanks for the explanation and improvement!

lib/saluki-components/src/sources/checks/execution_context.rs
fabbing: Makes sense! Thanks for explaining the convention :)

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -65,20 +60,37 @@ fn try_send_event(event: EventD) -> Result<(), GenericError> {
...
+    match GLOBAL_EXECUTION_CONTEXT.get() {
+        Some(execution_context) => execution_context
+            .configuration()
+            .get_typed_or_default::<bool>("tracemalloc_debug"),
+        None => false,
+    }
+}
+
+fn fetch_http_headers() -> &'static HashMap<String, String> {
+    static EMPTY: LazyLock<HashMap<String, String>> = LazyLock::new(|| HashMap::new());
```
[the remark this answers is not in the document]
GustavoCaso: Also, you can remove the closure: `static EMPTY: LazyLock<HashMap<String, String>> = LazyLock::new(HashMap::new);`

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -65,20 +60,37 @@ fn try_send_event(event: EventD) -> Result<(), GenericError> {
...
 }
 
+fn fetch_tracemalloc_enabled() -> bool {
+    match GLOBAL_EXECUTION_CONTEXT.get() {
+        Some(execution_context) => execution_context
+            .configuration()
+            .get_typed_or_default::<bool>("tracemalloc_debug"),
+        None => false,
+    }
+}
```
GustavoCaso, line 83: Maybe we can use a generic `get_config_key` and use it to fetch `tracemalloc_debug` instead of having a dedicated function for `fetch_tracemalloc_enabled`

Something like: 

```rust 
fn get_config_key<'a, T>(key: &str) -> T
where
    T: Default + Deserialize<'a>,
{
    match GLOBAL_EXECUTION_CONTEXT.get() {
        Some(execution_context) => execution_context.configuration().get_typed_or_default::<T>(key),
        None => T::default(),
    }
}
```

Note, that I'm not sure if this idiomatic Rust, better as @tobz 🙏

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
GustavoCaso: That is because we are using version 0.25 and the latest version is 0.26

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
fabbing: That sound like a very good idea! Implemented in the last commit.

fabbing:
Rebased on `main`.

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -56,29 +52,43 @@ fn try_send_service_check(service_check: ServiceCheck) -> Result<(), GenericErro
...
-    match GLOBAL_HOST.get() {
-        Some(host) => host.as_str(),
+    match GLOBAL_EXECUTION_CONTEXT.get() {
+        Some(execution_context) => execution_context.hostname(),
         None => "",
     }
 }
 
+fn fetch_http_headers() -> &'static HashMap<String, String> {
+    static EMPTY: LazyLock<HashMap<String, String>> = LazyLock::new(HashMap::new);
```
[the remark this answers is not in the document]
fabbing: This is fine, there will be only one instance of `EMPTY` for the process.

GustavoCaso (approved):
Awesome work my friend 🚀

lib/saluki-components/src/sources/checks/execution_context.rs
```diff
@@ -0,0 +1,87 @@
...
+/// This provides information to checks, either from the configuration or computed from the EnvironmentProvider if one
+/// is provided.
+#[derive(Clone)]
+pub struct ExecutionContext {
+    configuration: GenericConfiguration,
+    hostname: String,
+    http_headers: HashMap<String, String>,
+}
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
fabbing: Yes, because some fonctions are only used in tests at the moment :/

tobz (approved):

PR #894, opened 2025-09-03
Add log event type

andrewqian2001datadog:
## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

### DIFF
--- a/lib/saluki-core/src/data_model/event/log/mod.rs
+++ b/lib/saluki-core/src/data_model/event/log/mod.rs
@@ -1,0 +1,88 @@
+//! Logs.
+
+use std::num::NonZeroU64;
+
+use stringtheory::MetaString;
+
+// https://docs.datadoghq.com/api/latest/logs/#log-object
+/// Internal representation of a log event aligned to Datadog logs payload fields.
+#[derive(Clone, Debug, PartialEq)]
+pub struct Log {
+    /// Log message body.
+    message: MetaString,
+    /// Log status/severity (e.g., "info", "warn", "error").
+    status: Option<MetaString>,
+    /// Timestamp in seconds since Unix epoch.
+    timestamp: Option<NonZeroU64>,
+    /// Hostname associated with the log.
+    hostname: MetaString,
+    /// Service associated with the log.
+    service: MetaString,
+    /// Source of the log (ddsource).
+    source: MetaString,
+    // comma-separated tags: key:value,key:value
+    /// Comma-separated tags (ddtags), format: key:value,key:value.
+    tags: MetaString,
+}
+
+impl Log {
+    /// Creates a new `Log` with the given message.
+    pub fn new(message: impl Into<MetaString>) -> Self {
+        Self {
+            message: message.into(),
+            status: None,
+            timestamp: None,
+            hostname: MetaString::empty(),
+            service: MetaString::empty(),
+            source: MetaString::empty(),
+            tags: MetaString::empty(),
+        }
+    }
+
+    /// Sets the log status/severity.
+    pub fn with_status(mut self, status: impl Into<Option<MetaString>>) -> Self {
+        self.status = status.into(); self
+    }
+
+    /// Sets the timestamp in seconds since Unix epoch. `None` clears the timestamp.
+    pub fn with_timestamp_unix_s(mut self, ts: impl Into<Option<u64>>) -> Self {
+        self.timestamp = ts.into().and_then(NonZeroU64::new); self
+    }
+
+    /// Sets the hostname.
+    pub fn with_hostname(mut self, hostname: impl Into<Option<MetaString>>) -> Self {
+        self.hostname = hostname.into().unwrap_or_else(MetaString::empty); self
+    }
+
+    /// Sets the service name.
+    pub fn with_service(mut self, service: impl Into<Option<MetaString>>) -> Self {
+        self.service = service.into().unwrap_or_else(MetaString::empty); self
+    }
+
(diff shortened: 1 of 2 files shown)


andrewqian2001datadog:
https://gitlab.ddbuild.io/DataDog/saluki/-/jobs/1111289625

image

I ran the recommended command
`cargo update -p tracing-subscriber`

lib/saluki-components/src/sources/checks/builder/python/python_modules.rs
```diff
@@ -351,6 +351,7 @@ mod tests {
                         panic!("Unexpected event title: {}", ev.title());
                     }
                 }
+                saluki_core::data_model::event::Event::Log(_) => todo!(),
```
tobz, line 354: `saluki_core::data_model::event::Event` is already in scope as `Event` here, so this can just be `Event::Log(_) => todo!()`.

lib/saluki-core/src/data_model/event/log/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub fn message(&self) -> &str {
+        &self.message
+    }
+    /// Returns the timestamp in seconds, if set.
+    pub fn timestamp_unix_s(&self) -> Option<u64> {
+        self.timestamp.map(|v| v.get())
+    }
+}
+
+#[cfg(test)]
```
rayz, line 86: I think this can be removed, since it's not really useful for now

lib/saluki-core/src/data_model/event/log/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    /// Sets the log status/severity.
+    pub fn with_status(mut self, status: impl Into<Option<MetaString>>) -> Self {
+        self.status = status.into();
+        self
+    }
+
+    /// Sets the timestamp in seconds since Unix epoch. `None` clears the timestamp.
```
tobz, line 47: In Rust, doc comments have an implicit structure, which affects how they're used/displayed in IDE hover hints, in generated docs, and so on. The first line is meant to be a summary/high-level description, and any further detail should come as a new paragraph.

We already have an equivalent method for `EventD` which we should mirror here, in terms of its doc comment:

```rust
    /// Set the timestamp.
    ///
    /// Represented as a Unix timestamp, or the number of seconds since the Unix epoch.
```

lib/saluki-core/src/data_model/event/log/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+use saluki_context::tags::SharedTagSet;
+use stringtheory::MetaString;
+
+/// Internal representation of a log event aligned to Datadog logs payload fields https://docs.datadoghq.com/api/latest/logs/#log-object.
+#[derive(Clone, Debug, PartialEq)]
+pub struct Log {
+    /// Log message body.
+    message: MetaString,
+    /// Log status/severity (e.g., "info", "warn", "error").
+    status: Option<MetaString>,
```
rayz, line 14: Would it be better to use an enum for the status?

tobz: My question as well, unless this is specifically allowed to be free-form to cope with non-standard log levels.

lib/saluki-core/src/data_model/event/log/mod.rs
```diff
@@ -0,0 +1,85 @@
+//! Logs.
+
+use std::num::NonZeroU64;
+
+use saluki_context::tags::SharedTagSet;
+use stringtheory::MetaString;
+
+/// Internal representation of a log event aligned to Datadog logs payload fields https://docs.datadoghq.com/api/latest/logs/#log-object.
```
tobz, line 8: We should just say "A log event." here.

In general, we try to avoid language about exactly how one field maps to the equivalent field in the Agent codebase, or some specific Datadog feature, and so on. At the end of the day, yeah, we're using Saluki to build Agent Data Plane, but there's no reason that it couldn't be used to build something generic, and that's the mindset we want to be in when writing this stuff.

The exception here is when a field is truly Datadog-specific, in terms of the format of the data itself or something along those lines.

lib/saluki-core/src/data_model/event/log/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /// Sets the log source (ddsource).
+    pub fn with_source(mut self, source: impl Into<Option<MetaString>>) -> Self {
+        self.source = source.into().unwrap_or_else(MetaString::empty);
+        self
+    }
+
+    /// Sets the comma-separated Datadog tags string (ddtags).
+    pub fn with_ddtags(mut self, ddtags: impl Into<Option<SharedTagSet>>) -> Self {
```
tobz, line 72: We should replace all instances of `ddtags` with `tags`.

lib/saluki-core/src/data_model/event/log/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    /// Returns the message string slice.
+    pub fn message(&self) -> &str {
+        &self.message
+    }
+    /// Returns the timestamp in seconds, if set.
+    pub fn timestamp_unix_s(&self) -> Option<u64> {
+        self.timestamp.map(|v| v.get())
+    }
+}
```
tobz, line 85: Seems like we probably want getters for the rest of the fields as well?

tobz (changes requested):

PR #901, opened 2025-09-08
chore(agent-data-plane): add shared API client for interacting with (un)privileged APIs

tobz:
## Summary

This PR does some minor reworking of how we talk to the API endpoints exposed by ADP for the purposes of performing operations and debugging.

Where we before directly constructed an HTTP client, and specified the endpoint, and handled the possible errors in slightly different ways... we now have a somewhat more strongly-typed API client (called `APIClient`, heh) which exposes specific methods for these different endpoint paths.

The client is meant to at least harmonize the error handling aspect, but it doesn't go as far as necessarily deserializing the responses into strongly-typed structures, and so on. The idea is that this will _allow_ us to do that in the future, as well as centralize how the client is built so we can improve aspects such as not using a hardcoded endpoint/port, or switching to a different socket transport, etc, all in one place.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Built and ran ADP locally, and ensured that all existing subcommands still worked as expected.

## References

AGTMETRICS-233

### DIFF
--- a/bin/agent-data-plane/src/cli/utils.rs
+++ b/bin/agent-data-plane/src/cli/utils.rs
@@ -36,15 +36,15 @@ impl APIClient {
     /// If the request fails, or the server responds with an unexpected status code, an error is returned.
     pub async fn set_log_level(&self, filter_directives: String, duration_secs: u64) -> Result<(), GenericError> {
         let url = self.get_privileged_url("/logging/override");
-        let result = self
+        let response = self
             .inner
             .post(url)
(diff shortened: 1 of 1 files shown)


rayz (approved):

tobz:
### Merge activity

* **Sep 9, 4:59 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/DataDog/saluki/901).

PR #908, opened 2025-09-15
chore(saluki-config): add `ConfigurationLoader::for_tests`

rayz:
## Summary

This pr adds a method `ConfigurationLoader::for_tests` and also a bunch of unit tests that uses this method.

The dynamic updater has also been updated to handle nested keys that use "`.`". Previously, it would just create the key on the top level of the map object. 

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Unit tests

### DIFF
--- a/lib/saluki-config/src/lib.rs
+++ b/lib/saluki-config/src/lib.rs
@@ -442,6 +442,53 @@ impl ConfigurationLoader {
             })
         }
     }
+
+    /// Configures a [`ConfigurationLoader`] that is suitable for tests.
+    ///
+    /// This configures the builder with the following defaults:
+    ///
+    /// - configuration from a JSON file
+    /// - configuration from environment variables
+    ///
+    /// If `enable_dynamic_configuration` is true, a dynamic configuration sender is returned.
+    ///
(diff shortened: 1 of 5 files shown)


lib/saluki-config/src/lib.rs
```diff
@@ -442,6 +442,65 @@ impl ConfigurationLoader {
...
+            maybe_sender = Some(sender);
+        }
+
+        static ENV_MUTEX: OnceLock<std::sync::Mutex<()>> = OnceLock::new();
+
+        let guard = ENV_MUTEX.get_or_init(|| std::sync::Mutex::new(())).lock().unwrap();
+
+        if let Some(pairs) = env_vars.as_ref() {
+            for (k, v) in pairs.iter() {
+                std::env::set_var(k, v);
```
tobz, line 479: I think we'd want to prefix all of these with `TEST` so that the caller doesn't have to.

tobz (changes requested):

tobz (approved):

PR #910, opened 2025-09-18
implement preaggregation allowlist via metric router transform

lukesteensen:
## Summary

This adds configuration to allow routing certain metrics exclusively to the preaggregation pipeline. When this configuration is not present but preaggregation is still enabled, we will continue the current behavior of shipping all metrics to both pipelines.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## References

TPOT-154

### DIFF
--- a/bin/agent-data-plane/src/cli/debug/mod.rs
+++ b/bin/agent-data-plane/src/cli/debug/mod.rs
@@ -2,7 +2,7 @@ use tracing::{error, info};
 
 use crate::{
     cli::utils::APIClient,
-    config::{DebugConfig, SetLogLevelConfig},
+    config::{DebugConfig, SetLogLevelConfig, SetMetricLevelConfig},
 };
 
 mod workload;
@@ -13,6 +13,8 @@ pub async fn handle_debug_command(config: DebugConfig) {
     match config {
         DebugConfig::ResetLogLevel => reset_log_level().await,
         DebugConfig::SetLogLevel(config) => set_log_level(config).await,
+        DebugConfig::ResetMetricLevel => reset_metric_level().await,
+        DebugConfig::SetMetricLevel(config) => set_metric_level(config).await,
         DebugConfig::Workload(config) => handle_workload_command(config).await,
     }
 }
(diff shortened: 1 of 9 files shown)


lib/saluki-components/src/transforms/metric_router/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+        matched_dispatcher.flush().await?;
+        unmatched_dispatcher.flush().await?;
+
+        debug!(
+            matched_events = matched_count,
+            unmatched_events = unmatched_count,
+            "Successfully processed event batch"
+        );
+
+        Ok(())
```
tobz, line 95: I think could optimize this slightly for the cases where no metrics are being matched: use `FixedSizeEventBuffer::extract` to avoid allocating a brand new event buffer for the unmatched metrics, and then just pass the original/remaining event buffer via `context.dispatcher().dispatch_named(...)`.

In that case, we only pay the penalty of iterating over the metrics... but the iterator won't return anything, and so the buffered dispatcher will never allocate anything, and we just end up using the input event buffer as-is.

lib/saluki-components/src/transforms/metric_router/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+
+#[async_trait]
+impl Transform for MetricRouter {
+    async fn run(self: Box<Self>, mut context: TransformContext) -> Result<(), GenericError> {
+        let mut health = context.take_health_handle();
+
+        health.mark_ready();
+        debug!(
+            metric_names = ?self.metric_names,
+            "Metric router transform started"
```
tobz, line 158:
```suggestion
            "Metric Router transform started."
```

lib/saluki-components/src/transforms/metric_router/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+                        }
+                    }
+                    None => {
+                        debug!("Event stream terminated, shutting down metric router transform");
+                        break;
+                    }
+                }
+            }
+        }
+
```
tobz, line 179:
```suggestion

    debug!("Metric Router transform stopped.");
```

lib/saluki-components/src/transforms/metric_router/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+        );
+
+        loop {
+            select! {
+                _ = health.live() => continue,
+                maybe_events = context.events().next() => match maybe_events {
+                    Some(events) => {
+                        // Process the batch of events with proper error handling
+                        if let Err(e) = self.process_event_batch(events, &mut context).await {
+                            error!(error = %e, "Failed to process event batch");
```
tobz, line 168:
```suggestion
                            error!(error = %e, "Failed to process event batch.");
```

lib/saluki-components/src/transforms/metric_router/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+            }
+        }
+
+        matched_dispatcher.flush().await?;
+        unmatched_dispatcher.flush().await?;
+
+        debug!(
+            matched_events = matched_count,
+            unmatched_events = unmatched_count,
+            "Successfully processed event batch"
```
tobz, line 92:
```suggestion
            "Successfully processed event batch."
```

PR #919, opened 2025-09-24
enhancement(app, core): add support for dynamically filtering internal metrics by level

tobz:
## Summary

This PR adds support to dynamically filter internal metrics (at the point of collection) in order to allow reducing baseline internal metrics cardinality.

Saluki emits a lot of internal telemetry. Some of this telemetry is effectively "table stakes", required to understand basic behavior of the system and what the current workload looks like. Other portions of the telemetry are _interesting_ but not required: they can sometimes help to debug problems but aren't always relevant. Overall, the internal telemetry we currently emit translates into a lot of metrics to have to ingest and store. Given that this costs money, and we don't need _all_ of our internal telemetry _all_ of the time, we should have a way to expose optional internal telemetry only when required.

This PR adds the ability to filter specific levels of internal telemetry in a fashion similar to Saluki's dynamic log filtering support. We use the `metrics` crate for all internal metrics, which allows specifying a "level" as part of the [metric metadata](https://docs.rs/metrics/latest/metrics/#metadata). This means we can have "trace" or "debug" metrics for things that we don't _always_ need, and then classify everything else as "info", similar to how we generally think about verbosity when adding logging. However, in order to harness this, we need to a way to dynamically specify what level of metrics to emit.

Our approach is to track the level of a metric when we register it, and then during the flush period, filter out anything that is below the current filter level. This means that we effectively never emit it from the Internal Metrics source, but we're still tracking it in the internal metrics registry. Currently, we use a default filter level of DEBUG: debug metrics (and above) are passed through. This matches our current set of metrics, so there is no change to what we will emit by default.

In subsequent PRs, we'll adjust the level of various internal metrics that we emit in order to better tease apart what's required and what's optional, and then adjust the default filter level accordingly to stop emitting optional internal metrics by default.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Built and ran ADP locally and ensured all internal metrics were still emitted without being unintentionally filtered.

## References

AGTMETRICS-233

### DIFF
--- a/lib/saluki-app/src/metrics/api.rs
+++ b/lib/saluki-app/src/metrics/api.rs
@@ -117,7 +117,7 @@ async fn process_override_requests(filter_handle: FilterHandle, mut rx: mpsc::Re
         select! {
             maybe_override = rx.recv() => match maybe_override {
                 Some(Some((duration, new_level))) => {
-                    // TODO: Using the `Debug` representation of `Leve` is noisy, and we should add a method upstream to
+                    // TODO: Using the `Debug` representation of `Level` is noisy, and we should add a method upstream to
                     // just get the stringified representation of the level instead.
                     info!(level = ?new_level, "Overriding existing metric filtering directive for {} seconds...", duration.as_secs());
 


lib/saluki-app/src/metrics/api.rs
```diff
@@ -0,0 +1,157 @@
...
+    let mut override_active = false;
+    let override_timeout = sleep(Duration::MAX);
+
+    tokio::pin!(override_timeout);
+
+    loop {
+        select! {
+            maybe_override = rx.recv() => match maybe_override {
+                Some(Some((duration, new_level))) => {
+                    // TODO: Using the `Debug` representation of `Leve` is noisy, and we should add a method upstream to
```
rayz, line 120:
```suggestion
                    // TODO: Using the `Debug` representation of `Level` is noisy, and we should add a method upstream to
```

tobz:
### Merge activity

* **Sep 24, 8:13 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/DataDog/saluki/919).

PR #923, opened 2025-09-24
Add logs encoder

andrewqian2001datadog:
## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Run ADP with OTLP logs enabled
`DD_ADP_OTLP_ENABLED=true DD_OTLP_CONFIG='{}' RUST_LOG=info make run-adp-standalone`

Install grpcurl
`brew install grpcurl`

Send request
```
grpcurl -plaintext \
  -import-path /Users/andrew.qian/Documents/Code/saluki/lib/otlp-protos/proto \
  -proto opentelemetry/proto/collector/logs/v1/logs_service.proto \
  -d @ 127.0.0.1:4317 \
  opentelemetry.proto.collector.logs.v1.LogsService/Export <<'JSON'
{
  "resourceLogs": [
    {
      "resource": {
        "attributes": [
          { "key": "service.name", "value": { "stringValue": "saluki-adp" } },
          { "key": "service.version", "value": { "stringValue": "1.0.0" } },
          { "key": "deployment.environment", "value": { "stringValue": "production" } },
          { "key": "host.name", "value": { "stringValue": "my-host" } },
          { "key": "host.type", "value": { "stringValue": "server" } }
        ]
      },
      "scopeLogs": [
        {
          "scope": { 
            "name": "test-client", 
            "version": "0.1.0",
            "attributes": [
              { "key": "scope.attribute", "value": { "stringValue": "scope-value" } }
            ]
          },
          "logRecords": [
            {
              "timeUnixNano": "1735065600000000000",
              "observedTimeUnixNano": "1735065600000000000",
              "severityNumber": "SEVERITY_NUMBER_INFO",
              "severityText": "INFO",
              "body": { "stringValue": "hello from grpcurl (OTLP logs)" },
              "attributes": [
                { "key": "env", "value": { "stringValue": "dev" } },
                { "key": "user.id", "value": { "stringValue": "user123" } },
                { "key": "request.id", "value": { "stringValue": "req-456" } },
                { "key": "trace.id", "value": { "stringValue": "trace-789" } },
                { "key": "span.id", "value": { "stringValue": "span-abc" } },
                { "key": "http.method", "value": { "stringValue": "GET" } },
                { "key": "http.url", "value": { "stringValue": "https://example.com/api" } },
                { "key": "http.status_code", "value": { "intValue": "200" } },
                { "key": "error", "value": { "boolValue": false } },
                { "key": "custom.metric", "value": { "doubleValue": 42.5 } }
              ],
              "flags": 1,
              "traceId": "12345678901234567890123456789012",
              "spanId": "1234567890123456"
            }
          ]
        }
      ]
    }
  ]
}
JSON
```


lib/saluki-components/src/encoders/datadog/logs/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// Compressor level to use when the compressor kind is `zstd`. Defaults to 3.
+    #[serde(
+        rename = "serializer_zstd_compressor_level",
+        default = "default_zstd_compressor_level"
+    )]
+    zstd_compressor_level: i32,
+
+    /// Value for the `ddsource` field in Agent JSON envelope. Defaults to "otel".
+    #[serde(default = "default_ddsource")]
+    ddsource: String,
```
tobz, line 69: Seems like we should have a generic `source` field on the `Log` type itself, since we should be able to (theoretically) run any sort of log through this encoder and have an appropriate source identifier for them.

As is, we would need one encoder per log source, which is a big limitation.

lib/saluki-components/src/encoders/datadog/logs/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+
+impl LogsEndpointEncoder {
+    fn new(ddsource: String) -> Self {
+        Self {
+            ddsource,
+            tags_deduplicator: ReusableDeduplicator::new(),
+        }
+    }
+
+    fn status_to_str(status: LogStatus) -> &'static str {
```
tobz, line 182: This would be a good method to stick on `LogStatus` itself (called `as_str`, which is the idiomatic naming for this sort of thing) that way we have it if any other code needs it.

lib/saluki-components/src/encoders/datadog/logs/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+        for (k, v) in log.additional_properties() {
+            obj.insert(k.clone(), v.clone());
+        }
+
+        // If we have "otel.timestamp" as a numeric string, add a numeric "timestamp" field (nanoseconds since epoch).
+        if let Some(JsonValue::String(ns)) = obj.get("otel.timestamp") {
+            if let Ok(parsed) = ns.parse::<i64>() {
+                obj.insert("timestamp".to_string(), JsonValue::from(parsed));
+            }
+        }
```
tobz, line 237: Should we just have a `timestamp` field on `Log` itself? Do we have that for our internal log representation in the Agent?

We should try to be generic here where possible, and handle translations (like using `otel.timestamp` as the log timestamp) as far upstream as possible, ideally in the respective source.

PR #931, opened 2025-09-26
chore(smp): revamp SMP experiments for ADP

tobz:
## Summary

This PR revamps the SMP experiments we have for ADP. We've done a few things here:

- removed a lot of the one-off experiments (memory limit ones, etc) that were primarily used to originally demonstrate ADP behaviors, but have little to no value now as something to run for every change
- added many quality gate experiments to start better encoding ADP's expected resource usage for given workload profiles
- slight renaming of some existing experiments to indicate their target optimization goal
- renamed a few things from `saluki` to `agent-data-plane` to better align with the fact we're testing ADP specifically, and not just some generic Saluki thing

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

N/A

## References

AGTMETRICS-233


tobz:
* **#931** <a href="https://app.graphite.dev/github/pr/DataDog/saluki/931?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/saluki/931?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

test/smp/regression/adp/cases/dsd_uds_512kb_3k_contexts_throughput/experiment.yaml
```diff
@@ -0,0 +1,28 @@
...
+  memory_allotment: 2GiB
+
+  environment:
+    DD_API_KEY: foo00000001
+    DD_HOSTNAME: smp-regression
+    DD_DD_URL: http://127.0.0.1:9091
+
+    # Set the context limit in the aggregator to right above 3K, which matches the maximum number of contexts we expect
+    # to generate. We essentially don't want the aggregator to be a limiting factor.
+    DD_AGGREGATE_CONTEXT_LIMIT: "3100"
```
rayz, line 17: Is there a reason why we don't set `dogstatsd_cached_contexts_limit` as well? For my own experimenting, I was noticing that `adp.context_resolver_active_contexts` was higher than the number of contexts configured in my lading.yaml until I set the limit to match. (But also, I maybe be doing something wrong down stream 🤔 )

tobz: No _good_ reason to not set it, but the default is high enough that we never hit the limit... but the defaults for the aggregator _are_ low enough that we do hit it.

Makes me think that we should actually change the defaults of the aggregator to match the context resolver. 🤔

PR #937, opened 2025-09-29
chore: use `CacheBuilder` for OTLP metrics points cache

rayz:
## Summary

This pr updates the cache used in the OTLP source to use `CacheBuilder` which gives us TTL and sweep intervals. Previously, the cache (`HashMap`) just held onto the values indefinitely. 

The defaults   `delta_ttl: Duration::from_secs(60)`,  `sweep_interval: Duration::from_secs(30)` are taken from the agent code.

Main file updated: `lib/saluki-components/src/sources/otlp/metrics/cache.rs`.

The rest of the changes are updating tests to use a `for_test` implementation for the translator and cache.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Existing unit tests.


lib/saluki-components/src/sources/otlp/metrics/cache.rs
```diff
@@ -197,8 +221,8 @@ mod tests {
         message: &'static str,
     }
 
-    #[test]
-    fn test_monotonic_diff_unknown_start() {
+    #[tokio::test]
+    async fn test_monotonic_diff_unknown_start() {
```
tobz, line 225: Do these actually need to be async? My recollection was that `CacheBuilder::for_tests` should give us a cache with no expiration interval and no telemetry... which should lead us to not spawning any background tasks.

lib/saluki-components/src/sources/otlp/metrics/config.rs
```diff
@@ -122,6 +137,8 @@ impl Default for OtlpTranslatorConfig {
             instrumentation_library_metadata_as_tags: false,
             with_remapping: false,
             with_otel_prefix: false,
+            delta_ttl: Duration::from_secs(60),
+            sweep_interval: Duration::from_secs(30),
```
tobz, line 141: Do these come from somewhere in particular, or did we just come up with our own values here?

lib/saluki-components/src/sources/otlp/metrics/cache.rs
rayz: You're right, updated.

tobz (approved):

PR #950, opened 2025-10-08
enhancement(app): consolidate bootstrapping logic into `AppBootstrapper`

tobz:
## Summary

This PR consolidates all application initialization logic into a new helper, `AppBootstrapper`.

There's a lot of files touched here, so here's the high-level overview:

- we don't want people executing piecemeal `initialize_foo` functions, because it means things can get missed: we want an opinionated way to bootstrap a data plane application that we can centrally update/control
- `AppBootstrapper` acts as this new centralized mechanism, taking in a configuration file that controls how we configure various subsystems (namely logging) and ensuring we're always uniformly initializing logging, metrics, allocator telemetry, and TLS (and anything else we want to add in the future)
- since we're always initializing these subsystems now, we've removed all the feature flagging bits from `saluki-app`
- since we also want to configure logging more discretely (log to console? log to file? etc) we've moved the bootstrap configuration step earlier so it can be used in the bootstrap phase itself

## Notes

One other related change made here is how we differentiate the bootstrap configuration vs the "final" configuration. The bootstrap configuration consists of the normal YAML-based configuration file (passed via `agent-data-plane run --config <config file>` or the default config path e.g., `/etc/datadog-agent/datadog.yaml`) and environment variables. We use this for the bootstrap phase, and then when checking if ADP is in standalone mode and should reach out to the control plane for the full configuration... so far so good.

We've made a change here where we don't use that underlying configuration file for the _final_ configuration if we're sourcing things from the control plane (Core Agent): we take whatever we get from the control plane, plus any environment variable overrides, and that's it. The idea here is that we want to depend _fully_ on the control plane for configuration, with the only exception being environment variables since we need a way to override ADP-specific things like log level, and so on.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Built and ran ADP locally. I checked the following things:

- all the `run-adp`/`run-adp-standalone`/etc Make targets still work
- the logging-related configuration works (e.g., setting `DD_LOG_TO_CONSOLE=false` turns off console logging)
- tested that we properly handle using the specified config file _or_ the default path for the bootstrap config depending on which is provided
- tested that loading the configuration from the control plane still works when enabled

## References

AGTMETRICS-233

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -92,7 +92,6 @@ dependencies = [
  "bollard",
  "clap",
  "futures",
- "saluki-app",
  "saluki-error",
  "tokio",
  "tracing",
--- a/Makefile
+++ b/Makefile
@@ -242,7 +242,7 @@ endif
 	@DD_DOGSTATSD_PORT=9191 DD_DOGSTATSD_SOCKET=/tmp/adp-dogstatsd-dgram.sock DD_DOGSTATSD_STREAM_SOCKET=/tmp/adp-dogstatsd-stream.sock \
 	DD_TELEMETRY_ENABLED=true DD_PROMETHEUS_LISTEN_ADDR=tcp://127.0.0.1:5102 \
 	DD_AUTH_TOKEN_FILE_PATH=/etc/datadog-agent/auth_token \
-	target/debug/agent-data-plane
+	target/debug/agent-data-plane run
 
 .PHONY: run-adp-release
 run-adp-release: build-adp-release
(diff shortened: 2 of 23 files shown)


rayz (approved):
lgtm

tobz:
### Merge activity

* **Oct 8, 4:08 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/DataDog/saluki/950).

PR #996, opened 2025-11-06
[AGTMETRICS-440] enhancement: add new OTLP Relay component

rayz:
## Summary

This pr adds a new `Relay` component that forwards raw bytes as is to downstream components. This is needed so that we can create the OTLP Relay component which will serve as a proxy for OTLP metrics/logs/traces to the core-agent.

This pr also refactors some shared code used by the OTLP source, since we still want to keep that code for now even though we aren't doing the translation.

OTLP proxy mode is gated behind the `adp.otlp.proxy_mode` config.

Grpc Payload type has also been created since the trace-agent uses a grpc server to listen for traces.

A following pr will be made to the datadog agent to ensure that there aren't port conflicts since, the goal is for the user to send these OTLP signals to ADP which will use the default agent ports.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Manually by sending OTLP metrics,logs,traces and making sure it worked.

### DIFF
--- a/lib/saluki-components/src/common/otlp/mod.rs
+++ b/lib/saluki-components/src/common/otlp/mod.rs
@@ -6,6 +6,7 @@ pub mod config;
 
 use std::sync::Arc;
 
+use ::metrics::Counter;
 use async_trait::async_trait;
 use axum::body::Bytes;
 use axum::extract::State;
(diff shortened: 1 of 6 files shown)


lib/saluki-components/src/relays/otlp/config.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    /// The destination endpoint to forward OTLP data to.
+    ///
+    /// This is the base URL (e.g., `http://localhost:4319`) where OTLP data will be proxied.
+    /// The relay will append the appropriate paths (`/v1/metrics`, `/v1/logs`).
+    ///
+    /// Defaults to `http://localhost:4319`.
+    #[serde(default = "default_otlp_destination_endpoint")]
+    otlp_destination_endpoint: String,
+}
```
tobz: I'm not sure if I'm missing it, but this _looks_ like it would not de able to be deserialized as-is from an existing Agent configuration, just in terms of the way the fields are named.

lib/saluki-components/src/common/otlp/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+        Err(msg) => {
+            error!("Failed to handle OTLP traces: {}", msg);
+            (StatusCode::INTERNAL_SERVER_ERROR, "Internal processing error")
+        }
+    }
+}
+
+/// gRPC service implementation that delegates to the handler.
+struct GrpcServiceImpl<H> {
+    handler: std::sync::Arc<H>,
```
tobz: Import `Arc` so that we don't have all of the fully-qualified path usages.

lib/saluki-components/src/common/otlp/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+#[async_trait]
+impl<H: OtlpHandler> MetricsService for GrpcServiceImpl<H> {
+    async fn export(
+        &self, request: TonicRequest<ExportMetricsServiceRequest>,
+    ) -> Result<Response<ExportMetricsServiceResponse>, Status> {
+        self.memory_limiter.wait_for_capacity().await;
+
+        let raw_bytes = request.into_inner().encode_to_vec();
+
+        match self.handler.handle_metrics(Bytes::from(raw_bytes)).await {
```
tobz: This is fine for now, but since this is ultimately riding over HTTP... we probably want to explore trying to just manually set up the HTTP handlers so that we get the raw Protocol Buffers payloads without actually decoding.

This would let us avoid having to decode and then re-encode just to get back to `Bytes`.

lib/saluki-components/src/relays/otlp/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+                    // Create HTTP payload with destination endpoint
+                    let path = match otlp_payload.signal_type {
+                        OtlpSignalType::Metrics => "/v1/metrics",
+                        OtlpSignalType::Logs => "/v1/logs",
+                        OtlpSignalType::Traces => "/v1/traces",
+                    };
+                    let uri = format!("{}{}", self.destination_endpoint, path);
+                    let mut buffer = ChunkedBytesBuffer::new(4096);
+                    buffer.put_slice(&otlp_payload.data);
+                    let frozen_buffer = buffer.freeze();
```
tobz: We're going to end up doing a lot of allocating here.

What we should do instead is have `OtlpPayload` hold on to `Bytes` since that's what we're already getting from `OtlpHandler`. With that, we should then add an implementation of `From<Bytes>` to `FrozenChunkedBytesBuffer`, and then that way we can do a nearly zero-overhead conversion here to get things in the right shape for our HTTP request.

As noted in other comments, we can still improve on things even more than this, but this one is basically free to do _right now_.

lib/saluki-components/src/relays/otlp/config.rs
```diff
@@ -0,0 +1,83 @@
...
+    grpc_max_recv_msg_size_bytes: usize,
+
+    /// The destination endpoint to forward OTLP data to.
+    ///
+    /// This is the base URL (e.g., `http://localhost:4319`) where OTLP data will be proxied.
+    /// The relay will append the appropriate paths (`/v1/metrics`, `/v1/logs`).
+    ///
+    /// Defaults to `http://localhost:4319`.
+    #[serde(default = "default_otlp_destination_endpoint")]
+    otlp_destination_endpoint: String,
```
jszwedko: Note, I believe we'll need to be able to configure this by signal type since we will want to route traces to trace-agent and the rest to the Core Agent.

tobz (approved):

PR #1013, opened 2025-11-18
chore(correctness): add basic OTLP Traces correctness test case

tobz:
## Summary

This PR adds a basic correctness test case for OTLP Traces by running the Datadog Agent against itself.

We've done a few things here:

- added a new `traces` analysis mode
- added a new test case, `otlp-traces`, which uses the `traces` analysis mode and uses the Datadog Agent as both the baseline and comparison targets simply to ensure the test is deterministic
- exposed `lading_payload`'s existing support for generating (rudimentary) OpenTelemetry Traces payloads in `millstone`

Compared to metrics, we've taken a slightly different approach for checking equality of the emitted traces/spans. First, we've denormalized the data we receive from the targets, which is heavily chunked/nested. This means that you have to normally go through a number of layers of nesting/indirection before you get to the actual spans. Instead, we unwrap these layers and collect our own `Span` type which includes the span-specific fields, but also all of the metadata from the wrapper layers above, such as Agent and tracer metadata. Secondly, we don't _directly_ check the spans against each other for equality (i.e., `PartialEq`/`Eq` impls), but we instead of a generic diff algorithm that models the spans as tree structures. We do this because spans are highly nested objects and in order to more easily show _where_ spans differ from each other, we inherently need to be able to evaluate fields in spans in a way where we can get the calculated differences out without terminating the rest of the equality checks. Using a generic diff algorithm allows us to easily do those two things as part of one operation.

Overall, this works well and intuitively highlights the detected differences in a straightforward way. We've currently only got a single "special" comparison check in terms of coping with indeterminism, around "start" time.... but so far, things seem to be fairly deterministic which is good.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Ran all correctness test cases locally and ensured that existing test cases still passed, and that the new `otlp-traces` test case also passed.

## References

AGTMETRICS-393

### DIFF
--- a/test/correctness/otlp-traces/millstone.yaml
+++ b/test/correctness/otlp-traces/millstone.yaml
@@ -1,8 +1,8 @@
 seed: [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131]
 target: "grpc://target:4317/opentelemetry.proto.collector.trace.v1.TraceService/Export"
 aggregation_bucket_width_secs: 10
-volume: 10000
+volume: 1000
 corpus:
-  size: 10000
+  size: 1000
   payload:
     opentelemetry_traces:


bin/correctness/ground-truth/src/analysis/traces/mod.rs
```diff
@@ -0,0 +1,378 @@
...
+                        for span_diff in span_diff_recorder.diffs() {
+                            error!("- {}", span_diff);
+                        }
+                        error!("");
+
+                        error_count += 1;
+                    }
+                }
+                None => {
+                    error!("Failed to find matching comparison span ({}).", baseline_span_id);
```
rayz: Should we also error log the span id that is in comparison but not in baseline?

PR #1055, opened 2025-12-12
chore(ci): change how we build ADP container images to speed up CI

tobz:
## Summary

This PR changes how we build ADP container images in order to speed up CI and improve CI pipeline reliability.

In CI, we build ADP container images, we've got a number of constraints we have to contend with:

- we want to build ADP statically linked (this helps us get around `glibc` compatibility problems with some of the older platforms the Agent supports, without needing to use some super custom build environment/image that has an old-enough `glibc` version)
- we want to build ADP entirely in a single Dockerfile so that we can repeatably build it either locally or in CI
- Rust projects, and in particular ADP, are heavy to build and doubly so when we're trying to build maximally optimized versions in CI (`codegen-units = 1`, fat LTO, etc)

These constraints pushed us towards the current configuration:

- build ADP using an Alpine base image (gets us a native `musl`-backed build toolchain, allowing us to build ADP statically linked)
- use microVM-based runners, which let us get isolated CI runners that support building container images via Docker without depending on shared BuildKit runners that may encounter resource contention

However, the current configuration has a substantial performance issue due to the `musl`-based build toolchain just being.... _way_ slower than a comparable `glibc`-based toolchain. This seemingly comes down to the poor allocator performance of `musl`, which drags down everything about the compilation process: compiling individual crates, LTO, linking, etc. However, Alpine makes the cross-compilation to `musl` trivial compared to other approaches.

This PR reworks how we build ADP by switching to a more traditional cross-compile setup. We've done the following:

- we now use Ubuntu 24.04 as our base image for builds (this solves the performance problems we experienced with a `musl`-based build toolchain)
- we utilize Rust/Cargo's native support for cross-compilation to get us 90% of the way
- we've added some additional helpers to the Dockerfile and CI to solve the remaining problems (getting kernel headers in the right place when cross-compiling, disabling some compiler flags when targeting ARM64+`musl`, etc)

Ultimately, what we're left with is:

- we statically link ADP for both Linux AMD64/ARM64 when built in CI, but use the host target triple when building locally
- we do have to specify some additional settings into the build process, but these are hidden in the `make` targets or the CI configuration (with the ability for us to further clean up/abstract this in the future)

As a result of this, builds that previously took anywhere from 10 minutes to 15 minutes (our timeout for ADP build jobs in CI) now consistently take **4 to 5** minutes.

We've made some compromises along the way:

- we've fully disabled building in Python checks support for the moment (we weren't building it in before for ADP container images in CI, but we _were_ pulling it in for tests, etc... now we've removed it since it depends on dynamic linking and it was cleaner to ignore it wholesale for the moment)

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Ran a number of checks:

- ensured both `make build-adp-image` and `make build-adp-image-fips` still run to completion locally
- ensured the equivalent build jobs in CI pass
- manually checked the resulting ADP binaries in all images to ensure they were either dynamically or statically linked, depending on the target triple calculated due to either running locally or in CI

## References

AGTMETRICS-393

### DIFF
--- a/.gitlab/build.yml
+++ b/.gitlab/build.yml
@@ -91,7 +91,6 @@ build-adp-image-fips-amd64:
   variables:
     APP_IMAGE: ${ADP_PUBLIC_BASE_IMAGE}
     BUILD_FEATURES: "fips"
-    FIPS_ENABLED: "true"
 
 build-adp-image-fips-arm64:
   extends: [.build-common-variables, .build-arm64-variables, .build-adp-definition]
@@ -99,7 +98,6 @@ build-adp-image-fips-arm64:
   variables:
     APP_IMAGE: ${ADP_PUBLIC_BASE_IMAGE}
     BUILD_FEATURES: "fips"
-    FIPS_ENABLED: "true"
 
 build-adp-image-fips:
   extends: [.build-common-variables, .build-adp-multi-arch-definition]
@@ -150,6 +148,7 @@ build-adp-image-fips-internal-amd64:
       when: never
     - when: always
   variables:
+    APP_IMAGE: ${ADP_INTERNAL_BASE_IMAGE}
     BUILD_FEATURES: "fips"
 
 build-adp-image-fips-internal-arm64:


.gitlab/build.yml
```diff
@@ -141,47 +134,43 @@ build-adp-image-internal:
...
   rules:
     - if: !reference [.on_mq_branch, rules, if]
       when: never
     - when: always
-  needs:
-    - build-adp-image-fips-internal-amd64
-    - build-adp-image-fips-internal-arm64
+      needs:
+        - build-adp-image-fips-internal-amd64
+        - build-adp-image-fips-internal-arm64
```
tobz: The various spots here where we're moving the `needs` dictionary under the `rules` stuff above is that without doing this, it essentially lets the job run as soon as the `needs` jobs finish... even if they _fail_.

I don't know why the normal behavior of them having to  pass is ignored when `rules` comes into play, but this shift of where we apply the `needs` solves the problem. 🤷🏻

Makefile
```diff
@@ -4,43 +4,32 @@
...
-export TARGET_TRIPLE ?= $(shell command -v rustc 1>/dev/null && rustc -vV | sed -n 's|host: ||p')
-
-# High-level settings that ultimately get passed down to build-specific targets.
-export APP_FULL_NAME ?= Agent Data Plane
-export APP_SHORT_NAME ?= data-plane
-export APP_IDENTIFIER ?= adp
-export APP_GIT_HASH ?= $(shell git rev-parse --short HEAD 2>/dev/null || echo not-in-git)
-export APP_VERSION ?= $(shell cat bin/agent-data-plane/Cargo.toml | grep -E "^version = \"" | head -n 1 | cut -d '"' -f 2)
+# High-level settings that ultimately get passed down to build-specific targets and other spots.
+export TARGET_ARCH := $(shell uname -m | sed s/x86_64/amd64/ | sed s/aarch64/arm64/)
```
tobz: For many of the variables here, we're simply changing from `=?` to `:=`, and you may ask "why? what's that doing?".

In Make, variables defined via `=` are "recursively expanded", which means they're expanded every time they're referenced. `?=` is just an extension of this, where the variable isn't set to the right-hand side if it's already defined.. but it's still recursively expanded.

What this means is that it can end up running these commands we're using to get the value _many times_ when parsing the Makefile, to the point where `make` commands are noticeably slow.

For these commands where we really only need to set the value once, we use "simply expanded" variables, denoted by the use of `:=`.

Makefile
```diff
@@ -4,43 +4,32 @@
...
+export BUILD_TARGET := $(or $(BUILD_TARGET),$(shell rustc -vV | sed -n 's|host: ||p'))
+export APP_GIT_HASH := $(or $(CI_COMMIT_SHA),$(shell git rev-parse --short HEAD 2>/dev/null || echo not-in-git))
+export APP_BUILD_TIME := $(or $(CI_PIPELINE_CREATED_AT),0000-00-00T00:00:00-00:00)
```
tobz: We now prefer taking these values from environment variables set in CI, otherwise falling back to suitable values for local development.

Specifically, we don't bother with getting an accurate build time for local builds because it ends up leading to is invalidating the Cargo build cache even if nothing has changed code-wise.

Makefile
```diff
@@ -4,43 +4,32 @@
...
-export CONTAINER_TOOL ?= auto
-ifeq ($(CONTAINER_TOOL),auto)
-	ifeq ($(shell docker version >/dev/null 2>&1 && echo docker), docker)
-		override CONTAINER_TOOL = docker
-	else ifeq ($(shell podman version >/dev/null 2>&1 && echo podman), podman)
-		override CONTAINER_TOOL = podman
-	else
-		override CONTAINER_TOOL = unknown
-	endif
-endif
```
tobz: Removed this (and switched all usages of it to `docker`) because Docker is ubiquitous enough that it was just leading to clutter in the file.

If people feel strongly about being able to use another compatible equivalent like `podman`, or something else... we can always add it back. 🤷🏻

PR #1056, opened 2025-12-12
fix(ci): support profiling again in SMP experiments via direct `ddprof` usage

tobz:
## Summary

This PR reintroduces profiling of ADP during SMP experiment runs.

When we previously switched to statically linking ADP, we lost the ability for SMP to natively profile ADP during experiments, as SMP brings in `ddprof` via `LD_PRELOAD`, which is not available if the binary isn't dynamically linked. :)

This PR updates our SMP experiments to use "internal profiling", which sets up the necessary bits -- Trace Agent socket, profiling tags, etc -- but relies on the target itself to handle actually doing the profiling and sending of profile data. We accomplish this by using a wrapper script that, when it detects profiling is enabled, runs ADP under `ddprof` directly (instead of via `LD_PRELOAD`), which allows us to get profiling output.

We've added a few things to support this:
- an "internal build" flag used by the SMP-specific ADP image build jobs to cause the Docker builds to add the wrapper script and install `ddprof`
- the wrapper script itself which all SMP experiments use to invoke ADP
- updated our SMP configuration to switch to internal profiling
- updated every SMP experiment to specify the necessary "profiling only" environment variables on profiling replicated to trigger the wrapper script logic

One aspect of this is that we still won't get allocation/heap profiling, as `ddprof` depends on hooking allocator-related functions to capture that information, which is another mechanism unavailable to us with a statically linked build. We're looking into ways to tackle this in the future, perhaps by taking advantage of the Full Host Profiler.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

N/A

## References

AGTMETRICS-393

### DIFF
--- a/test/smp/regression/adp/cases/dsd_uds_100mb_3k_contexts_throughput/experiment.yaml
+++ b/test/smp/regression/adp/cases/dsd_uds_100mb_3k_contexts_throughput/experiment.yaml
@@ -3,7 +3,7 @@ erratic: false
 
 target:
   name: agent-data-plane
-  command: /usr/local/bin/agent-data-plane --config /etc/agent-data-plane/empty.yaml run
+  command: /maybe-profile.sh /usr/local/bin/agent-data-plane --config /etc/agent-data-plane/empty.yaml run
   cpu_allotment: 4
   memory_allotment: 2GiB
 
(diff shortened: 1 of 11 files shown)


test/smp/regression/adp/cases/dsd_uds_1mb_3k_contexts_throughput/experiment.yaml
```diff
@@ -34,6 +34,12 @@ target:
     DD_TELEMETRY_ENABLED: "true"
     DD_PROMETHEUS_LISTEN_ADDR: tcp://127.0.0.1:5102
 
+  profiling_environment:
+    SMP_PROFILING_ENABLED: "true"
+    DD_TRACE_AGENT_URL: "unix:///var/run/datadog/apm.socket"
+    DD_PROFILING_NATIVE_PRESET: "cpu_live_heap"
+    DD_PROFILING_INLINED_FUNCTIONS: "true"
```
GeorgeHahn: Profiles are showing up named [`myservice`](https://app.datadoghq.com/profiling/explorer?query=env%3Asingle-machine-performance%20job_id%3Ae422040e-774c-46de-a405-3f6b22f61821%20experiment%3Adsd_uds_1mb_3k_contexts_throughput&agg_m=%40prof_native_cpu_cores&agg_m_source=base&agg_t=sum&extra_search_fields=%7B%22filters_query%22%3A%22%22%2C%22sample_type%22%3A%22cpu-time%22%7D&fromUser=true&selected_tf=1765570752668.0667%2C1765573261673.1216&viz=flame_graph&start=1765491577295&end=1765577977295&paused=true); setting `DD_SERVICE` will fix that.

rayz (approved):

PR #1074, opened 2025-12-19
enhancement(core): add new "decoder" component and add named outputs to relays

tobz:
## Summary

This PR adds support for a new "decoder" component, and updates the relay component to allow for having named outputs.

As part of our continued OTLP work, we've now reached a point where to effectively add _incremental_ OTLP support, we need to be able to conditionally handle specific OTLP signals: if we can, for example, only handle metrics natively... then we can take these payloads and process them but proxy the rest. However, our two "ingest" components for OTLP -- OTLP Relay and OTLP Source -- conflict with each other: we can't run them both at the same time. This means that to do something like the above, we'd have to make sources be able to emit both events _and/or_ payloads, or we'd have to have a way to get all the raw payloads and conditionally choose to decode some of them while proxying the rest. Given this PR, we've obviously chosen to go with the latter. :)

Firstly, we've introduced a new "decoder" component. This is meant to be the ingest-side equivalent of encoders, and pairs specifically with relays. Similar to encoders and forwarders, using a relay and decoder is meant for situations where we need more control/flexibility than what is possible with just a source alone. There's not much more to say about decoders here... I think the code is simple enough that it speaks for itself.

The second thing we've done, which is the slightly bigger chunk of this PR, is add the ability for relays to have named outputs. This is necessary, going back to what was said above, to be able to conditionally take some payloads and send them to a decoder, and take _other_ payloads and send them directly to a forwarder. This necessitated a number of changes:

- `OutputDefinition` now has a generic parameter that is meant to carry either `EventType` or `PayloadType`, depending on the component type, in order to support the fact that named outputs aren't exclusively for events anymore
- due to the overall change, we had to also rework the OTLP Relay, OTLP Source, and Trace Agent Forwarder components at the same time

We've also reworked some of the prior hard-coded logic around the OTLP Relay and Trace Agent Forwarder components:

- OTLP Relay now exclusively emits gRPC payloads without any endpoint information: all that we include is the gRPC service path for differentiating what type of payload it is
- Trace Agent Forwarder now specifically checks that the payload is an OTLP traces export request

We've made a number of other small changes, mostly stylistic as things got refactored and/or moved around. I tried to stop myself from going too wild with stylistic changes, though. 😅 

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Existing unit/integration/correctness tests.

## References

AGTMETRICS-393

### DIFF
--- a/bin/agent-data-plane/src/components/remapper/mod.rs
+++ b/bin/agent-data-plane/src/components/remapper/mod.rs
@@ -43,8 +43,8 @@ impl TransformBuilder for AgentTelemetryRemapperConfiguration {
         EventType::Metric
     }
 
-    fn outputs(&self) -> &[OutputDefinition] {
-        static OUTPUTS: &[OutputDefinition] = &[OutputDefinition::default_output(EventType::Metric)];
+    fn outputs(&self) -> &[OutputDefinition<EventType>] {
+        static OUTPUTS: &[OutputDefinition<EventType>] = &[OutputDefinition::default_output(EventType::Metric)];
         OUTPUTS
     }
 
--- a/lib/saluki-common/src/buf/chunked.rs
+++ b/lib/saluki-common/src/buf/chunked.rs
@@ -114,6 +114,21 @@ impl FrozenChunkedBytesBuffer {
     pub fn len(&self) -> usize {
         self.chunks.iter().map(|chunk| chunk.len()).sum()
     }
+
+    /// Consumes the frozen buffer and returns a single `Bytes` value representing all chunks.
+    ///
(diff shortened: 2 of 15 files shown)


tobz:
* **#1074** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1074?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1074?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1075, opened 2025-12-19
enhancement(correctness): add support for checking APM stats in "trace" analysis mode

tobz:
## Summary

This PR adds support for validating the correctness of APM statistics when running the "traces" analysis mode.

There's a lot here, but it's not necessarily _complex_, so just listing out what I did in rough order:

- updated `datadog-protos` to add `serde` derives/annotations to the _generated_ structs that we use, in order to support deserializing MessagePack-based versions of the APM stats payloads sent by the Agent
- updated the Datadog Traces Encoder as we had to switch away from using the `bytes` feature of the `protobuf` crate to allow for implementing `serde` on the types (this is actually a no-op, performance-wise, since  we always had to allocate for converting to `Bytes`/`Chars` for bytes/string fields)
- updated the log format for `millstone`/`datadog-intake` since it was too noisy
- added support to `stele` to represent APM stats: recreating the nested structure and handling merging of the statistics, all in a way that they can be checked for naive equality in the actual analysis, similar to spans
- updated `datadog-intake` to capture these stats and expose an endpoint for dumping them
- updated the "traces" analysis mode to actually compare the statistics :)

I made some other ancillary changes -- reducing log format verbosity, etc -- during the course of getting this work, and it made things nicer so I've left those changes in.

There's still some work to do on the correctness logic. A non-exhaustive list:

- we naively compare peer tags (we need to add support for passing in which peer tags we care about, which then filters the peer tags that get used for generating the aggregation key)
- we don't handle merging the "ok"/"error" sketches together
- we don't do anything to abstract away the time component of the stats buckets: since all spans will have a timestamp when ingested, the buckets _should_ always be the same... but if we hit logic that caused timestamps to be clamped or updated to the wall clock time, this assumption would break
- we compare the client statistics naively via `PartialEq`, but we should use the structural diff approach that we use for spans so that it's easier to highlight what any actual differences are

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Ran the `otlp-traces` correctness test locally, and validated that stats are being collected, dumped, and analyzer. Ensured that multiple consecutive runs passed without issue.

## References

AGTMETRICS-393

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ ifeq ($(CI),true)
 endif
 export CARGO_TOOL_VERSION_cargo-binstall ?= 1.16.2
 export CARGO_TOOL_VERSION_dd-rust-license-tool ?= 1.0.3
-export CARGO_TOOL_VERSION_cargo-deny ?= 0.18.3
+export CARGO_TOOL_VERSION_cargo-deny ?= 0.18.9
 export CARGO_TOOL_VERSION_cargo-hack ?= 0.6.30
 export CARGO_TOOL_VERSION_cargo-nextest ?= 0.9.99
 export CARGO_TOOL_VERSION_cargo-autoinherit ?= 0.1.5


bin/correctness/stele/src/traces/stats.rs
```diff
@@ -0,0 +1,370 @@
...
+impl FromStr for BucketTimeframe {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let parts: Vec<&str> = s.trim_matches(['[', ']']).split(',').collect();
+        if parts.len() != 2 {
+            return Err(format!("expected two elements, found {}", parts.len()));
+        }
+        let start_time_ns = parts[0].parse::<u64>().map_err(|e| e.to_string())?;
+        let duration_ns = parts[1].parse::<u64>().map_err(|e| e.to_string())?;
```
rayz: `parts[1]` includes the starts time

tobz: Oooh, good call. 👍🏻

PR #1080, opened 2025-12-26
[AGTMETRICS-439] enhancement(components): add otlp decoder

rayz:
## Summary

This PR creates the OTLP Decoder.
The topology is updated using the new named outputs for the `Relay` such that metrics and logs will be proxied while traces will go to the OTLP Decoder.
When `data_plane.otlp.proxy.traces.enabled` is true, traces will be proxied to the trace agent on port 5003
```
data_plane:
  enabled: true
  otlp:
    enabled: true
    proxy:
      enabled: true
      receiver:
        protocols:
          grpc:
            endpoint: 127.0.0.1:4319
          http:
            endpoint: 127.0.0.1:4320
      traces: 
        enabled: false 
      metrics:
        enabled: true
      logs:
        enabled: true
      
```

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Run ADP and enable OTLP + proxy:
```
data_plane:
  enabled: true
  otlp:
    enabled: true
    proxy:
      enabled: true
      receiver:
        protocols:
          grpc:
            endpoint: 127.0.0.1:4319
          http:
            endpoint: 127.0.0.1:4320
      traces: 
        enabled: false 
      metrics:
        enabled: true
      logs:
        enabled: true
```

Send OTLP traces,metrics,logs and see that they show up in datadog.

### DIFF
--- a/lib/saluki-components/src/common/otlp/config.rs
+++ b/lib/saluki-components/src/common/otlp/config.rs
@@ -200,6 +200,17 @@ pub struct TracesConfig {
     /// Corresponds to `otlp_config.traces.probabilistic_sampler` in the Agent.
     #[serde(default)]
     pub probabilistic_sampler: ProbabilisticSampler,
+
+    /// The internal port on the Core Agent to forward traces to.
+    ///
+    /// Defaults to 5003.
+    #[serde(default = "default_internal_port")]
+    #[allow(unused)]
+    pub internal_port: u16,
+}
+
+const fn default_internal_port() -> u16 {
+    5003
 }
 
 /// Configuration for OTLP traces probabilistic sampling.
@@ -243,6 +254,7 @@ impl Default for TracesConfig {
             ignore_missing_datadog_fields: false,
             enable_otlp_compute_top_level_by_span_kind: default_enable_otlp_compute_top_level_by_span_kind(),
             probabilistic_sampler: ProbabilisticSampler::default(),
+            internal_port: default_internal_port(),
         }
     }
 }


lib/saluki-components/src/decoders/otlp/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+                        None => {
+                            debug!("Payloads stream closed, shutting down decoder.");
+                            break;
+                        }
+                    };
+
+                    let grpc_payload = match payload.try_into_grpc_payload() {
+                        Some(grpc) => grpc,
+                        None => {
+                            error!("Received non-gRPC payload in OTLP decoder.");
```
tobz: This could be a warning, I'd say. It's not quite an error since we're able to effectively filter out non-gRPC payloads, but we still want to have something higher than info so that we can be alerted to this mismatch.

We should also update the message to indicate that the payload is being dropped, like so:

`Received non-gRPC payload in OTLP decoder. Dropping.`

(Not that we should do it here, but it'd be good to turn this into a metric as well, like `components_receive_errors_total` or something. Doing it by hand would be the simplest way, but maybe there's some nice API we could add add to `Consumer` based on the generic type, like payload-specific `next_*` methods when the type is `Consumer<Payload>`, and `Consumer` would be the one to handle incrementing the relevant error metrics, etc.)

lib/saluki-components/src/decoders/otlp/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+                            if let Some(event_buffer) = event_buffer_manager.try_push(trace_event) {
+                                if let Err(e) = context.dispatcher().dispatch(event_buffer).await {
+                                    error!(error = %e, "Failed to dispatch trace events.");
+                                }
+                            }
+                        }
+                    }
+                }
+                path if path == &*OTLP_METRICS_GRPC_SERVICE_PATH => {
+                    error!("OTLP metrics decoding not yet implemented.");
```
tobz: Same thing here about saying that we're dropping the payload:

`OTLP metrics decoding not yet implemented. Dropping metrics payload.`

lib/saluki-components/src/relays/otlp/mod.rs
```diff
@@ -144,9 +145,10 @@ impl Relay for OtlpRelay {
                     break;
                 },
                 Some(otlp_payload) = payload_rx.recv() => {
+                    let output_name = otlp_payload.signal_type.to_string();
```
tobz: Can you make `OtlpSignalType` have an `as_str` method so we can return a `&'static str` and avoid an allocation every time we dispatch?

lib/saluki-components/src/decoders/otlp/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+
+                    let grpc_payload = match payload.try_into_grpc_payload() {
+                        Some(grpc) => grpc,
+                        None => {
+                            error!("Received non-gRPC payload in OTLP decoder.");
+                            continue;
+                        }
+                    };
+
+                    match grpc_payload.service_path() {
```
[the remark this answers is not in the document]
rayz: Oops yea.

lib/saluki-components/src/decoders/otlp/mod.rs
```diff
@@ -110,42 +110,42 @@ impl Decoder for OtlpTracesDecoder {
...
-                    };
-
-                    for resource_spans in request.resource_spans {
-                        let trace_events = self.translator.translate_resource_spans(resource_spans, &self.metrics);
-                        for trace_event in trace_events {
-                            if let Some(event_buffer) = event_buffer_manager.try_push(trace_event) {
-                                if let Err(e) = context.dispatcher().dispatch(event_buffer).await {
-                                    error!(error = %e, "Failed to dispatch trace events.");
+                        path if path == &*OTLP_TRACES_GRPC_SERVICE_PATH => {
+                            let body = grpc_payload.body().clone();
```
tobz: We should be destructuring the payload by calling `GrpcPayload::into_parts` and then we don't have to clone anything.

bin/agent-data-plane/src/config.rs
```diff
@@ -198,23 +199,74 @@ pub struct DataPlaneOtlpProxyConfiguration {
...
+            .unwrap_or(true);
+        let proxy_metrics = config
+            .try_get_typed("data_plane.otlp.proxy.metrics.enabled")?
+            .unwrap_or(true);
+        let proxy_logs = config
+            .try_get_typed("data_plane.otlp.proxy.logs.enabled")?
+            .unwrap_or(true);
+
+        if enabled {
+            tracing::info!(
```
tobz: We should be importing `info!`.

bin/agent-data-plane/src/config.rs
```diff
@@ -198,23 +199,74 @@ pub struct DataPlaneOtlpProxyConfiguration {
...
+                proxy_enabled = enabled,
+                core_agent_otlp_grpc_endpoint = %core_agent_otlp_grpc_endpoint,
+                core_agent_otlp_http_endpoint = %core_agent_otlp_http_endpoint,
+                proxy_traces = proxy_traces,
+                proxy_metrics = proxy_metrics,
+                proxy_logs = proxy_logs,
+                "OTLP proxy mode configuration loaded"
```
tobz:
```suggestion
                "OTLP proxy mode enabled. Select OTLP payloads will be proxied to the Core Agent."
```

tobz (changes requested):

lib/saluki-components/src/forwarders/otlp/mod.rs
```diff
@@ -0,0 +1,336 @@
...
+}
+
+#[async_trait]
+impl ForwarderBuilder for OtlpForwarderConfiguration {
+    fn input_payload_type(&self) -> PayloadType {
+        PayloadType::Http | PayloadType::Grpc
+    }
+
+    async fn build(&self, context: ComponentContext) -> Result<Box<dyn Forwarder + Send>, GenericError> {
+        let trace_agent_endpoint = format!("http://localhost:{}", 5003);
```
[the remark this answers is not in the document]
rayz: Hmm, updated it to be `let core_agent_traces_internal_port = config
            .try_get_typed("otlp_config.traces.internal_port")?
            .unwrap_or(5003);` since that was easier than making `OtlpForwarderConfiguration` derive serde Deserialize and pulling in the entire `OtlpConfig` for this one field.

tobz (approved):

PR #1135, opened 2026-01-29
chore(agent-data-plane): pass session ID from RAR when subscribing to `configstream` updates

rahulkaukuntla:
## Summary

This PR will help Datadog Agent PR [#45147](https://github.com/DataDog/datadog-agent/pull/45147), which adds RAR-gated config streaming from the agent. Without these changes, ADP doesn't send the RAR `session_id` to the core-agent via grpc metadata. The Agent is then unable to verify that ADP was registered with RAR, causing this error message to emit:

> 2026-01-29 09:57:48 EST | DATAPLANE | ERROR | (lib/saluki-env/src/configstream/stream.rs:69) | Error while reading config event stream: status: 'The request does not have valid authentication credentials', self: "session_id required in metadata: remote agent must register with RAR before subscribing to config stream".

This implementation enables synchronous RAR registration by calling `register_once()` early in run.rs before creating the config stream, storing the session ID in a shared SessionIdHandle. The session ID is passed directly to `create_config_stream()` as a parameter, eliminating the need for environment variable polling. The background refresh loop detects the existing session ID and only performs refreshes, avoiding double registration. 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

I ran an agent with the [#45147](https://github.com/DataDog/datadog-agent/pull/45147) changes, with RAR, ADP, and dogstatsd enabled. The config was streamed properly, as ADP (running with these changes) properly registered itself with RAR and sent the `session_id` in the grpc metadata.


bin/agent-data-plane/src/cli/run.rs
```diff
@@ -91,15 +128,8 @@ pub async fn handle_run_command(
...
 
-        info!("Waiting for initial configuration from Datadog Agent...");
-        dynamic_config.ready().await;
-        info!("Initial configuration received.");
-
-        // Reload our data plane configuration based on the dynamic configuration.
-        let dynamic_dp_config = DataPlaneConfiguration::from_configuration(&dynamic_config)
-            .error_context("Failed to load data plane configuration.")?;
-
-        (dynamic_config, dynamic_dp_config)
```
tobz: This can't be removed.

The reason it exists here to ensure we have the full configuration from the Core Agent before moving forward.

tobz (changes requested):

bin/agent-data-plane/src/cli/run.rs
rahulkaukuntla: Nice catch! Thanks for the context, I re-added these lines.

PR #1139, opened 2026-01-30
chore(otlp): use retain for single span sampling

rayz:
## Summary

This pr updates the trace sampler to filter SSS/analytic spans in place to avoid cloning.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Performance

### DIFF
--- a/lib/saluki-components/src/transforms/trace_sampler/mod.rs
+++ b/lib/saluki-components/src/transforms/trace_sampler/mod.rs
@@ -227,6 +227,26 @@ impl TraceSampler {
         false
     }
 
+    /// Apply analyzed span sampling to the trace.
+    ///
+    /// Returns `true` if the trace was modified.
+    fn analyzed_span_sampling(&self, trace: &mut Trace) -> bool {
+        let retained = trace.retain_spans(|span| span.metrics().contains_key(KEY_ANALYZED_SPANS));
+        if retained > 0 {
+            // Mark trace as kept with high priority
+            let sampling = TraceSampling::new(
+                false,
+                Some(PRIORITY_USER_KEEP),
+                None,
+                Some(MetaString::from(format!("{:.2}", self.sampling_rate))),
+            );
+            trace.set_sampling(Some(sampling));
+            true
+        } else {
+            false
+        }
+    }
+
     /// Returns `true` if the given trace has any analyzed spans.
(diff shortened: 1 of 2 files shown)


lib/saluki-core/src/data_model/event/trace/mod.rs
```diff
@@ -95,6 +95,25 @@ impl Trace {
...
+        }
+
+        let mut kept = Vec::new();
+        for span in self.spans.drain(..) {
+            if f(&span) {
+                kept.push(span);
+            }
+        }
+        kept.shrink_to_fit();
+        self.spans = kept;
```
tobz: Why not just use `Vec::retain`? Less code, avoids allocating a new vector.

tobz (changes requested):

lib/saluki-components/src/transforms/trace_sampler/mod.rs
```diff
@@ -261,10 +238,12 @@ impl TraceSampler {
...
-        let ss_spans = self.get_single_span_sampled_spans(trace);
-        if !ss_spans.is_empty() {
-            // Span sampling has kept some spans -> update the trace
-            trace.set_spans(ss_spans);
+        let has_sss = trace
+            .spans()
+            .iter()
+            .any(|span| span.metrics().contains_key(KEY_SPAN_SAMPLING_MECHANISM));
+        if has_sss {
+            trace.retain_spans(|span| span.metrics().contains_key(KEY_SPAN_SAMPLING_MECHANISM));
```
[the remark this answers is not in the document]
rayz: I think we still need the pre-check to avoid dropping the spans when there are no matches.

lib/saluki-core/src/data_model/event/trace/mod.rs
```diff
@@ -95,6 +95,34 @@ impl Trace {
...
+            return 0;
+        }
+
+        let mut has_match = false;
+        for span in self.spans.iter() {
+            if f(span) {
+                has_match = true;
+                break;
+            }
+        }
```
tobz: This is changing the behavior compared to before... where now we only retain matching spans so long as there's at least _one_ matching span.

That's fine, but just to make sure: is this new behavior the actual intended behavior?

PR #1145, opened 2026-02-01
enhancement(core): add support for spawning supervisors on a dedicated runtime

tobz:
## Summary

This PR adds support to the runtime system to configure/spawn supervisors on a dedicated asynchronous runtime.

As part of the goal to supervisor-ify the various tasks that compromise the control plane and data plane functionality of ADP, a major consideration is that we spawn various tasks on dedicated runtimes: we have a dedicated control plane runtime, internal observability runtime, the main multi-threaded runtime where the topology runs, and then the dedicated multi-threaded runtime we use for compute-heavy tasks. `Supervisor` was designed with nesting in mind, but everything assumed that we'd spawn tasks on the exact same runtime across the entire supervision tree... which doesn't align with our needs in practice.

This PR adds support for customizing a `Supervisor` to specify what runtime it should execute on: the ambient runtime (wherever its run from at that moment), or a dedicated runtime. When using a dedicated runtime, we can configure whether or not a single-threaded or multi-threaded runtime is created behind the scenes. A dedicated background OS thread is spawned in either case to drive the supervisor to completion, with the difference being that for single-threaded runtimes, all work happens on that thread, while multi-threaded runtimes have a pool of dedicated worker threads where all meaningful task execution occurs.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Existing unit tests.

## References

AGTMETRICS-393

### DIFF
--- a/lib/saluki-core/src/runtime/dedicated.rs
+++ b/lib/saluki-core/src/runtime/dedicated.rs
@@ -127,7 +127,7 @@ impl Future for DedicatedRuntimeHandle {
 
 /// Spawns a supervisor in a dedicated runtime on a new OS thread, returning a handle to await for the result.
 ///
-/// The returned handle will resolve if initialization of the of the runtime failed, or after the supervisor completes,
+/// The returned handle will resolve if initialization of the runtime failed, or after the supervisor completes,
 /// whether due to shutdown or failure.
 ///
 /// A background OS thread is spawned to run the supervisor, named after the supervisor's ID: `<id>-sup-rt`. For


tobz:
* **#1145** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1145?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1145?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1150, opened 2026-02-02
chore(otlp): use `StringBuilder` in OTLP trace translation 

rayz:
## Summary

This pr adds `StringBuilder` to the translator to be used whenever we want to incrementally build a string in the OTLP traces translator logic.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Existing tests/ci


lib/saluki-components/src/common/otlp/traces/translator.rs
```diff
@@ -33,15 +33,16 @@ pub fn convert_span_id(span_id: &[u8]) -> u64 {
 
 pub fn resource_attributes_to_tagset(attributes: &[otlp_common::KeyValue], interner: &GenericMapInterner) -> TagSet {
     let mut tags = TagSet::with_capacity(attributes.len());
+    let mut string_builder = StringBuilder::new();
```
tobz: You can end up reusing a single, long-lived `StringBuilder` if you pass around `&mut StringBuilder` instead of just a reference to the interner. `StringBuilder` allows embedding a reference to the interner so that when the `I` in `StringBuilder<I>` is `I: Interner`, you get `StringBuilder::try_intern`. We could add a new method to that impl block that wraps `MetaString::from_interner` to complete the refactor.

tobz (changes requested):

tobz (approved):

PR #1154, opened 2026-02-03
chore(agent-data-plane): rework remote agent logic to use 'bootstrap' regime

tobz:
## Summary

This PR updates ADP's remote agent logic to follow the "bootstrap" regime that we have in place for going from the baseline configuration available to us all the way to communicating with the Core Agent to receive our final configuration.

After #1135, we were left in the position where in order to receive configstream updates, we needed to register as a remote agent with RAR... but we also had to initialize other subsystems (like the control plane) to actually expose the necessary services for RAR, and we did somewhat of a song-and-dance to get all the right pieces in place, pre-registering and then manually passing around session IDs and so on. This was suboptimal for a few reasons:

- the session ID is an internal implementation detail that shouldn't leak to the caller
- having to know that we already registered and that we shouldn't register again is an internal implementation detail that shouldn't leak to the caller, and is _also_ a footgun

This PR attempts to remedy the situation by applying the same "bootstrap" regime to remote agent bring-up as we did in `saluki-app` for general app bootstrapping. We now try to hide all of the relevant orchestration of the remote agent registration -- when to do it, how to get the session ID and pass it around, etc -- in a new type, `RemoteAgentBootstrap`.

The new flow looks something like:

- we determine whether or not we need to register as a remote agent (remote agent enabled, or "use config stream" enabled)
- if so, we create `RemoteAgentBootstrap` which:
  - immediately spawns the registration task (register/refresh/reregister as needed, updates session ID)
  - waits for the first registration call to complete, whether it succeeds or fails
  - returns an error if initial registration fails, otherwise returns itself
- if we created a remote agent bootstrap object, we proceed with normal dynamic config loading:
  - create a config stream and attach it to our configuration
  - wait until the config is ready
  - reload data plane configuration based on new dynamic configuration
- pass our remote agent bootstrap object to control plane initialization to be used to attach gRPC services related to RAR

Essentially, we treat `RemoteAgentBootstrap` as the one and only thing that matters. We use it to attempt our initial registration as a remote agent, failing quickly if that initial registration fails. We continue using the same object for getting any necessary bits, like config stream or RAR gRPC services, which encapsulates all of the implementation details like the dynamically updated session ID, and so on. This ensures that the caller only needs focus on the functional aspect: make sure we're registered before moving on with configuration loading.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Existing unit and integration tests.

## References

AGTMETRICS-393

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -497,9 +497,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.11.0"
+version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b35204fbdc0b3f4446b89fc1ac2cf84a8a68971995d0bf2e925ec7cd960f9cb3"
+checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 
 [[package]]
 name = "bytesize"


tobz:
* **#1154** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1154?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1154?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1164, opened 2026-02-04
enhancement(otlp source): add SIMD optimization for normalizing tag values

aqian01:
## Summary

This PR adds SIMD optimization to for normalizing tags and tag values (inspired by this [comment](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/traceutil/normalize/normalize.go#L382)). 

This PR optimizes the byte comparison by comparing 16 bytes at once instead of 1 at a time for certain cpu architectures (`x86`, `x86_64`, `aarch64`) that have registers enabling parallel operations. 

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Added bench marks and had codex generate "realistic" tag values to test with. I also had it run the bench marks 15 times to get the average of 6.1038% faster cpu time. Planning to remove all the bench mark code if we decide to merge this change.

The bench marks compare the "simd" algorithm to the "previous algorithm" that validated a tag. 

The [SMP](https://app.datadoghq.com/dashboard/4br-nxz-khi?fromUser=false&fullscreen_end_ts=1770666794000&fullscreen_paused=true&fullscreen_refresh_mode=paused&fullscreen_section=overview&fullscreen_start_ts=1770666194000&fullscreen_widget=4615193821777026&refresh_mode=paused&tpl_var_adp-run-id%5B0%5D=b07fe56e-e9c2-46ac-b455-2bb079a41a60&tpl_var_experiment%5B0%5D=otlp_ingest_traces_5mb_memory&view=spans&from_ts=1770666194000&to_ts=1770666794000&live=false) results show no improvements though, when I was chatting with some SMP people, I was advised to use the Rust bench marks to test performance improvements for optimizing a function.

### DIFF
--- a/test/integration/cases/adp-disabled-exit/config.yaml
+++ b/test/integration/cases/adp-disabled-exit/config.yaml
@@ -7,7 +7,6 @@ container:
   env:
     DD_API_KEY: "test-api-key"
     DD_HOSTNAME: "integration-test"
-    DD_LOG_LEVEL: "info"
     DD_DATA_PLANE_ENABLED: "false"
     DD_DATA_PLANE_STANDALONE_MODE: "false"
 


tobz (changes requested):
Overall, I like the idea here and think we should push it forward, but I have some structuring changes I want to see.

Basically, I think we should turn `normalize.rs` into its own directory (`normalize`), where we move `normalize.rs` into `normalize/mod.rs`, and then under that directory, add instruction set-specific files: `normalize_sse2.rs` and `normalize_neon.rs`. The respective implementations would live in those files, and in `normalize/mod.rs`, we would conditionally import those modules with the relevant target architecture flags.

Beyond that, I'm not sold on all of the bench-specific code/variants that we have. It seems like we should have the single exposed function that has the multiple implementations to select the right instruction set-specific backend, rather than exposing them individually.. and then update the benchmarks to use that.

PR #1165, opened 2026-02-05
Go like Scheduler, NativeCheck, http-check

fabbing:
## Summary
This PR bundles multiples changes around check execution:
- Add a reimplementation of the Go core-agent scheduler
  - The behavior is different in `send_to_worker()`, we would need `sync:mpmc` to exactly match the Go one. In the meantime, a pattern similar to what is done in work-stealing queue is used.
  - Need unit tests
- Add the `lib/core-check` library
  - Contain an interface for implementing Rust based checks, uncoupled from Saluki internals
  - Allow to also run these checks in the Go core-agent thanks to the `Sink` trait
  - Will be modified to implemented the Slim Check interface
- Add the `NativeCheck` type to run Rust based check, compiled in Saluki
- Modify the check trait
  - To allow use of `async` code in `run()`
  - Use interior mutability to access immutable fields via const reference (plays nicer with the scheduler)
- Add a reimplementation of the Python `http_chunk`, not exhaustive
- Add a feature flags for `checks` and `python-checks`
  - Needs to be threaded in the Makefile

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

### DIFF
--- a/lib/saluki-components/src/sources/checks/scheduler/job.rs
+++ b/lib/saluki-components/src/sources/checks/scheduler/job.rs
@@ -114,7 +114,7 @@ impl JobQueue {
                 }
 
                 _ = interval.tick() => {
-                    let elasped = Instant::now().duration_since(last_tick);
+                    let elasped = last_tick.elapsed();
                     if elasped > Duration::from_secs(2) {
                         warn!(
                             elapsed_secs = elasped.as_secs(),


lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -215,39 +222,38 @@ impl Source for ChecksSource {
...
                                 AutodiscoveryEvent::CheckUnscheduled { config } => {
                                     for instance in &config.instances {
                                         let check_id = instance.id();
-                                        if !check_ids.contains(check_id) {
-                                            warn!("Unscheduling check {} not found, skipping.", check_id);
-                                            continue;
-                                        }
-
-                                        scheduler.unschedule(check_id);
+                                        scheduler.clone().unschedule(check_id).await;
```
tobz: Why are we cloning here?

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -197,12 +203,13 @@ impl Source for ChecksSource {
         );
 
         let mut autodiscovery_rx = self.autodiscovery_rx;
+        runner.clone().run().await;
```
tobz: Why are we cloning here?

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -215,39 +222,38 @@ impl Source for ChecksSource {
                                     let mut runnable_checks: Vec<Arc<dyn Check + Send + Sync>> = vec![];
                                     for instance in &config.instances {
                                         let check_id = instance.id();
-                                        if check_ids.contains(check_id) {
+                                        if scheduler.clone().is_check_scheduled(check_id) {
```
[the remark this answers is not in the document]
fabbing: Good catch! Fixed in https://github.com/DataDog/saluki/pull/1165/commits/10e002d73942c58a1ab5698b37d73d18a787581c

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -215,39 +222,38 @@ impl Source for ChecksSource {
...
                                                 runnable_checks.push(check);
-                                                check_ids.insert(check_id.clone());
                                                 break;
                                             }
                                         }
                                     }
 
                                     for check in runnable_checks {
-                                        scheduler.schedule(check);
+                                        scheduler.clone().schedule(check).await;
```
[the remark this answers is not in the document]
fabbing: This is one is needed, because the `Scheduler` passes an `Arc<Scheduler>` to the `JobQueue` so that they can later schedule the firing checks.

lib/saluki-components/src/sources/checks/mod.rs
fabbing: Fixed in https://github.com/DataDog/saluki/commit/10e002d73942c58a1ab5698b37d73d18a787581c!

lib/saluki-components/src/sources/checks/mod.rs
```diff
@@ -197,12 +203,13 @@ impl Source for ChecksSource {
...
 
         let mut autodiscovery_rx = self.autodiscovery_rx;
+        runner.clone().run().await;
 
         loop {
             select! {
                 _ = &mut global_shutdown => {
                     debug!("Checks source received shutdown signal.");
-                    scheduler.shutdown().await;
+                    scheduler.clone().shutdown().await;
```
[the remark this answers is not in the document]
fabbing: Fixed in https://github.com/DataDog/saluki/commit/10e002d73942c58a1ab5698b37d73d18a787581c!

lib/saluki-components/src/sources/checks/mod.rs
fabbing: I think I can't avoid it because run spawns a tokio task. Is it possible to avoid or would you to design it differently?

PR #1166, opened 2026-02-05
fix(agent-data-plane): allow check config files without `init_config`

Enzu83:
## Summary

This PR allows check configs that don't have `init_config` explicitly defined. Check configs like these are valid in the Agent, so should be in ADP to keep full compatibility between the Agent and ADP.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

New unit test parsing a config file without `init_config`


lib/saluki-env/src/autodiscovery/providers/local.rs
```diff
@@ -78,6 +78,7 @@ impl LocalAutodiscoveryProvider {
 
 #[derive(Debug, Deserialize)]
 struct LocalCheckConfig {
+    #[serde(default = "HashMap::new")]
```
tobz: Should just be able to do `#[serde(default)]` since `HashMap<K, V>` implements `Default`.

PR #1178, opened 2026-02-09
enhancement(core): add dataspaces for decoupled, cross-process state

tobz:
## Summary

This PR introduces support for _dataspaces_, a process-centric approach for eventually consistent sharing of data.

### Preface

Dataspaces are a novel approach to communicating between processes in an actor system. While pure message passing is simple and generally straightforward to understand, it falls down in a number of areas: of note, specificity (what part of the "conversation" a process cares about), discoverability (joining the conversation as a late comer), and context (knowing what has already been said prior to joining). In the [literature](https://syndicate-lang.org/papers/conversational-concurrency-201712310922.pdf), _dataspaces_ make up the foundation of what the author calls "conversational concurrency." In Saluki, we very much intend for it to provide much of the same thing: message passing without message passing.

In `saluki-core`, we have a burgeoning runtime system that, if you squint, looks a lot like Erlang/OTP, but importantly: we are not Erlang. Rust does not, and never will, have the requisite parts -- language _and_ VM -- to elegantly provide an actor system, including message passing. Attempts to build actor systems in Rust generally stumble (IMO) when it comes to the message passing/coordination bit because it looks weird, and requires a lot of contortions to get around the fact that the language and runtime are just not designed for message passing actors. We still have a need for processes to coordinate with each other, though.

### What are dataspaces?

_Dataspaces_ are a way to meet our message passing-style needs in the runtime system. In simple terms, dataspaces represent a collection of tuples -- owner, identifier, data -- that are pushed into the dataspace by processes and then queried by other processes. There are a number of specific terms, features, and constraints of dataspaces that make them particularly powerful:

- all data in the dataspace is strongly typed (multiple types are supported)
- processes **assert** (add) data into the dataspace with an associated **identifier** (string or integer)
- only the process that owns an assertion can **retract** (remove) that assertion
- assertions can be overwritten, but again, only by the owning process
- assertions are automatically retracted when the process dies (either gracefully or non-gracefully)
- processes subscribe to assertions/retractions in a pub/sub-style fashion
- subscriptions can "filter" for specific assertions by matching a single identifier, or matching identifiers in a glob-style approach ("give me assertion `1`", or "give me assertions `api-routes-*`", etc)

### A real example: dynamic API routes

This has all been a bit theoretical, so let's ground dataspaces in a concrete use case: _dynamic API routes_. This is actually something that upstack PRs add support for, but let's chat about it briefly here.

ADP has API endpoints which expose routes tied to various subsystems: memory bounds/allocation tracking, health status, DSD stats tracking, and more. These API routes need to be registered when we're setting up the process very early on, and have no way to be changed later on. This means we have to often expose specific bits of data from subsystems deep in the program just to be able to properly attach their API routes, leading to a more fragile design than necessary. This doesn't change even if we reverse the flow from collecting all API routes and adding them at process start to pushing the API route add logic down to all the places that care about it. Additionally, we would also need to add in all of the synchronization for handling concurrent calls to add/remove API routes, as well as consider how to handle scenarios where a component that added API routes has gone away and is no longer able to handle those API calls.

Dataspaces allow us to elegantly solve this problem:

- we create a new shared type designed explicitly for defining dynamic API routes (this essentially becomes the API boundary)
- subsystem processes assert/retract their API routes as they come into and go out of existence
- API server processes subscribe to assertions/retractions of the new dynamic API route type, and update their route tables as those assertions/retractions come in

At no point do the API server processes or the subsystem processes have to know (or care) about where the other one lives, or how to get to them: they communicate through the dataspace, and all of the comings and goings are tracked there. They don't need to build their own signaling mechanisms for handling failure. Additionally, we now also get a new side benefit: if the API server processes die, and get restarted, they can subscribe and immediately get a snapshot of the current state of the world in terms of asserted API routes, and almost immediately recover.

Decoupled, eventually consistent, simple.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

A slew of new unit tests.

## References

AGTMETRICS-400


webern (approved):
Neat 😎

PR #1207, opened 2026-02-19
chore(ci): speed up integration tests by nearly 40%

tobz:
## Summary

This PR optimizes both how integration tests are executed as well as streamlines the assertions in our existing tests in order to significantly reduce the total time to run the integration test suite.

We've made two main changes: allow for parallel assertions, and update the test cases to use them while also reducing the duration/timeout of various assertions.

The first part is sort of obvious: we ran assertions sequentially, so if you have two assertions back to back that monitor a condition for 10 seconds, you spend 20 seconds total to get through them... even if they could have both run in parallel. We updated the test case configuration schema, and `panoramic` itself, to support defining a set of assertions in a `parallel` block. All assertions still run sequentially, but `parallel` blocks run their subassertions in parallel. This means we can have an assertion that ensures we'd reach steady state -- like the topology being healthy -- and then run other assertions in parallel afterwards, such as ensuring the process stays healthy _and_ doesn't emit any error logs, and so on.

The second part is a follow-on: we had to update our existing integration test cases to use the `parallel` block support, and we also tweaked many of the timeouts/duration of various assertions. We don't need to wait for 30 seconds of the process being stable after we've already asserted another condition that implies it's healthy... 10 or 15 seconds will suffice.

The result of these changes is that a full integration test suite run now takes roughly 40-45 seconds, where it would previously take between 65-70 seconds. This represents an average reduction of around 40%. 🚀 

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Ran the integration test suite locally and ensured it still ran correctly.

## References

AGTMETRICS-393

### DIFF
--- a/bin/correctness/panoramic/src/runner.rs
+++ b/bin/correctness/panoramic/src/runner.rs
@@ -375,8 +375,14 @@ impl TestRunner {
         if let Err(e) = self.write_logs(&test_name).await {
             warn!(test = %test_name, error = %e, "Failed to write container logs to disk.");
         }
+        debug!(test = %test_name, "Wrote container logs to disk.");
+        phase_timings.push(PhaseTiming {
+            phase: "write_logs".to_string(),
+            duration: phase_start.elapsed(),
+        });
 
         // Cleanup.
+        let phase_start = Instant::now();
         debug!(test = %test_name, "Cleaning up container and resources...");
         if let Err(e) = self.cleanup(&driver).await {
             warn!(test = %test_name, error = %e, "Failed to clean up resources.");


andrewqian2001datadog (approved):
NIT: 

the `[logs_contains](https://github.com/DataDog/saluki/blob/tobz/speed-up-integration-tests-20260219/bin/correctness/panoramic/src/config.rs#L182)` config has no during field in the schema, however it exists [here](https://github.com/DataDog/saluki/blob/tobz/speed-up-integration-tests-20260219/test/integration/cases/adp-no-pipelines-exit/config.yaml#L20C1-L20C15)

Also I think this block of [code](https://github.com/DataDog/saluki/blob/tobz/speed-up-integration-tests-20260219/bin/correctness/panoramic/src/runner.rs#L373-L388) is a little suspicious? Is `write_logs` influencing the timer duration ` duration: phase_start.elapsed()`?

tobz:
> Also I think this block of [code](https://github.com/DataDog/saluki/blob/tobz/speed-up-integration-tests-20260219/bin/correctness/panoramic/src/runner.rs#L373-L388) is a little suspicious? Is write_logs influencing the timer duration  duration: phase_start.elapsed()?

It is influencing it, yes, although I don't think it's suspicious by itself? It's definitely a sort of post-run/cleanup activity, although it's not _strictly_ related to cleaning up the actual containers.

I'll split it out into a distinct phase.

PR #1218, opened 2026-02-26
chore(agent-data-plane): OTTL filtering correctness test

RiantZ:
## Summary
This PR adds correctness test for OTTL filtering processor component using DDOT as reference.
We also using special version of SMP lading to generate human-readable traffic instead of random data.
The PR also add modification to ground-truth to add option to exclude APM statistics from comparison and as well add option to configure ignored fields using yaml config file.

**N.B.:** _Special attention of reviewers_: I did local modification of Saluki [transform](https://github.com/DataDog/saluki/blob/30a68fdab3faa0a6c0a4629e50f941867bbce519/lib/saluki-components/src/common/otlp/traces/transform.rs#L683) (switch off the normalization) - it was required to pass the test with new lading (realistic data), otherwise correctness test for traces were failed: resources were not matching GET vs get, etc.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?
By changing OTTL filtering predicate for DDOT or ADP to check that correctness test is failed
By monitored number of filtered out spans

[OTAGENT-836](https://datadoghq.atlassian.net/browse/OTAGENT-836)

[OTAGENT-836]: https://datadoghq.atlassian.net/browse/OTAGENT-836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitlab/e2e.yml
+++ b/.gitlab/e2e.yml
@@ -18,11 +18,17 @@
     GROUND_TRUTH_ALPINE_IMAGE: registry.ddbuild.io/alpine:latest
     GROUND_TRUTH_MILLSTONE__IMAGE: ${SALUKI_IMAGE_REPO_BASE}/millstone:${CI_COMMIT_SHA}
     GROUND_TRUTH_DATADOG_INTAKE__IMAGE: ${SALUKI_IMAGE_REPO_BASE}/datadog-intake:${CI_COMMIT_SHA}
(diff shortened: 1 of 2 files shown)


test/correctness/otlp-traces/config.yaml
```diff
@@ -1,4 +1,6 @@
 analysis_mode: traces
+additional_span_ignore_fields:
+  - meta.error.msg
```
rayz: This should probably be fixed right? I'll open an [issue](https://github.com/DataDog/saluki/issues/1221)

PR #1224, opened 2026-03-05
enhancement(agent-data-plane): switch to `HttpClient` from `saluki-io` for data plane API client to reduce binary size

tobz:
## Summary

This PR switches from `reqwest` to our own `HttpClient` in `agent-data-plane` in order to meaningfully reduce the ADP binary size by eliminating a ton of duplicate codegen due to monomorphization.

Quoting the [Rust Compiler Development Guide](https://rustc-dev-guide.rust-lang.org/backend/monomorph.html):

> Rust takes a different approach: it monomorphizes all generic types. This means that compiler stamps out a different copy of the code of a generic function for each concrete type needed. For example, if I use a `Vec<u64>` and a `Vec<String>` in my code, then the generated binary will have two copies of the generated code for `Vec`: one for `Vec<u64>` and another for `Vec<String>`. The result is fast programs, but it comes at the cost of compile time (creating all those copies can take a while) and binary size (all those copies might take a lot of space).

The bulk of monomorphization cost in ADP, outside of standard library types, is related to the `hyper` stack, and by extension, things that depend on `hyper`: `tonic`, `reqwest`, etc. In particular, `reqwest` is very heavy in terms of binary size because of all the additional things it pulls in and how _they_ monomorphize.

This PR replaces `reqwest` with our own `HttpClient` in order to shrink the binary size by reducing the penalty we pay through monomorphization: while the dependency graph stays almost exactly the same, we drop so many additional generated copies of monomorphized code that we drastically shrink the binary, to the tune of ~1MiB, or roughly 5%.

We had to do some light refactoring to switch to `HttpClient`, as well as implementing some additional logic to match what `reqwest` provided, but this was minimal: support for UDS in `HttpClient`, and a no-verify certificate verifier to allow self-signed certificates.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

Existing tests.

## References

AGTMETRICS-400

### DIFF
--- a/bin/agent-data-plane/src/cli/utils.rs
+++ b/bin/agent-data-plane/src/cli/utils.rs
@@ -183,14 +183,11 @@ impl DataPlaneAPIClient {
     pub async fn config(&mut self) -> Result<String, GenericError> {
         let uri = self.build_uri("/config", None);
         let req = Request::get(uri).body(String::new()).expect("valid request");
-        let resp = self.client.send(req).and_then(process_response_body).await?;
-        if resp.status() == StatusCode::NOT_FOUND {
-            return Err(generic_error!(
-                "Workload provider not configured: no External Data available."
-            ));
-        }
-
-        body_when_success(resp)
+        self.client
+            .send(req)
+            .and_then(process_response_body)
+            .await
+            .and_then(body_when_success)
     }
 
     /// Retrieves the tags from the workload provider.
@@ -252,7 +249,7 @@ async fn process_response_body(response: Response<Incoming>) -> Result<Response<
 }
 
 fn body_when_success(resp: Response<String>) -> Result<String, GenericError> {
-    if !resp.status().is_success() {
+    if resp.status().is_success() {
         Ok(resp.into_body())
     } else {
         Err(generic_error!(
@@ -264,7 +261,7 @@ fn body_when_success(resp: Response<String>) -> Result<String, GenericError> {
 }
 
 fn empty_when_success(resp: Response<String>) -> Result<(), GenericError> {
-    if !resp.status().is_success() {
+    if resp.status().is_success() {
         Ok(())
     } else {
         Err(generic_error!(


lib/saluki-components/src/forwarders/otlp/mod.rs
```diff
@@ -64,7 +66,9 @@ impl ForwarderBuilder for OtlpForwarderConfiguration {
         let normalized_endpoint = normalize_endpoint(&self.core_agent_otlp_grpc_endpoint);
         let core_agent_grpc_channel = Channel::from_shared(normalized_endpoint)
             .error_context("Failed to construct gRPC channel due to an invalid endpoint.")?
+            .connect_timeout(Duration::from_secs(5))
```
tobz: This is a one-off change that is unrelated to the PR title, yes... but what it's doing here is matching the usage of another Tonic callsite in terms of the _generated_ code: since everything _inside_ of `Channel` is monomorphized, this code was previously something like `Foo<Bar<Blah<...>>>` and other usages were `TimeoutLayer<FooBar<Blah<...>>>`. This meant a ton of duplicated code generation.

Is this a fragile optimization? Yes. Nothing breaks if the other Tonic usages change the options they enable, though... the code size just goes back up slightly. 

This is also just a good change to make, since timeouts are good. :shrug:

bin/agent-data-plane/src/cli/utils.rs
```diff
@@ -221,25 +220,57 @@ impl DataPlaneAPIClient {
...
+    if !status.is_server_error() {
+        Ok(Response::from_parts(parts, body))
     } else {
-        let body = response.text().await.unwrap_or_else(|_| String::from("<no body>"));
         Err(generic_error!("Received non-success response ({}): {}.", status, body))
     }
 }
+
+fn body_when_success(resp: Response<String>) -> Result<String, GenericError> {
+    if !resp.status().is_success() {
```
rayz:
```suggestion
    if resp.status().is_success() {
```

tobz: Fixed in 58d384c1e4.

bin/agent-data-plane/src/cli/utils.rs
```diff
@@ -184,13 +180,17 @@ impl DataPlaneAPIClient {
...
-    pub async fn config(&self) -> Result<String, GenericError> {
-        let url = self.get_privileged_url("/config");
-        let response = self.privileged_api_client.get(url).send().await?;
+    pub async fn config(&mut self) -> Result<String, GenericError> {
+        let uri = self.build_uri("/config", None);
+        let req = Request::get(uri).body(String::new()).expect("valid request");
+        let resp = self.client.send(req).and_then(process_response_body).await?;
+        if resp.status() == StatusCode::NOT_FOUND {
+            return Err(generic_error!(
+                "Workload provider not configured: no External Data available."
```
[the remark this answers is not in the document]
tobz: Fixed in 58d384c1e4.

PR #1227, opened 2026-03-06
feat(agent-data-plane): Add OTTL transform processor component skeleton

RiantZ:
## Summary
* Add a new ottl_transform_processor component that executes OTTL transformation statements (e.g. set(attributes["key"], "value")) against trace spans, following the [OpenTelemetry transform processor spec](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/release/v0.144.x/processor/transformprocessor).
* Register the transform processor in the traces pipeline blueprint, chained after the existing OTTL filter processor and before APM onboarding.
* Rename the filter processor's YAML configuration section from ottl_config to ottl_filter_config to disambiguate it from the new ottl_transform_config section.
* Add documentation
* Unit-tests
* Performance tests

**What it does**:
The transform processor runs as a synchronous transform in the traces pipeline. For each span in each trace, it evaluates a list of OTTL statements in order. Currently supported:
* set editor function: set(attributes["key"], "value") — adds or overwrites a span attribute. Supports string, int, float, and bool literal values (converted to string in span meta). Setting to Nil (e.g. referencing a non-existent attribute) removes the key.
* where clauses: set(attributes["x"], "v") where attributes["env"] == "prod" — conditional execution based on span or resource attributes.
* resource.attributes read access: conditions can reference trace-level resource tags (read-only; SharedTagSet does not yet expose mutable access).
* error_mode: ignore (log and continue), silent (continue without logging), propagate (log and stop processing further statements for that span).

**Limitations**:
* Only the set editor function is implemented; other OTTL functions (e.g. delete_key, replace_match) are not yet supported.
* resource.attributes is read-only — writing to resource attributes is not possible because SharedTagSet does not expose mutable access.
* Only trace/span context is supported (no metrics, logs, or span events).

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

[OTAGENT-891](https://datadoghq.atlassian.net/browse/OTAGENT-891)

[OTAGENT-891]: https://datadoghq.atlassian.net/browse/OTAGENT-891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ export CARGO_BINSTALL_STRATEGIES ?= crate-meta-data,compile
 ifeq ($(CI),true)
 	override CARGO_BINSTALL_STRATEGIES = compile
 endif
-export CARGO_TOOL_VERSION_cargo-binstall ?= 1.16.7
+export CARGO_TOOL_VERSION_cargo-binstall ?= 1.17.7
 export CARGO_TOOL_VERSION_dd-rust-license-tool ?= 1.0.3
 export CARGO_TOOL_VERSION_cargo-deny ?= 0.18.9
 export CARGO_TOOL_VERSION_cargo-hack ?= 0.6.30


bin/agent-data-plane/src/components/ottl_transform_processor/mod.rs
```diff
@@ -0,0 +1,947 @@
...
+
+        let mut editors = CallbackMap::new();
+        editors.insert(
+            "set".to_string(),
+            Arc::new(|args: &mut dyn ottl::Args| {
+                let value = args.get(1)?;
+                args.set(0, &value)?;
+                Ok(Value::Nil)
+            }),
+        );
```
tobz: It doesn't have to be done here, but I suspect we may want to create a set of standard editors/converters/etc that live in `ottl` for basic functions like these.

This could let us do something like:

```rust
// Create a `CallbackMap` with editors like `set`, etc.
let mut editors = ottl::editors::standard();

// Add our own editors to `editors` here, if we want.
```

tobz (approved):
This looks good to me, given the TODOs we have in place for actually supporting mutability of resource tags.

PR #1257, opened 2026-03-24
enhancement(context): add support for efficient mutable tag sets

tobz:
## Summary

This PR refactors `TagSet` to support efficient mutation of tag sets.

Today, `TagSet` and `SharedTagSet` form the basis for creating tag sets and cheaply sharing them. In components and helpers where we're building tag sets, we start by creating a `TagSet`, which supports mutability, and once these tagsets are "done", we can freeze them to create a `SharedTagSet`, which is then able to be cheaply cloned/shared. We use `SharedTagSet` as the tag storage for all event types: metrics, logs, and traces. However, this means that events cannot have their tags mutated. Adding support for mutation of tags can be done easily with a naive implementation, but we don't want to lose the things that make `SharedTagSet` so nice, namely being able to cheaply clone and share those chunks of tags that are constructed early in the pipeline, which are potentially interned, and so on.

This PR introduces changes to `TagSet` that _effectively_ give it the ability to function like `SharedTagSet` -- the ability to cheaply extend it from another `SharedTagSet` -- while also efficiently allowing the "mutation" of the tag set by utilizing an "overlay" design: `TagSet` holds a "base" set of shared tag sets, a set of removals (which base tags to ignore), and a set of additions (new tags to add). These removals and additions are "overlaid" on the base tags during iteration, leading to the tags being emitted reflecting the intended state.

We've designed removal and addition to follow the existing semantics of `TagSet` but be as efficient as we can _reasonably_ be. As tag _removal_ is generally the more desired operation, we've tried to optimize this case a little more by making it efficient to do so: we use a bitset to store removal indexes so that determining which tags to remove during iteration is O(1).

In a sense, what we're really doing is preparing ourselves to be able to switch to using `TagSet` in events instead of `SharedTagSet`: `TagSet` today can already be mutated, but it has no support for structural sharing like `SharedTagSet` does.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Added a bevy of new unit tests for our changes, as well as property tests to ensure the optimized removal/addition logic functions as intended.

## References

AGTMETRICS-400

### DIFF
--- a/lib/saluki-context/src/tags/tagset/frozen.rs
+++ b/lib/saluki-context/src/tags/tagset/frozen.rs
@@ -1,7 +1,5 @@
 use std::fmt;
 
-use serde::{Deserialize, Serialize};
-
 use crate::tags::Tag;
 
 /// Flat, immutable tag storage.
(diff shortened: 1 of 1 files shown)


lib/saluki-context/src/tags/tagset/frozen.rs
```diff
@@ -0,0 +1,95 @@
...
+/// This is a private type used as the backing storage inside `SharedTagSet`. It holds a simple
+/// `Vec<Tag>` and provides read-only access. External callers work with `TagSet` (mutable) and
+/// `SharedTagSet` (shared/frozen) instead.
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[serde(transparent)]
+pub(crate) struct FrozenTagSet(Vec<Tag>);
+
+impl FrozenTagSet {
+    /// Creates a new `FrozenTagSet` from the given vector of tags.
+    pub fn new(tags: Vec<Tag>) -> Self {
```
webern: the type is `pub(crate)` but these functions are `pub`.

webern: I would err on the side of less visibility whenever possible...

PR #1274, opened 2026-03-27
chore(config): switch to `facet` for deserialization

tobz:
## Summary

This PR switches `saluki-config` to use `facet` over `figment` in order to support future improvements around configuration loading and error diagnostics.

Currently, `saluki-config` is based on `figment`, which is in turn based on `serde`. This works well 99% of the time, and aligns with the ecosystem's general choice to reach for `serde` for (de)serializing YAML and JSON data. However, it comes with some notable pitfalls:

- error diagnostics are very dependent on using `serde` in basic/simple ways, and become useless or non-existent even when only slightly straying off the happy path
- `figment` is heavily based around `serde_json::Value`, and so shuffling data into your configuration object generally requires being able to give it an actual nested map-like structure, which isn't _always_ feasible (more on this below)

The main source of these issues is that `serde` is geared around procedural macro-based code generation and incoming serialized flowing in one direction towards the deserialization code. Essentially, if you have a non-standard format for incoming data, or you're doing things where the procedural macro loses visibility (like `#[serde(flatten)]`) into the types involved, you're kind of stuck with _runtime_-based approach which is where things very quickly become unhelpful.

In contrast to `serde`, `facet` is a library ecosystem essentially designed around reflection. Once you have reflection, things like deserializing arbitrary data formats into structs becomes almost free. On top of this, `facet` has a number of helper libraries and patterns to support our goals: completely custom parsing logic for deserialization, maintaining insight into types/fields when flattened, and so on. Our intent is to switch over to `facet` with the current functionality staying the same, and then slowly move things over to `facet` completely such that we can then start taking advantage of these benefits.

To that end, this PR switches entirely to `facet` inside `saluki-config` but leaves a bridge to `serde` as part of the typed getter methods on `GenericConfiguration`, to avoid having to convert over everything all at once.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

## References

AGTMETRICS-400


lib/saluki-components/src/common/datadog/config.rs
```diff
@@ -133,6 +133,11 @@ mod tests {
     const PROXY_B: &str = "http://proxy-b.example.com:3128";
     const PROXY_B_URI: &str = "http://proxy-b.example.com:3128/";
 
+    fn serde_to_facet(v: serde_json::Value) -> facet_value::Value {
+        let s = serde_json::to_string(&v).unwrap();
+        facet_json::from_str(&s).unwrap()
```
webern: 🙃

webern (approved):

Issue #806, opened 2025-07-14
Add support for `no_proxy`


rayz:
#530 Added support for proxies but not for `no_proxy`.

We will probably need to create a custom [Intercept](https://docs.rs/hyper-http-proxy/latest/hyper_http_proxy/enum.Intercept.html).

PR #1289, opened 2026-03-31
enhancement(config): add `no_proxy` support

thieman:
Adds support for the following proxy configuration fields through YAML or environment variables:

- `proxy.no_proxy`: List of hosts, subdomains, IP addresses, CIDRs, or a `*` wildcard which will bypass any configured HTTP or HTTPS proxies
- `no_proxy_nonexact_match`: Default False, in which case only exact hosts and IPs are considered for `no_proxy`. This seems to be mostly legacy behavior and is recommended to be set to True in our public docs.
- `use_proxy_for_cloud_metadata`: Default False, in which case requests to two known cloud provider metadata IPs (`169.254.169.254` and `100.100.100.200`) will bypass the proxy.

Relevant sources:
- https://docs.datadoghq.com/agent/configuration/proxy/
- https://github.com/DataDog/datadog-agent/blob/4eae10d668564ee2bc0681a172ce7f878e4e82e5/pkg/util/http/transport.go
- https://github.com/DataDog/datadog-agent/blob/4eae10d668564ee2bc0681a172ce7f878e4e82e5/pkg/config/setup/config.go

Fixes https://github.com/DataDog/saluki/issues/806.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3595,6 +3595,7 @@ dependencies = [
  "saluki-io",
  "saluki-metadata",
  "saluki-metrics",
+ "saluki-tls",
  "serde",
  "serde_json",
  "serde_with",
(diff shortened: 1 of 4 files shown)


tobz (approved):
This seems reasonable to me. 👍🏻

PR #1290, opened 2026-03-31
enhancement(common): extend `PermissiveBool` to match Go's `strconv.ParseBool`

thieman:
Noticed this as part of working on config support for `DD_NO_PROXY_NONEXACT_MATCH` which is a boolean config value. Within the existing Agent there is a wide variety of bool parsing going on, but it looks like `nodetreemodel` is eventually calling out to `spf13/cast` which eventually gets to `strconv.ParseBool`. This Go stdlib function allows some string values which will currently error out in Saluki:

> ParseBool returns the boolean value represented by the string. It accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False. Any other value returns an error.

This PR extends `PermissiveBool` to accommodate all of these cases, in addition to its current support for case-insensitive "true" and "false".

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3619,6 +3619,7 @@ dependencies = [
  "saluki-error",
  "saluki-metrics",
  "serde",
+ "serde_json",
  "serde_with",
  "sha3",
  "smallvec",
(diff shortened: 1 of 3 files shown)


tobz (changes requested):
Seems like we're missing basic unit tests that it actually accepts the short forms?

tobz (approved):

tobz (approved):

PR #1305, opened 2026-04-02
fix(ddsketch): correct trim_left bin overflow and handle oversized metrics in encoder

### LINKED ISSUE #1119: Investigate why correctness tests don't seem to exercise sample rate handling difference between Agent and ADP for DogStatsD.
## Context

I recently discovered that it seems that the Datadog Agent does not actually enforce a bin limit any longer on its sketch structures used for distributions coming over DogStatsD. This shows up in terms of how we now handle distributions in ADP which involves clamping the incoming sample rate to avoid trying to add more samples to a `DDSketch` than would be allowed by the maximum theoretical count based on the configured bin limit and integral types.

While the Agent will happily track more samples than the theoretical limit, we do not in ADP. This is only visible in some internal deployments where distributions with _very_ high sample rates are present.

The main question is: why do we not seem to exercise this code path in correctness tests given that we're using `lading`, which often generates payload in a very fuzzer-like fashion. We should investigate if something on the `lading` side is preventing sufficiently high sample rates from being generated in these payloads, or if something else is at play.

thieman:
Result of an investigation into https://github.com/DataDog/saluki/issues/1119. Closes https://github.com/DataDog/saluki/issues/1119.

To answer the original question from that issue: the correctness tests do not exercise this distribution case because it is only triggered when the provided sample rate is less than the supported minimum of `0.000000003845` ([source](https://github.com/DataDog/saluki/blob/d80c47e3bda0f1515bba2461433ecfa39349e3da/lib/saluki-components/src/sources/dogstatsd/mod.rs#L126-L128)). This never happens in the correctness test because Lading [defaults the sampling range](https://github.com/DataDog/lading/blob/bd60f9813012278a87a30516379805841c2e089f/lading_payload/src/dogstatsd.rs#L220) to a uniform distribution in the range `[0.1, 1.0]`. When I created a new correctness test which forced the sampling rate to below the minimum, I did observe the expected deviation in DDA and Saluki results. However, I also observed a panic and a logic error, both of which are fixed by this PR.

This PR fixes and provides regression tests for these problems, described below. It does _not_ add any new correctness tests, as we confirmed that sketch behavior at very low sample rates is divergent. After speaking with @tobz we agreed to keep the Saluki behavior, which enforces a limit on maximum bin count, in place as it aligns with our larger goals around bounding.

## Change Summary

Two related bugs triggered by distributions with very large per-sample weights (weight > `u16::MAX` = 65535, i.e. sample rates below ~1.5e-5):

**Bug 1 — `trim_left` bin count explosion (`lib/ddsketch`)**

When a single `insert_n` call generates multiple bins for the same key (because the weight exceeds `MAX_BIN_WIDTH = 65535`), the collapse loop in `trim_left` accumulated overflow bins into the output size budget via `overflow.truncate(bin_limit + overflow_len)`. This prevented any actual reduction: the bin array grew without bound across insertions rather than staying at `DDSKETCH_CONF_BIN_LIMIT` (4096). The fix truncates to `bin_limit` unconditionally.

**Bug 2 — encoder panic on oversized sketch (`lib/saluki-components`)**

The unbounded sketch from bug 1 serializes to an uncompressed encoded size larger than the encoder's 3.2 MiB compressed payload limit. `encode_inner` detected this via `would_write_exceed_threshold`'s early `len > threshold` guard and returned `Ok(false)` — the signal to flush and retry. But on an empty builder, `flush()` returns an empty vec, hitting the panic in `run_request_builder`: `"builder told us to flush, but gave us nothing"`. The fix adds `InputExceedsCompressedSizeLimit` to `RequestBuilderError` and returns it from `encode_inner` when the single metric's encoded size already exceeds the compressed limit, so the caller drops the metric via the existing error path rather than looping forever.

## Test plan

- [ ] `trim_left_respects_bin_limit_with_large_weights` — inserts 10 values at weight 260,078,024 (ADP's minimum safe sample rate), asserts bin count never exceeds 4096; fails after the 2nd insert without the fix (bins = 7938)
- [ ] `input_exceeds_compressed_size_limit` — encodes an input larger than the configured compressed limit into an empty builder, asserts `InputExceedsCompressedSizeLimit` is returned (not a panic); confirms builder remains usable after the error
- [ ] Manually reproduced original panic via correctness test harness with sample rate `1e-9`; confirmed no crash with both fixes applied

### DIFF
--- a/lib/ddsketch/src/agent/sketch.rs
+++ b/lib/ddsketch/src/agent/sketch.rs
@@ -688,6 +688,9 @@ fn trim_left(bins: &mut SmallVec<[Bin; 4]>, bin_limit: u16) {
     // above (when collapsed counts exceed MAX_BIN_WIDTH) are prepended before bins_end, and together the
     // combined slice is capped at bin_limit. This may discard some higher-key bins from bins_end when
     // overflow is large, which is the expected precision trade-off for a bounded sketch.
+    //
+    // As of April 2026, this is an intentional divergence from the Datadog Agent implementation,
+    // which does not truncate bins to stay under a limit.
     overflow.truncate(bin_limit);
 
     mem::swap(bins, &mut overflow);
--- a/lib/ddsketch/src/agent/sketch.rs
+++ b/lib/ddsketch/src/agent/sketch.rs
@@ -771,35 +771,38 @@ mod tests {
         assert_eq!(total_before, total_after);
     }
 
-    /// When collapsed mass overflows u16::MAX, generate_bins is called with the first kept bin's
-    /// key — not the keys of the removed bins. This is the core correctness property: all mass,
-    /// including overflow, is attributed to the first kept key.
+    /// When collapsed mass overflows u16::MAX, all overflow is attributed to the first kept bin's
+    /// key, and the first kept bin absorbs the maximum possible count before drain.
     ///
-    /// The old code pushed overflow bins with `bin.k` (the removed bin's key) during accumulation,
-    /// which would leave bins with removed keys in the output. This test verifies that every bin
-    /// in the output has a key >= the first kept key before trimming.
+    /// The old code subtracted MAX_BIN_WIDTH from `missing` during the accumulation loop and
+    /// stashed it in a wrong-key overflow bin. Even though that bin was later drained, the
+    /// subtraction meant less mass reached `bin_remove.increment(...)`, leaving the first kept
+    /// bin with a lower `n` than it should have.
     ///
     /// Input:  [(0,50000), (1,50000), (2,1)]  limit=1  →  remove 2 bins
     /// missing = 50000 + 50000 = 100000
     /// first_kept_k = 2; bins[2].increment(100000): n → 65535, returns 34466
-    /// generate_bins(k=2, 34466): overflow = [(2, 34466)]
-    /// overflow + bins_end = [(2,34466), (2,65535)]  →  2 bins > limit=1
-    /// drain 1 from front: [(2, 65535)]
-    /// All output keys must be 2; no keys 0 or 1 may appear.
+    /// generate_bins(k=2, 34466): overflow = [(2,34466)]
+    /// overflow + bins_end = [(2,34466),(2,65535)]  →  2 bins > limit=1
+    /// drain 1 from front: [(2,65535)]
+    ///
+    /// Old code: loop emits {k=1,n=65535} and reduces missing to 34465; bins[2].increment(34465)
+    /// gives n=34466. After drain the result is [(2,34466)] — correct key but wrong (lower) count.
     #[test]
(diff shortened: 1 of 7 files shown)


lib/saluki-components/src/common/datadog/request_builder.rs
```diff
@@ -333,6 +342,16 @@ where
...
 
+        // If the input's encoded size already exceeds the compressed payload limit, it can never fit regardless of
+        // what else is in the payload, so there is no point in asking the caller to flush and retry. Return an error
+        // so the input is dropped rather than looping forever.
+        if self.scratch_buf.len() > self.compressed_len_limit {
+            return Err(RequestBuilderError::InputExceedsCompressedSizeLimit {
+                encoded_len: self.scratch_buf.len(),
+                compressed_len_limit: self.compressed_len_limit,
+            });
+        }
```
tobz: This actually... isn't right, now that I think about it.

The whole thing is that we really _can't_ know if we've exceed the compressed size limit or not until we actually attempt compressing the request payload. I think what's going on, more likely than not, is that we're triggering the "can't encode this metric" pathway (`Ok(false)`) because `CompressionEstimator::would_write_exceed_threshold` returns `true`: naively, it will end up checking if `encoded_len` is greater than `self.compressed_len_limit`, and return `true` if so. That's why your "fix" appears to work, because it's preventing us from reaching that faulty logic.

I think we would actually just want to remove the conditional I mentioned from `CompressionEstimator::would_write_exceed_threshold` since, realistically, it shouldn't be able to estimate overflow without any compressed data have been written out yet. I don't know why I added that check initially but it's feeling wrong in hindsight.

lib/ddsketch/src/agent/sketch.rs
```diff
@@ -681,13 +681,17 @@ fn trim_left(bins: &mut SmallVec<[Bin; 4]>, bin_limit: u16) {
...
     let (_, bins_end) = bins.split_at(num_to_remove);
     overflow.extend_from_slice(bins_end);
 
-    // I still don't yet understand how this works, since you'd think bin limit should be the overall limit of the
-    // number of bins, but we're allowing more than that.. :thinkies:
-    overflow.truncate(bin_limit + overflow_len);
+    // Truncate to bin_limit so the total bin count stays within the configured limit. Overflow bins created
+    // above (when collapsed counts exceed MAX_BIN_WIDTH) are prepended before bins_end, and together the
+    // combined slice is capped at bin_limit. This may discard some higher-key bins from bins_end when
+    // overflow is large, which is the expected precision trade-off for a bounded sketch.
```
tobz: Technically, I think we actually want to trim the _lower_ bins, since lowest-collapsing is the default behavior for the official DDSketch implementation, IIRC: it means you end up trading accuracy at lower percentiles where it matters far less.

As is, we would be harming the upper percentiles: p95, p99, yadda yadda.

PR #1307, opened 2026-04-03
fix(ddsketch): correct `trim_left` collapse behavior and widen bin count type to `u32`

thieman:
## Summary

Fixes a series of bugs in the agent DDSketch `trim_left` function, discovered while investigating bin count explosion at ADP's minimum sample rate (~260M observations per sample).

- **Bin count not actually capped**: `trim_left` used `truncate(bin_limit + overflow_len)` instead of `truncate(bin_limit)`, meaning the overflow bins inflated the threshold and no reduction occurred. Bin count grew without bound, eventually producing serialized payloads that caused a panic in the request builder (handled separately).
- **Collapsed mass attributed to wrong key**: When accumulated mass from removed bins exceeded `MAX_BIN_WIDTH` mid-loop, overflow bins were emitted using the removed bin's key instead of the first kept bin's key. Per `CollapsingLowestDenseStore` in sketches-go, all removed mass must collapse into the first kept bin's index.
- **Mass loss from u16 overflow**: With `Bin.n: u16` (max 65535), a single `insert_n` at weight ~260M generated ~3969 bins per key. Even with the above fixes, collapsing those bins saturated at 65535 and discarded real observations. Widening `Bin.n` to `u32` (max ~4.3B) means the ADP weight fits in a single bin, eliminating mass loss for all realistic workloads.
- **Dead code in overflow path**: After widening to u32, the `generate_bins` call in `trim_left` was dead — any bins it produced were always prepended and immediately drained. Simplified to: accumulate mass, `increment` into first kept bin (saturating at u32::MAX), `drain` the prefix.

The collapse semantics are validated against the `CollapsingLowestDenseStore` implementation in [sketches-go](https://github.com/DataDog/sketches-go), which is the reference implementation. The one intentional divergence: sketches-go uses `float64` counts and never loses mass; we use `u32` and lose mass only if a single collapsed bin exceeds ~4.3B observations, which is not reachable in practice.

## Test plan

- `trim_left_collapses_removed_mass_into_first_kept_bin` — exact bin counts after multi-bin collapse
- `trim_left_preserves_total_count_when_no_overflow` — sum of `n` preserved exactly
- `trim_left_preserves_exact_count_with_u32_bins` — u32 absorbs 100001-count collapse without saturation (with u16 this lost 34466 observations)
- `trim_left_no_op_when_within_limit` — no-op boundary
- `trim_left_respects_bin_limit_with_large_weights` — bin count never exceeds limit at ADP minimum sample rate weight
- Allocation tests updated to reflect `Bin` size growing from 4→8 bytes

### DIFF
--- a/lib/ddsketch/Cargo.toml
+++ b/lib/ddsketch/Cargo.toml
@@ -22,13 +22,13 @@ smallvec = { workspace = true, features = ["union"] }
 
 [dev-dependencies]
 criterion = { workspace = true }
-proptest = { workspace = true }
 dhat = { workspace = true }
 ndarray = { version = "0.17", default-features = false }
 ndarray-stats = { version = "0.7", default-features = false }
 noisy_float = { workspace = true }
 num-traits = { workspace = true }
 ordered-float = { workspace = true }
+proptest = { workspace = true }
 rand = { workspace = true, features = ["small_rng", "thread_rng"] }
 rand_distr = { workspace = true }
 
--- a/lib/ddsketch/Cargo.toml
+++ b/lib/ddsketch/Cargo.toml
@@ -22,6 +22,7 @@ smallvec = { workspace = true, features = ["union"] }
 
 [dev-dependencies]
 criterion = { workspace = true }
+proptest = { workspace = true }
 dhat = { workspace = true }
 ndarray = { version = "0.17", default-features = false }
 ndarray-stats = { version = "0.7", default-features = false }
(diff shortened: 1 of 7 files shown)


tobz (approved):
This actually increase memory usage much less than I thought it would. Our SMP workloads aren't super distribution-heavy, but distributions _are_ represented well enough so this seems reasonable to push through. 👍🏻

Issue #1134, opened 2026-01-29
Implement missing samplers in Trace Sampler transform


aqian01:
Implement the missing samplers 
rare samplers - ([code](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/agent.go#L1071))
error tracking standalone ([code](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/agent.go#L1061))

andrewqian2001datadog:
Closing issue since we have open PRs for both. 

https://github.com/DataDog/saluki/pull/1311

https://github.com/DataDog/saluki/pull/1314/s

PR #1311, opened 2026-04-03
feat(trace sampler): implement rare sampler

thieman:
## Summary

Implements the rare sampler from `datadog-agent/pkg/trace/sampler/rare_sampler.go` matching the Go V0 behavior (`runSamplers`). Ensures low-traffic trace shapes that would otherwise be dropped by probabilistic or priority sampling remain visible.

## Changes

- **`rare_sampler.rs`** — new module with `RareSampler`, a per-(env,service)-shard `SeenSpans` with cardinality-bounded TTL tracking and modular-hash shrinking on overflow
- **`signature.rs`** — expose `span_hash_for_rare` helper (`compute_span_hash` with `env=""`, `with_resource=true`)
- **`apm.rs`** — add `RareSamplerConfig` with `tps`, `cooldown`, `cardinality` fields; `enable_rare_sampler` lives at the `ApmConfiguration` wrapper level to support env var remapping (`APM_ENABLE_RARE_SAMPLER`); disabled by default
- **`mod.rs`** — wire rare sampler into `run_samplers` before all other samplers in both probabilistic and legacy (priority) paths
- **`saluki-common/src/rate.rs`** — new `TokenBucket` (tokens-per-second refill, burst capacity) extracted as a reusable primitive

## Behavioral notes

- Rare sampler runs before all other samplers on every trace
- In the probabilistic path: rare short-circuits probabilistic/error sampling (else-if chain matching Go V0)
- In the legacy path: rare is checked after manual-drop but before priority/no-priority samplers
- When rare wins: `_dd.rare = 1` set on the first rare span; no `_dd.p.dm` set (matches Go)
- Default: disabled (`apm_config.enable_rare_sampler: false`), 5 TPS, 300s cooldown, 200 cardinality per shard

## Test plan

- [ ] `rare_sampler.rs`: 11 unit tests covering disabled mode, new signature kept, TTL drop, expiry re-sample, measured spans, non-top-level ignored, multi-top-level flag placement, rate limiting, cardinality shrink, span eligibility
- [ ] `saluki-common/src/rate.rs`: 4 unit tests for `TokenBucket`
- [ ] `apm.rs`: 5 config tests (disabled by default, enable via YAML, enable via env var, env var overrides YAML, tuning via YAML)
- [ ] `mod.rs`: 7 interaction tests (rare catches 0% probabilistic drop, `_dd.rare=1` set, no resample within TTL, rare disabled drops, legacy path PriorityAutoDrop caught, 100%/0% probabilistic baseline)

Partial fix for #1134 (rare sampler only; error tracking standalone and priority/no-priority sampler gaps remain).

### DIFF
--- a/lib/saluki-components/src/common/datadog/apm.rs
+++ b/lib/saluki-components/src/common/datadog/apm.rs
@@ -85,6 +85,10 @@ impl Default for RareSamplerConfig {
 struct ApmConfiguration {
     #[serde(default)]
     apm_config: ApmConfig,
+
+    /// Enables the rare sampler. This needs to live up here rather than nested
+    /// within `apm_config` so that we can remap the environment variable path
+    /// using the ConfigurationLoader::with_key_aliases.
     #[serde(default = "default_rare_sampler_enabled", rename = "apm_enable_rare_sampler")]
     enable_rare_sampler: bool,
 }
--- a/lib/saluki-components/src/common/datadog/apm.rs
+++ b/lib/saluki-components/src/common/datadog/apm.rs
@@ -35,7 +35,7 @@ const fn default_rare_sampler_tps() -> f64 {
     5.0
 }
 
-const fn default_rare_sampler_cooldown_period_secs() -> f64 {
+const fn default_rare_sampler_cooldown_secs() -> f64 {
     300.0 // 5 minutes
 }
 
@@ -55,18 +55,12 @@ fn default_env() -> MetaString {
     MetaString::from("none")
 }
 
-/// Rare sampler configuration.
+/// Rare sampler tuning configuration (`apm_config.rare_sampler.*`).
+///
+/// The enabled flag lives separately at `apm_config.enable_rare_sampler` to match the Datadog
+/// Agent's config layout and allow `DD_APM_ENABLE_RARE_SAMPLER` env var override.
 #[derive(Clone, Debug, Deserialize)]
 struct RareSamplerConfig {
-    /// Enables the rare sampler.
-    ///
-    /// When enabled, the rare sampler catches traces for (env, service, name, resource, error type, http status)
-    /// combinations that are not seen by the priority sampler.
-    ///
-    /// Defaults to `false`.
-    #[serde(default = "default_rare_sampler_enabled")]
-    enabled: bool,
-
     /// Target traces per second for the rare sampler.
     ///
     /// Defaults to 5.0.
@@ -76,8 +70,8 @@ struct RareSamplerConfig {
     /// Cooldown period in seconds before a rare signature can be sampled again.
     ///
     /// Defaults to 300.0 (5 minutes).
-    #[serde(default = "default_rare_sampler_cooldown_period_secs")]
-    cooldown_period_secs: f64,
+    #[serde(default = "default_rare_sampler_cooldown_secs")]
+    cooldown: f64,
 
     /// Max number of span signatures tracked per (env, service) shard before shrinking.
     ///
@@ -89,9 +83,8 @@ struct RareSamplerConfig {
 impl Default for RareSamplerConfig {
     fn default() -> Self {
         Self {
-            enabled: default_rare_sampler_enabled(),
             tps: default_rare_sampler_tps(),
-            cooldown_period_secs: default_rare_sampler_cooldown_period_secs(),
+            cooldown: default_rare_sampler_cooldown_secs(),
             cardinality: default_rare_sampler_cardinality(),
         }
     }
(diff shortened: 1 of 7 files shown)


lib/saluki-components/src/common/datadog/apm.rs
```diff
@@ -27,6 +27,22 @@ const fn default_error_tracking_standalone_enabled() -> bool {
...
+    5.0
+}
+
+const fn default_rare_sampler_cooldown_period_secs() -> f64 {
+    300.0 // 5 minutes
+}
+
+const fn default_rare_sampler_cardinality() -> usize {
+    200
+}
```
thieman: Confirmed these against https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/config/config.go#L615-L617

aqian01 (commented):
For this [code](https://github.com/DataDog/saluki/blob/thieman/rare-sampler/lib/saluki-components/src/transforms/trace_sampler/mod.rs#L356-L370)

I think we want to add a rare check?

In the [agent](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/agent.go#L1129-L1140), there is a rare check before the no priority "legacy" path.

Added a test to show that the we decide to drop even when rare matches
```
/// Demonstrates the current ADP/agent divergence for OTLP traces without a sampling priority.
    ///
    /// The Datadog Agent still honors the rare decision for this shape, but ADP currently routes
    /// OTLP no-priority traces through the upstream OTLP sampling branch and drops them when that
    /// rate check fails.
    #[test]
    fn rare_sampler_otlp_no_priority_differs_from_agent() {
        let mut sampler = create_sampler_with_rare_enabled();
        sampler.probabilistic_sampler_enabled = false;
        sampler.error_sampling_enabled = false;
        sampler.otlp_sampling_rate = 0.0;

        let mut meta = saluki_common::collections::FastHashMap::default();
        meta.insert(
            MetaString::from_static(OTEL_TRACE_ID_META_KEY),
            MetaString::from("00000000000000000000000000000001"),
        );
        let span = create_top_level_span(888, 1).with_meta(meta);
        let mut trace = create_test_trace(vec![span]);

        let (keep, priority, decision_maker, root_idx) = sampler.run_samplers(&mut trace);
        assert!(!keep, "ADP currently drops this OTLP trace even though rare matched");
        assert_eq!(priority, PRIORITY_AUTO_DROP);
        assert_eq!(decision_maker, "");
        assert_eq!(
            trace.spans()[root_idx.unwrap()]
                .metrics()
                .get(rare_sampler::RARE_KEY)
                .copied(),
            Some(1.0),
            "the rare sampler ran, but its keep decision was not honored"
        );
    }
```

aqian01 (approved):

lib/saluki-components/src/transforms/trace_sampler/mod.rs
```diff
@@ -327,10 +346,19 @@ impl TraceSampler {
                 return (false, priority, "", Some(root_span_idx));
             }
 
+            if rare {
+                return (true, PRIORITY_AUTO_KEEP, "", Some(root_span_idx));
```
webern: (Claude Opus 4.6) Minor divergence from Go V0: In the Go legacy path ([agent.go:1129-1131](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/agent.go#L1129-L1131)), `if rare { return true, true }` does **not** modify the trace chunk's priority — it preserves whatever the tracer originally set (could be AutoDrop=0, AutoKeep=1, or UserKeep=2). Here, we explicitly return `PRIORITY_AUTO_KEEP`, which means a UserKeep(2) trace caught by rare gets downgraded to AutoKeep(1).

The no-priority path (line 377) and OTLP path (line 359) already follow this same pattern, so this is consistent within the Rust codebase. Just flagging the Go delta — could matter if downstream consumers distinguish UserKeep from AutoKeep.

lib/saluki-components/src/transforms/trace_sampler/mod.rs
```diff
@@ -282,33 +290,44 @@ impl TraceSampler {
...
+        let now = std::time::SystemTime::now();
         let contains_error = self.trace_contains_error(trace, false);
         let Some(root_span_idx) = self.get_root_span_index(trace) else {
             return (false, PRIORITY_AUTO_DROP, "", None);
         };
 
+        // Run the rare sampler early, before all other samplers. This mirrors the Go agent behavior
+        // where the rare sampler runs first to catch traces that would otherwise be dropped entirely.
+        // logic taken from: https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/agent.go#L1078
+        let rare = self.rare_sampler.sample(trace, root_span_idx);
```
webern: (Claude Opus 4.6) Pre-existing issue (not introduced by this PR), but worth noting now that the rare sampler amplifies its impact: In Go V0 ([agent.go:1070-1077](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/agent.go#L1070-L1077)), `ErrorTrackingStandalone` is checked **before** the rare sampler call, gating all samplers from running on non-error traces. In Rust, `run_samplers` has no ETS guard — the rare sampler runs unconditionally, and the ETS check at `process_trace:454` only fires on the `keep=false` fallback path. This means with ETS enabled, the rare sampler (and other samplers) can keep non-error traces that Go would drop, violating the ETS contract.

webern (approved):
Assisted code review. Looks good.

It may have identified some testing gaps if they are of interest:

  Go `agent_test.go` TestSampling — the interesting gaps are in the integration-level interactions. Three scenarios from Go aren't covered in the PR:

  1. Manual drop short-circuits rare — Go's priority < 0 check fires before rare. A UserDrop(-1) trace should be dropped even if rare=true. The PR has no test for this. Relevant because the Rust code gates rare behind priority < PRIORITY_AUTO_DROP.
  2. Rare wins over probabilistic when both would keep — Go's probabilistic-rare-100 tests that when probabilistic is 100% and rare is enabled, the trace is kept via the rare path (samplerName=rare, no _dd.p.dm tag). The PR's rare_sampler_catches_unsampled_trace only tests the case where probabilistic is 0%. A test at 100% would verify decision_maker stays empty when rare takes precedence over probabilistic.
  3. Rare + error trace interaction — Go tests (error-sampled-prio-unsampled, error-sampled) show that rare catches error traces before the error sampler runs. The PR has no test where contains_error=true and rare is enabled — verifying the else-if chain means error sampler is never consulted when rare fires.

All three are run_samplers-level interaction tests (they'd go in mod.rs tests).

PR #1314, opened 2026-04-06
feat(trace sampler): implement error tracking standalone mode

thieman:
## Summary

Implements Error Tracking Standalone (ETS) sampling mode, matching [`datadog-agent/pkg/trace/agent/agent.go` `runSamplers`](https://github.com/DataDog/datadog-agent/blob/be33ac1490c4a34602cbc65a211406b73ad6d00b/pkg/trace/agent/agent.go#L976).

## Changes

- **`transforms/trace_sampler/mod.rs`** — ETS check runs at the top of `run_samplers` before all other samplers. Traces containing errors (including exception span events) are routed to the error sampler; non-error traces are dropped immediately. Dropped non-error ETS traces are forwarded to intake with `DroppedTrace=true` (suppressing SSS and analytics events), matching Go agent behavior.
  - For OTLP traces when the probabilistic sampler is disabled, `_dd.p.dm` and `_sampling_priority_v1` are pre-assigned inside the ETS block via `otlp_pre_sample()` — mirroring `OTLPReceiver.createChunks` in the Go agent. User-set priorities receive `dm="-4"` (manual); probabilistically-sampled traces receive `dm="-9"`.
- **`encoders/datadog/traces/mod.rs`** — emits `_dd.error_tracking_standalone.error = "true"` as a chunk tag (only for traces that actually contain an error span or exception span event) and sets `X-Datadog-Error-Tracking-Standalone: true` on outbound requests when ETS is enabled. The HTTP header is pre-built at construction time to avoid per-request allocation.
- **`common/datadog/mod.rs`** — adds `DECISION_MAKER_MANUAL` constant (`"-4"`) for user/manual sampling decisions.
- **`common/datadog/apm.rs`** — ETS enabled via `apm_config.error_tracking_standalone.enabled` (YAML) or `DD_APM_ERROR_TRACKING_STANDALONE_ENABLED` (env var), following the same alias pattern as the rare sampler. Bool method docstrings standardized to "Returns `true` if ...".
- **`common/datadog/request_builder.rs`** — adds `additional_headers()` hook to `EndpointEncoder` trait for encoder-specific request headers.
- **`test/correctness/otlp-traces-ets/`** — correctness test comparing ETS behavior against a DDA baseline, with `error_rate: 0.1` to ensure a meaningful number of error traces pass through.

## Behavioral notes

- ETS takes priority over all other samplers (rare, probabilistic, priority)
- Error detection includes both `span.error != 0` and `_dd.span_events.has_exception = "true"`
- Error traces: routed through error sampler (TPS-limited); if kept, tagged with ETS chunk tag and forwarded with ETS request header
- Non-error traces: forwarded with `DroppedTrace=true`, no SSS or analytics event fallback; ETS chunk tag is not written (tag is only present on error traces)
- OTLP pre-sampling (dm + priority) is computed inside the ETS block via a dedicated `otlp_pre_sample()` method, keeping the computation co-located with its only consumer

## Test plan

- [x] `ets_keeps_trace_with_error` — error trace kept by error sampler
- [x] `ets_drops_trace_without_error` — non-error trace dropped by run_samplers
- [x] `ets_forwards_dropped_trace_with_dropped_flag` — non-error ETS trace forwarded with `DroppedTrace=true`; SSS not applied
- [x] `ets_keeps_trace_with_exception_span_event` — exception span events count as errors
- [x] `ets_disabled_uses_normal_sampling` — ETS disabled falls through to normal sampling path
- [x] `ets_otlp_non_error_gets_presample_priority_and_dm` — non-error OTLP trace gets `priority=AutoKeep, dm="-9"` before ETS drop
- [x] `ets_otlp_error_gets_presample_priority_and_dm` — error OTLP trace gets `priority=AutoKeep, dm="-9"` when kept
- [x] `ets_otlp_probabilistic_path_skips_presample` — probabilistic sampler path: no pre-sampling applied
- [x] `ets_non_otlp_unaffected_by_presample` — non-OTLP traces unaffected by OTLP pre-sampling logic
- [x] `ets_otlp_user_priority_gets_manual_dm` — user-set priority gets `dm="-4"` (manual)
- [x] `ets_header_present_when_enabled` / `ets_header_absent_when_disabled` — HTTP header behavior
- [x] `ets_chunk_tag_present_for_error_trace` — chunk tag written for error traces
- [x] `ets_chunk_tag_absent_for_non_error_trace` — chunk tag not written for non-error traces
- [x] `ets_chunk_tag_absent_when_disabled` — chunk tag not written when ETS is disabled
- [x] Config tests: disabled by default, enabled via YAML, enabled via env var, env var overrides YAML
- [x] Correctness test (`otlp-traces-ets`): DDA vs ADP output matches with no differences detected

Stacked on #1311.

### DIFF
--- a/lib/saluki-components/src/common/datadog/apm.rs
+++ b/lib/saluki-components/src/common/datadog/apm.rs
@@ -226,7 +226,7 @@ impl ApmConfig {
         self.errors_per_second
     }
 
-    /// Returns if probabilistic sampling is enabled.
+    /// Returns `true` if probabilistic sampling is enabled.
     pub const fn probabilistic_sampler_enabled(&self) -> bool {
         self.probabilistic_sampler.enabled
     }
@@ -236,22 +236,22 @@ impl ApmConfig {
         self.probabilistic_sampler.sampling_percentage
     }
 
-    /// Returns if error sampling is enabled.
+    /// Returns `true` if error sampling is enabled.
     pub const fn error_sampling_enabled(&self) -> bool {
         self.error_sampling_enabled
     }
 
-    /// Returns if error tracking standalone mode is enabled.
+    /// Returns `true` if error tracking standalone mode is enabled.
     pub const fn error_tracking_standalone_enabled(&self) -> bool {
         self.error_tracking_standalone
     }
 
-    /// Returns if stats computation by span kind is enabled.
+    /// Returns `true` if stats computation by span kind is enabled.
     pub const fn compute_stats_by_span_kind(&self) -> bool {
         self.compute_stats_by_span_kind
     }
 
-    /// Returns if peer tags aggregation is enabled.
+    /// Returns `true` if peer tags aggregation is enabled.
     pub const fn peer_tags_aggregation(&self) -> bool {
         self.peer_tags_aggregation
     }
@@ -278,7 +278,7 @@ impl ApmConfig {
         }
     }
 
-    /// Returns whether the rare sampler is enabled.
+    /// Returns `true` if the rare sampler is enabled.
     pub const fn rare_sampler_enabled(&self) -> bool {
         self.enable_rare_sampler
     }
--- a/lib/saluki-components/src/common/datadog/apm.rs
+++ b/lib/saluki-components/src/common/datadog/apm.rs
@@ -91,6 +91,15 @@ struct ApmConfiguration {
     /// using the ConfigurationLoader::with_key_aliases.
     #[serde(default = "default_rare_sampler_enabled", rename = "apm_enable_rare_sampler")]
     enable_rare_sampler: bool,
+
+    /// Enables Error Tracking Standalone mode. Lives here (rather than nested within `apm_config`)
+    /// so that the env var path (`DD_APM_ERROR_TRACKING_STANDALONE` → `apm_error_tracking_standalone`)
+    /// can be remapped via ConfigurationLoader::with_key_aliases.
+    #[serde(
+        default = "default_error_tracking_standalone_enabled",
+        rename = "apm_error_tracking_standalone"
+    )]
+    enable_error_tracking_standalone: bool,
 }
 
 #[derive(Clone, Debug, Deserialize)]
@@ -124,26 +133,6 @@ impl Default for ProbabilisticSamplerConfig {
 }
 
 #[derive(Clone, Debug, Deserialize)]
-struct ErrorTrackingStandaloneConfig {
-    /// Enables Error Tracking Standalone mode.
-    ///
-    /// When enabled, error tracking standalone mode suppresses single-span sampling and analytics
-    /// events for dropped traces.
-    ///
-    /// Defaults to `false`.
-    #[serde(default = "default_error_tracking_standalone_enabled")]
-    enabled: bool,
-}
-
-impl Default for ErrorTrackingStandaloneConfig {
-    fn default() -> Self {
-        Self {
-            enabled: default_error_tracking_standalone_enabled(),
-        }
-    }
-}
-
-#[derive(Clone, Debug, Deserialize)]
 pub struct ApmConfig {
     /// Target traces per second for priority sampling.
     ///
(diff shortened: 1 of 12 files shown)


thieman:
[Claude Sonnet 4.6] Related to #1134.

lib/saluki-components/src/common/datadog/apm.rs
```diff
@@ -91,6 +91,15 @@ struct ApmConfiguration {
     /// using the ConfigurationLoader::with_key_aliases.
     #[serde(default = "default_rare_sampler_enabled", rename = "apm_enable_rare_sampler")]
     enable_rare_sampler: bool,
+
+    /// Enables Error Tracking Standalone mode. Lives here (rather than nested within `apm_config`)
+    /// so that the env var path (`DD_APM_ERROR_TRACKING_STANDALONE_ENABLED` → `apm_error_tracking_standalone_enabled`)
+    /// can be remapped via ConfigurationLoader::with_key_aliases.
```
tobz, line 97: I missed that we also did this pattern with the rare sampler enabled field... falls right into that very narrow space between what key aliases and env remappings give us. 😭

lib/saluki-components/src/common/datadog/apm.rs
```diff
@@ -256,7 +243,7 @@ impl ApmConfig {
 
     /// Returns if error tracking standalone mode is enabled.
     pub const fn error_tracking_standalone_enabled(&self) -> bool {
-        self.error_tracking_standalone.enabled
+        self.error_tracking_standalone
```
tobz, line 246: Can we actually update the doc comment for this method to say:

> Returns `true` if error tracking standalone mode is enabled.

lib/saluki-components/src/transforms/trace_sampler/mod.rs
```diff
@@ -453,8 +465,12 @@ impl TraceSampler {
             return true;
         }
 
+        // ETS: forward dropped traces with DroppedTrace=true, suppressing SSS/analytics.
         if self.error_tracking_standalone {
-            return false;
+            if let Some(root_idx) = root_span_idx {
+                self.apply_sampling_metadata(trace, false, priority, decision_maker, root_idx);
+            }
+            return true;
```
tobz: My brain is a little mushy trying to think this one through...

On the Agent side, this method is equivalent to `sample` ([here](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/agent.go#L967-L990)), where the boolean return value is whether or not to keep the trace.

Above this line, we check if `keep` is true and then return `true` if so... so if we're here, `keep` is false. Nowhere in `sample` is `keep` mutated after the call to `a.traceSampling(now, ts, pt)`, so why do _we_ return `true` even though we know `keep` is `false`? 🤔

tobz (changes requested):

lib/saluki-components/src/common/datadog/apm.rs
thieman: It's pattern matching here, I'll have it update this for the other bool-returning methods here as well

aqian01 (approved):

lib/saluki-components/src/transforms/trace_sampler/mod.rs
thieman: If we return `false` here then the Trace is removed from the buffer and ultimately dropped. In ETS mode we need to forward _all_ traces, and non-error traces get `DroppedTrace=true` metadata on them. Verified this behavior with the new correctness test added in this PR.

lib/saluki-components/src/transforms/trace_sampler/mod.rs
```diff
@@ -244,6 +244,35 @@ impl TraceSampler {
...
     }
 
+    /// Computes the OTLP pre-sampling priority and decision maker for a trace, mirroring
+    /// `OTLPReceiver.createChunks` in DDA which runs before `runSamplersV1`.
+    ///
+    /// Returns `Some((priority, dm))` for OTLP traces when the probabilistic sampler is disabled,
+    /// or `None` if pre-sampling does not apply.
+    ///
+    /// See: https://github.com/DataDog/datadog-agent/blob/be33ac1490c4a34602cbc65a211406b73ad6d00b/pkg/trace/api/otlp.go#L561-L585
+    fn otlp_pre_sample(&mut self, trace: &mut Trace, root_span_idx: usize) -> Option<(i32, &'static str)> {
```
thieman, line 254: This was a missing piece, in DDA some metadata is applied on incoming OTLP traces before the samplers are run. The ETS branch in runSamplers needs that metadata. In ADP, that same metadata is applied _after_ `run_samplers` is called, so we don't have it available here. This block allows us to source the OTLP-specific information we need without changing up the order of sampling vs applying metadata, which would cause us to waste cycles on otherwise-dropped traces.

lib/saluki-components/src/transforms/trace_sampler/mod.rs
tobz: Ahhh, OK. That tracks. 👍🏻

tobz (approved):

PR #1316, opened 2026-04-06
feat(trace sampler): emit sampler.seen/kept metrics

thieman:
## Summary

Implements all 6 sampler metrics from [`datadog-agent/pkg/trace/sampler/metrics.go`](https://github.com/DataDog/datadog-agent/blob/448ea066c365d4c6ecc74ed6c7e41e3aad64d452/pkg/trace/sampler/metrics.go) and [`rare_sampler.go`](https://github.com/DataDog/datadog-agent/blob/448ea066c365d4c6ecc74ed6c7e41e3aad64d452/pkg/trace/sampler/rare_sampler.go), part of #1134.

| Metric | Type | Tags |
|---|---|---|
| `datadog.trace_agent.sampler.seen` | Counter | `sampler`, `target_service`, `target_env`¹, `sampling_priority`² |
| `datadog.trace_agent.sampler.kept` | Counter | same |
| `datadog.trace_agent.sampler.size` | Gauge | `sampler` (priority / no_priority / error) |
| `datadog.trace_agent.sampler.rare.hits` | Counter | none |
| `datadog.trace_agent.sampler.rare.misses` | Counter | none |
| `datadog.trace_agent.sampler.rare.shrinks` | Gauge | none |

¹ `target_env` only emitted for priority, no_priority, rare, and error samplers  
² `sampling_priority` only emitted for the priority sampler

**Implementation details:**
- `run_samplers` returns a `SamplingDecision` struct instead of a bare tuple, carrying sampler name and priority for metric recording
- `seen`/`kept` counter handles are lazily registered and cached in a `FastHashMap` — hot path pays only a hash-map lookup after the first observation of each tag combination
- `rare.hits`/`rare.misses` are incremented inline in `RareSampler::sample()`; `rare.shrinks` is updated each time `SeenSpans::shrink()` fires
- `sampler.size` gauges are updated after each `transform_buffer` batch by polling `size()` on each score-based sampler

## Test plan

- [ ] All 43 existing trace sampler unit tests pass
- [ ] `cargo clippy` clean

### DIFF
--- a/lib/saluki-components/src/transforms/trace_sampler/metrics.rs
+++ b/lib/saluki-components/src/transforms/trace_sampler/metrics.rs
@@ -3,19 +3,23 @@
 //! Mirrors the `Metrics` / `MetricsKey` types from
 //! [`datadog-agent/pkg/trace/sampler/metrics.go`](https://github.com/DataDog/datadog-agent/blob/be33ac1490c4a34602cbc65a211406b73ad6d00b/pkg/trace/sampler/metrics.go).
 //!
-//! For each sampling decision, two counters are emitted:
+//! Per-decision counters (updated inline on every trace):
 //! - `datadog.trace_agent.sampler.seen` — every trace that passed through the sampler
 //! - `datadog.trace_agent.sampler.kept` — traces that were kept
 //!
 //! Both are tagged with `sampler`, `target_service`, and (when relevant) `target_env` and
 //! `sampling_priority`.
+//!
+//! Per-sampler gauges (updated after each batch in `transform_buffer`):
+//! - `datadog.trace_agent.sampler.size` — current signature-table size for priority/no_priority/error samplers
 
-use metrics::Counter;
+use metrics::{Counter, Gauge};
 use saluki_common::collections::FastHashMap;
 use saluki_metrics::MetricsBuilder;
 
 const METRIC_SEEN: &str = "datadog.trace_agent.sampler.seen";
 const METRIC_KEPT: &str = "datadog.trace_agent.sampler.kept";
+const METRIC_SIZE: &str = "datadog.trace_agent.sampler.size";
 
 /// The sampler that made a sampling decision.
 ///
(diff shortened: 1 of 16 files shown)


lib/saluki-components/src/transforms/trace_sampler/metrics.rs
```diff
@@ -0,0 +1,165 @@
...
+        let key = CounterKey {
+            sampler,
+            service: service.to_owned(),
+            env: env.to_owned(),
+            sampling_priority: if sampler == SamplerName::Priority {
+                sampling_priority
+            } else {
+                None
+            },
+        };
```
tobz, line 135: We don't want to allocate every single we make a sampling decision. 😵‍💫 

What we should do here is switch `self.counters` to be `FastIndexMap`, and then have a borrowed _and_ owned counter key, and implement `indexmap::Equivalent` between them.

This would look something like:

```rust
struct CounterKey {
   sampler: SamplerName,
    service: MetaString,
    env: MetaString,
    sampling_priority: Option<i32>,
}

struct BorrowedCounterKey<'a> {
    sampler: SamplerName,
    service: &'a str,
    env: &'a str,
    sampling_priority: Option<u32>,
}

impl BorrowedCounterKey<'_> {
    fn into_owned(self) -> CounterKey {
        CounterKey {
            sampler: self.sampler,
            service: self.service.into(),
            ...
        }
    }
}

impl<'a> indexmap::Equivalent<BorrowedCounterKey<'a>> for CounterKey {
    fn equivalent(&self, key: &BorrowedCounterKey<'a>) -> bool {
        self.sampler == key.sampler && ...
    }
}
```

With that, you would construct `BorrowedCounterKey<'a>` first and do a manual check for the entry, creating it if it didn't exist:

```rust
let (seen, kept) = match self.counters.get_mut(&borrowed_key) {
    Some((seen, kept)) => (seen, kept),
    None => {
        let key = borrowed_key.into_owned();
        self.counters.entry(key).or_insert_with_key(|key| {
            let seen = register_counter(&self.builder, METRIC_SEEN, &key);
            let kept = register_counter(&self.builder, METRIC_KEPT, &key);
            (seen, kept)
        })
    }
};
```

You might need to tweak some things, but essentially, the gist is that we defer allocating anything until we know we actually need to.

lib/saluki-components/src/transforms/trace_sampler/metrics.rs
```diff
@@ -0,0 +1,165 @@
...
+//! Per-sampler gauges (updated after each batch in `transform_buffer`):
+//! - `datadog.trace_agent.sampler.size` — current signature-table size for priority/no_priority/error samplers
+
+use metrics::{Counter, Gauge};
+use saluki_common::collections::FastHashMap;
+use saluki_metrics::MetricsBuilder;
+
+const METRIC_SEEN: &str = "datadog.trace_agent.sampler.seen";
+const METRIC_KEPT: &str = "datadog.trace_agent.sampler.kept";
+const METRIC_SIZE: &str = "datadog.trace_agent.sampler.size";
```
tobz, line 22: These should all just be `trace_sampler.*`.

We avoid Datadog Agent-specific telemetry naming in `saluki-*` crates where possible, and we _remap_ them in ADP ([here](https://github.com/DataDog/saluki/blob/main/bin/agent-data-plane/src/state/metrics/mod.rs#L283-L306), [here](https://github.com/DataDog/saluki/tree/main/bin/agent-data-plane/src/state/metrics/rules)) to their Agent-specific names.

tobz (changes requested):

lib/saluki-components/src/transforms/trace_sampler/metrics.rs
thieman: Just reiterating, I haven't reviewed this PR at all so there's probably quite a bit of slop in it, I'll move PRs out of draft to ready to review and post them in our Slack channel once I've reviewed and tweaked. Will incorporate your feedback just the same though, thanks.

PR #1344, opened 2026-04-10
fix(trace sampler): correctness test for probabilistic sampler + fixes

thieman:
## Summary

- Adds a new correctness test \`otlp-traces-probabilistic\` that verifies ADP makes identical keep/drop decisions as the DD Agent baseline when the APM-level probabilistic sampler is enabled at 50%
- Fixes \`_dd.p.dm\` tag placement: when \`probabilistic_sampler_enabled=true\`, the DD Agent writes this to **trace chunk metadata tags** only (not span meta); ADP was incorrectly writing it to span meta
- Fixes the SSI metadata check in \`ground-truth\` to skip both sides when \`otlp_direct_analysis_mode: true\`, since probabilistic sampling may legitimately drop the first-per-service spans that \`_dd.install.id\` is attached to
- Adds \`test-correctness-otlp-traces-probabilistic\` CI job so the new test runs on every PR pipeline

## DD Agent behavior references

- **Probabilistic sampler decision**: [\`pkg/trace/sampler/probabilistic_sampler.go\`](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/sampler/probabilistic_sampler.go) — deterministic \`fnv1a_32(trace_id) & 0x3FFF < rate * 0x4000\`
- **\`_dd.p.dm\` written to chunk tags (not span meta) for OTLP+probabilistic**: [\`pkg/trace/agent/agent.go#runSamplersV1\`](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/agent/agent.go) — the probabilistic sampler path calls \`setChunkAttributes\` which writes DM to chunk-level metadata, distinct from the span meta path used in legacy sampling
- **OTLP receiver pre-sampling (legacy path)**: [\`pkg/trace/api/otlp.go#L561-L585\`](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/api/otlp.go) — when \`ProbabilisticSamplerEnabled=false\`, the OTLP receiver pre-assigns \`priority=AutoKeep\` and \`dm=-9\` before the sampler runs; this path does write DM to span meta
- **\`otlp_direct_analysis_mode\`**: Used in other OTLP trace correctness tests because ADP computes APM stats post-sampling while the DD Agent computes them pre-sampling; stats comparison is intentionally skipped in this mode

## Test plan

- [x] \`make test-correctness-otlp-traces-probabilistic\` passes
- [x] \`make test-correctness-otlp-traces\` still passes (legacy OTLP path unaffected)
- [x] \`make test-correctness-otlp-traces-ets\` still passes
- [x] All \`trace_sampler\` unit tests pass (53/53)

🤖 Generated with [Claude Sonnet 4.6](https://claude.com/claude-code)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -534,7 +534,7 @@ test-all: test test-property test-docs test-miri test-loom
 
 .PHONY: test-correctness
 test-correctness: ## Runs the complete correctness suite
-test-correctness: test-correctness-dsd-plain test-correctness-dsd-origin-detection test-correctness-otlp-metrics test-correctness-otlp-traces test-correctness-otlp-traces-ets test-correctness-otlp-traces-ottl-filtering test-correctness-otlp-traces-ottl-transform
+test-correctness: test-correctness-dsd-plain test-correctness-dsd-origin-detection test-correctness-otlp-metrics test-correctness-otlp-traces test-correctness-otlp-traces-ets test-correctness-otlp-traces-ottl-filtering test-correctness-otlp-traces-ottl-transform test-correctness-otlp-traces-probabilistic
 
 .PHONY: test-correctness-dsd-plain
 test-correctness-dsd-plain: build-ground-truth
@@ -577,6 +577,12 @@ test-correctness-otlp-traces-ottl-transform: build-ground-truth
 test-correctness-otlp-traces-ottl-transform: ## Runs the 'otlp-traces-ottl-transform' E2E test (OTel Collector + OTTL transform vs ADP + OTTL transform)
 	@echo "[*] Running 'otlp-traces-ottl-transform' correctness test case..."
 	@target/release/ground-truth $(shell pwd)/test/correctness/otlp-traces-ottl-transform/config.yaml
+
+.PHONY: test-correctness-otlp-traces-probabilistic
+test-correctness-otlp-traces-probabilistic: build-ground-truth
+test-correctness-otlp-traces-probabilistic: ## Runs the 'otlp-traces-probabilistic' correctness test (probabilistic sampler at 50%)
+	@echo "[*] Running 'otlp-traces-probabilistic' correctness test case..."
+	@target/release/ground-truth $(shell pwd)/test/correctness/otlp-traces-probabilistic/config.yaml
 
 .PHONY: build-panoramic
 build-panoramic: check-rust-build-tools
--- a/bin/correctness/ground-truth/src/analysis/traces/mod.rs
+++ b/bin/correctness/ground-truth/src/analysis/traces/mod.rs
@@ -299,14 +299,17 @@ impl TracesAnalyzer {
         }
 
         // Ensure that we observe at least one span on each side where Single Step Instrumentation-related metadata is
-        // present (when not in OTLP-direct mode).
-        if !self.options.otlp_direct_analysis_mode && !self.baseline_ssi_metadata_present {
-            error!("No Single Step Instrumentation metadata found in baseline spans.");
-            error_count += 1;
-        }
-        if !self.comparison_ssi_metadata_present {
-            error!("No Single Step Instrumentation metadata found in comparison spans.");
-            error_count += 1;
+        // present (when not in OTLP-direct mode). In OTLP-direct mode, probabilistic sampling may legitimately drop
+        // the first-per-service spans that SSI metadata is attached to, so the check is skipped on both sides.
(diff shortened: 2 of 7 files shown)


bin/correctness/ground-truth/src/analysis/traces/mod.rs
```diff
@@ -299,14 +299,17 @@ impl TracesAnalyzer {
...
-        if !self.options.otlp_direct_analysis_mode && !self.baseline_ssi_metadata_present {
-            error!("No Single Step Instrumentation metadata found in baseline spans.");
-            error_count += 1;
-        }
-        if !self.comparison_ssi_metadata_present {
-            error!("No Single Step Instrumentation metadata found in comparison spans.");
-            error_count += 1;
+        // present (when not in OTLP-direct mode). In OTLP-direct mode, probabilistic sampling may legitimately drop
+        // the first-per-service spans that SSI metadata is attached to, so the check is skipped on both sides.
+        if !self.options.otlp_direct_analysis_mode {
```
thieman, line 303: This fix seems legitimate to me as it better aligns with what the comment was saying. Still a bit wary to change the tests rather than fixing the impl though, so check me please

lib/saluki-components/src/transforms/trace_sampler/mod.rs
```diff
@@ -464,7 +464,11 @@ impl TraceSampler {
         };
 
         let meta = root_span_value.meta_mut();
-        if priority > 0 {
+        // When the APM-level probabilistic sampler is used with OTLP traces, the DD Agent writes
+        // _dd.p.dm to trace chunk tags only (not span meta). For the legacy OTLP sampling path,
+        // it is written to both. We match that behavior by skipping the span meta write only when
+        // both conditions hold; the DM value still flows through TraceSampling to the encoder.
+        if priority > 0 && !(is_otlp && self.probabilistic_sampler_enabled) {
```
thieman, line 471: [Claude Sonnet 4.6] DDA references for this behavior:

- **Probabilistic path (chunk tags)**: when `ProbabilisticSamplerEnabled=true`, DDA writes `_dd.p.dm` to `pt.TraceChunk.Tags` (chunk-level), not span meta: https://github.com/DataDog/datadog-agent/blob/a90340956b9eaab7e5c16119cc45cfc408901d49/pkg/trace/agent/agent.go#L1090
- **Legacy OTLP pre-sampling path (span meta)**: when `ProbabilisticSamplerEnabled=false`, the OTLP receiver writes `_dd.p.dm` directly to `spans[0].Meta` via `traceutil.SetMeta`: https://github.com/DataDog/datadog-agent/blob/d01d85293503f17dad8c2da1b9f00ec0ccc5556e/pkg/trace/api/otlp.go#L568-L585

bin/correctness/ground-truth/src/analysis/traces/mod.rs
tobz: Yeah, no, I agree: OTLP direct mode is for testing against OpenTelemetry Collector, and so indeed it's always going to be missing from whatever side the OpenTelemetry Collector is running as, which will needlessly fail things overall.

Makefile
```diff
@@ -534,7 +534,7 @@ test-all: test test-property test-docs test-miri test-loom
 
 .PHONY: test-correctness
 test-correctness: ## Runs the complete correctness suite
-test-correctness: test-correctness-dsd-plain test-correctness-dsd-origin-detection test-correctness-otlp-metrics test-correctness-otlp-traces test-correctness-otlp-traces-ets test-correctness-otlp-traces-ottl-filtering test-correctness-otlp-traces-ottl-transform
+test-correctness: test-correctness-dsd-plain test-correctness-dsd-origin-detection test-correctness-otlp-metrics test-correctness-otlp-traces test-correctness-otlp-traces-ets test-correctness-otlp-traces-ottl-filtering test-correctness-otlp-traces-ottl-transform test-correctness-otlp-traces-probabilistic
```
tobz, line 537: Note to self: we might just want to update `ground-truth` to work like `panoramic`, where it can run these in parallel. Not only is this target just unwieldy, but it makes running them so much slower.

Like... it's just containers with their own network and what not, so there's no risk of cross-contamination or anything, we can just limit concurrency like `panoramic` does.

Anyways.

tobz (approved):

PR #1346, opened 2026-04-11
chore(common): add a helper type for human-friendly byte size parsing with `Facet` support

tobz:
## Summary

This PR adds a new helper type, `ByteSize`, for handling human-friendly byte size values in a more `Facet`-friendly way.

Currently, we use the `bytesize` crate, and that works fine... but as we push to move towards `facet`, we need all types used in our configuration to support `facet`. That's sometimes hard because it's fairly new, and `ByteSize` falls into that area. However, it's so simple that doing newtype wrappers and clunky trait impls to make it usable in `facet` felt weird... so I decided to just implement it directly.

This PR adds our own implementation of `ByteSize`. It has all the things we care about for it:

- holds the total number of bytes internally
- can be created from helper methods with either decimal or binary unit suffixes
- support for deserialization via `serde` or `facet`

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

New unit tests.

## References

AGTMETRICS-400

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -555,15 +555,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 
 [[package]]
-name = "bytesize"
-version = "2.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


tobz:
Closing this as we've shifted focus from trying to switch over to `facet` as of a few weeks ago.

PR #1358, opened 2026-04-13
enhancement(ground-truth): add parallel test runner

thieman:
## Summary

- Replaces all individual sequential Makefile targets with a single `test-correctness` target that discovers and runs all correctness test cases in parallel
- Adds `run-all` subcommand to `ground-truth` that mirrors how `panoramic` manages integration tests: uses `stream::iter` + `buffer_unordered` for semaphore-free concurrency limiting, with a default parallelism of 4
- Adds `run` subcommand to `ground-truth` for single test case invocation (replaces old bare positional arg); updates GitLab CI jobs accordingly
- New test cases are automatically picked up without any Makefile or CI changes

## Details

`ground-truth` previously took a single config file path as a positional argument. It now has two subcommands:
- `run <config.yaml>` — runs a single test case (existing behavior, used by GitLab CI jobs)
- `run-all -d <dir> [-p <parallelism>]` — discovers all `config.yaml` files in subdirectories and runs them in parallel (default parallelism: 4)

The Makefile gains:
- `make test-correctness` — runs the full suite via `run-all` (parallelism overridable with `GROUND_TRUTH_PARALLELISM=N`)
- `make test-correctness-case CASE=<name>` — escape hatch for a single named case

Motivated by tobz's note in #1344.

## Test plan

- [ ] `cargo build --package ground-truth` compiles cleanly
- [ ] `make test-correctness` runs all cases in parallel
- [ ] `make test-correctness GROUND_TRUTH_PARALLELISM=8` overrides parallelism
- [ ] `make test-correctness-case CASE=dsd-plain` runs a single case
- [ ] Adding a new test case directory is auto-discovered without Makefile/CI changes

### DIFF
--- a/.gitlab/e2e.yml
+++ b/.gitlab/e2e.yml
@@ -21,7 +21,7 @@
     GROUND_TRUTH_COMPARISON__IMAGE: ${SALUKI_IMAGE_REPO_BASE}/bundled-agent-adp:${CI_COMMIT_SHA}
   script:
     - make build-ground-truth
-    - target/release/ground-truth ./test/correctness/${CORRECTNESS_TEST_CASE}/config.yaml
+    - target/release/ground-truth run ./test/correctness/${CORRECTNESS_TEST_CASE}/config.yaml
 
 # Mixin for correctness tests where baseline is the CI-built bundled ADP image.
 # Tests with a different baseline (for example, DDOT) should NOT extend this and instead
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1504,6 +1504,7 @@ version = "0.1.0"
 dependencies = [
  "airlock",
  "argh",
+ "futures",
  "rand 0.9.3",
  "rand_distr",
  "reqwest",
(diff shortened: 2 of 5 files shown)


tobz (commented):
I like that this makes things go a little faster, but the output is a bit hard to comb through because it's all interleaved without any identifying information about which log lines belong to which test case, etc.

I do think we really ought to strive to model this more on how `panoramic` does it, where it's got a nice TUI mode that hides a lot of the details at info level but then has a non-TUI mode for CI. It would also be interesting to explore just merging the two together, so that `panoramic` runs everything, and we just have two classes of tests it supports (and the ability to add more in the future), etc...

thieman:
👍 I'll look more closely at the panoramic model

thieman:
[Claude Sonnet 4.6] Closing in favor of a fuller merge of ground-truth into panoramic.

thieman:
Tackled this here https://github.com/DataDog/saluki/pull/1379

PR #1375, opened 2026-04-14
chore(airlock): auto-detect non-standard Docker socket paths on macOS

### LINKED ISSUE #1374: Improve support for running containerized tests on macOS.
In multiple cases, users new to the Saluki repo have spent time debugging an issue that occurs because they are running Lima, or Colima, or, basically anything else, that doesn't put the docker socket at the canonical location. `airlock` and/or `ground-truth` expect the socket to be at `/var/run/docker.sock`.

One user has reported that setting `DOCKER_HOST` works, another has fixed it with a symlink, but let's try to make it "just work" the first time for non-standard socket locations.

webern:
This commit improves developer onboarding experience when using macOS as the developement environment by searching for the docker socket at common locations. Affects: airlock and panoramic

## Change Type
- [X] Bug fix
- [ ] New feature
- [X] Non-functional (chore, refactoring, docs)
- [ ] Performance

### Baseline

I unset DOCKER_HOST on my macOS machine (using Lima) and made sure there was no symlink (or anything else) at `/var/run/docker.sock`.

I ran this to build the images:

```bash
make build-datadog-intake-image build-millstone-image build-datadog-agent-image
```

Then, on a check-out that did not include the new changes, I ran

```bash
make test-correctness-dsd-plain
```

Result:

```text
2026-04-14T13:53:42.922776Z ERROR ground_truth: Failed to run test to completion.

Caused by:
    Socket not found: /var/run/docker.sock
make: *** [test-correctness-dsd-plain] Error 1
```

### With the Fix

`make test-correctness-dsd-plain` worked and passed.

### On Linux

Let's check CI!

## References

- Closes: #1374

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -180,6 +180,7 @@ crossbeam-queue = { version = "0.3", default-features = false, features = [
 float-cmp = { version = "0.10", default-features = false }
 tower-http = { version = "0.6", default-features = false }
 bollard = { version = "0.18", default-features = false }
+home = { version = "0.5", default-features = false }
 reqwest = { version = "0.13", default-features = false }
 lading-payload = { git = "https://github.com/DataDog/lading", rev = "bd60f9813012278a87a30516379805841c2e089f" }
 serde_yaml = { version = "0.9", default-features = false }
--- a/bin/correctness/airlock/src/docker.rs
+++ b/bin/correctness/airlock/src/docker.rs
@@ -1,0 +1,76 @@
+//! Docker daemon connection with non-standard socket discovery.
+
+use std::{env, path::PathBuf};
+
+use bollard::{Docker, API_DEFAULT_VERSION};
+use home::home_dir;
+use saluki_error::{ErrorContext as _, GenericError};
+use tracing::debug;
+
+/// Default connection timeout, in seconds, matching bollard.
+const DEFAULT_TIMEOUT: u64 = 120;
+
+/// The standard Docker socket path on Linux.
+const STANDARD_SOCKET: &str = "/var/run/docker.sock";
+
+/// Connect to the Docker daemon, searching non-standard socket locations on macOS.
+///
+/// Resolution order:
+///
+/// 1. If `DOCKER_HOST` is set **or** `/var/run/docker.sock` exists on disk, defer to bollard's
+///    default connection logic — no extra work needed.
+/// 2. Otherwise, try each candidate socket path in order and use the first that exists.
+/// 3. If no candidate is found, fall back to `Docker::connect_with_defaults()` so the error
+///    message comes from bollard rather than from us.
+///
+/// # Candidate paths
+///
+/// These are checked (in order) when the standard socket is absent:
+///
+/// | Path | Environment |
+/// |------|-------------|
+/// | `$HOME/.docker/run/docker.sock` | Docker Desktop |
+/// | `$HOME/.orbstack/run/docker.sock` | OrbStack |
+/// | `$HOME/.rd/docker.sock` | Rancher Desktop |
+/// | `$HOME/.lima/default/sock/docker.sock` | Lima (default instance) |
+/// | `$HOME/.lima/docker/sock/docker.sock` | Lima (docker instance) |
+/// | `$HOME/.colima/default/docker.sock` | Colima |
+/// | `$HOME/.colima/docker/docker.sock` | Colima (docker profile) |
+/// | `$HOME/.local/share/containers/podman/machine/podman.sock` | Podman Desktop (macOS) |
+///
+/// # Errors
+///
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -180,6 +180,7 @@ crossbeam-queue = { version = "0.3", default-features = false, features = [
 float-cmp = { version = "0.10", default-features = false }
 tower-http = { version = "0.6", default-features = false }
 bollard = { version = "0.18", default-features = false }
+home = "0.5"
```
tobz: Make this match our standard pattern of starting with default features disabled.

bin/correctness/airlock/src/docker.rs
```diff
@@ -0,0 +1,81 @@
...
+            if path.exists() {
+                let path = path.display();
+                debug!("using non-standard Docker socket: {path}");
+                return Docker::connect_with_unix(&path.to_string(), DEFAULT_TIMEOUT, API_DEFAULT_VERSION).map_err(
+                    |e| {
+                        GenericError::msg(format!(
+                            "failed to connect to Docker using socket discovered at {path}: {e}"
+                        ))
+                    },
+                );
```
tobz: The more idiomatic approach here would be -- assuming the error returned by `Docker::connect_with_unix` actually implements `std::error::Error` -- would be:

```suggestion
                return Docker::connect_with_unix(&path.to_string(), DEFAULT_TIMEOUT, API_DEFAULT_VERSION)
                    .with_error_context(|| format!("failed to connect to Docker using socket discovered at {path}"))?
```

This way we get an error _chain_ instead. You do have to import the `ErrorContext` trait for that, though: `use saluki_error::ErrorContext as _`.

tobz (changes requested):

bin/correctness/airlock/src/docker.rs
webern: Oops, sorry, kinda new to the claude thing. I didn't catch that.

bin/correctness/airlock/src/docker.rs
```diff
@@ -0,0 +1,76 @@
...
+///
+/// 1. If `DOCKER_HOST` is set **or** `/var/run/docker.sock` exists on disk, defer to bollard's
+///    default connection logic — no extra work needed.
+/// 2. Otherwise, try each candidate socket path in order and use the first that exists.
+/// 3. If no candidate is found, fall back to `Docker::connect_with_defaults()` so the error
+///    message comes from bollard rather than from us.
+///
+/// # Candidate paths
+///
+/// These are checked (in order) when the standard socket is absent:
```
thieman, line 28: is there an option to use `docker context` here? i think that should work without knowing the socket path ahead of time, and respect other user config wrt context choice rather than trying to guess and have our own precedence

webern: TBH, I didn't know about `docker context` when I came up with the quick fix idea. The idea here is to make it likely to just work on a developer's mac, but I suppose more complex setups would not? The use of DOCKER_HOST would override this behavior.

Anyway, wrt to `docker context`, the crate we are using, `bollard` doesn't support it. I don't think Lima sets up a docker context automatically on a machine so, maybe that wouldn't have solved the "just works" goal?

I'm not too familiar with this. If it seems important and useful we can look into it. But we just had three developers go down the troubleshooting path of not having their socket where bollard default's to so I decided to just find it with a minimal hotfix.

thieman: Ah ok I'm using `colima` not just Lima, my context seems legit:

```
❯ docker context ls
NAME            DESCRIPTION                               DOCKER ENDPOINT                                            ERROR
colima *        colima                                    unix:///Users/travis.thieman/.colima/default/docker.sock
default         Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
desktop-linux   Docker Desktop                            unix:///Users/travis.thieman/.docker/run/docker.sock
```

Fine to merge what you have here as an improvement on how it worked before, can always improve it more later.

webern: Cool, thanks. Google's AI said this:

> The Rust crate Bollard does not natively support Docker Contexts.Bollard is an asynchronous client that communicates directly with the Docker API via sockets or HTTP/SSL. Since Docker Contexts are a feature of the Docker CLI—stored in local configuration files (usually ~/.docker/config.json) to manage multiple daemon endpoints—they are transparent to API-level libraries like Bollard.

tobz: Yeah, I agree that this PR as-is is a meaningful step up, and if we decide we _need_ to support `docker context` for some reason, can always do it as a follow-up.

tobz (approved):

PR #1389, opened 2026-04-15
fix(agent-data-plane): fix `test_render_rar_telemetry` test

tobz:
## Summary

This PR fixes an issue with the `test_render_rar_telemetry` test in the `agent-data-plane` crate related to a bad assertion.

I realized that this test was failing sometimes when using Claude, and Claude was identifying correctly that it was a pre-existing failure so it was fine to ignore during those runs... but it was still broken. Why? Well, it turns out that we don't run tests defined in _binary_ crate, so we never ran it when doing `make test`, but if you ran the tests specifically for the `agent-data-plane` crate, it would run and inevitably fail.

We've updated the `test` and `test-property` targets to now include all binary crates as it searches for tests to run, which now causes this test (and others) to be run.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Ran `make test` locally to ensure the test is now being hit, and ensured it also _passed_ after applying the fix to the test itself.

## References

AGTMETRICS-400

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -502,13 +502,13 @@ sync-docs-config: ## Synchronizes the Vale configuration, updating configured st
 test: check-rust-build-tools cargo-install-cargo-nextest
 test: ## Runs all unit tests
 	@echo "[*] Running unit tests..."
-	cargo nextest run --lib -E 'not test(/property_test_*/)'
+	cargo nextest run --lib --bins --no-fail-fast -E 'not test(/property_test_*/)'
 
 .PHONY: test-property
 test-property: check-rust-build-tools cargo-install-cargo-nextest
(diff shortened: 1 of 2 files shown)


tobz:
* **#1389** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1389?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1389?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

aqian01 (approved):

Issue #1329, opened 2026-04-10
DogStatsD feature and configuration audit (tracking issue)


webern:
We want to introduce customers to the benefits of the `agent-data-plane` DogStatsD pipeline soon!

To that end, we are conducting an audit of configuration values and features that might still need to be implemented, investigated or modified. The following issues are a result of that audit.

## Documentation

- [x] #1416 <!-- documentation and skill -->

## Phase 1

- [x] #1645 <!-- cmd_port -->
- [x] #1640 <!-- provider_kind -->
- [x] #1641 <!-- use_api_v2 misclassified -->
- [x] #1331 <!-- bind_host -->
- [x] #1460 <!-- expected_tags_duration NOTE: not updated in the ledger when the bug was discovered -->
- [x] #1332 <!-- extra_tags -->
- [x] #1333 <!-- tags / DD_TAGS -->
- [x] #1334 <!-- use_dogstatsd -->
- [x] #1337 <!-- log_to_syslog -->
- [x] #1341 <!-- dogstatsd_so_rcvbuf -->
- [x] #1352 <!-- dogstatsd_metrics_stats_enable / dogstatsd_stats_enable -->
- [x] #1356 <!-- dogstatsd_logging_enabled / dogstatsd_log_file -->
- [x] #1371 <!-- skip_ssl_validation -->
- [x] #1380 <!-- logging_frequency / log verbosity controls -->
- [x] #1421 <!-- dogstatsd_buffer_size -->
- [x] #1422 <!-- dogstatsd_mapper_profiles -->
- [x] #1423 <!-- dogstatsd_non_local_traffic -->
- [x] #1424 <!-- dogstatsd_origin_detection -->
- [x] #1425 <!-- dogstatsd_packet_buffer_flush_timeout -->
- [x] #1426 <!-- dogstatsd_origin_detection_client -->
- [x] #1427 <!-- dogstatsd_packet_buffer_size -->
- [x] #1428 <!-- dogstatsd_pipeline_autoadjust -->
- [x] #1429 <!-- dogstatsd_socket -->
- [x] #1430 <!-- dogstatsd_tags -->
- [x] #1431 <!-- dogstatsd_port -->
- [x] #1432 <!-- dogstatsd_queue_size -->
- [x] #1433 <!-- statsd_metric_blocklist -->
- [x] #1434 <!-- statsd_metric_blocklist_match_prefix -->
- [x] #1435 <!-- dogstatsd_tag_cardinality -->
[...]

PR #1395, opened 2026-04-15
enhancement(aggregate): raise context limit warn on dropped metrics

### LINKED ISSUE #1330: Raise `aggregate_context_limit` default to prevent silent metric drops.
The `aggregate_context_limit` default is 5,000. The core agent has no explicit limit on unique contexts in the aggregator, though in practice an unbounded context count degrades performance there too. When ADP exceeds this limit, metrics are silently dropped.

Raise the default to a high but bounded value (e.g. 1,000,000). Log and emit telemetry when the limit is breached so operators can detect the condition before data is lost.

webern:
## Summary

Raise the default aggregate context limit from 5,000 to 1,000,000 and add logging to make context limit breaches
  observable.

Previously, when the aggregate transform hit its context limit, metrics were silently dropped with only a trace-level log per drop. This made it difficult to diagnose unexpected metric loss in production.

Now, a single warn-level message is emitted on the first dropped metric after the limit is reached, and an info-level recovery message is emitted when a flush brings the context count back below the limit. The logging uses a context_limit_breached flag on AggregationState that tracks whether we are currently in a breached state, ensuring we log exactly once per breach/recovery cycle rather than per dropped metric.

This was surfaced during a comparison audit of DogStatsD behavior in the core agent vs ADP. See #1329 for the list of issues that stemmed from that audit. This PR closes #1330

## Change Type
- [X] Bug fix (?)
- [X] New feature (?)
- [ ] Performance

## How did you test this PR?

I was able to add assertions to the existing unit tests in addition to the following manual test plan.

### Verified Existing Behavior

Before making code changes I build ADP and ran it in standalone mode setting the context limit to 10:

```bash
DD_DATA_PLANE_STANDALONE_MODE=true \
DD_API_KEY=api-key-adp-standalone \
DD_HOSTNAME=adp-standalone \
DD_DOGSTATSD_PORT=9191 \
DD_AGGREGATE_CONTEXT_LIMIT=10 \
DD_DATA_PLANE_TELEMETRY_ENABLED=true \
DD_DATA_PLANE_TELEMETRY_LISTEN_ADDR=tcp://127.0.0.1:5102 \
DD_IPC_CERT_FILE_PATH=/tmp/adp-ipc-cert.pem \
target/devel/agent-data-plane --config /tmp/adp-empty-config.yaml run
```

In another terminal I sent a loop of metrics that would exceed the context limit:

```bash
for i in {1..20}; do
  echo "test.metric.$i:1|c" | nc -u -w0 localhost 9191
done
```

I confirmed that the logs did not inform me about dropped metrics.

I then set the log level to trace for the affected component:

```
DD_LOG_LEVEL="info,saluki_components::transforms::aggregate=trace"
```

And did it all again. This, time I could see that, yes, metrics are being dropped:

```
2026-04-15 17:11:17 CEST | DATAPLANE | TRACE | (lib/saluki-components/src/transforms/aggregate/mod.rs:374) | Dropping metric due to context limit.
2026-04-15 17:11:17 CEST | DATAPLANE | TRACE | (lib/saluki-components/src/transforms/aggregate/mod.rs:374) | Dropping metric due to context limit.
2026-04-15 17:11:17 CEST | DATAPLANE | TRACE | (lib/saluki-components/src/transforms/aggregate/mod.rs:374) | Dropping metric due to context limit.
2
```

**This Verified**: the behavior we want to change. Namely you would need noisy trace logging otherwise you would not know about dropped metrics.

## Verifying the Changes

After making the code changes, (BTW I temporarily set the flush interval to 10 seconds in the code, setting it back before PR time)

I ran it again without setting the log level (i.e. defaulting to `info`):

```bash
DD_DATA_PLANE_STANDALONE_MODE=true \
DD_API_KEY=api-key-adp-standalone \
DD_HOSTNAME=adp-standalone \
DD_DOGSTATSD_PORT=9191 \
DD_AGGREGATE_CONTEXT_LIMIT=10 \
DD_DATA_PLANE_TELEMETRY_ENABLED=true \
DD_DATA_PLANE_TELEMETRY_LISTEN_ADDR=tcp://127.0.0.1:5102 \
DD_IPC_CERT_FILE_PATH=/tmp/adp-ipc-cert.pem \
target/devel/agent-data-plane --config /tmp/adp-empty-config.yaml run
```

I saw the desired `warn` and `info` logs without the `trace`-level logs:

```
2026-04-15 18:15:38 CEST | DATAPLANE | WARN | (lib/saluki-components/src/transforms/aggregate/mod.rs:374) | context_limit:10 | Context limit reached, dropping metrics. Consider increasing `aggregate_context_limit`.
2026-04-15 18:15:50 CEST | DATAPLANE | INFO | (lib/saluki-components/src/transforms/aggregate/mod.rs:321) | Context limit no longer exceeded, metrics are being accepted again.
```

## References

Closes #1330

### DIFF
--- a/lib/saluki-components/src/transforms/aggregate/mod.rs
+++ b/lib/saluki-components/src/transforms/aggregate/mod.rs
@@ -25,7 +25,7 @@ use tokio::{
     select,
     time::{interval, interval_at},
 };
-use tracing::{debug, error, trace};
+use tracing::{debug, error, info, trace, warn};
 
 mod telemetry;
 use self::telemetry::Telemetry;
@@ -44,7 +44,7 @@ const fn default_primary_flush_interval() -> Duration {
 }
 
 const fn default_context_limit() -> usize {
-    5000
+    1_000_000
 }
 
 const fn default_counter_expiry_seconds() -> Option<u64> {
@@ -108,7 +108,7 @@ pub struct AggregateConfiguration {
     /// When the maximum number of contexts is reached in the current aggregation window, additional metrics are dropped
     /// until the next window starts.
     ///
-    /// Defaults to 1000.
+    /// Defaults to 1,000,000.
     #[serde(rename = "aggregate_context_limit", default = "default_context_limit")]
     context_limit: usize,
 
@@ -306,12 +306,20 @@ impl Transform for Aggregate {
 
                         let should_flush_open_windows = final_primary_flush && self.flush_open_windows;
 
+                        // Remember if the context limit had been surpassed before this flush.
+                        let was_breached = self.state.context_limit_breached();
+
                         let mut dispatcher = context.dispatcher().buffered().expect("default output should always exist");
                         if let Err(e) = self.state.flush(get_unix_timestamp(), should_flush_open_windows, &mut dispatcher).await {
                             error!(error = %e, "Failed to flush aggregation state.");
                         }
 
                         self.telemetry.increment_flushes();
+
+                        // If flush recovered us from a breach, log the recovery.
+                        if was_breached && !self.state.context_limit_breached() {
+                            info!("Context limit no longer exceeded, metrics are being accepted again.");
+                        }
 
                         match dispatcher.flush().await {
                             Ok(aggregated_events) => debug!(aggregated_events, "Dispatched events."),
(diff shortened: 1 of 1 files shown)


tobz (approved):
Nice, makes sense to me. 👍🏻

PR #1399, opened 2026-04-15
feat(dogstatsd): enable ADP for dogstatsd with only data_plane.dogstatsd.enabled

jszwedko:
## Summary

Mirrors the change made in DataDog/datadog-agent#49409 for the saluki side.

- Previously, enabling Agent Data Plane for the dogstatsd workload required setting both `data_plane.enabled: true` **and** `data_plane.dogstatsd.enabled: true`.
- Now `data_plane.dogstatsd.enabled: true` alone is sufficient — `data_plane.enabled` no longer needs to be set explicitly.
- If `data_plane.enabled` is explicitly set to `false` (via config file or `DD_DATA_PLANE_ENABLED=false`), it still acts as a global kill switch and disables ADP regardless of per-feature flags.

### Before

```yaml
data_plane:
  enabled: true        # required
  dogstatsd:
    enabled: true
```

### After

```yaml
data_plane:
  dogstatsd:
    enabled: true      # sufficient on its own
```

## Implementation

The key change is in `DataPlaneConfiguration` in `config.rs`:

- `enabled` is now stored as `Option<bool>` instead of `bool`, preserving the distinction between "not configured" and "explicitly set to false".
- `enabled()` returns `true` when `data_plane.enabled` is absent and at least one data pipeline is enabled — mirroring the `IsConfigured` check in the Go implementation.
- `enabled()` still returns `false` immediately when `data_plane.enabled` is explicitly `false`.

### Why `Option<bool>`?

When ADP runs in **standalone mode** or when the **Core Agent config stream is disabled**, the full Agent configuration is not pushed to ADP, so `data_plane.enabled` may be absent. In that case, ADP infers enablement from whether any data pipeline flag is set. When the config stream is active, the Core Agent sends the full configuration and `data_plane.enabled` is always present, so the `Option` resolves to `Some` and is used directly.

No changes are needed in `run.rs` — the existing guard `if !in_standalone_mode && !dp_config.enabled()` works correctly with the updated method.

## Test plan

- [x] `test_dogstatsd_only_enables_adp` — `data_plane.dogstatsd.enabled: true` alone enables ADP
- [x] `test_global_disabled_overrides_dogstatsd` — `data_plane.enabled: false` disables ADP even when dogstatsd is enabled
- [x] `test_nothing_configured_not_enabled` — no config → ADP not enabled
- [x] `test_explicit_enabled_true_with_no_pipelines` — `data_plane.enabled: true` alone still enables ADP (existing behavior preserved)
- [x] New integration test case `dogstatsd-only-enabled` — verifies DSD starts with only `DD_DATA_PLANE_DOGSTATSD_ENABLED=true`

### DIFF
--- a/test/integration/cases/dogstatsd-only-enabled/config.yaml
+++ b/test/integration/cases/dogstatsd-only-enabled/config.yaml
@@ -1,0 +1,32 @@
+name: "dogstatsd-only-enabled"
+description: "Verifies DogStatsD pipeline starts with only data_plane.dogstatsd.enabled set (data_plane.enabled not required)"
+timeout: 60s
+
+container:
+  image: "saluki-images/datadog-agent:testing-devel"
+  env:
+    DD_API_KEY: "test-api-key"
+    DD_HOSTNAME: "integration-test-dsd-only"
+    # DD_DATA_PLANE_ENABLED is intentionally omitted to verify that DD_DATA_PLANE_DOGSTATSD_ENABLED
+    # alone is sufficient to enable ADP for DogStatsD.
+    DD_DATA_PLANE_STANDALONE_MODE: "true"
+    DD_DATA_PLANE_DOGSTATSD_ENABLED: "true"
+    DD_DOGSTATSD_PORT: "8125"
+    DD_DOGSTATSD_NON_LOCAL_TRAFFIC: "true"
+  exposed_ports:
+    - "8125/udp"
+
+assertions:
+  # Make sure the process becomes healthy, and stays up without errors, listening for DSD (UDP),
+  # for ~10 seconds after.
+  - parallel:
+      - type: process_stable_for
+        duration: 10s
+      - type: port_listening
+        port: 8125
+        protocol: udp
+        timeout: 10s
+      - type: log_not_contains
+        pattern: "panic|PANIC"
+        regex: true
+        during: 10s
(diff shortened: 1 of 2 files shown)


test/integration/cases/dogstatsd-only-enabled/config.yaml
```diff
@@ -0,0 +1,32 @@
...
+timeout: 60s
+
+container:
+  image: "saluki-images/datadog-agent:testing-devel"
+  env:
+    DD_API_KEY: "test-api-key"
+    DD_HOSTNAME: "integration-test-dsd-only"
+    # DD_DATA_PLANE_ENABLED is intentionally omitted to verify that DD_DATA_PLANE_DOGSTATSD_ENABLED
+    # alone is sufficient to enable ADP for DogStatsD.
+    DD_DATA_PLANE_STANDALONE_MODE: "true"
```
webern, line 12: Is standalone what you want? Do you want to create a second version with standalone set to false to get more coverage?

```text
    DD_DATA_PLANE_STANDALONE_MODE: "true"
```

webern (approved):

jszwedko:
Doh, I forgot I still had this open. We actually changed tact to, instead of enabling `data_plane.enabled` if any pipeline is enabled to instead default `data_plane.dogstatsd.enabled` to `true` so that only `data_plane.enabled` needs to be set. This was done in https://github.com/DataDog/saluki/pull/1474

PR #1401, opened 2026-04-16
chore(dogstatsd): properly handle `dogstatsd_string_interner_size` counting entries instead of bytes

### LINKED ISSUE #1367: Rename `dogstatsd_string_interner_size` to avoid semantic clash with the core agent.
The core agent and ADP both use the key name `dogstatsd_string_interner_size` for the string interner capacity, but they interpret the value differently.

| | Core agent | ADP |
|---|---|---|
| Key | [`dogstatsd_string_interner_size`][a1] | [`dogstatsd_string_interner_size`][s1] |
| Default | `4096` | `2 MiB` |
| Unit | **entry count** ([compared against `len(map)`][a2]) | **byte size** ([typed as `ByteSize`][s1]) |

A customer migrating with `dogstatsd_string_interner_size: 4096` would get 4096 entries in the core agent but 4096 bytes (4 KiB) in ADP — an undersized interner. The key should be renamed on the ADP side to avoid the clash and the differences documented.

[a1]: https://github.com/DataDog/datadog-agent/blob/141f1206279/pkg/config/setup/common_settings.go#L1607
[a2]: https://github.com/DataDog/datadog-agent/blob/141f1206279/comp/dogstatsd/server/intern.go#L63
[s1]: https://github.com/DataDog/saluki/blob/e9373ef98013/lib/saluki-components/src/sources/dogstatsd/mod.rs#L246-L250

tobz:
## Summary

This PR solves an issue with `dogstatsd_string_interner_size` representing the _number_ of interner entries to limit to, rather than the number of bytes total.

In ADP, we treat `dogstatsd_string_interner_size` as a byte size limit for the string interner, while in the Core Agent, it's treated as an entry count limit. While the byte size interpretation is the superior one, we need to ensure we can coexist with the Core Agent, and so must interpret it in this way.

This PR updates the DogStatsD source configuration to support a new configuration field -- `dogstatsd_string_interner_size_bytes` -- which specifically supports byte sizes while `dogstatsd_string_interner_size` is changed to indicate the number of entries. We preferentially use `dogstatsd_string_interner_size_bytes` when set, falling back to `dogstatsd_string_interner_size` otherwise. In order to derive an actual byte size when using `dogstatsd_string_interner_size`, we simply multiply the number by a fixed amount -- 512 bytes -- which leads to the default of 4096 for `dogstatsd_string_interner_size` giving us 2 MiB overall.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

New unit tests.

## References

DADP-11

Closes #1367.

### DIFF
--- a/bin/agent-data-plane/src/sizing_guide_template.html
+++ b/bin/agent-data-plane/src/sizing_guide_template.html
@@ -104,7 +104,7 @@
 
                 // Set the string interner size to a linear interpolation between 2MB and 96MB at 5000 and 250k cardinality, respectively
                 document.getElementById(
-                    "dsd_in.dogstatsd_string_interner_size",
+                    "dsd_in.dogstatsd_string_interner_size_bytes",
                 ).value = Math.max(
                     2097152,
                     Math.floor(
--- a/lib/saluki-components/src/sources/dogstatsd/mod.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/mod.rs
@@ -107,9 +107,15 @@ const fn default_no_aggregation_pipeline_support() -> bool {
     true
 }
 
-const fn default_context_string_interner_size() -> ByteSize {
-    ByteSize::mib(2)
+const fn default_context_string_interner_entry_count() -> u64 {
+    4096
 }
+
+/// Baseline byte cost per interner entry, used to convert the Core Agent's entry-count-based
+/// `dogstatsd_string_interner_size` to a byte size.
+///
+/// 4096 entries × 512 bytes = 2 MiB, matching ADP's previous default.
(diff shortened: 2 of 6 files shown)


aqian01 (commented):
https://github.com/DataDog/saluki/blob/tobz/dsd-string-interner-size-1367/test/smp/regression/adp/experiments.yaml#L353

Think we should change this to 
```
DD_DOGSTATSD_STRING_INTERNER_SIZE_BYTES: 32MiB
```

Theres a few more

Image

tobz:
Ah yeah, good call. Will fix. 👍🏻

aqian01 (approved):

PR #1413, opened 2026-04-17
enhancement(dogstatsd): implement `bind_host` config

### LINKED ISSUE #1331: Read `bind_host` from config instead of hardcoding the listen address.
ADP hardcodes the DogStatsD UDP listen address to localhost. The core agent exposes `bind_host` in configuration, allowing users to bind to a different interface (e.g. `0.0.0.0`).

Read `bind_host` from config instead of hardcoding.

webern:
## Summary

  Reads `bind_host` from DogStatsD config, matching core agent precedence: `dogstatsd_non_local_traffic=true` →
  `0.0.0.0`, else `bind_host` if set, else `127.0.0.1`. `bind_host` accepts either an IP literal or a hostname
  (resolved via `tokio::net::lookup_host`). Unresolvable values surface an error with the underlying `io::Error`
  attached.

  ## Change Type
  - [X] New feature

  ## How did you test this PR?

Four new tests, red→green (two fail pre-fix). Two use only static config. The other two need the container's
runtime IP to assert a meaningful listener binding — without it, the tests can't distinguish "bind_host was read
from config" from "the hardcoded default happened to match 127.0.0.1/0.0.0.0." That's why the PR also adds a
templating mechanism to panoramic for injecting runtime values into both ADP config and assertion patterns. 

Those panoramic commits are isolated.

  ```bash
  # build
  make build-datadog-agent-image build-panoramic

  # run the baseline pre-existing test
  ./target/release/panoramic run -d test/integration/cases --tests basic-startup

  # run the new tests. two were red/green
  ./target/release/panoramic run -d test/integration/cases --tests dogstatsd-bind-host
  ./target/release/panoramic run -d test/integration/cases --tests dogstatsd-default-bind
  ./target/release/panoramic run -d test/integration/cases --tests dogstatsd-non-local-overrides-bind-host
  ./target/release/panoramic run -d test/integration/cases --tests dogstatsd-bind-custom-hostname
  ```

  ## References

  Closes #1331

### DIFF
--- a/lib/saluki-components/src/sources/dogstatsd/mod.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/mod.rs
@@ -81,6 +81,12 @@ enum Error {
 
     #[snafu(display("No listeners configured. Please specify a port (`dogstatsd_port`) or a socket path (`dogstatsd_socket` or `dogstatsd_stream_socket`) to enable a listener."))]
     NoListenersConfigured,
+
+    #[snafu(display("Could not resolve bind_host '{}': {}", host, source))]
+    UnresolvableBindHost { host: String, source: std::io::Error },
+
+    #[snafu(display("bind_host '{}' resolved to zero IP addresses.", host))]
+    BindHostHasNoAddresses { host: String },
 }
 
 const fn default_buffer_size() -> usize {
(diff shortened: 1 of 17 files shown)


lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -81,6 +81,9 @@ enum Error {
 
     #[snafu(display("No listeners configured. Please specify a port (`dogstatsd_port`) or a socket path (`dogstatsd_socket` or `dogstatsd_stream_socket`) to enable a listener."))]
     NoListenersConfigured,
+
+    #[snafu(display("Invalid bind_host '{}': not a valid IP address.", address))]
+    InvalidBindHost { address: String },
```
webern: Oh wait, could something other than an IP address be valid? 🤔

tobz (approved):

PR #1414, opened 2026-04-17
chore(ci): optimize size/usage of pre-built CI images

tobz:
## Summary

As stated in the PR title, really.

Every CI job starts off by pulling the configured container image, which means bigger images on a cold runner node (image not cached) will take longer before the actual job itself can start.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

The images build cleanly in CI, and seem to work for dependent jobs. 🤞🏻 

## References

DADP-11

### DIFF
--- a/.ci/images/build/Dockerfile
+++ b/.ci/images/build/Dockerfile
@@ -55,10 +55,14 @@ RUN CI=true make cargo-preinstall && \
     rm -rf /root/.cargo/registry /tmp/Makefile /tmp/bin
 
 # Final sweep: remove anything leftover in /tmp, apt metadata, docs/man pages, and ephemeral caches.
+# The nightly toolchain's rust-std (in lib/rustlib/*/lib) is only used when compiling Rust code;
+# since we only use nightly to run rustfmt, we can drop it. rustfmt itself still links dynamically
+# against libLLVM / librustc_driver in lib/, so only the rustlib subtree is safe to remove.
 RUN find /tmp -mindepth 1 -delete && \
     rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*.deb && \
     rm -rf /root/.cache && \
     rm -rf /root/.cargo/registry && \
+    rm -rf /root/.rustup/toolchains/nightly-*/lib/rustlib/*/lib && \
     rm -rf /usr/share/doc/* /usr/share/man/* /usr/share/locale/*
 
 COPY .ci/images/build/entrypoint.sh /
--- a/.ci/images/build/Dockerfile
+++ b/.ci/images/build/Dockerfile
@@ -1,54 +1,65 @@
-FROM registry.ddbuild.io/images/docker:27.3.1
+FROM registry.ddbuild.io/images/base/gbi-ubuntu_2204:release
 
 ARG RUST_VERSION=1.93.0
 ARG TARGETARCH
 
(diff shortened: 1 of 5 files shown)


.ci/images/build/Dockerfile
```diff
@@ -1,54 +1,69 @@
...
-# Install Rust and common Cargo tooling that we depend on.
+# Install Rust and common Cargo tooling that we depend on. Nightly is installed with a minimal profile
+# since we only use it to run rustfmt with Edition 2024.
 RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --profile minimal --default-toolchain ${RUST_VERSION}
 ENV PATH="/root/.cargo/bin:${PATH}"
-RUN rustup toolchain add nightly && \
-    rustup component add clippy && \
-    rustup component add --toolchain nightly rustfmt
+RUN rustup component add clippy && \
+    rustup toolchain install nightly --profile minimal --component rustfmt
```
aqian01, line 46: https://github.com/DataDog/saluki/blob/tobz/optimize-build-ci-image/.gitlab/fuzz.yml#L14
`- cargo +nightly install cargo-fuzz`

Since we also use it to compile the fuzz testing, does the minimal nightly cause an issue?

.ci/images/smp/Dockerfile
```diff
@@ -1,26 +1,54 @@
...
+
+# Install pr-commenter for posting analysis reports back to GitHub.
+COPY .ci/install-pr-commenter.sh /
 RUN chmod +x /install-pr-commenter.sh && /install-pr-commenter.sh
 
-# Build and install bloaty for binary size analysis.
-COPY .ci/images/smp/install-bloaty.sh /
-RUN chmod +x /install-bloaty.sh && /install-bloaty.sh
+# Install AWS CLI for retrieiving credentials and settings for the SMP image registry location.
+COPY .ci/install-awscli.sh /
```
aqian01, line 47: The script `/install-awscli.sh` only installs the cli for x86, do we want other architectures as well?

```
install_awscli() {
    local url="https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip"
```

aqian01 (approved):

.ci/images/build/Dockerfile
tobz: The minimal profile just limits what components get installed by default via Rustup to the _bare_ minimum: `rustc`, the stdlib, and Cargo, and then we're additionally adding `rustfmt`... so that's still perfectly sufficient for the fuzz target binaries to get compiled.

To be sure, though, I'll manually trigger the `fuzz:infra` job.

.ci/images/smp/Dockerfile
tobz: This is true, but we only run the jobs in `benchmark.yml` on AMD64 runners.

tobz: Although saying that makes me realize... we still technically build the SMP CI image for `linux/arm64`, even if we never end up using that version. Kind of a footgun... I'll add it as a follow-up issue to address just to avoid someone else stumbling over that in the future.

.ci/images/build/Dockerfile
tobz: For posterity, the `fuzz:infra` job still compiles the fuzz target binaries correctly.

PR #1419, opened 2026-04-20
chore(ci): build/install `buildcache` in build/SMP CI images

tobz:
## Summary

This PR builds and installs [`buildcache`](https://gitlab.com/bits-n-bites/buildcache), an advanced build caching tool, in our build and SMP CI images.

The goal of `buildcache` is to allow for cross-build caching of build artifacts, like individual crates, so that we can speed up builds across CI jobs, whether in the same CI pipeline or across them. Pretty straightforward.

This PR focuses solely on getting `buildcache` into our CI images, but not at all on actually configuring it since we're still working on the supporting bits for those changes (the remote backend to actually store the cached objects in, etc). Also of note is that we're building this from a fork that includes support for accessing S3 over HTTPS. We intend to build from upstream if/when the change gets merged.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Will test by running the image builds in CI.

## References

DADP-11

### DIFF
--- a/.ci/images/build/Dockerfile
+++ b/.ci/images/build/Dockerfile
@@ -1,3 +1,18 @@
+# Builder stage: compile buildcache from source. The final image already ships cmake/g++/ninja
+# (needed by Rust C/C++ build scripts), but keeping this in a separate stage prevents the cloned
+# source tree and CMake build directory from landing in the final image layers.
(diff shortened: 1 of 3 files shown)


aqian01 (approved):

webern (approved):
LGTM

PR #1436, opened 2026-04-20
enhancement(ci): build zstd-compressed CI helper images

tobz:
## Summary

This PR adds the ability to generate CI helper images that are compressed via Zstandard in an attempt to improve the image pull times for our CI runners.

We've added new build jobs specifically for Zstandard which skip flattening the image as flattening and then compressing involves a long chain of commands to get things in the right shape. We've also changed the behavior of the build jobs to not automatically push the images via the `latest` tag, but instead move that to a new manually-triggered job to allow for building the images in CI to test without inherently promoting them.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Will test in CI.

## References

DADP-11

### DIFF
--- a/.gitlab/internal.yml
+++ b/.gitlab/internal.yml
@@ -28,12 +28,75 @@
       .
     # Now flatten the image, realizing any space savings from duplicate layers, cleaned up cruft, etc.
     - crane flatten --tag ${IMAGE_REF} ${IMAGE_REF}
-    # Tag this as `latest` so we can immediately take advatange of it.
-    - crane tag ${IMAGE_REF} latest
     # Recalculate the image digest post-flattening and update the build metadata so we can sign the flattened image.
     - NEW_DIGEST=$(crane digest ${IMAGE_REF})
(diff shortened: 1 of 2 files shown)


.gitlab/internal.yml
```diff
@@ -28,13 +28,76 @@
...
+  id_tokens:
+    DDSIGN_ID_TOKEN:
+      aud: image-integrity
+  needs: []
+  rules:
+    - if: $CI_PIPELINE_SOURCE == "web"
+      when: manual
+      allow_failure: true
+    - if: $CI_PIPELINE_SOURCE == "schedule" && $BUILD_HELPER_IMAGES == "true"
+      allow_failure: true
```
webern, line 52: Allow the scheduled run to fail?

webern (approved):
😎

.gitlab/internal.yml
tobz: It just lets these individual jobs fail without causing the overall build to fail, and without short-circuiting subsequent jobs in the stage... so we could still manage to update/publish a passing image build even if another one failed.

PR #1440, opened 2026-04-22
chore(otlp): add support for OTLP semantic mappings

tobz:
## Summary

This PR adds support for mapping OTLP attribute "concepts" to their relevant attributes with the ability to specify priority/ordering to more cleanly handle looking up specific things (for example, HTTP status code) regardless of the specific semantic convention version/scheme used to generate it.

To quote the upstream Agent PR that introduced this support (https://github.com/DataDog/datadog-agent/pull/46222):

> This PR introduces a **semantics library** (`pkg/trace/semantics`) for OTLP trace processing and migrates OTel-to-DD attribute lookups to use it where appropriate. The library provides a registry of semantic concepts and their equivalent attribute keys across OTel versions and conventions, so lookups can follow a single precedence order (e.g. old key then new key, span then resource) instead of ad-hoc key lists.

Essentially, prior to this, finding a specific "concept" -- such as the HTTP status code -- might involve checking for multiple attribute names or types: look for `http.status_code` as an integer _or_ string, then try looking for `http.response.status_code` as an integer _or_ string. Extrapolated out against 40-50 of these "concepts", that's a lot of potential code -- code that is specific to that concept -- for looking up all of those concepts. This PR (and the upstream PR it's based on) make these "concepts" a first-class citizen by defining them centrally and then tying them to a definition of all of the underlying attributes that they can be resolved from in priority order. Looking up these concepts then simply becomes a matter of loading the relevant attribute mappings and trying them in order... which can be hidden behind generic and reusable lookup helpers.

This PR implements all of the same machinery present in the Agent PR, but notably does not yet switch over all concept-style lookups to use it. We've migrated only the usages necessary to get our correctness tests passing again to try and scope down the size of the PR, and will do the remainder in a follow-up PR. Similarly, we have a stacked PR on top of this one (#1441) which actually updates the Agent version used to 7.78.0, where the upstream Agent PR is present, to demonstrate that this PR actually fixes our correctness tests when implemented.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Existing unit tests + stacked PR using Agent 7.78.0.

The way to view this:
- see that in _this_ PR, our OTLP traces-related correctness tests are failing (because we're changing how we process certain attributes to match 7.78.0)
- see that in #1441, by simply switching to testing against Agent 7.78.0, those tests now pass when based on _this_ PR

## References

DADP-11


tobz:
* **#1441** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1441?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1440** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1440?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1440?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

aqian01 (approved):
LGTM

PR #1457, opened 2026-04-22
feat(components): config key registry and smoke tests

thieman:
## Goal

Configuration keys in the Datadog Agent are often misconfigured in the field because the connection between YAML paths, environment variables, and the Rust structs that consume them is implicit and untested. A key can be renamed, an env var alias can be dropped, or a new field can be added to a struct without any test catching the drift.

This PR introduces a programmatic config key registry and a smoke test runner that makes these connections explicit and verifiable. Every config struct that directly consumes YAML/env var configuration now has a test that will fail if a key stops working, if an env var stops mapping to the right field, or if a new field is added without registering an annotation for it.

### Config key registry (`lib/saluki-components/src/config_registry/`)

A hand-written catalog of configuration keys, one \`SalukiAnnotation\` constant per key. Each annotation records:
- **Canonical YAML path** (from the vendored Agent schema)
- **Environment variables** that deliver this value
- **Value type** (String, Bool, Integer, Float, StringList)
- **Schema default** (pulled from the vendored YAML via codegen — used to auto-flip Bool test values when the generic test value would match the default and leave the struct unchanged)
- **Which config structs consume this key** (\`used_by\`)

The generated \`schema.rs\` is produced by \`cargo xtask gen-config-schema\` from the vendored \`core_schema.yaml\`, and now includes \`default\` values alongside the existing \`yaml_path\`, \`env_vars\`, and \`value_type\` fields. Hand-written ADP-specific keys (not in the Agent schema) live alongside the generated ones as custom \`SchemaEntry\` statics with the same structure.

### \`run_config_smoke_tests\` runner (`test_support.rs`)

A single async function that drives three verifications for any config struct \`T: PartialEq + Serialize\`:

1. **Supported keys**: For each annotation registered to the struct, loads it via the canonical \`yaml_path\` and then independently via each declared env var. All must produce identical structs, and each must differ from the default (unset-config) struct.

2. **Unsupported keys**: For every other annotation in \`ALL_ANNOTATIONS\`, asserts the struct is unaffected — catches the case where a key leaks across struct boundaries.

3. **Full field coverage**: Loads the struct with all supported keys set simultaneously and verifies every serialized leaf field differs from the default. Catches fields that are wired up in the struct but have no registered annotation, or annotations with the wrong YAML path.

A \`base_config\` parameter supplies any fields that are required for deserialization (e.g. \`api_key\` for \`ForwarderConfiguration\`). A \`non_config_fields\` slice excludes runtime-injected fields from the coverage check (e.g. the \`nanos\` sub-fields of Duration values when only whole-second injection is practical).

### Escape hatches

Two optional annotation fields handle edge cases cleanly without requiring changes to production types:

- \`value_type_override\`: For fields where the Agent schema declares \`Float\` but the Rust type is an integer (\`i32\`, \`u64\`, etc.), overriding to \`Integer\` injects \`42\` instead of \`1.5\`.
- \`test_json\`: For fields requiring structured input (JSON arrays, objects, or specific enum strings) that the generic type-derived test values cannot satisfy (e.g. mapper profiles, tag cardinality, Duration objects).

### Annotation invariants test

A separate \`annotation_invariants\` test asserts that \`Full\`/\`Partial\` annotations have a non-empty \`used_by\` list and \`Incompatible\` annotations have an empty one — catching support-level miscategorisations at the registry level.

### Mutation testing

Verified that meaningful mutations are caught:

| Mutation | Caught by |
|---|---|
| Remove struct from \`used_by\` | Coverage check + unsupported-key check |
| Wrong \`yaml_path\` | Coverage check |
| Add struct to \`used_by\` that doesn't use the key | Supported-key check |
| \`Incompatible\` with non-empty \`used_by\` | \`annotation_invariants\` test |
| \`Incompatible\` + empty \`used_by\` (orphans key) | Coverage check |
| Remove annotation entirely | Coverage check |

## Struct coverage

All 16 top-level config structs that directly call \`config.as_typed()\` in a \`from_configuration\` method are covered. The field count is the number of registered annotations exercising that struct.

| Struct | Crate location | Fields tested |
|---|---|---|
| \`ProxyConfiguration\` | \`common::datadog::proxy\` | 5 |
| \`ForwarderConfiguration\` | \`common::datadog::config\` | 23 |
| \`DogStatsDConfiguration\` | \`sources::dogstatsd\` | 25 |
| \`OtlpConfiguration\` | \`sources::otlp\` | 17 |
| \`AggregateConfiguration\` | \`transforms::aggregate\` | 10 |
| \`DogStatsDMapperConfiguration\` | \`transforms::dogstatsd_mapper\` | 2 |
| \`DogStatsDPrefixFilterConfiguration\` | \`transforms::dogstatsd_prefix_filter\` | 6 |
| \`DatadogMetricsConfiguration\` | \`encoders::datadog::metrics\` | 4 |
| \`DatadogTraceConfiguration\` | \`encoders::datadog::traces\` | 4 |
| \`DatadogLogsConfiguration\` | \`encoders::datadog::logs\` | 2 |
| \`DatadogEventsConfiguration\` | \`encoders::datadog::events\` | 2 |
| \`DatadogServiceChecksConfiguration\` | \`encoders::datadog::service_checks\` | 2 |
| \`DatadogApmStatsEncoderConfiguration\` | \`encoders::datadog::stats\` | 2 |
| \`OtlpDecoderConfiguration\` | \`decoders::otlp\` | 6 |
| \`OtlpRelayConfiguration\` | \`relays::otlp\` | 5 |
| \`TraceObfuscationConfiguration\` | \`transforms::trace_obfuscation\` | 25 |
| **Total** | | **140** |

## Test plan

- [x] \`cargo test -p saluki-components -- config_smoke\` — all 16 smoke tests pass
- [x] \`cargo test -p saluki-components -- annotation_invariants\` — registry invariants pass
- [x] Mutation testing confirms all three verification checks and the invariants test catch meaningful annotation errors


thieman:
## Smoke test failure modes

The runner detects four distinct failure categories. Here's what each looks like in practice, using `ProxyConfiguration` as the example.

---

### 1. yaml_path not wired up

The key is registered and the test tries to load it, but the struct field has a wrong (or missing) `serde(rename)` so the value never reaches the field.

```
config smoke tests for 'ProxyConfiguration' failed with 2 error(s):

  [1] yaml_path 'proxy.http': struct did not change from its default —
      is the test value the same as the default, or is the key not wired up?

  [2] 1 serialized field(s) are never changed by any registered config key: [proxy_http_BROKEN]
      Fix: register a ConfigKey for each field and add it to the test set.
      Fix: if a field is intentionally not config-driven (e.g. injected at runtime),
      add its serialized name to the `non_config_fields` slice in this test call.
```

Both errors fire together: the key changed nothing (error 1), and the struct now has an orphaned field name that nothing exercises (error 2).

---

### 2. env var name wrong

The yaml_path works, but one of the `env_vars` entries is misspelled — so loading via that env var produces a different (or default) struct.

```
config smoke tests for 'ProxyConfiguration' failed with 1 error(s):

  [1] env var 'HTTP_PROXYYY' produced a different struct than yaml_path 'proxy.http'
```

---

### 3. Struct field with no registered ConfigKey

A field exists in the struct and is deserialized from config, but no `ConfigKey` was registered for it (i.e. it was added to the struct without being added to the registry or the test set).

```
config smoke tests for 'ProxyConfiguration' failed with 1 error(s):

  [1] 1 serialized field(s) are never changed by any registered config key: [sneaky_unregistered_field]
      Fix: register a ConfigKey for each field and add it to the test set.
      Fix: if a field is intentionally not config-driven (e.g. injected at runtime),
      add its serialized name to the `non_config_fields` slice in this test call.
```

If the field is genuinely not config-driven (e.g. injected at runtime), the fix is to pass its serialized name in the `non_config_fields` slice:

```rust
run_config_smoke_tests(structs::FOO, foo_reg::ALL, &["workload_provider"], |cfg| { ... }).await
```

---

### 4. Key registered in ALL_KEYS but missing from the test call

A key has `used_by` pointing at this struct (so it appears in `ALL_KEYS`) but was not included in the `keys` slice passed to the runner. This fires as an immediate panic before any config loading, since it's a test setup error.

```
key 'proxy.https' is registered for 'ProxyConfiguration' in ALL_KEYS but is missing from the test set
```

tobz (changes requested):
Can we just fix all of the intradoc link errors? Otherwise, API docs are gonna blow up when we try to generate them. 😅

thieman:
Tracked in DADP-64 (https://datadoghq.atlassian.net/browse/DADP-64) — ASTA Preview blocker.

lib/saluki-components/src/config_registry/datadog/aggregate.rs
```diff
@@ -0,0 +1,158 @@
+//! Annotations for the aggregate transform configuration keys.
+use crate::config_registry::{generated::schema, structs, SalukiAnnotation, SchemaEntry, SupportLevel, ValueType};
+
+// ADP-specific keys not present in the vendored Agent schema.
```
thieman, line 4: Disagreements between the configuration schema and ADP are called out in a few different ways:

1. Adding SchemaEntrys manually like we do on this line
2. Any annotations with `additional_yaml_paths` or `env_var_override`
3. Annotations with `value_type_override`, in particular the schema has everything as a `number` but ADP needs to know the difference between floats and ints

I'll bring these up with the configuration team.

lib/saluki-components/src/config_registry/datadog/dogstatsd.rs
```diff
@@ -0,0 +1,358 @@
...
+    };
+
+    /// `dogstatsd_port` — UDP port. Schema says Float; field is u16.
+    DOGSTATSD_PORT = SalukiAnnotation {
+        schema: &schema::DOGSTATSD_PORT,
+        support_level: SupportLevel::Full,
+        additional_yaml_paths: &[],
+        env_var_override: None,
+        used_by: &[structs::DOGSTATSD_CONFIGURATION],
+        value_type_override: Some(ValueType::Integer),
```
thieman, line 174: Example `number` to `Integer` value type override, turns out ports can't be floats

lib/saluki-components/src/config_registry/datadog/dogstatsd.rs
```diff
@@ -0,0 +1,368 @@
...
+// The Agent schema defines these as nested properties under an `enable_payloads` section:
+//   enable_payloads.events, enable_payloads.series, etc.
+// (see ENABLE_PAYLOADS_EVENTS / ENABLE_PAYLOADS_SERIES etc. in generated/schema.rs)
+//
+// Saluki's DogStatsDConfiguration uses flat underscore-separated field names with no `#[serde(rename)]`,
+// so the yaml_path for deserialization is `enable_payloads_events` (not `enable_payloads.events`).
+// The two naming conventions are incompatible at the config-loader level — setting the dotted path
+// produces a nested object that the flat field cannot consume. Custom statics with the flat paths
+// are required until the struct field naming is aligned with the Agent schema.
+// TODO: https://github.com/DataDog/saluki/issues — align enable_payloads field names with schema
```
thieman: (1st one) Highlighting specific deviation from the schemas here that may be a bug on the ADP side, need to investigate further

lib/saluki-components/src/config_registry/datadog/trace_obfuscation.rs
```diff
@@ -0,0 +1,223 @@
...
+//! See the constants `APM_CONFIG_OBFUSCATION_*` in `generated/schema.rs`.
+//!
+//! However, [`TraceObfuscationConfiguration`] contains a `config: ObfuscationConfig` field,
+//! so `cfg.as_typed::<TraceObfuscationConfiguration>()` reads from `config.*` paths, not
+//! `apm_config.obfuscation.*`. Custom statics with the `config.*` prefix are required for
+//! the smoke tests to exercise the right yaml paths.
+//!
+//! TODO: evaluate whether TraceObfuscationConfiguration should be wired to read from
+//! `apm_config.obfuscation.*` directly (like `from_apm_configuration` does in production),
+//! and whether the smoke test should use that path instead.
```
thieman, line 16: (2nd one) Highlighting specific deviation from the schemas here that may be a bug on the ADP side, need to investigate further

lib/saluki-components/src/config_registry/datadog/dogstatsd_prefix_filter.rs
```diff
@@ -0,0 +1,91 @@
...
+// ADP renamed the field to `statsd_metric_namespace_blocklist` for inclusive language, but the
+// struct has NO serde alias for the old spelling. This means Agent config files that set
+// `statsd_metric_namespace_blacklist` are silently ignored by ADP — the customized list is
+// dropped and ADP falls back to the hardcoded default. The default lists happen to be identical
+// today, so this only affects users who have customized the key in their Agent config.
+//
+// Fix: add `#[serde(alias = "statsd_metric_namespace_blacklist")]` to the struct field in
+// `transforms/dogstatsd_prefix_filter/mod.rs` and switch this static to reference the generated
+// STATSD_METRIC_NAMESPACE_BLACKLIST schema entry with the old key as additional_yaml_paths.
+// TODO: https://github.com/DataDog/saluki/issues — add serde alias for blacklist → blocklist
```
thieman: (3rd one) Highlighting specific deviation from the schemas here that may be a bug on the ADP side, need to investigate further

lib/saluki-components/src/config_registry/generated/mod.rs
```diff
@@ -0,0 +1,6 @@
+// @generated — DO NOT EDIT
+// Regenerate with: cargo xtask gen-config-schema
+
+#![allow(missing_docs)]
+
+pub mod schema;
```
thieman: So you don't have to crash your browser, the members inside `schema.rs` are `SchemaEntry` that look like:

```
pub const APM_CONFIG_CLIENT_STATS_FLUSH_INTERVAL: SchemaEntry = SchemaEntry {
      yaml_path: "apm_config.client_stats_flush_interval",
      env_vars: &["DD_APM_CLIENT_STATS_FLUSH_INTERVAL"],
      value_type: ValueType::Float,
      default: Some("1"),
  };

  pub const APM_CONFIG_COMPUTE_STATS_BY_SPAN_KIND: SchemaEntry = SchemaEntry {
      yaml_path: "apm_config.compute_stats_by_span_kind",
      env_vars: &["DD_APM_COMPUTE_STATS_BY_SPAN_KIND"],
      value_type: ValueType::Bool,
      default: Some("true"),
  };
```

lib/saluki-components/src/config_registry/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+    pub const DATADOG_APM_STATS_ENCODER_CONFIGURATION: &str = "DatadogApmStatsEncoderConfiguration";
+    /// Identifier for [`crate::decoders::otlp::OtlpDecoderConfiguration`].
+    pub const OTLP_DECODER_CONFIGURATION: &str = "OtlpDecoderConfiguration";
+    /// Identifier for [`crate::relays::otlp::OtlpRelayConfiguration`].
+    pub const OTLP_RELAY_CONFIGURATION: &str = "OtlpRelayConfiguration";
+    /// Identifier for [`crate::transforms::trace_obfuscation::TraceObfuscationConfiguration`].
+    pub const TRACE_OBFUSCATION_CONFIGURATION: &str = "TraceObfuscationConfiguration";
+}
+
+/// How well saluki supports a given configuration key.
```
thieman, line 95: Idea is to flesh out this functionality a bit more as part of runtime configuration analysis. For now anything that makes it into any of our config structs is support `Full`

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -352,6 +353,16 @@ pub struct DogStatsDConfiguration {
     additional_tags: Vec<String>,
 }
 
+#[cfg(test)]
+impl PartialEq for DogStatsDConfiguration {
+    fn eq(&self, other: &Self) -> bool {
+        // `workload_provider: Option<Arc<dyn WorkloadProvider>>` is #[serde(skip)] and trait
+        // objects can't implement PartialEq. JSON comparison naturally excludes skip fields
+        // and stays correct as new fields are added.
+        serde_json::to_value(self).ok() == serde_json::to_value(other).ok()
```
thieman: There are three config structs we couldn't derive PartialEq for so we drop down to JSON comparison, I figure this should work? Could also do a custom PartialEq implementation with a bunch of field comparisons but I was worried that'd drift

xtask/src/main.rs
```diff
@@ -0,0 +1,17 @@
+use std::env;
```
thieman: Not sure if we had any existing patterns for either code gen or separate tooling binaries, happy to change this up to whatever you'd like

lib/saluki-components/build.rs
```diff
@@ -0,0 +1,268 @@
+use std::fmt::Write as _;
```
thieman, line 1: Planning on adding some Makefile targets for the code generation and probably a Claude skill to handle updating and reconciling a new schema version with the config registry

xtask/src/main.rs
webern: When I was working with Claude on a config gen related design before, I found that it was obsessed with `xtast` which isn't a concept I had come across before. I think `build.rs` can be powerful solution because it just integrates fully with `cargo build`, `cargo test`, etc injecting itself into the build graph. `build.rs` is pretty powerful and canonical for code gen.

thieman: Done, that's much nicer, thanks. The generated Rust code for the schemas is no longer committed.

lib/saluki-components/src/sources/dogstatsd/mod.rs
thieman: Tried alternative approach here, adds `derivative` as a dev-dependency so we can exclude certain fields from PartialEq https://github.com/DataDog/saluki/pull/1457/commits/2d7f32d961ca46659dfd187521c1d4e09e390526

lib/saluki-components/src/common/datadog/config.rs
```diff
@@ -205,3 +206,23 @@ mod tests {
...
+    use super::ForwarderConfiguration;
+    use crate::config_registry::structs;
+    use crate::config_registry::test_support::run_config_smoke_tests;
+
+    #[tokio::test]
+    async fn smoke_test() {
+        run_config_smoke_tests(
+            structs::FORWARDER_CONFIGURATION,
+            &[],
+            json!({ "api_key": "smoke-test-api-key" }),
```
tobz, line 226: Hmm, what's the point of setting this?

lib/saluki-components/src/config_registry/datadog/dogstatsd.rs
tobz: Should this actually be an issue that we open so it can be referred to here? 😅

tobz (approved):

lib/saluki-components/src/common/datadog/config.rs
thieman: If `api_key` isn't provided then deserialization panics, added an explanatory comment above this line https://github.com/DataDog/saluki/pull/1457/commits/8006656f6ace47a3f570acaeb4b5cc1a5c7ab505

lib/saluki-components/src/config_registry/datadog/dogstatsd.rs
thieman: Created this issue and pointed the three sets of known deviations to it https://github.com/DataDog/saluki/issues/1480

PR #1462, opened 2026-04-23
chore(ci): use `buildcache` for ADP builds in CI

tobz:
## Summary

This PR introduces `buildcache` to the actual ADP build jobs in CI as a way to try and amortize the cost of compiling slowly-changing dependencies and thereby speed up/smooth out the build time for repeated CI pipeline runs.

As a brief refresh, `buildcache` is a build caching tool, invoked as a wrapper around the compiler (`rustc`) which will use a cached version of a given dependency if available, which lets us avoid having to recompile it. Applied to all dependencies in the build, many of which hardly change between CI pipeline runs, and you can often achieve high cache hit rates and reduce the amount of work done overall during building.

This PR introduces a number of pieces to make this possible:

- a new `buildcache` CI image that we use to amortize the cost of building/compiling `buildcache` itself (we pull the `buildcache` binary from this image as part of the ADP container image build process)
- updating our `general` CI image and install script for the AWS CLI to make sure it's available for multi-arch image build scenarios
- adding a new helper script for generating temporary AWS credentials, and updating our ADP build jobs to use the `general` CI image to generate those credentials before invoking the ADP container image build so that `buildcache` has access to our shared S3 cache bucket
- updates to `Dockerfile.agent-data-plane` to optionally use `buildcache` as a wrapper around `rustc` (when configured), activating only in CI and otherwise not being used for local image builds

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Non-functional (chore, refactoring, docs)
- [ ] Performance

## How did you test this PR?

- [x] Ensured that all `build-adp-image*` Make targets built cleanly when run locally
- [x] Ensured all build jobs in CI passed

## References

DADP-11

PR #1463, opened 2026-04-23
chore(ci): split more of ADP docker build into clean-vs-prepped stages

tobz:
## Summary

This PR chunks up more of the Dockerfile for ADP to allow for using more fine-grained, pre-built images in lieu of doing things from scratch.

Our Dockerfile for ADP is designed to work seamlessly between local development _and_ CI, but we also want to prioritize CI because, well, we probably run 50-100x more ADP builds via CI than we do locally. Reducing that overhead is meaningful, and helps tighten the CI loop. To this end, we want to amortize the cost of repeated steps that don't often (or ever) change from one CI pipeline to another.

This PR introduces additional "clean vs prepped" stages to further amortize things, namely with license building/collection and the additional of profiling helpers. We've started adopting a "clean vs prepped" approach, where for chunks of work in the Dockerfile, we model them as a stage where we run all the steps on-demand (local development), _or_ providing an input image that already ran the exact same steps and we can just use it immediately as if we the layer was already cached (CI). In particular, we've applied this treatment to the "license" and "profiling" stage.

The license stage involves download the reference collection of SPDX licenses, and then extracting the ones that are listed in our third-party license file, and then including them in the final ADP image. Ironically, doing this from scratch can take about 20-30s because we have to install `curl` to get the license tarball. These files almost never change, and so it makes a lot of sense to just stick them in a slim image and avoid the whole `apt-get install curl` preamble where possible.

Similarly, the profiling stage involves pulling down `ddprof`, and adding a special wrapper script to run ADP via `ddprof`, and including them in builds that require profiling, like SMP. We didn't quite do the full "clean vs prepped" approach, but we've separated this into its own stage so that it can happen in parallel rather than having to wait until the very end after the ADP build finishes.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

- [x] Ensured that all `build-adp-image*` Make targets ran cleanly when used locally
- [x] Ensured that all ADP build jobs in CI ran cleanly 

## References

DADP-11

### DIFF
--- a/.ci/images/spdx-licenses/Dockerfile
+++ b/.ci/images/spdx-licenses/Dockerfile
@@ -1,0 +1,15 @@
+# Builder stage: download the SPDX license text archive and pull the text license files out of it.
+FROM registry.ddbuild.io/images/mirror/ubuntu:24.04 AS builder
+
+ENV DEBIAN_FRONTEND=noninteractive \
+    TZ=Etc/UTC
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends ca-certificates curl
+
+RUN curl -s -L -o /tmp/spdx.tar.gz https://github.com/spdx/license-list-data/archive/refs/tags/v3.28.0.tar.gz && \
+    tar -C /tmp -xzf /tmp/spdx.tar.gz && \
+    mv /tmp/license-list-data-3.28.0 /tmp/spdx-licenses
+
+FROM registry.ddbuild.io/images/mirror/ubuntu:24.04 AS final
+COPY --from=builder /tmp/spdx-licenses/text /tmp/spdx-licenses/text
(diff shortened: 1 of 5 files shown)


tobz:
* **#1463** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1463?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1463?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1462** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1462?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

aqian01 (approved):

PR #1471, opened 2026-04-24
chore(ci): optimize build CI image further + converge on zstd-compressed CI images

tobz:
## Summary

As stated in the PR title.

We're doing two main things here:

- switch completely to zstd-compressed CI images (they work just fine, no reason to build both and have all the extra jobs/boilerplate)
- optimize and reduce the layer count of the build CI image

For the build CI image, we've shrunk it even further and, while doing so, captured more of the logic in new scripts which should hopefully make the Dockerfile much easier to read... isolating the various individual commands to distinct and intuitively namedscripts.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Manually built the CI images and ensured that they built correctly and that they worked for an actual CI pipeline run.

## References

DADP-11


webern (approved):
Looks great and well organized. Pre-approving even though a problem needs to be fixed before merging:

```
$ docker pull ${SALUKI_IMAGE_REPO_BASE}/bundled-agent-adp:${CI_COMMIT_SHA}
/scripts-6743-1626885429/step_script: line 227: docker: command not found
```

tobz:
> Looks great and well organized.

Thanks! :)

> Pre-approving even though a problem needs to be fixed before merging:

Yeah, was looking into this one but I'm confused because pulling the `latest` tag manually (and I even double checked based on the image digest itself) shows `docker` being available on `$PATH` but in `test-integration`, it's not agreeing with me... 🤔

EDIT: Actually, maybe my local setup was lying to me about `latest` being the right one :suspect:

PR #1461, opened 2026-04-23
docs(dogstatsd): reclassify expected_tags_duration as divergent

webern:
Update the dogstatsd configuration document and ledger to reflect a bug and issue with expected_tags_duration.

## Summary

If I end up fixing this immediately then maybe we won't need to merge this.

## Change Type
- X Non-functional (chore, refactoring, docs)

## How did you test this PR?

Just need to read it.

## References

Found while working on #1332.
Related to #1460

PR #1472, opened 2026-04-24
fix(config): correctly parse `expected_tags_duration`

### LINKED ISSUE #1460: Implement `expected_tags_duration` time string parsing (e.g. `30s`) to match the Agent.
I think I found a bug. The core agent declares `expected_tags_duration` as a Go
`time.Duration` and viper accepts duration strings via `time.ParseDuration`. ADP reads
it as a bare `u64`, so `30s` is rejected at startup.

## Repro

`datadog.yaml`:
```yaml
expected_tags_duration: 30s
```

ADP log:
```
2026-04-23 17:36:44 UTC | DATAPLANE | ERROR | (bin/agent-data-plane/src/main.rs:100) | Expected value for field 'expected_tags_duration' to be 'u64', got 'string "30s"' instead.
```

## References

- ADP reader (`u64`): https://github.com/DataDog/saluki/blob/main/lib/saluki-components/src/transforms/host_tags/mod.rs#L31-L39
- Core agent declaration (`time.Duration`): https://github.com/DataDog/datadog-agent/blob/1fa422e0ed9dee6da49b02da0947a1a6fb570b0e/pkg/config/setup/common_settings.go#L1217
- Viper string-to-duration path: https://github.com/DataDog/cast/blob/v1.8.0/caste.go#L79-L85

A related subtlety: a bare integer like `5` means 5 nanoseconds in the core agent but
5 seconds in ADP, so numeric values copied from core agent configs behave differently
even when ADP does not error.

## Suggested action

Accept the same duration-string format as the core agent, or document the deviation.

webern:
Add DurationString, a config value type that accepts the duration formats the Agent produces via viper/cast: strings with Go time-unit suffixes ("30s", "1h30m"), bare numeric strings (nanoseconds), and integer or float scalars (nanoseconds). Wire it into host_tags for expected_tags_duration, which rejected "30s" by reading the key as u64. Bare integers now mean nanoseconds, matching the Agent; ADP had been interpreting them as seconds. Adds correctness and integration tests.

## Change Type
- [X] Bug fix

## How did you test this PR?

I used two tests (not included in the PR), one correctness and one integration, both of which failed before the fix. The assumptions themselves were rigorously researched and re-researched with AI-assistance including a verification that we are not internally using this config currently (i.e. we are not relying on the incorrect behavior)

## References

Closes #1460.
Invalidates/Closes PR #1461


bin/correctness/panoramic/src/config.rs
```diff
@@ -408,11 +335,10 @@ pub fn discover_tests(dirs: &[PathBuf]) -> Result<Vec<DiscoveredTest>, GenericEr
...
                         Err(e) => {
-                            tracing::warn!(
-                                path = %config_path.display(),
-                                error = %e,
-                                "Failed to load test case, skipping"
-                            );
+                            // Previously we had a warning here that cannot be seen in TUI-mode. It is better to fail
+                            // loudly and fast when we have a bad test configuration than to falsely believe our test is
+                            // working when we see that all tests passed.
+                            panic!("Failed to load test case, bad configuration: {e}");
```
jszwedko, line 414: 👍

jszwedko (approved):
I did find myself wondering if we should rely on the Core Agent to normalize these values rather than need to maintain a parallel implementation of Go's duration string representation to avoid future regressions, but the Go implementation is unlikely to change so this seems ok to have a copy of it.

lib/saluki-config/src/duration_string.rs
```diff
@@ -0,0 +1,742 @@
...
+            serialized == actual_serialized,
+            "Expected the input '{input}' to be serialized as '{serialized}' but got '{actual_serialized}'"
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn duration_string_success_01() {
+        run_success_case("0", Duration::default(), "0s0ns").unwrap();
+    }
```
tobz: Can you collapse these all down to a single test method? The sheer number of functions here makes it hard to actually sort through and see the relevant cases being tested. 😅

lib/saluki-config/src/duration_string.rs
```diff
@@ -0,0 +1,742 @@
...
+                    "Expected the error message when parsing '{input}' to contain {expected_msg:?}, but the message is {e}"
+                );
+            }
+        }
+
+        Ok(())
+    }
+
+    #[test]
+    fn duration_string_failure_01() {
```
tobz: Same comment here about collapsing.

lib/saluki-config/src/duration_string.rs
```diff
@@ -0,0 +1,742 @@
...
+/// A configuration value like `expected_tags_duration: 30` means 30 **nanoseconds**, not 30 seconds. Use `"30s"` for
+/// 30 seconds. This matches the Agent's `time.Duration` coercion.
+///
+/// # Serialization
+///
+/// Serializes as `"{seconds}s{nanoseconds}ns"`. For example, 30 seconds becomes `"30s0ns"` and 30.5 seconds becomes
+/// `"30s500000000ns"`. Whole seconds are maximized and the nanosecond component is always less than `1_000_000_000`,
+/// so the form is unambiguous and round-trips through this parser and would be accepted by the Agent as well.
+#[derive(Clone, Copy, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct DurationString(Duration);
```
tobz, line 49: Can we actually turn this into a deserialization helper, as done in `PermissiveBool`? That would mean moving this to `saluki_common::deser` and then slightly tweaking it to be usable with `serde_with` instead of using `DurationString` directly as the type in the various spots where we need this.

test/correctness/dsd-expected-tags-duration/datadog.yaml
```diff
@@ -0,0 +1,26 @@
...
+dogstatsd_origin_detection: true
+
+# Gauges can be processed out-of-order when multiple workers are used, while ADP does not use multiple workers, so ADP
+# always ends up with the correct (last seen) value, while DSD might return the last seen value... or the value seen
+# four updates ago, etc etc.
+dogstatsd_workers_count: 1
+
+# 10s is a valid value for this key, but prior to fixing #1460 we would have errored on this value because we expected
+# an integer. This test proves that we accept duration strings.
+expected_tags_duration: 10s
```
tobz: I think this is the wrong way to think about this test.

Testing whether or not we can deserialize `10s` is doable with a unit test. If we're going to keep it, what this correctness test should strive to show is that we actually apply host tags in the same way as the Agent... which should be reflected in the name of the test, like `dsd-initial-host-tags` or something.

test/integration/cases/dogstatsd-expected-tags-duration/config.yaml
```diff
@@ -0,0 +1,37 @@
...
+  env:
+    DD_API_KEY: "test-api-key"
+    DD_HOSTNAME: "integration-test-dsd"
+    DD_DATA_PLANE_ENABLED: "true"
+    DD_DATA_PLANE_STANDALONE_MODE: "false"
+    DD_DATA_PLANE_DOGSTATSD_ENABLED: "true"
+    DD_DOGSTATSD_PORT: "8125"
+    DD_DOGSTATSD_NON_LOCAL_TRAFFIC: "true"
+    # This is the test input. It is a valid DurationString that would not be accepted before issue #1460 was fixed.
+    DD_EXPECTED_TAGS_DURATION: "0m13ms"
```
tobz: Overkill. We're already testing this in a correctness test _and_ unit tests.

tobz (changes requested):

lib/saluki-config/src/duration_string.rs
tobz: Chatted with @webern about this and we're going to leave it as-is for now, because switching it to a deserialization helper would require contortions in some of the places it's used that would make the complexity reduction a wash.

webern:
Unfortunately the [push](https://github.com/DataDog/saluki/compare/fbfc23b82a8ce45feaf1289a963bae35e6ff7901..5c2d8b802497a73d3724cc8715fc9795b9e66532) included a force with rebase, but the [commit](https://github.com/DataDog/saluki/pull/1472/commits/5c2d8b802497a73d3724cc8715fc9795b9e66532) does have the isolated changes requested.

tobz (approved):

PR #1473, opened 2026-04-24
feat(dogstatsd): added capture and replay functionality

lucastemb:
## Summary

Adds DogStatsD capture and replay support to ADP, modeled after the Go Agent `dogstatsd-capture` / `dogstatsd-replay` flow.

This PR adds:
- the ability to start DogStatsD traffic captures through the control-plane API and CLI
- persistence of captured UDS packets to Datadog capture files
- replay tagger state in the capture trailer
- support for reading `.dog` and zstd-compressed capture files
- replay back into DogStatsD with original packet cadence and synthetic UDS credentials, so captured PID-based origin enrichment can be restored during replay

## Change Type
- [ ] Bug fix
- [X] New feature
- [ ] Performance

## How did you test this PR?

Unit tests and a custom Python/bash script. 

```bash
 cargo test -p saluki-components dogstatsd::replay
 cargo test -p saluki-components dogstatsd::origin
 cargo test -p saluki-components dogstatsd::tests::
 cargo test -p saluki-io unix
 cargo test -p agent-data-plane dogstatsd
 cargo test -p agent-data-plane control_plane
```

## References
 Related to the Go Agent DogStatsD capture/replay implementation used as the parity reference.


bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -49,6 +52,48 @@ struct StatsCommand {
...
+    1
+}
+
+/// Starts a DogStatsD traffic capture.
+#[derive(FromArgs, Debug)]
+#[argh(subcommand, name = "capture")]
+struct CaptureCommand {
+    /// how long the traffic capture should run for, using Go-style duration syntax such as `10s` or `1m0s`
+    #[argh(option, short = 'd', long = "duration", default = "default_capture_duration()")]
+    capture_duration: String,
```
webern: After I merge #1472 please rebase and use `DurationString` for this.

```suggestion
    capture_duration: DurationString,
```

lucastemb:
@codex Review

lib/saluki-components/src/sources/dogstatsd/replay/writer.rs
```diff
@@ -0,0 +1,837 @@
...
+    } else {
+        Ok(CaptureFileWriter::Plain(writer))
+    }
+}
+
+fn run_capture_loop(
+    state: Arc<Mutex<WriterState>>, traffic_rx: Receiver<CaptureRecord>, mut writer: CaptureFileWriter,
+    duration: Duration, workload_provider: Option<Arc<dyn WorkloadProvider + Send + Sync>>,
+) {
+    let mut pid_map = HashMap::new();
```
aqian01: Should we use our custom type FastHashMap instead?

lib/saluki-components/src/sources/dogstatsd/replay/injector.rs
```diff
@@ -0,0 +1,225 @@
...
+fn replay_process_credentials(captured_pid: i32) -> Result<ProcessCredentials, GenericError> {
+    let captured_pid =
+        u32::try_from(captured_pid).map_err(|_| generic_error!("Replay packet contained a negative captured PID."))?;
+    let replay_pid = i32::try_from(std::process::id())
+        .map_err(|_| generic_error!("Current process ID did not fit in the replay credential format."))?;
+
+    Ok(ProcessCredentials {
+        pid: replay_pid,
+        uid: captured_pid,
+        gid: REPLAY_CREDENTIALS_GID,
```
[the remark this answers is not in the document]
lucastemb: `dogstatsd replay` being run by a non-root user  is not something that we expect to encounter in practice. ADP follows the Core Agent's design. 

The auth token file would only be readable by root (or dd-agent) anyways. A non-root user would be stopped much earlier at the auth token read step.

lib/saluki-components/src/sources/dogstatsd/replay/reader.rs
```diff
@@ -0,0 +1,234 @@
...
+                e
+            )
+        })?;
+
+        Ok(Some(state))
+    }
+}
+
+fn read_capture_contents(path: &Path) -> Result<Vec<u8>, GenericError> {
+    let raw_contents = fs::read(path)
```
aqian01: Non blocking. 

How big are capture files?

`    let raw_contents = fs::read(path)`

We load the entire capture file into memory here and i'm wondering if streaming it instead would be more performant?

As in we used a buffered reader and process one record at a time?

Same with this line, we decompress the entire file into memory

```
    let contents = zstd::stream::decode_all(Cursor::new(raw_contents.as_slice())).map_err(|e| {
```

Codex says the downsides of a buffered reader approach are
```
rewind() becomes fallible and slightly slower because it must seek/reopen.
read_state() needs seek logic to read the trailer.
```
So not too sure if it's worth it.

aqian01 (commented):
Looking good, left a nit/question for now but ill take a better look tmrw

lib/saluki-components/src/sources/dogstatsd/replay/reader.rs
lucastemb: Hm interesting. That's actually a good point. This is how the Core Agent reader works, but I think you're right there is room for a bit of optimization. 

Currently, for `.dog.zstd` files we have two buffers one for compressed and another for the decompressed bytes, but for our intents and purposes we only need the decompressed bytes.

After a bit of rubber ducking with Claude and Codex, it seems like way to reduce peak memory usage is to decode the `capture.dog.zstd` file **directly** and store the contents in only one buffer as opposed to two.

I'm not convinced that the pros outweigh the cons for a full streaming solution though.

lib/saluki-components/src/sources/dogstatsd/replay/file.rs
```diff
@@ -0,0 +1,151 @@
+#![allow(dead_code)]
```
webern, line 1: I'm wondering if `file_header.rs` is a better name.

I'm assuming the dead-code allowance this is left-over from previous working states. It might help agents a bit to put doc comments in models

```suggestion
//! Provides parsing and writing of the header and version information of a Datadog replay file.
```

lib/saluki-components/src/sources/dogstatsd/replay/file.rs
```diff
@@ -0,0 +1,151 @@
...
+pub(super) fn file_version(buf: &[u8]) -> Result<u8, GenericError> {
+    if !datadog_matcher(buf) {
+        return Err(generic_error!(
+            "Cannot verify file version from invalid capture header."
+        ));
+    }
+
+    let version = 0xF0 ^ buf[VERSION_INDEX];
+    if version > DATADOG_FILE_VERSION {
+        return Err(generic_error!("Unsupported capture file version: {}.", version));
```
webern, line 50: Does this implementation support older versions? i.e. is `>` correct vs `==`?

lib/saluki-components/src/sources/dogstatsd/replay/state.rs
```diff
@@ -0,0 +1,445 @@
...
+    origin::OriginTagCardinality,
+    tags::{SharedTagSet, Tag, TagSet},
+};
+use saluki_env::workload::EntityId;
+use saluki_error::{generic_error, GenericError};
+use stringtheory::MetaString;
+use tracing::warn;
+
+#[derive(Clone, Default)]
+struct ReplayEntityTags {
```
webern, line 18: What is a replay entity tag?

```suggestion
/// A tag blah blah
#[derive(Clone, Default)]
struct ReplayEntityTags {
```

lib/saluki-components/src/sources/dogstatsd/replay/state.rs
```diff
@@ -0,0 +1,445 @@
...
+                }
+
+                saw_tags.then_some(merged)
+            }
+        }
+    }
+}
+
+#[derive(Clone)]
+struct LoadedReplayState {
```
webern, line 97:
```suggestion
/// The inner state of `DogStatsDReplayState` which it wraps with a mutex guard.
#[derive(Clone)]
struct ReplayStateInner {
```

lib/saluki-components/src/sources/dogstatsd/replay/state.rs
```diff
@@ -0,0 +1,445 @@
...
+impl DogStatsDReplayState {
+    /// Creates a new, empty replay-state store.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Returns whether replay state is currently loaded.
+    pub fn is_loaded(&self) -> bool {
+        self.with_live_state(|_| ()).is_some()
+    }
```
webern, line 128: Constructing an unusable object is a bit of an anti-pattern. Can't tell right now, but is there a reason you can't construct it in a usable state?

```suggestion
impl DogStatsDReplayState {
    /// Creates a new, ready, replay-state store.
    pub async fn new(path: impl AsRef<Path>) -> Result<Self> {
        Self{
            inner: Arc::new(RwLock::new(load_the_thing(path.as_ref()).await?))
        }
    }
```

Hmmm, well Claude is telling me that it's because the lifecycle is that this is created at startup. I suppose you could have `Option<DogStatsDReplayState>` there. But... oh well.

lib/saluki-components/src/sources/dogstatsd/replay/state.rs
```diff
@@ -0,0 +1,445 @@
...
+
+    /// Loads replay state into memory.
+    ///
+    /// If both the saved entity map and PID map are empty, the existing replay state is cleared and `false` is
+    /// returned. This preserves the Go "empty request clears replay state" behavior while still allowing pid-map-only
+    /// captures to load successfully.
+    pub fn load(&self, replay_state: TaggerState) -> Result<bool, GenericError> {
```
webern, line 135: This is weird. I guess it's mirroring the Go implementation?

lib/saluki-components/src/sources/dogstatsd/replay/state.rs
```diff
@@ -0,0 +1,445 @@
...
+
+fn shared_tags_from_strings(tags: Vec<String>) -> Option<SharedTagSet> {
+    if tags.is_empty() {
+        return None;
+    }
+
+    Some(TagSet::from_iter(tags.into_iter().map(Tag::from)).into_shared())
+}
+
+fn replay_expiry_window(duration_ms: i64) -> Duration {
```
webern, line 241: Document rationale. Later when future someone is fixing a bug related to how long the expiry_window is, it's helpful to know that twice duration was just an arbitrary guess at what a safe value would be.

lib/saluki-components/src/sources/dogstatsd/replay/state.rs
```diff
@@ -0,0 +1,445 @@
...
+        Ok(true)
+    }
+
+    /// Clears any loaded replay state from memory.
+    pub fn clear(&self) {
+        let mut state = self.inner.write().expect("dogstatsd replay state lock poisoned");
+        *state = None;
+    }
+
+    pub(crate) fn get_tags_for_entity(
```
webern, line 196: I think it can be helpful to LLMs and humans alike to keep visibility as tight as possible. It's an enforced sort of documentation: this function is designed for this module only, not to be called by other modules.

```suggestion
    /// Gets the tags for a workload entity (container, pod, or global).
    pub(super) fn get_tags_for_entity(
```

lib/saluki-components/src/sources/dogstatsd/replay/state.rs
```diff
@@ -0,0 +1,445 @@
...
+
+            loaded_state
+                .entity_tags
+                .get(resolved_entity)
+                .and_then(|entity_tags| entity_tags.merge_for_cardinality(cardinality))
+        })
+        .flatten()
+    }
+
+    pub(crate) fn resolve_container_entity_for_pid(&self, process_id: u32) -> Option<EntityId> {
```
webern, line 217: Ultimately it's a `get` from a map and `resolve` makes it sound like it does something more fancy. Also, EntityID can be for other things too, right? Not just a container? like, Pod or... other stuff (sorry 😅 ), I think "for_pid" is implicit with the process_id arg.

```suggestion
    pub(super) fn get_entity_id(&self, process_id: u32) -> Option<EntityId> {
```

lib/saluki-components/src/sources/dogstatsd/replay/state.rs
```diff
@@ -0,0 +1,445 @@
...
+        self.with_live_state(|loaded_state| loaded_state.pid_map.get(&process_id).cloned())
+            .flatten()
+    }
+
+    fn with_live_state<T>(&self, f: impl FnOnce(&LoadedReplayState) -> T) -> Option<T> {
+        let mut state = self.inner.write().expect("dogstatsd replay state lock poisoned");
```
webern, line 223: @tobz what are we really supposed to do with poisoned locks. It's something I've never really known how to deal with. Is this right? take down the system?

lib/saluki-components/src/sources/dogstatsd/replay/state.rs
```diff
@@ -0,0 +1,445 @@
...
+        })
+        .flatten()
+    }
+
+    pub(crate) fn resolve_container_entity_for_pid(&self, process_id: u32) -> Option<EntityId> {
+        self.with_live_state(|loaded_state| loaded_state.pid_map.get(&process_id).cloned())
+            .flatten()
+    }
+
+    fn with_live_state<T>(&self, f: impl FnOnce(&LoadedReplayState) -> T) -> Option<T> {
```
webern, line 222: I'm not super confident in this rename suggestion, but there's definitely cognitive load trying to understand this function from its name. The part that's kind of weird is that it drops the state when done, so I suggested `read_or_drop` but Claude hated it. It also told me `with_` is `a Rust convention for "acquire a resource, run a closure against it, release the resource"`... hmm, I hate that. Anyway, take it or leave it.

```suggestion
    /// Acquires the lock and runs `f` against the state. If the state is expired, drops it.
    fn read_state<T>(&self, f: impl FnOnce(&LoadedReplayState) -> T) -> Option<T> {
```

lib/saluki-components/src/sources/dogstatsd/replay/mod.rs
```diff
@@ -0,0 +1,24 @@
+//! DogStatsD capture and replay support.
+//!
+//! This module contains the capture writer, capture reader, replay-state store, and replay injector used by the ADP
+//! operator commands.
+#![allow(dead_code)]
```
webern, line 5: Perhaps there's a follow up to wire this up, but it not, it needs a scrub of dead_code annotations.

```suggestion
```

lib/saluki-components/src/sources/dogstatsd/replay/mod.rs
```diff
@@ -0,0 +1,24 @@
+//! DogStatsD capture and replay support.
+//!
+//! This module contains the capture writer, capture reader, replay-state store, and replay injector used by the ADP
+//! operator commands.
```
webern, line 4: You may be right, I may be wrong? But operator? Does that mean the K8s operator?

```suggestion
//! CLI commands.
```

lib/saluki-components/src/sources/dogstatsd/replay/capture.rs
```diff
@@ -0,0 +1,205 @@
...
+
+use super::writer::{CaptureRecord, CaptureTargetDir, TrafficCaptureWriter};
+
+const UNAVAILABLE_CAPTURE_CONTROL_ERROR: &str =
+    "DogStatsD capture control is unavailable because the source is not running.";
+
+/// Owns capture lifecycle for a running DogStatsD source instance.
+#[derive(Clone)]
+pub(crate) struct TrafficCapture {
+    inner: Arc<TrafficCaptureInner>,
```
webern, line 18: Is there no mutex needed on this one? Only one thread writes at a time?

lib/saluki-components/src/sources/dogstatsd/replay/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+mod state;
+mod writer;
+
+/// Magic Unix group ID used to identify replayed DogStatsD traffic.
+///
+/// This mirrors the Go Agent replay constant so replayed UDS credentials can be distinguished from live traffic.
+pub const REPLAY_CREDENTIALS_GID: u32 = 999_888_777;
+
+pub use self::capture::DogStatsDCaptureControl;
+pub(crate) use self::capture::TrafficCapture;
```
webern, line 20: Rust-analyzer confirms `TrafficCapture` is only consumed by `sources/dogstatsd/mod.rs`, the direct parent of this module. `pub(super)` is sufficient.

```suggestion
pub(super) use self::capture::TrafficCapture;
```

lib/saluki-components/src/sources/dogstatsd/replay/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+pub use self::injector::DogStatsDReplayInjector;
+pub use self::reader::TrafficCaptureReader;
+pub use self::state::DogStatsDReplayState;
+pub(crate) use self::writer::CaptureRecord;
```
webern, line 24: Same — `CaptureRecord` is only used in `sources/dogstatsd/mod.rs`.

```suggestion
pub(super) use self::writer::CaptureRecord;
```

lib/saluki-components/src/sources/dogstatsd/replay/capture.rs
```diff
@@ -0,0 +1,205 @@
...
+}
+
+impl DogStatsDCaptureControl {
+    /// Creates a new, unbound capture control handle.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Binds the control handle to a running capture runtime.
+    pub(crate) fn bind(&self, capture: TrafficCapture) {
```
webern, line 89: Unless there is a good reason to do otherwise, order functions in the order of most-visible to least-visible.

- pub
- pub(crate)
- pub(super)
- [private]

lib/saluki-components/src/sources/dogstatsd/replay/reader.rs
```diff
@@ -0,0 +1,243 @@
...
+    path::Path,
+};
+
+use datadog_protos::agent::{TaggerState, UnixDogstatsdMsg};
+use prost::Message;
+use saluki_error::{generic_error, GenericError};
+
+use super::file::{datadog_matcher, file_version, DATADOG_HEADER, MIN_STATE_VERSION};
+
+/// Reads DogStatsD capture files written by the Go agent.
```
webern, line 13: Is this implementation specific to files written by the Go agent? Or does this implementation also read files written by ADP?

lib/saluki-components/src/sources/dogstatsd/replay/reader.rs
```diff
@@ -0,0 +1,243 @@
...
+///
+/// The reader keeps the full capture file contents in memory so packet replay can make multiple passes over the same
```
webern, line 15: Will this be Ok for memory usage? Is this what the Go agent does as well? I'm wondering how big these files can be. It would be much more complex if it needs to be a buffered reader to never have it all in memory...

lib/saluki-components/src/sources/dogstatsd/replay/capture.rs
```diff
@@ -0,0 +1,205 @@
...
+use saluki_error::{generic_error, GenericError};
+
+use super::writer::{CaptureRecord, CaptureTargetDir, TrafficCaptureWriter};
+
+const UNAVAILABLE_CAPTURE_CONTROL_ERROR: &str =
+    "DogStatsD capture control is unavailable because the source is not running.";
+
+/// Owns capture lifecycle for a running DogStatsD source instance.
+#[derive(Clone)]
+pub(crate) struct TrafficCapture {
```
webern, line 17: All `pub(crate)` items in this struct — the struct itself and all its methods (`new`, `with_workload_provider`, `is_ongoing`, `start_capture`, `stop_capture`, `enqueue`) — can be `pub(super)`. Callers are all within `sources::dogstatsd`.

```suggestion
pub(super) struct TrafficCapture {
```

lib/saluki-components/src/sources/dogstatsd/replay/capture.rs
webern, line 89:
```suggestion
    pub(super) fn bind(&self, capture: TrafficCapture) {
```

lib/saluki-components/src/sources/dogstatsd/replay/capture.rs
```diff
@@ -0,0 +1,205 @@
...
+}
+
+struct TrafficCaptureInner {
+    writer: TrafficCaptureWriter,
+    default_capture_dir: PathBuf,
+}
+
+impl TrafficCapture {
+    /// Creates a new capture controller with the given default directory and queue depth.
+    pub(crate) fn new(default_capture_dir: PathBuf, queue_depth: usize) -> Self {
```
webern, line 28:
```suggestion
    pub(super) fn new(default_capture_dir: PathBuf, queue_depth: usize) -> Self {
```

lib/saluki-components/src/sources/dogstatsd/replay/capture.rs
```diff
@@ -0,0 +1,205 @@
...
+        Self::with_workload_provider(default_capture_dir, queue_depth, None)
+    }
+
+    /// Creates a new capture controller with the given default directory, queue depth, and optional workload provider.
+    pub(crate) fn with_workload_provider(
```
webern, line 33:
```suggestion
    pub(super) fn with_workload_provider(
```

lib/saluki-components/src/sources/dogstatsd/replay/capture.rs
```diff
@@ -0,0 +1,205 @@
...
+        Self {
+            inner: Arc::new(TrafficCaptureInner {
+                writer: TrafficCaptureWriter::with_workload_provider(queue_depth, workload_provider),
+                default_capture_dir,
+            }),
+        }
+    }
+
+    /// Returns whether a capture session is currently active.
+    pub(crate) fn is_ongoing(&self) -> bool {
```
webern, line 46:
```suggestion
    pub(super) fn is_ongoing(&self) -> bool {
```

lib/saluki-components/src/sources/dogstatsd/replay/capture.rs
```diff
@@ -0,0 +1,205 @@
...
+        self.inner.writer.is_ongoing()
+    }
+
+    /// Starts a new capture session.
+    pub(crate) fn start_capture(
```
webern, line 51:
```suggestion
    pub(super) fn start_capture(
```

lib/saluki-components/src/sources/dogstatsd/replay/capture.rs
```diff
@@ -0,0 +1,205 @@
...
+        let target_dir = match requested_dir {
+            Some(path) => CaptureTargetDir::Explicit(path.to_path_buf()),
+            None => CaptureTargetDir::Implicit(self.inner.default_capture_dir.clone()),
+        };
+
+        self.inner.writer.start_capture(target_dir, duration, compressed)
+    }
+
+    /// Stops the current capture session, if one is running.
+    pub(crate) fn stop_capture(&self) {
```
webern, line 63:
```suggestion
    pub(super) fn stop_capture(&self) {
```

lib/saluki-components/src/sources/dogstatsd/replay/capture.rs
```diff
@@ -0,0 +1,205 @@
...
+        self.inner.writer.stop_capture();
+    }
+
+    /// Enqueues a captured packet for persistence.
+    pub(crate) fn enqueue(&self, record: CaptureRecord) -> bool {
```
webern, line 68:
```suggestion
    pub(super) fn enqueue(&self, record: CaptureRecord) -> bool {
```

lib/saluki-components/src/sources/dogstatsd/replay/writer.rs
```diff
@@ -0,0 +1,840 @@
...
+
+/// One raw DogStatsD packet captured before framing and decoding.
+#[derive(Debug)]
+pub(crate) struct CaptureRecord {
+    pub(crate) timestamp_ns: i64,
+    pub(crate) payload: Vec<u8>,
+    pub(crate) pid: Option<i32>,
+    pub(crate) ancillary: Vec<u8>,
+    pub(crate) container_id: Option<String>,
+}
```
webern, line 51:
```suggestion
pub(super) struct CaptureRecord {
    pub(super) timestamp_ns: i64,
    pub(super) payload: Vec<u8>,
    pub(super) pid: Option<i32>,
    pub(super) ancillary: Vec<u8>,
    pub(super) container_id: Option<String>,
}
```

lib/saluki-components/src/sources/dogstatsd/replay/reader.rs
```diff
@@ -0,0 +1,243 @@
...
+
+        Ok(Self {
+            contents,
+            version,
+            offset: DATADOG_HEADER.len(),
+        })
+    }
+
+    /// Returns the parsed capture file version.
+    pub(crate) fn version(&self) -> u8 {
```
webern, line 44:
```suggestion
    pub(super) fn version(&self) -> u8 {
```

lib/saluki-components/src/sources/dogstatsd/replay/reader.rs
```diff
@@ -0,0 +1,243 @@
...
+        self.version
+    }
+
+    pub(crate) fn rewind(&mut self) {
```
webern, line 48:
```suggestion
    pub(super) fn rewind(&mut self) {
```

lib/saluki-components/src/sources/dogstatsd/replay/reader.rs
```diff
@@ -0,0 +1,243 @@
...
+        self.offset = DATADOG_HEADER.len();
+    }
+
+    /// Reads the next captured packet record.
+    ///
+    /// Returns `Ok(None)` when the reader reaches the trailing replay-state separator.
+    pub(crate) fn read_next(&mut self) -> Result<Option<UnixDogstatsdMsg>, GenericError> {
```
webern, line 55:
```suggestion
    pub(super) fn read_next(&mut self) -> Result<Option<UnixDogstatsdMsg>, GenericError> {
```

webern (commented):
Didn't quite have time to get through all of it. A lot of suggestions might be off but there are some principles worth instilling, I think:

- Name what it does: function names are very important for reducing cognitive load
- Use tight visibility - pub(super) over pub(crate) when possible. It documents intent and gives a signal in the diff later if an agent or human changes the visibility (we see our internals leaking degrading the abstraction layer)
- Comments help an uninitiated reader. Even on private functions etc

Nice work!

lib/saluki-components/src/sources/dogstatsd/replay/state.rs
tobz: It really does depend because it boils down to whether or not the state being protected can ever be corrupted/invalid due to a panic while locked... or if the lock is just being used to provide thread safety in lieu of a concurrent data structure, etc.

tobz (commented):
Before we commit to going down the rabbit hole of adding all of this plumbing in various spots, I want to understand: is there a reason we can't simply fake the origin detection by... capturing the _original_ set of of origin tags and replaying them directly?

It's not immediately clear to me what the value in doing so is, and I think it's worth asking AMP to figure that out.

lib/saluki-components/src/sources/dogstatsd/replay/file.rs
lucastemb: Yes it does. 

It looks at the 5th spot of the header byte array where we list the current `.dog` file version. For example, version 2 is represented by`0xF2`. There is a comment in the Go Agent that I should probably include in Saluki as well that reads: `(F0 | datadogFileVersion) for different file versions support`. So, since 2 > 3 (the current `DATADOG_FILE_VERSION`), we proceed as normal. 

It's just there to protect against any file formats that may be above what we currently support.

lib/saluki-components/src/sources/dogstatsd/replay/mod.rs
lucastemb: Yeah, in retrospect, the wording is a little strange. It was supposed to describe a user (or operator) of ADP that would be running these CLI commands.

lib/saluki-components/src/sources/dogstatsd/replay/reader.rs
lucastemb: This was actually a concern that @aqian01 shared in his review of this PR. I did make an incremental improvement over the Core Agent for compressed (`.dog.zstd`) files by making it such that we don't store both the compressed and decompressed file in memory at a given time. 

A buffered reader would increase complexity by a lot, so I think this small optimization should be enough unless we happen to encounter any major issues down the line. 

Reference: https://github.com/DataDog/datadog-agent/blob/main/comp/dogstatsd/replay/impl/reader_creator.go#L19

lib/saluki-components/src/sources/dogstatsd/replay/reader.rs
lucastemb: Good catch. The wording is misleading. It's supposed to mean the capture files in the format that was introduced in the Go Agent (which we now subsequently ported to ADP). Every thing after written could just be axed tho

lib/saluki-components/src/sources/dogstatsd/replay/capture.rs
lucastemb: Nope. Inside of `TrafficCaptureInner` we have the `TrafficCaptureWriter` which has a `WriterState` protected by a mutex.

lib/saluki-components/src/sources/dogstatsd/replay/state.rs
lucastemb: I think I'm going to opt for `with_unexpired_state` since I think that describes the check fairly while still balancing the Rust convention.

lib/saluki-config/src/duration_string.rs
```diff
@@ -24,7 +24,7 @@ use snafu::Snafu;
 /// Accepted inputs:
 ///
 /// - Strings with Go time-unit suffixes: `"30s"`, `"1h30m"`, `"250ms"`, `"2h45m30s"`, `"1.5h"`. Valid suffixes: `ns`,
-///   `us`, `µs`, `ms`, `s`, `m`, `h`.
+///   `us`, `µs`, `μs`, `ms`, `s`, `m`, `h`.
```
tobz, line 27: Looks like we have a duplicate unit suffix here?

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -49,6 +52,50 @@ struct StatsCommand {
     limit: Option<usize>,
 }
 
+fn default_capture_duration() -> DurationString {
+    "1m0s"
+        .parse()
+        .expect("default DogStatsD capture duration must be valid")
+}
```
tobz, line 59: We should just add a const function (`new`) to `DurationString` that takes `Duration`, and then _this_ can be a `const fn` that does something like `DurationString::new(Duration::from_secs(60))`

bin/agent-data-plane/src/internal/control_plane.rs
```diff
@@ -48,6 +67,170 @@ fn get_cert_path_from_config(config: &GenericConfiguration) -> Result<PathBuf, G
...
+    fn new(capture_control: DogStatsDCaptureControl, replay_state: DogStatsDReplayState) -> Self {
+        Self {
+            capture_control,
+            replay_state,
+        }
+    }
+}
+
+#[async_trait]
+impl AgentSecure for DogStatsDCaptureApi {
```
tobz, line 85: Get rid of this and just manually define the two API routes we need for starting a capture and setting the tagger state.

We should do it how the DogStatsD Statistics destination does it, where it defines its own API handler type that we then get a copy of and install in the API builder. We'd likely want to grab that when constructing the topology and return it in `create_topology` instead of returning the capture control struct like we're doing now.

bin/agent-data-plane/src/internal/control_plane.rs
```diff
@@ -58,6 +241,7 @@ pub struct PrivilegedApiWorker {
     dp_config: DataPlaneConfiguration,
     env_provider: ADPEnvironmentProvider,
     dsd_stats_config: DogStatsDStatisticsConfiguration,
+    _dsd_capture_control: Option<DogStatsDCaptureControl>,
```
tobz, line 244: Don't prefix this with an underscore if we actually use it.

bin/agent-data-plane/src/env_provider.rs
```diff
@@ -31,6 +32,7 @@ pub struct ADPEnvironmentProvider {
     host_provider: BoxedHostProvider,
     workload_provider: Option<RemoteAgentWorkloadProvider>,
     autodiscovery_provider: Option<BoxedAutodiscoveryProvider>,
+    replay_state: DogStatsDReplayState,
```
tobz, line 35: No reason for this to live here, especially since we can hide it entirely if we do the trick of getting back the API handler for capture/replay control from the source's configuration type instead.

lib/protos/datadog/src/lib.rs
```diff
@@ -59,7 +59,9 @@ pub mod traces {
 pub mod agent {
     pub use super::agent_include::datadog::api::v1::agent_client::AgentClient;
     pub use super::agent_include::datadog::api::v1::agent_secure_client::AgentSecureClient;
+    pub use super::agent_include::datadog::api::v1::agent_secure_server;
```
tobz, line 62: Undo this.

lib/protos/datadog/src/lib.rs
```diff
@@ -59,7 +59,9 @@ pub mod traces {
...
     pub use super::agent_include::datadog::autodiscovery::*;
+    pub use super::agent_include::datadog::config;
```
tobz, line 64: Undo this.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1042,6 +1195,136 @@ async fn drive_stream(
...
+                workload_provider,
+                process_id_from_peer_addr(peer_addr),
+                ancillary_data,
+                payload,
+            ));
+        }
+        ListenAddress::Unix(_) => {
+            let Some(stream_capture) = stream_capture else {
+                return;
+            };
```
tobz, line 1220: You don't need any of this stream capture stuff. Just use local variables.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1042,6 +1195,136 @@ async fn drive_stream(
...
+}
+
+fn resolve_capture_container_id(
+    codec: &DogStatsDCodec, workload_provider: Option<&(dyn WorkloadProvider + Send + Sync)>, process_id: Option<i32>,
+    payload: &[u8],
+) -> Option<String> {
+    payload_local_container_id(codec, payload).or_else(|| {
+        let process_id = u32::try_from(process_id?).ok()?;
+        workload_provider
+            .and_then(|provider| provider.resolve_container_entity_for_pid(process_id))
```
tobz, line 1261: I'm not seeing why we've added `resolve_container_entity_for_pid` to `WorkloadProvider`.

As best as I can tell, `provider` here is the real workload provider -- `RemoteAgentWorkloadProvider` -- which then will use `OnDemandPIDResolver` under the hood -- which is never going to match anything in a capture because the "process ID" we've got is whatever it was at the time of the capture.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1042,6 +1195,136 @@ async fn drive_stream(
...
+    while let Ok(Some(frame)) = framer.next_frame(&mut remaining, true) {
+        if let Ok(parsed) = codec.decode_packet(&frame) {
+            if let Some(container_id) = parsed_packet_local_container_id(parsed) {
+                return Some(container_id);
+            }
+        }
+    }
+
+    None
+}
```
tobz, line 1279: From looking at the Agent source code, I'm not sure I can actually find where it tries to go through the payload itself, only trying to resolve the spoofed PID and doing a single mapping lookup for PID to container ID:

- `processUDSOrigin` is called to get PID and container ID ([here](https://github.com/DataDog/datadog-agent/blob/main/comp/dogstatsd/listeners/uds_common.go#L329))
- `processUDSOrigin` takes spoofed process ID and uses it to look up container ID from PID map ([here](https://github.com/DataDog/datadog-agent/blob/main/comp/dogstatsd/listeners/uds_linux.go#L77-L87))
- resolved container ID is attached to the capture buffer ([here](https://github.com/DataDog/datadog-agent/blob/main/comp/dogstatsd/listeners/uds_common.go#L338))

Nothing else.

tobz (changes requested):

lib/saluki-config/src/duration_string.rs
lucastemb: A little strange, but they are both distinct characters. `(µ != μ)`. They look identical, but one is supposed to be the micro sign while the other is supposed to be mu. Go also explicitly accepts both.

PR #1495, opened 2026-04-27
fix(io): add observability for HTTP errors and permanently dropped queue items

thieman:
## Summary

Adds two new observability signals to the Datadog forwarder:

**HTTP error metric by status code**
- `component_errors_total{error_type:http_send}` already existed for all HTTP failures
- Now also emits `component_errors_total{error_type:http_send, error_code:<code>}` alongside it when we have a status code, using a lazily-registered per-code counter (following the `PerEndpointTelemetry` pattern)
- Transport-level failures (no HTTP response) still only emit the untagged variant

**Permanently dropped queue items metric**
- New `component_items_dropped_total` counter — fires when items are permanently lost from the retry queue
- Permanent drops are: disk evictions (disk full) and in-memory drops when disk persistence is not enabled
- In-memory entries that spill to disk are not counted (not actually dropped)
- `PushResult` already only set `items_dropped` at permanent drop sites, so no structural changes were needed
- `track_dropped_events` split into `track_dropped_items` / `track_dropped_events` for symmetry

Jira: [DADP-60](https://datadoghq.atlassian.net/browse/DADP-60)

## Test plan

- [ ] Verify `component_errors_total{error_code:403}` fires on 403 response from intake
- [ ] Verify `component_items_dropped_total` fires when retry queue drops items (disk full or memory-only mode)

[DADP-60]: https://datadoghq.atlassian.net/browse/DADP-60?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/saluki-components/src/common/datadog/telemetry.rs
+++ b/lib/saluki-components/src/common/datadog/telemetry.rs
@@ -19,8 +19,8 @@ pub struct ComponentTelemetry {
     events_dropped_http: Counter,
     events_dropped_encoder: Counter,
     events_dropped_queue: Counter,
-    http_send_failed: Counter,
-    http_errors_map: Arc<Mutex<HashMap<StatusCode, Counter>>>,
+    http_failed_send: Counter,
+    http_errors_by_code: Arc<Mutex<HashMap<StatusCode, Counter>>>,
 }
 
 impl ComponentTelemetry {
@@ -43,9 +43,9 @@ impl ComponentTelemetry {
                 "component_events_dropped_total",
                 ["intentional:true", "drop_reason:queue_limit"],
             ),
-            http_send_failed: builder
-                .register_debug_counter_with_tags("component_errors_total", ["error_type:send_failed"]),
-            http_errors_map: Arc::new(Mutex::new(HashMap::new())),
+            http_failed_send: builder
+                .register_debug_counter_with_tags("component_errors_total", ["error_type:http_send"]),
+            http_errors_by_code: Arc::new(Mutex::new(HashMap::new())),
         }
     }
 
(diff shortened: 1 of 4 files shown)


webern (approved):

lib/saluki-components/src/common/datadog/telemetry.rs
```diff
@@ -59,12 +68,34 @@ impl ComponentTelemetry {
...
     /// Tracks a failed transaction.
-    pub fn track_failed_transaction(&self, metadata: &Metadata) {
-        self.http_failed_send.increment(1);
+    ///
+    /// When `status` is `Some`, the metric is additionally emitted with an `error_code` tag for the specific status.
+    pub fn track_failed_transaction(&self, metadata: &Metadata, status: Option<StatusCode>) {
+        match status {
+            None => self.http_failed_send.increment(1),
+            Some(status) => {
+                let mut map = self.http_errors_by_code.lock().unwrap();
```
webern, line 77: Is it intended that both code paths would increment this count?

```suggestion
            Some(status) => {
                self.http_failed_send.increment(1);
                let mut map = self.http_errors_by_code.lock().unwrap();
```

tobz: I don't think so, no.

If we always increment the non-tagged _and_ tagged version, we end up overcounting if you do a sum over all `component_errors_total` series... whereas this way, the sum is accurate.

tobz (approved):
LGTM. 👍🏻

PR #1497, opened 2026-04-27
docs(dogstatsd): document ADP DogStatsD stats behavior

### LINKED ISSUE #1352: Document how ADP's on-demand approach to DogStatsD statistics collection differs from `dogstatsd_stats_enable` and `dogstatsd_metrics_stats_enable`.
The core agent has two independent stats keys at different granularities:

- [`dogstatsd_stats_enable`][1] (default `false`) — packet-level throughput. Counts packets/sec via a ring buffer and publishes the result as a Go expvar on [`dogstatsd_stats_port`][2] (default 5000). To actually collect this data, operators must configure an OpenMetrics check pointed at the expvar endpoint.
- [`dogstatsd_metrics_stats_enable`][3] (default `false`) — metric-level debug statistics. Tracks count and last_seen per unique metric+tags combination in a persistent map. Powers the `dogstatsd-stats` CLI command and HTTP endpoint. Runtime-toggleable.

ADP collapses both into a single on-demand mechanism with no config toggle and no scraper configuration required. The [`DogStatsDStatisticsConfiguration`][4] destination is always wired into the topology but has near-zero cost when idle — stats are only collected during a time-bounded window triggered via the [privileged API handler][5] ([`/dogstatsd/stats`][7][`?collection_duration_secs=N`][8]). The handler blocks for the requested duration, then [returns the collected stats inline as JSON][10] — no persistent endpoint to scrape. The [CLI][6] calls [the same API][9] and displays the results directly as summary or cardinality analysis.

Neither key has any effect on ADP. Document this as a known divergence — customers expecting the config-toggle-plus-scraper approach should understand that ADP provides equivalent functionality on-demand without configuration.

[1]: https://github.com/DataDog/datadog-agent/blob/472741e8926e/pkg/config/setup/common_settings.go#L1580
[2]: https://github.com/DataDog/datadog-agent/blob/472741e8926e/pkg/config/setup/common_settings.go#L1579
[3]: https://github.com/DataDog/datadog-agent/blob/472741e8926e/pkg/config/setup/common_settings.go#L1598
[4]: https://github.com/DataDog/saluki/blob/9b4b516ffd3d/lib/saluki-components/src/destinations/dsd_stats/mod.rs#L84
[5]: https://github.com/DataDog/saluki/blob/9b4b516ffd3d/lib/saluki-components/src/destinations/dsd_stats/mod.rs#L206
[6]: https://github.com/DataDog/saluki/blob/9b4b516ffd3d/bin/agent-data-plane/src/cli/dogstatsd.rs#L30
[7]: https://github.com/DataDog/saluki/blob/9b4b516ffd3d/lib/saluki-components/src/destinations/dsd_stats/mod.rs#L261
[8]: https://github.com/DataDog/saluki/blob/9b4b516ffd3d/lib/saluki-components/src/destinations/dsd_stats/mod.rs#L201-L202
[9]: https://github.com/DataDog/saluki/blob/9b4b516ffd3d/bin/agent-data-plane/src/cli/utils.rs#L161-L163
[10]: https://github.com/DataDog/saluki/blob/9b4b516ffd3d/lib/saluki-components/src/destinations/dsd_stats/mod.rs#L234-L235

lucastemb:
## Summary

This PR documents intentional divergence between ADP and the Agent for the DogStatsD stats collection. In the Agent,  `dogstatsd_stats_enable` and `dogstatsd_metric_stats_enable` enable config-driven stats collection paths. In ADP, those keys have no effect because DogStatsD stats are collected on demand through privileged API/CLI for a bounded duration, then return directly to the caller.  

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

N/A 
## References

- Closes: https://github.com/DataDog/saluki/issues/1352

### DIFF
--- a/docs/agent-data-plane/configuration/dogstatsd.md
+++ b/docs/agent-data-plane/configuration/dogstatsd.md
@@ -95,17 +95,29 @@ default values.
 | `statsd_metric_namespace_blacklist`  | Prefixes exempt from namespace   | `_blacklist` key          | Use `_blocklist` key ([#1353])                  |
 | `telemetry.enabled`                  | Global telemetry toggle          | Agent toggle              | Use `data_plane.telemetry_enabled` ([#1338])    |
 
-### DogStatsD Statistics (`dogstatsd_stats_enable` / `dogstatsd_metrics_stats_enable`)
+### DogStatsD statistics (`dogstatsd_stats_enable` / `dogstatsd_metrics_stats_enable`)
 
-The core agent's DogStatsD stats feature collects per-metric statistics (name cardinality, tag
-counts, etc.). You enable it via `dogstatsd_stats_enable`, wait a sampling period, then query the
-`dogstatsd_stats_port` endpoint. `dogstatsd_metrics_stats_enable` is a related toggle for an older
-stats path.
+The core agent has two DogStatsD statistics mechanisms with different scopes.
+`dogstatsd_stats_enable` enables packet-level throughput statistics from a ring buffer, exposed as
+Go expvar data on `dogstatsd_stats_port` (default `5000`). Operators must configure an OpenMetrics
+check to scrape that endpoint before the data is submitted. `dogstatsd_metrics_stats_enable`
+enables runtime-toggleable metric-level debug statistics that track count and last-seen time per
+unique metric and tag combination. That data powers the core agent's `dogstatsd-stats` CLI command
+and HTTP endpoint.
 
-ADP exposes equivalent capability through its privileged API on demand — no config change is
-required. You trigger collection directly via the ADP binary and retrieve results via the privileged
-endpoint. The config keys `dogstatsd_stats_enable`, `dogstatsd_stats_buffer`,
-`dogstatsd_stats_port`, and `dogstatsd_metrics_stats_enable` are not read by ADP. See [#1352].
+ADP does not mirror either config-toggle path. Instead, ADP provides an on-demand metric-level view
+through a DogStatsD statistics destination that is always wired into the topology, but only collects
+data during a time-bounded request. To collect statistics, run
+`agent-data-plane dogstatsd stats --duration-secs N` or call the privileged
+`/dogstatsd/stats?collection_duration_secs=N` API. The handler waits for the requested collection
+window, then returns count and last-seen time per metric context inline as JSON. The CLI uses the
+same API and renders the result as either summary or cardinality analysis.
+
+ADP does not expose the core agent's packet-per-second expvar endpoint, a persistent DogStatsD
+statistics endpoint to scrape, or a runtime toggle for metric-level collection. You do not need to
+set up scraper configuration for this data. The config keys `dogstatsd_stats_enable`,
+`dogstatsd_stats_buffer`, `dogstatsd_stats_port`, and `dogstatsd_metrics_stats_enable` have no
+effect in ADP. See [#1352].
 
 ### `dogstatsd_flush_incomplete_buckets`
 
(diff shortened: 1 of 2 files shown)


jszwedko (commented):
I recognize that it wouldn't make sense to expose goexpvar-style stats from ADP, but I think we should make it possible to scrape the stats (including `PacketsLastSecond`) since customers may be relying on this telemetry in Datadog. Would it makes sense to:

1. Expose this in ADP's local openmetrics endpoint
2. Send the telemetry to the Core Agent via the Remote Agent Interface so it is scrapable there? Technically we could even expose it as goexpvars there 🤔 

?

The `dogstatsd_stats_enable` config setting should also be read/reloadable via the Core Agent (this may already happen for free?).

I'm ok with merging this PR and following up on the above (though we should track the above as issues first).

tobz:
> I recognize that it wouldn't make sense to expose goexpvar-style stats from ADP, but I think we should make it possible to scrape the stats (including `PacketsLastSecond`) since customers may be relying on this telemetry in Datadog. Would it makes sense to:
> 
> 1. Expose this in ADP's local openmetrics endpoint

We already effectively do this, but under ADP-specific metric names (`component_packets_received_total`, etc) that users could do their own rate calculations over.

> 2. Send the telemetry to the Core Agent via the Remote Agent Interface so it is scrapable there? Technically we could even expose it as goexpvars there 🤔 

We also already do something like this, at least in terms of sending `dogstatsd.udp_packets` and `dogstatsd.uds_packets`.. but they're exposed as internal telemetry (have to scraped via OpenMetrics check), not expvars.

docs/agent-data-plane/configuration/dogstatsd/known-configs.json
```diff
@@ -688,7 +688,7 @@
     "feature_state": "DIVERGENT",
     "action": "DOCUMENT",
```
webern: Sorry for the design of this thing. My fault. But anyway the schema does allow for `DOCUMENTED` as an end state. However, ignore this if @jszwedko and @tobz decide that we should implement different behavior here...

```suggestion
    "action": "DOCUMENTED",
```

jszwedko:
> > I recognize that it wouldn't make sense to expose goexpvar-style stats from ADP, but I think we should make it possible to scrape the stats (including `PacketsLastSecond`) since customers may be relying on this telemetry in Datadog. Would it makes sense to:
> > 
> > 1. Expose this in ADP's local openmetrics endpoint
> 
> We already effectively do this, but under ADP-specific metric names (`component_packets_received_total`, etc) that users could do their own rate calculations over.
> 
> > 2. Send the telemetry to the Core Agent via the Remote Agent Interface so it is scrapable there? Technically we could even expose it as goexpvars there 🤔
> 
> We also already do something like this, at least in terms of sending `dogstatsd.udp_packets` and `dogstatsd.uds_packets`.. but they're exposed as internal telemetry (have to scraped via OpenMetrics check), not expvars.

👍 I'm ok with this for the first phase 1 where we can document divergences. It sounds like it is still possible to query for packets/s which I'm definitely interested in preserving for phase 1.

docs/agent-data-plane/configuration/dogstatsd.md
```diff
@@ -95,17 +95,29 @@ default values.
...
+and HTTP endpoint.
+
+ADP does not mirror either config-toggle path. Instead, ADP provides an on-demand metric-level view
+through a DogStatsD statistics destination that is always wired into the topology, but only collects
+data during a time-bounded request. To collect statistics, run
+`agent-data-plane dogstatsd stats --duration-secs N` or call the privileged
+`/dogstatsd/stats?collection_duration_secs=N` API. The handler waits for the requested collection
+window, then returns count and last-seen time per metric context inline as JSON. The CLI uses the
+same API and renders the result as either summary or cardinality analysis.
+
```
jszwedko, line 115: Can we update this section to mention the stats are available via the OpenMetrics endpoint?

jszwedko (approved):

PR #1498, opened 2026-04-27
fix(config): align enable_payloads and flush_incomplete_buckets keys with Agent schema

### LINKED ISSUE #1366: Config key naming mismatches for `enable_payloads.*` and `dogstatsd_flush_incomplete_buckets`.
Two config keys have naming mismatches between the core agent and ADP. In each case the feature works, but a customer migrating with their existing config could get silent misconfiguration.

### 1. `enable_payloads.*` (dots) vs `enable_payloads_*` (underscores)

Controls which payload types are sent to the backend. Same behavior, same defaults, different separator.

| Core agent key | ADP key | Default |
|---|---|---|
| [`enable_payloads.events`][a1] | [`enable_payloads_events`][s1] | `true` |
| [`enable_payloads.series`][a1] | [`enable_payloads_series`][s1] | `true` |
| [`enable_payloads.service_checks`][a1] | [`enable_payloads_service_checks`][s1] | `true` |
| [`enable_payloads.sketches`][a1] | [`enable_payloads_sketches`][s1] | `true` |

A customer setting `enable_payloads.series: false` in their config will have no effect on ADP — it expects `enable_payloads_series`. This may resolve with the Facet config migration (same class of issue as proxy config key mapping).

### 2. `dogstatsd_flush_incomplete_buckets` vs `aggregate_flush_open_windows`

Controls whether to flush aggregation buckets that haven't completed their full time window. Same behavior, same default (`false`), different key names.

| Core agent | ADP |
|---|---|
| [`dogstatsd_flush_incomplete_buckets`][a2] | [`aggregate_flush_open_windows`][s2] |

A customer setting `dogstatsd_flush_incomplete_buckets: true` will have no effect on ADP.

[a1]: https://github.com/DataDog/datadog-agent/blob/141f1206279/pkg/config/setup/common_settings.go#L1466-L1470
[s1]: https://github.com/DataDog/saluki/blob/e9373ef98013/lib/saluki-components/src/sources/dogstatsd/mod.rs#L305-L324
[a2]: https://github.com/DataDog/datadog-agent/blob/141f1206279/pkg/config/setup/common_settings.go#L1597
[s2]: https://github.com/DataDog/saluki/blob/e9373ef98013/lib/saluki-components/src/transforms/aggregate/mod.rs#L126-L127

### LINKED ISSUE #1480: config: three schema deviations in ADP that may be bugs
Three places in the config key registry expose schema incompatibilities between the Agent and ADP that need investigation. All three were surfaced while building the config smoke tests in #1457.

---

### 1. `enable_payloads.*` — flat vs. nested field naming

**Code:** https://github.com/DataDog/saluki/pull/1457#discussion_r3139962787
**File:** \`lib/saluki-components/src/config_registry/datadog/dogstatsd.rs\`

The Agent schema defines these as nested properties: \`enable_payloads.events\`, \`enable_payloads.series\`, \`enable_payloads.service_checks\`, \`enable_payloads.sketches\` (see \`ENABLE_PAYLOADS_EVENTS\` etc. in \`generated/schema.rs\`).

ADP's \`DogStatsDConfiguration\` declares them as flat underscore-separated fields with no \`#[serde(rename)]\`: \`enable_payloads_events\`, \`enable_payloads_series\`, etc. The two naming conventions are incompatible at the config-loader level — an Agent config file that sets \`enable_payloads.events: false\` produces a nested object that the flat field cannot consume, and the value is silently ignored.

**Fix:** Either rename the struct fields to use \`#[serde(rename = "enable_payloads.events")]\` etc., or add a \`#[serde(flatten)]\` sub-struct that mirrors the Agent's nested \`enable_payloads\` section.

---

### 2. Trace obfuscation config path prefix mismatch

**Code:** https://github.com/DataDog/saluki/pull/1457#discussion_r3139970331
**File:** \`lib/saluki-components/src/config_registry/datadog/trace_obfuscation.rs\`

The Agent schema defines obfuscation keys under \`apm_config.obfuscation.*\` (e.g. \`apm_config.obfuscation.credit_cards.enabled\`). See \`APM_CONFIG_OBFUSCATION_*\` constants in \`generated/schema.rs\`.

\`TraceObfuscationConfiguration\` contains a \`config: ObfuscationConfig\` field, so \`cfg.as_typed::<TraceObfuscationConfiguration>()\` reads from \`config.*\` paths, not \`apm_config.obfuscation.*\`. This means the struct is effectively unreachable via the Agent's documented config keys unless the caller goes through \`from_apm_configuration\` (which does read from \`apm_config.obfuscation.*\` and populates the struct correctly).

**Fix:** Evaluate whether \`TraceObfuscationConfiguration\` should be wired to deserialize directly from \`apm_config.obfuscation.*\` so it works consistently regardless of which constructor is used.

---

### 3. `statsd_metric_namespace_blacklist` → `blocklist` — silent config drop

**Code:** https://github.com/DataDog/saluki/pull/1457#discussion_r3139976003
**File:** \`lib/saluki-components/src/config_registry/datadog/dogstatsd_prefix_filter.rs\`

ADP renamed \`statsd_metric_namespace_blacklist\` to \`statsd_metric_namespace_blocklist\` for inclusive language, but the struct field has **no \`#[serde(alias)\`** for the old spelling. An Agent config file that customizes \`statsd_metric_namespace_blacklist\` will be silently ignored by ADP — the customized namespace exclusion list is dropped and ADP falls back to the hardcoded default.

The default lists happen to be identical today, so this only affects users who have customized the key. It is a silent data-loss bug for any such user.

**Fix:** Add \`#[serde(alias = "statsd_metric_namespace_blacklist")]\` to the \`metric_prefix_blocklist\` field in \`transforms/dogstatsd_prefix_filter/mod.rs\`. This is a one-line fix with no risk.

---

All three are tracked in the annotation files with `TODO` comments pointing here.

thieman:
## Summary

Fixes two silent config-drop bugs where Agent config keys were silently ignored by ADP:

- **`enable_payloads.*`**: The Agent uses nested YAML (`enable_payloads.events: false`), but ADP struct fields used flat underscore names with no alias. Values from Agent config files were silently dropped. Fixed by adding four `KEY_ALIASES` entries that bridge the nested dotted path to the flat key at file-load time (same mechanism already used for `proxy.http` → `proxy_http`). The config-registry annotations are updated to reference the generated `schema::ENABLE_PAYLOADS_*` entries, removing four custom schema statics. Closes #1480 (sub-issue 1) and #1366.
- **`dogstatsd_flush_incomplete_buckets`**: The Agent uses this key; ADP only recognized `aggregate_flush_open_windows`. Fixed by adding a `#[serde(alias = "dogstatsd_flush_incomplete_buckets")]` on the struct field (same pattern as `dogstatsd_expiry_seconds`) and updating the config-registry annotation to use `schema::DOGSTATSD_FLUSH_INCOMPLETE_BUCKETS` as the primary schema entry with `aggregate_flush_open_windows` as an additional path, removing a custom schema static. Closes #1366.

## Test plan

- [ ] All 16 config smoke tests pass (`cargo test -p saluki-components -- config_smoke`)
- [ ] All pre-commit checks pass (fmt, clippy, license, docs)

### DIFF
--- a/lib/saluki-components/src/config.rs
+++ b/lib/saluki-components/src/config.rs
@@ -23,6 +23,13 @@ pub const KEY_ALIASES: &[(&str, &str)] = &[
         "apm_config.error_tracking_standalone.enabled",
         "apm_error_tracking_standalone_enabled",
     ),
+    // The Agent schema defines `enable_payloads.*` as nested YAML keys, while ADP struct fields
+    // use flat underscore-separated names. These aliases bridge the two representations so that
+    // an Agent config file setting `enable_payloads.events: false` is correctly consumed by ADP.
+    ("enable_payloads.events", "enable_payloads_events"),
+    ("enable_payloads.series", "enable_payloads_series"),
+    ("enable_payloads.service_checks", "enable_payloads_service_checks"),
+    ("enable_payloads.sketches", "enable_payloads_sketches"),
 ];
 
 /// Remappings from environment variable names to canonical config keys.
--- a/lib/saluki-components/src/config_registry/datadog/aggregate.rs
+++ b/lib/saluki-components/src/config_registry/datadog/aggregate.rs
@@ -23,13 +23,6 @@ static AGGREGATE_CONTEXT_LIMIT_SCHEMA: SchemaEntry = SchemaEntry {
     default: None,
 };
 
-static AGGREGATE_FLUSH_OPEN_WINDOWS_SCHEMA: SchemaEntry = SchemaEntry {
-    yaml_path: "aggregate_flush_open_windows",
-    env_vars: &[],
-    value_type: ValueType::Bool,
-    default: None,
-};
-
 static COUNTER_EXPIRY_SECONDS_SCHEMA: SchemaEntry = SchemaEntry {
     yaml_path: "counter_expiry_seconds",
     env_vars: &[],
@@ -79,11 +72,11 @@ crate::declare_annotations! {
         test_json: None,
     };
 
-    /// `aggregate_flush_open_windows` — flush open buckets on shutdown.
+    /// `dogstatsd_flush_incomplete_buckets` — flush open buckets on shutdown.
     AGGREGATE_FLUSH_OPEN_WINDOWS = SalukiAnnotation {
-        schema: &AGGREGATE_FLUSH_OPEN_WINDOWS_SCHEMA,
+        schema: &schema::DOGSTATSD_FLUSH_INCOMPLETE_BUCKETS,
         support_level: SupportLevel::Full,
-        additional_yaml_paths: &[],
+        additional_yaml_paths: &["aggregate_flush_open_windows"],
         env_var_override: None,
         used_by: &[structs::AGGREGATE_CONFIGURATION],
         value_type_override: None,
--- a/docs/agent-data-plane/configuration/dogstatsd.md
+++ b/docs/agent-data-plane/configuration/dogstatsd.md
@@ -79,21 +79,16 @@ architecture is fundamentally different or the feature is platform-specific.
 The following settings are recognized by both ADP and the core agent, but with different behavior or
 default values.
 
-| Config Key                           | Description                      | Agent Behavior            | ADP Behavior                                    |
-|--------------------------------------|----------------------------------|---------------------------|-------------------------------------------------|
-| `dogstatsd_context_expiry_seconds`   | Context cache TTL (seconds)      | Default 20s, configurable | Hardcodes 30s ([#1340])                         |
-| `dogstatsd_flush_incomplete_buckets` | Flush open buckets on shutdown   | Standard key              | Key is `aggregate_flush_open_windows` ([#1366]) |
-| `dogstatsd_metrics_stats_enable`     | Enable per-metric debug stats    | Config toggle             | On-demand via API ([#1352])                     |
-| `dogstatsd_stats_enable`             | Enable internal stats endpoint   | Config toggle             | On-demand via API ([#1352])                     |
-| `dogstatsd_stats_buffer`             | Internal stats buffer size       | Configurable              | On-demand via API ([#1352])                     |
-| `dogstatsd_stats_port`               | Internal stats endpoint port     | Configurable port         | On-demand via API ([#1352])                     |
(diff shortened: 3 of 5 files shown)


lib/saluki-components/src/config_registry/datadog/aggregate.rs
```diff
@@ -79,11 +72,11 @@ crate::declare_annotations! {
...
     };
 
-    /// `aggregate_flush_open_windows` — flush open buckets on shutdown.
+    /// `dogstatsd_flush_incomplete_buckets` — flush open buckets on shutdown.
     AGGREGATE_FLUSH_OPEN_WINDOWS = SalukiAnnotation {
-        schema: &AGGREGATE_FLUSH_OPEN_WINDOWS_SCHEMA,
+        schema: &schema::DOGSTATSD_FLUSH_INCOMPLETE_BUCKETS,
         support_level: SupportLevel::Full,
-        additional_yaml_paths: &[],
+        additional_yaml_paths: &["aggregate_flush_open_windows"],
```
thieman, line 79: ADP-specific alias is here

lib/saluki-components/src/config.rs
```diff
@@ -23,6 +23,13 @@ pub const KEY_ALIASES: &[(&str, &str)] = &[
         "apm_config.error_tracking_standalone.enabled",
         "apm_error_tracking_standalone_enabled",
     ),
+    // The Agent schema defines `enable_payloads.*` as nested YAML keys, while ADP struct fields
+    // use flat underscore-separated names. These aliases bridge the two representations so that
+    // an Agent config file setting `enable_payloads.events: false` is correctly consumed by ADP.
+    ("enable_payloads.events", "enable_payloads_events"),
+    ("enable_payloads.series", "enable_payloads_series"),
+    ("enable_payloads.service_checks", "enable_payloads_service_checks"),
+    ("enable_payloads.sketches", "enable_payloads_sketches"),
```
thieman, line 32: These don't take in env vars so I think we can get away without KEY_ALIASES if we move them into a nested sub-struct, I'll try that

webern (approved):
Not sure if this comment is blocking, but if not, LGTM:

> These don't take in env vars so I think we can get away without KEY_ALIASES if we move them into a nested sub-struct, I'll try that

PR #1515, opened 2026-04-28
chore(dev): throttle PR volume from Renovate

webern:
## Summary

Replace the bare `config:recommended` config with a stack that batches non-major updates into a single weekly PR, requires a 7-day soak after a release before Renovate proposes it, and caps open and hourly PR counts. Enables OSV alerts so security advisories - which bypass the schedule, soak, and rate limits - are caught earlier for Rust crates than the GitHub advisory feed alone would catch them.

Migrates the file from JSON to JSONC so each setting can be documented inline.

## Change Type

## How did you test this PR?

`npx --yes --package renovate -- renovate-config-validator`

Also, Claude claims this will happen if the config is bad:

```
  When the bot's next scan runs against main after merge, if it can't parse the config or finds an
  invalid setting it didn't catch in static validation (e.g. a typo'd preset name, a deprecated option),
  it auto-opens an issue titled something like "Action Required: Fix Renovate Configuration" in the repo.
   No issue = no problem the bot found. Watch the Issues tab for ~an hour after merge.
```

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,57 @@
+// Renovate configuration for the Saluki repo.
+//
+// This file is JSONC (JSON with Comments). Can be validated with:
+//     npx --yes --package renovate -- renovate-config-validator renovate.json
+//
+// The configuration below is designed to spam us with fewer PRs:
+//
+//   - Updates land in a single weekly window instead of continuously.
+//   - Non-major updates collapse into one combined PR.
+//   - Brand-new releases soak for a week before Renovate proposes them, which
+//     filters out same-week re-bumps and gives upstream maintainers time to
+//     yank broken releases.
+//   - Hard caps prevent surprise floods from a busy ecosystem week.
+//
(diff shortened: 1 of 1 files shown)


tobz (approved):

renovate.json
```diff
@@ -1,6 +1,57 @@
...
+  //   - `schedule` (opens immediately, not just on Monday)
+  //   - `minimumReleaseAge` (no 7-day soak)
+  //   - `prConcurrentLimit` and `prHourlyLimit` (does not count against caps)
+  //   - grouping (gets its own focused PR rather than being bundled in the
+  //     weekly non-major group)
+  //
+  // The block below customizes how those security PRs look. The bypass
+  // behavior itself is on by default and does not need to be enabled.
+
+  // Enable the OSV.dev advisory database in addition to GitHub's feed.
```
webern:
```suggestion
```

Issue #1380, opened 2026-04-14
Design and document ADP's log verbosity controls, and warn on unused `logging_frequency`.


webern:
The core agent uses [`logging_frequency`][1] (default `500`) to throttle repetitive INFO-level log messages. After an initial burst, the forwarder only logs every Nth successful transaction POST ([`transaction.go:459-472`][2]). Customers set this when transaction success logs are too noisy.

ADP does not have this problem — successful forwarder operations are logged at `debug` level, which is invisible at the default `info` level. The `logging_frequency` config key is not read by ADP.

However, ADP's log verbosity configuration may need attention:

1. **Design the user-facing log level interface.** ADP currently exposes `log_level` backed by Rust's [`tracing_subscriber::EnvFilter`][3], which accepts complex per-module filter strings (e.g. `saluki_components::common::datadog=trace`). Decide whether to keep this power-user interface, wrap it in something simpler (e.g. a plain `debug`/`info`/`warn`/`error` enum like the core agent's `log_level`), or something else. Implement and document the chosen design.

2. **Document the difference.** ADP's logging is less noisy than the core agent by design. Document that `logging_frequency` is not used, and explain how to configure log verbosity in ADP instead.

3. **Optional:** If ADP detects that `logging_frequency` is set in configuration, consider emitting a warning log directing the user to the correct ADP log verbosity settings.

[1]: https://github.com/DataDog/datadog-agent/blob/1fa422e0ed9d/pkg/config/setup/common_settings.go#L1102
[2]: https://github.com/DataDog/datadog-agent/blob/1fa422e0ed9d/comp/forwarder/defaultforwarder/transaction/transaction.go#L459
[3]: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html

thieman:
Tracked in DADP-28 (https://datadoghq.atlassian.net/browse/DADP-28) — ASTA Preview blocker.

lucastemb:
I'm leaning towards a hybrid approach where by default we can wrap it in something simpler much like the Agent. For example:`log_level: info`. Then for more advanced log filtering, we can have an additional config option like `log_filter: saluki_components::common::datadog=trace,info` that could give customers a little bit more control and would override `log_level` when set.

webern:
[...]

Issue #1356, opened 2026-04-13
Implement DogStatsD debug logging (`dogstatsd_logging_enabled`, `dogstatsd_log_file`).


webern:
The core agent's [`serverDebug`][1] component can write per-metric debug information (metric name, tags, count, last seen) to a dedicated rotating log file. This is controlled by four keys ([definitions][2]):

| Key | Default | Purpose |
|---|---|---|
| `dogstatsd_logging_enabled` | `true` | Enable writing to the debug log file (also requires `dogstatsd_metrics_stats_enable`) |
| `dogstatsd_log_file` | platform default | Path to the debug log file |
| `dogstatsd_log_file_max_size` | `"10Mb"` | Max size before rotation |
| `dogstatsd_log_file_max_rolls` | `3` | Number of rotated files to keep |

This is a companion feature to `dogstatsd_metrics_stats_enable` — when both are enabled, every metric sample processed by the stats collector is [logged][3] with its name, tags, count, and last-seen timestamp. It is distinct from `dogstatsd_capture_*` (which captures raw traffic).

ADP does not support this. Trace-level logging of full DogStatsD frames is an acceptable workaround for now. This is purely a debug feature — should be implemented eventually for support tooling parity but not immediately needed.

[1]: https://github.com/DataDog/datadog-agent/blob/b81c3e5a7d73/comp/dogstatsd/serverDebug/serverdebugimpl/debug.go#L156
[2]: https://github.com/DataDog/datadog-agent/blob/b81c3e5a7d73/pkg/config/setup/common_settings.go#L1584-L1587
[3]: https://github.com/DataDog/datadog-agent/blob/b81c3e5a7d73/comp/dogstatsd/serverDebug/serverdebugimpl/debug.go#L192-L194

thieman:
Tracked in DADP-26 (https://datadoghq.atlassian.net/browse/DADP-26) — ASTA Preview blocker.

PR #1516, opened 2026-04-28
fix(agent-data-plane): better align logging behavior with Datadog Agent

tobz:
## Summary

This PR fixes a number of aspects around logging capabilities of the Datadog Agent to ensure the desired logging behavior is where it should be.

The Datadog Agent (DDA) logs stuff... like most programs do. It has a number of knobs to tweak -- how to log, where to log, and so on -- and it follows those. However, DDA is a heterogeneous collection of processes using a shared set of helpers/packages to ensure that their logging behavior is consistent: everyone uses the same logging format, but each process should have its own log file, and so on.

This PR fixes up a number of differences in how ADP was handling logging in order to get us closer to the Datadog Agent. In no particular order:

- update `AppBootstrapper` to allow passing in a dedicated logging configuration, _and_ to allow it to reload the logging configuration after the fact
- switch to a concept of using a logging configuration "translator" for ADP that lets us read our logging configuration and then translate it into the more generic Saluki-based logging configuration
- adds a number of integration tests to show that we're matching the Agent's behavior for logging

### Reloadable logging configuration

One particular problem we have, based on our architecture, is that our configuration can change between what we load during the bootstrap phase and what we end up with after connecting to DDA. While most of that configuration doesn't matter until we build the topology, there's still the matter of things like the logging configuration. Without a way to reload the logging configuration after getting our fully resolved configuration, we might be using the wrong log format or log level, and so on.

We've introduced the ability to reload the logging configuration after the initial bootstrap. What this allows us to do, as you can surmise, is to first initialize logging using whatever we get from the bootstrap configuration so that we can actually log stuff (a lot happens between the process starting and getting the fully resolved configuration from DDA!) and then once we're pretty sure we have the final configuration, we reinitialize the logging layer.

This takes the form of using the `Reload` layer from `tracing-subscriber` which we already use for dynamic log levels, and uses it for reloading the entire logging layer _stack_. This means we can start logging to new outputs, or change the log format, after the fact.

This PR doesn't currently support responding to dynamic changes of logging settings over the entire lifetime of the process -- just the one reload after we get the fully resolved configuration -- but this work supports adding that in the future.

### Logging configuration translator

Another problem we had was that we defined a single logging configuration as part of `saluki-app`, which deserialized itself from a `GenericConfiguration`. This was designed to line up with the configuration settings in DDA, and generally worked. However, one particular flaw with this is that there is _some_ dynamic logic required for certain fields, in particular the "log file" field.

Essentially, when we should be logging to file, we need to figure out a log file to write to. That log file is subagent-specific, which means defining a default. Picking the default is very easy, but we would have had to bake it into `saluki-app` which means anything else using Saluki inherits that choice and has to hack around it in some way. Not great.

Following some internal discussions we've had around the idea that eventually we'll need mechanisms to translate custom configuration schemas into a generic Saluki form -- logging, telemetry, topology, etc -- I took this as the opportunity to start introducing that pattern through a "logging configuration translator".

Essentially, ADP has a new type (literally `LoggingConfigurationTranslator`) that takes `GenericConfiguration` and spits out `LoggingConfiguration`. It handles all the logic of figuring out which Agent configuration fields matter, along with any logic for choosing things like the log file to write to, whether or not a particular log output is enabled, and so on... and it "translates" that to `LoggingConfiguration`. This lets us decouple `saluki-app` from `agent-data-plane` while still ensuring we can do what we need to do to establish parity in logging behavior between DDA and ADP.

### Actual changes to logging behavior (and tests to verify those changes)

Practically speaking, what this PR was designed to do is:

- ensure that ADP logs to file based on the configured `log_file` (or default log file path, if unset), unless disabled (`disable_file_logging`)
- use the same logging settings for other log outputs / formats (console/syslog outputs, JSON format, etc)
- ensure that all of these settings are respected during the bootstrap phase _and_ after getting the fully resolved configuration from DDA

We've added three new integration tests to demonstrate that this is now happening, and happening correctly:

- `adp-logging-default-path`
- `adp-logging-ignores-core-agent-log-file`
- `adp-logging-respects-data-plane-log-file`

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

New and existing tests.

## References

DADP-51


webern:
You may have drive-by fixed a few open issues:
#1356
#1380

tobz:
> You may have drive-by fixed a few open issues: #1356 #1380

Ah, this definitely doesn't touch #1356 but starts to edge into #1380 for sure.

lib/saluki-app/src/logging/mod.rs
```diff
@@ -63,23 +82,43 @@ impl LoggingGuard {
 /// # Errors
 ///
 /// If the logging subsystem was already initialized, an error will be returned.
-pub(crate) async fn initialize_logging(config: &LoggingConfiguration) -> Result<LoggingGuard, GenericError> {
+pub(crate) async fn initialize_logging(config: LoggingConfiguration) -> Result<LoggingGuard, GenericError> {
     // TODO: Support for logging to syslog.
```
lucastemb, line 86: Non-blocking nit: syslog support still looks out of scope. Once it's implemented we'll need tests to ensure ADP inherits the correct Agent config.

lucastemb (approved):

lib/saluki-app/src/logging/mod.rs
lucastemb: Whoops I referenced the wrong line 💔

bin/correctness/panoramic/src/assertions/file_contains.rs
```diff
@@ -0,0 +1,183 @@
...
+use futures::TryStreamExt as _;
+use tracing::trace;
+
+use crate::assertions::{Assertion, AssertionContext, AssertionResult};
+
+/// Assertion that checks a file exists in the container, and optionally that its contents match a pattern.
+///
+/// The assertion polls the file via `docker exec cat <path>` until the deadline. If `pattern` is `None`, the
+/// assertion passes as soon as the file is readable. If `pattern` is `Some`, the assertion waits until the file
+/// is readable AND its contents match the pattern.
```
webern, line 17: Nice documentation. 👍

test/integration/cases/adp-logging-default-path/config.yaml
```diff
@@ -0,0 +1,27 @@
...
+    DD_DOGSTATSD_PORT: "8125"
+  exposed_ports:
+    - "8125/udp"
+
+assertions:
+  # Wait for ADP to receive its initial configuration from the Core Agent and reload logging.
+  - type: log_contains
+    pattern: "Initial configuration received."
+    timeout: 30s
+  # The default per-subagent log file path is used.
```
webern, line 23: 👍

test/integration/cases/adp-logging-ignores-core-agent-log-file/config.yaml
```diff
@@ -0,0 +1,33 @@
...
+  # Wait for ADP to receive its initial configuration from the Core Agent and reload logging.
+  - type: log_contains
+    pattern: "Initial configuration received."
+    timeout: 30s
+  # The Core Agent honored DD_LOG_FILE; its own log file has CORE entries.
+  - type: file_contains
+    path: "/tmp/coreagent-only.log"
+    pattern: "CORE"
+    timeout: 30s
+  # ADP did NOT honor DD_LOG_FILE; it still wrote to its per-subagent default path.
```
webern, line 29: 👍

test/integration/cases/adp-logging-respects-data-plane-log-file/config.yaml
```diff
@@ -0,0 +1,28 @@
...
+    DD_DOGSTATSD_PORT: "8125"
+  exposed_ports:
+    - "8125/udp"
+
+assertions:
+  # Wait for ADP to receive its initial configuration from the Core Agent and reload logging.
+  - type: log_contains
+    pattern: "Initial configuration received."
+    timeout: 30s
+  # ADP writes to the explicitly-configured per-subagent path.
```
webern, line 24: 👍

webern (approved):

PR #1519, opened 2026-04-28
fix(config): wire obfuscation to apm_config path; add blacklist alias (#1480 issues 2+3)

### LINKED ISSUE #1480: config: three schema deviations in ADP that may be bugs

thieman:
## Summary

Fixes issues 2 and 3 from #1480. Issue 1 was fixed in #1515.

**Issue 2 — Trace obfuscation config path prefix mismatch:**
- `TraceObfuscationConfiguration.from_configuration` was reading from `config.*` paths instead of the Agent's `apm_config.obfuscation.*` paths. It now delegates to `from_apm_configuration`.
- Annotations updated to reference generated `APM_CONFIG_OBFUSCATION_*` schema entries with correct YAML paths.
- `DD_APM_OBFUSCATION_*` env vars (which Figment produces as flat keys without the `_config_` segment) are now wired up via 25 `KEY_ALIASES` entries. `ObfuscationConfig` deserializes directly from those flat keys using `#[serde(flatten)]` on each sub-struct field and flat `apm_obfuscation_*` serde renames on each sub-struct's fields. No intermediate copy or conversion struct needed.

**Issue 3 — `statsd_metric_namespace_blacklist` → silent config drop:**
- `DogStatsDPrefixFilterConfiguration.metric_prefix_blocklist` had no `#[serde(alias)]` for the Agent's original `statsd_metric_namespace_blacklist` spelling, silently dropping customized namespace exclusion lists.
- Added `alias = "statsd_metric_namespace_blacklist"` and updated the annotation to use the generated schema entry.

**Refactoring along the way:**
- Extracted `deserialize_space_separated_or_seq` (previously only in `proxy.rs`) to `saluki_config::space_separated`, where it sits alongside `duration_string` as a shared config deserialization utility. Added an `Option<Vec<String>>` variant.
- `ObfuscationConfig` now stores typed sub-structs as fields (`credit_cards: CreditCardObfuscationConfig`, `http: HttpObfuscationConfig`, etc.) rather than flat scalar fields. Each sub-struct carries its own flat serde renames; `#[serde(flatten)]` propagates them up. `JsonObfuscationConfig` is split into `EsObfuscationConfig` / `MongoObfuscationConfig` / `OpenSearchObfuscationConfig` so each can carry the correct rename prefix. Callers access fields directly — no accessor methods on pure data types.

## Test plan

- [ ] `cargo test -p saluki-components --lib -- trace_obfuscation` passes
- [ ] `cargo test -p saluki-components --lib -- dogstatsd_prefix_filter` passes
- [ ] CI green

Closes #1480 (issues 2 and 3)

### DIFF
--- a/lib/saluki-components/src/common/datadog/obfuscation.rs
+++ b/lib/saluki-components/src/common/datadog/obfuscation.rs
@@ -13,30 +13,39 @@ use serde::Deserialize;
 #[derive(Clone, Debug, Default, Deserialize, Facet)]
 #[cfg_attr(test, derive(PartialEq, serde::Serialize))]
 pub struct ObfuscationConfig {
+    /// Credit card obfuscation settings.
     #[serde(flatten)]
     pub(crate) credit_cards: CreditCardObfuscationConfig,
 
+    /// HTTP URL obfuscation settings.
     #[serde(flatten)]
     pub(crate) http: HttpObfuscationConfig,
 
+    /// Memcached obfuscation settings.
     #[serde(flatten)]
     pub(crate) memcached: MemcachedObfuscationConfig,
 
+    /// Redis obfuscation settings.
     #[serde(flatten)]
     pub(crate) redis: RedisObfuscationConfig,
 
+    /// Valkey obfuscation settings.
     #[serde(flatten)]
     pub(crate) valkey: ValkeyObfuscationConfig,
 
+    /// SQL obfuscation settings.
     #[serde(flatten)]
     pub(crate) sql: SqlObfuscationConfig,
 
+    /// MongoDB obfuscation settings.
     #[serde(flatten)]
     pub(crate) mongo: MongoObfuscationConfig,
 
+    /// Elasticsearch obfuscation settings.
     #[serde(flatten)]
     pub(crate) es: EsObfuscationConfig,
 
+    /// OpenSearch obfuscation settings.
     #[serde(flatten)]
     pub(crate) open_search: OpenSearchObfuscationConfig,
 }
@@ -45,9 +54,11 @@ pub struct ObfuscationConfig {
 #[derive(Clone, Debug, Default, Deserialize, Facet)]
 #[cfg_attr(test, derive(PartialEq, serde::Serialize))]
 pub struct HttpObfuscationConfig {
+    /// Whether to remove query strings from HTTP URLs.
     #[serde(default, rename = "apm_obfuscation_http_remove_query_string")]
     pub(crate) remove_query_string: bool,
 
+    /// Whether to obfuscate path segments containing digits.
     #[serde(default, rename = "apm_obfuscation_http_remove_paths_with_digits")]
     pub(crate) remove_path_digits: bool,
 }
@@ -56,9 +67,11 @@ pub struct HttpObfuscationConfig {
 #[derive(Clone, Debug, Default, Deserialize, Facet)]
 #[cfg_attr(test, derive(PartialEq, serde::Serialize))]
 pub struct MemcachedObfuscationConfig {
+    /// Whether memcached obfuscation is enabled.
     #[serde(default, rename = "apm_obfuscation_memcached_enabled")]
     pub(crate) enabled: bool,
 
+    /// Whether to keep the command (if false, entire tag is removed).
     #[serde(default, rename = "apm_obfuscation_memcached_keep_command")]
     pub(crate) keep_command: bool,
 }
@@ -67,12 +80,15 @@ pub struct MemcachedObfuscationConfig {
 #[derive(Clone, Debug, Default, Deserialize, Facet)]
 #[cfg_attr(test, derive(PartialEq, serde::Serialize))]
 pub struct CreditCardObfuscationConfig {
+    /// Whether credit card obfuscation is enabled.
     #[serde(default, rename = "apm_obfuscation_credit_cards_enabled")]
     pub(crate) enabled: bool,
 
+    /// Whether to use Luhn checksum validation (reduces false positives, increases CPU cost).
     #[serde(default, rename = "apm_obfuscation_credit_cards_luhn")]
     pub(crate) luhn: bool,
 
+    /// Tag keys that are known to not contain credit cards and can be kept.
     #[serde(
         default,
         deserialize_with = "deserialize_space_separated_or_seq",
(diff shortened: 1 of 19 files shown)


lib/saluki-components/src/common/datadog/obfuscation.rs
```diff
@@ -276,16 +165,77 @@ impl SqlObfuscationConfig {
     }
 }
 
-impl JsonObfuscationConfig {
-    pub fn enabled(&self) -> bool {
-        self.enabled
-    }
+/// Elasticsearch obfuscation configuration.
```
thieman, line 168: Changed these to be dumb structs rather than sharing `JsonObfuscationConfig` and having to call methods

lib/saluki-components/src/common/datadog/proxy.rs
```diff
@@ -299,43 +300,6 @@ fn ip_in_cidr(network: IpAddr, prefix_len: u8, addr: IpAddr) -> bool {
     }
 }
 
-/// Deserializes a `Vec<String>` from either a sequence or a space-separated string.
```
thieman, line 302: This is moved into shared code in saluki-config so it can also be used by obfuscation config

lib/saluki-components/src/transforms/trace_obfuscation/json.rs
```diff
@@ -84,31 +79,21 @@ impl JsonObfuscator {
...
-    }
-
     fn default_sql_config() -> SqlObfuscationConfig {
         SqlObfuscationConfig::default()
     }
 
     /// Helper for tests - creates obfuscator and obfuscates in one call.
     fn obfuscate_json_string(
-        json_str: &str, config: &JsonObfuscationConfig, sql_config: &SqlObfuscationConfig,
+        json_str: &str, keep_values: &[String], obfuscate_sql_values: &[String], sql_config: &SqlObfuscationConfig,
```
thieman, line 88: [Claude Sonnet 4.6] The changes in `lib/saluki-components/src/transforms/trace_obfuscation/json.rs` are purely mechanical test updates driven by a 4-line production signature change.

**Production change:** `JsonObfuscator::new` previously took `config: &JsonObfuscationConfig`. Since `JsonObfuscationConfig` was split into three separate types (`EsObfuscationConfig`, `MongoObfuscationConfig`, `OpenSearchObfuscationConfig`) so each can carry its own flat serde rename prefix, the constructor now takes the two relevant slices directly:

```rust
// before
pub fn new(config: &JsonObfuscationConfig, sql_config: &SqlObfuscationConfig)

// after
pub fn new(keep_values: &[String], obfuscate_sql_values: &[String], sql_config: &SqlObfuscationConfig)
```

**Test changes:** The tests had a helper `obfuscate_json_string(json, &config, &sql_config)` that built a `JsonObfuscationConfig` struct literal and passed it in. With the new signature each call site becomes `obfuscate_json_string(json, keep_values, obfuscate_sql_values, &sql_config)`. Every change is a mechanical substitution of the same values — no test logic, inputs, or assertions were modified.

tobz (approved):

PR #1524, opened 2026-04-29
fix(dogstatsd): respect `dogstatsd_origin_detection_client=false`

### LINKED ISSUE #1426: Investigate / implement `dogstatsd_origin_detection_client`
A cursory glance surfaced the following:

The Agent gates parsing of client-provided origin fields on `dogstatsd_origin_detection_client` (default: `false`). When disabled (the default), the parser ignores the container ID extension (`c:`/local data), external data, and per-metric `|card:` cardinality hints sent by DogStatsD clients. The stated reason is to avoid silently changing tag cardinality for existing installations.

ADP has no equivalent gate — it always parses and honors these fields unconditionally.

**User-visible consequence:** a DogStatsD client sending container ID enrichment or `|card:` hints will have those fields honored on ADP but silently ignored on the Agent by default. Operators migrating from an Agent deployment where this key was never set may see different tag sets on origin-enriched metrics after switching to ADP.

**Code pointers:**
- Agent gate: [`comp/dogstatsd/server/parse.go:63`](https://github.com/DataDog/datadog-agent/blob/main/comp/dogstatsd/server/parse.go#L63)
- ADP codec (unconditional): [`lib/saluki-io/src/deser/codec/dogstatsd/`](https://github.com/DataDog/saluki/tree/main/lib/saluki-io/src/deser/codec/dogstatsd)

**Question to resolve:** should ADP adopt the opt-in gate for migration compatibility, or is always-on the correct long-term behavior?

webern:
## Summary

Prior to this commit, DogStatsD packet parsing unconditionally parsed `c:`, `e:` and `card:` fields. When
`dogstatsd_origin_detection_client=false` it should ignore these.

Like this: https://github.com/DataDog/datadog-agent/blob/feb4ecd13f/comp/dogstatsd/server/parse.go#L210-L216

## Change Type
- [x] Bug fix
- [x] New feature (?)

## How did you test this PR?

Unit tests only. It appears a new form of packet generation other than millstone/lading would be required to send the real container information in the packets for a real integration test.

## References

Closes #1426

### DIFF
--- a/lib/saluki-io/src/deser/codec/dogstatsd/metric.rs
+++ b/lib/saluki-io/src/deser/codec/dogstatsd/metric.rs
@@ -111,7 +111,6 @@ pub fn parse_dogstatsd_metric<'a>(
                     maybe_tags = Some(tags);
                 }
                 // Local Data: client-provided data used for resolving the entity ID that this metric originated from.
-
                 b'c' if chunk.len() > 1 && chunk[1] == b':' => {
                     if config.client_origin_detection {
                         let (_, local_data) = all_consuming(preceded(tag("c:"), local_data)).parse(chunk)?;
@@ -126,7 +125,6 @@ pub fn parse_dogstatsd_metric<'a>(
                     }
                 }
                 // External Data: client-provided data used for resolving the entity ID that this metric originated from.
-
                 b'e' if chunk.len() > 1 && chunk[1] == b':' => {
                     if config.client_origin_detection {
                         let (_, external_data) = all_consuming(preceded(tag("e:"), external_data)).parse(chunk)?;
@@ -134,7 +132,6 @@ pub fn parse_dogstatsd_metric<'a>(
                     }
                 }
                 // Cardinality: client-provided cardinality for the metric.
-
                 b'c' if chunk.starts_with(CARDINALITY_PREFIX) => {
                     if config.client_origin_detection {
                         let (_, cardinality) = cardinality(chunk)?;
--- a/lib/saluki-io/src/deser/codec/dogstatsd/metric.rs
+++ b/lib/saluki-io/src/deser/codec/dogstatsd/metric.rs
@@ -23,17 +23,41 @@ enum MetricType {
 }
 
 /// A DogStatsD metric packet.
+///
+/// See the [DogStatsD datagram format][datagram] reference for the wire format and the protocol versions that
+/// introduced each field.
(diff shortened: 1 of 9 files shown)


lib/saluki-io/src/deser/codec/dogstatsd/metric.rs
```diff
@@ -29,8 +29,11 @@ pub struct MetricPacket<'a> {
     pub values: MetricValues,
     pub num_points: u64,
     pub timestamp: Option<u64>,
+    /// Extension to the statsd protocol. Prefixed with `c:` and used to carry a container ID.
     pub local_data: Option<&'a str>,
+    /// Extension to the statsd protocol. Prefixed with `e:` and used to carry a richer blob of workload identity data.
     pub external_data: Option<&'a str>,
+    /// Extension to the statsd protocol. Prefixed with `card:` and used to specify which origin fields should be used.
     pub cardinality: Option<OriginTagCardinality>,
```
tobz: I would update each of these doc comments to have the first line actually describe the field (`Local Data attached to the metric.`, etc) and then a super brief explanation with a link to the actual DogStatsD reference documentation: https://docs.datadoghq.com/extend/dogstatsd/datagram_shell?tab=metrics

In particular, there's subsections for each protocol version bump that adds support for these distinct extensions.

lib/saluki-io/src/deser/codec/dogstatsd/mod.rs
```diff
@@ -140,6 +141,17 @@ impl DogStatsDCodecConfiguration {
         self.minimum_sample_rate = minimum_sample_rate;
         self
     }
+
+    /// Sets whether client-provided origin detection fields are parsed.
+    ///
+    /// When disabled, the `c:` (local data), `e:` (external data), and `card:` (cardinality) fields
```
tobz: Both here and anywhere else we make references like this, we should use `Local Data` and `External Data`, just to match the capitalization elsewhere in the code and in the Agent codebase itself.

tobz (changes requested):

lib/saluki-io/src/deser/codec/dogstatsd/metric.rs
webern: Since it's a fundamental type I went ahead with full documentation.

tobz (approved):

PR #1525, opened 2026-04-29
feat(dogstatsd): Add DogStatsD SO_RCVBUF support

### LINKED ISSUE #1341: Support `dogstatsd_so_rcvbuf` socket receive buffer size configuration.
The core agent supports `dogstatsd_so_rcvbuf` to set the UDP socket receive buffer size. ADP does not expose this setting. Add support.

aqian01:
## Summary

[Source](https://github.com/DataDog/saluki/issues/1341) 
Adds DogStatsD dogstatsd_so_rcvbuf compatibility support in ADP.

- Adds dogstatsd_so_rcvbuf / DD_DOGSTATSD_SO_RCVBUF config support
- Applies receive buffer sizes to DogStatsD UDP and UDS sockets
- Preserves OS default socket receive buffer behavior when set to 0
- Registers the key in the config registry

This setting increases the OS receive buffer for the DogStatsD socket. That means during a burst of incoming metrics, the operating system can hold more unread UDP datagrams before Saluki has a chance to read them.

The flow is roughly:
network packet
  -> OS socket receive buffer   <-- dogstatsd_so_rcvbuf controls this
  -> Saluki reads from socket
  -> Saluki internal queues/buffers
  -> processing/aggregation/forwarding

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests

## References

Closes #1341

### DIFF
--- a/lib/saluki-components/src/sources/dogstatsd/mod.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/mod.rs
@@ -1319,7 +1319,6 @@ const fn get_adjusted_buffer_size(buffer_size: usize) -> usize {
 mod tests {
     use std::{
         env,
-        ffi::OsString,
         net::{IpAddr, Ipv4Addr, SocketAddr, SocketAddrV4},
         sync::{Mutex, OnceLock},
     };
@@ -1441,7 +1440,11 @@ mod tests {
             .into_typed::<DogStatsDConfiguration>()
             .expect("configuration should deserialize");
 
-        restore_env_var(ENV_VAR, previous);
+        if let Some(value) = previous {
+            env::set_var(ENV_VAR, value);
+        } else {
+            env::remove_var(ENV_VAR);
+        }
 
         assert_eq!(config.socket_receive_buffer_size, 262_144);
     }
@@ -1728,14 +1731,6 @@ mod tests {
         ];
         let mut actual = config.build_addresses(bind_host);
         address_list_eq(&mut expected, &mut actual).unwrap();
-    }
-
-    fn restore_env_var(name: &str, previous: Option<OsString>) {
-        if let Some(value) = previous {
-            env::set_var(name, value);
-        } else {
-            env::remove_var(name);
-        }
     }
 }
 
--- a/lib/saluki-components/src/sources/dogstatsd/mod.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/mod.rs
@@ -495,30 +495,35 @@ impl DogStatsDConfiguration {
             (self.socket_receive_buffer_size != 0).then_some(self.socket_receive_buffer_size);
         for address in addresses {
             // We use a match here so that we can add context to the error message.
-            let listener = match &address {
+            let mut listener = match &address {
                 ListenAddress::Tcp(_) => Listener::from_listen_address(address)
                     .await
                     .context(FailedToCreateListener { listener_type: "TCP" })?,
-                ListenAddress::Udp(_) => {
-                    Listener::from_listen_address_with_socket_receive_buffer_size(address, socket_receive_buffer_size)
-                        .await
-                        .context(FailedToCreateListener { listener_type: "UDP" })?
-                }
+                ListenAddress::Udp(_) => Listener::from_listen_address(address)
+                    .await
+                    .context(FailedToCreateListener { listener_type: "UDP" })?,
                 ListenAddress::Unixgram(_) => {
-                    Listener::from_listen_address_with_socket_receive_buffer_size(address, socket_receive_buffer_size)
+                    Listener::from_listen_address(address)
                         .await
                         .context(FailedToCreateListener {
                             listener_type: "UDS (datagram)",
                         })?
                 }
                 ListenAddress::Unix(_) => {
-                    Listener::from_listen_address_with_socket_receive_buffer_size(address, socket_receive_buffer_size)
+                    Listener::from_listen_address(address)
                         .await
                         .context(FailedToCreateListener {
                             listener_type: "UDS (stream)",
                         })?
                 }
             };
+
+            if !matches!(listener.listen_address(), ListenAddress::Tcp(_)) {
+                if let Some(size) = socket_receive_buffer_size {
+                    listener = listener.with_receive_buffer_size(size);
+                }
+            }
+
             listeners.push(listener);
         }
         Ok(listeners)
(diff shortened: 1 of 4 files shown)


lib/saluki-io/src/net/listener.rs
```diff
@@ -232,6 +267,72 @@ impl Listener {
     }
 }
 
+fn bind_udp_socket(
+    listen_address: &ListenAddress, addr: SocketAddr, recv_buffer_size: Option<usize>,
+) -> Result<TokioUdpSocket, ListenerError> {
```
tobz: So I think we can avoid all of this here and actually just use `configure_stream_socket_receive_buffer_size` because all of the socket types we end up working with here -- `UdpSocket`, `TcpStream`, `UnixStream`, and `UnixDatagram` -- support [`AsFd`](https://doc.rust-lang.org/nightly/std/os/fd/trait.AsFd.html), which means we can create `SockRef<&S>` from them.

The trick is that we would just need to use `configure_stream_socket_receive_buffer_size` once we've gotten the socket/stream, so for TCP/UDP, in `Listener::accept`, we'd just call it to configure the socket before handling the socket back to the caller.

lib/saluki-io/src/net/listener.rs
```diff
@@ -232,6 +267,72 @@ impl Listener {
...
+        SockRef::from(socket)
+            .set_recv_buffer_size(size)
+            .context(FailedToConfigureListener {
+                address: listen_address.clone(),
+                setting: SOCKET_RECV_BUFFER_SIZE_SETTING,
+            })?;
+    }
+
+    Ok(())
+}
```
tobz: I don't think we need a full-blown separate method here, totally fine to just use `configure_stream_socket_receive_buffer_size`. Like, practically speaking, the "UDP" listener is in fact just a UDP socket, etc.

lib/saluki-io/src/net/listener.rs
```diff
@@ -106,6 +115,21 @@ impl Listener {
...
+
+    /// Creates a new `Listener` from the given listen address and optional socket receive buffer size.
+    ///
+    /// The receive buffer size applies only to UDP and Unix domain socket listeners. A value of `None` keeps the OS
+    /// default.
+    ///
+    /// ## Errors
+    ///
+    /// If the listen address cannot be bound, or if the listener cannot be configured correctly, an error is returned.
+    pub async fn from_listen_address_with_socket_receive_buffer_size(
```
tobz: If we do the changes to apply the socket receive buffer size in `accept`, then we can drop this constructor method and just add a new setting method called `with_receive_buffer_size` to update `self.socket_receive_buffer_size`.

tobz (changes requested):

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -503,6 +517,13 @@ impl DogStatsDConfiguration {
                         })?
                 }
             };
+
+            if !matches!(listener.listen_address(), ListenAddress::Tcp(_)) {
+                if let Some(size) = socket_receive_buffer_size {
+                    listener = listener.with_receive_buffer_size(size);
+                }
+            }
+
```
tobz, line 502: We should also set the socket receive buffer size even for TCP listeners, and I think by doing so, we could also clean up this code a lot:

```suggestion
            let listener_type = address.listener_type();
            let listener = Listener::from_listen_address(address)
                .await
                .context(FailedToCreateListener { listener_type })?
                .with_receive_buffer_size(socket_receive_buffer_size);
```

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1385,6 +1411,44 @@ mod tests {
...
+            .expect("configuration should deserialize");
+
+        if let Some(value) = previous {
+            env::set_var(ENV_VAR, value);
+        } else {
+            env::remove_var(ENV_VAR);
+        }
+
+        assert_eq!(config.socket_receive_buffer_size, 262_144);
+    }
```
tobz: I think we can ditch this test since we know that we're going to cover whether or not the setting is picked up when this environment variable is specified by virtue of the config registry smoke tests.

(Manually setting/removing environment variables in tests is also very fragile which makes these sorts of tests a flake risk in some scenarios.)

lib/saluki-io/src/net/listener.rs
```diff
@@ -205,6 +221,11 @@ impl Listener {
...
             ListenerInner::Unixgram(unix) => {
                 if let Some(socket) = unix.take() {
+                    configure_stream_socket_receive_buffer_size(
+                        &socket,
+                        self.socket_receive_buffer_size,
+                        "UDS (datagram)",
+                    )?;
                     enable_uds_socket_credentials(&socket).context(FailedToConfigureStream {
                         setting: "SO_PASSCRED",
                         stream_type: "UDS (datagram)",
```
tobz: Can we also change all of these locations where we set `stream_type` to a fixed string to use `self.listen_address.listener_type()` instead? Just so that the values match the proposed change to how we set `listener_type` in the error in the DSD source.

lib/saluki-io/src/net/listener.rs
```diff
@@ -222,6 +243,11 @@ impl Listener {
                     address: self.listen_address.clone(),
                 })
                 .and_then(|(socket, _)| {
+                    configure_stream_socket_receive_buffer_size(
+                        &socket,
+                        self.socket_receive_buffer_size,
+                        "UDS (stream)",
```
tobz: Can we specify this via `self.listen_address.listener_type()` instead?

lib/saluki-io/src/net/stream.rs
```diff
@@ -155,6 +155,22 @@ impl Stream {
             StreamInner::Connectionless { socket } => socket.receive(buf).await,
         }
     }
+
+    #[cfg(test)]
+    pub(crate) fn recv_buffer_size_for_tests(&self) -> io::Result<usize> {
```
tobz: Just call this `recv_buffer_size`, we already know it's for tests only from the `#[cfg(test)]`.

lib/saluki-io/src/net/listener.rs
tobz: We should also be setting it for TCP.

tobz (changes requested):

tobz (approved):

PR #1526, opened 2026-04-29
feat(panoramic): add DogStatsD events correctness test

### LINKED ISSUE #1528: fix(dogstatsd): backfill current time for events without an explicit timestamp
## Problem

When a DogStatsD event packet does not include a \`d:\` timestamp field, ADP forwards the event to the platform with \`ts = 0\` (the protobuf default). The reference Datadog Agent backfills the current time in this case.

### Reference agent behavior

In \`pkg/aggregator/aggregator.go\`, \`addEvent()\`:

\`\`\`go
func (agg *BufferedAggregator) addEvent(e event.Event) {
    if e.Ts == 0 {
        e.Ts = time.Now().Unix()
    }
    // ...
}
\`\`\`

Events without \`d:\` enter the aggregator with \`Ts = 0\` (set during parsing in \`comp/dogstatsd/server/parse_events.go\`, passed through unchanged by \`enrichEvent()\` in \`enrich.go\`), and \`addEvent()\` stamps them with the current Unix time before they are buffered and flushed.

### ADP behavior

In \`lib/saluki-components/src/sources/dogstatsd/mod.rs\`, \`handle_event_packet()\`:

\`\`\`rust
let eventd = EventD::new(packet.title, packet.text)
    .with_timestamp(packet.timestamp)  // None when d: absent
    // ...
\`\`\`

In \`lib/saluki-components/src/encoders/datadog/events/mod.rs\`, \`encode_eventd()\`:

\`\`\`rust
if let Some(timestamp) = eventd.timestamp() {
    event.set_ts(timestamp as i64);
}
// No else branch — ts defaults to 0 in protobuf when d: is absent
\`\`\`

ADP only sets \`ts\` when \`d:\` was present in the packet. Events without a timestamp are forwarded with \`ts = 0\`.

## Fix

In \`handle_event_packet()\` or in a new enrichment step (analogous to the \`events_enrich\` transform added in #1526 for hostname), fall back to \`SystemTime::now()\` when \`packet.timestamp\` is \`None\`, mirroring \`addEvent()\` in the reference agent.

## Context

Discovered while building the \`dsd-events\` correctness test (PR #1526). Timestamp is excluded from the correctness comparison because the two pipelines assign timestamps at different points (ADP immediately at ingestion, stock agent at aggregation flush time), making direct comparison unstable. The fix is correctness work independent of the test.

thieman:
## Summary

Adds end-to-end correctness test infrastructure for DogStatsD events, comparing stock Datadog Agent output against ADP. Along the way, found and fixed three ADP correctness gaps.

**New infrastructure across the correctness testing stack:**
- **stele**: new `Event` type covering all 9 lading-generated event fields including timestamp
- **datadog-intake**: two new event ingestion paths share a single `EventsState`:
  - `POST /api/v1/events_batch` — ADP's protobuf path
  - `POST /intake/` — stock agent's JSON path (events nested under `{"events": {"<source_type>": [...]}}`)
- **panoramic**: new `analysis_mode: events` with `EventsAnalyzer`; `CollectedData` fetches from `/events/dump`
- **test/correctness/dsd-events**: new test case sending 1000 all-event millstone payloads (100 contexts, fixed seed), exercising all 9 lading-generated event fields with ~50% probability each
- **CI**: `test-correctness-dsd-events` job added to GitLab e2e pipeline; gated in release pipeline alongside `dsd-plain`

**Key finding during development:** The stock agent sends DogStatsD events to `/intake/` as JSON, not to any dedicated events API endpoint. The existing `handle_intake` handler was silently swallowing these with `debug!`-level logging, making them invisible during diagnosis.

## Timestamp handling

Lading draws timestamps as random `u32` values across a 136-year range (year 1970–2106). When a DogStatsD event has no `d:` field, both pipelines backfill the current time, but start at slightly different moments so the values diverge by a few seconds. Any timestamp within one hour of wall-clock time is normalized to `i64::MAX` before sorting and diffing — the probability a lading value lands that close to now is ~0.000084%, so near-now timestamps are reliably identified as fill-ins. Explicit `d:` timestamps are compared exactly.

## ADP fixes included

**Hostname injection** (`fix(dogstatsd): inject default hostname into events missing h: field`):
- DogStatsD events without an explicit `h:` field were going out with no hostname set
- Reference: DDA `comp/dogstatsd/server/enrich.go` `enrichEvent()`
- Fix: extend `HostEnrichment` transform with `enrich_eventd`; wire a dedicated `events_enrich` transform into the DSD pipeline: `dsd_in.events → events_enrich → dd_events_encode`

**source_type_name default** (`fix(dogstatsd): default source_type_name to "api" for events without s: field`):
- DogStatsD events without `s:` were going out with `source_type_name = ""`
- Reference: DDA `pkg/serializer/internal/metrics/events.go` `writeItem()` groups no-source events under `"api"`
- Fix: default `source_type_name` to `"api"` in `handle_event_packet()` when `s:` is absent

**Timestamp backfill** (`fix(dogstatsd): backfill current time for events without explicit d: timestamp`):
- DogStatsD events without a `d:` field were going out with `ts = 0`
- Reference: DDA `pkg/aggregator/aggregator.go` `addEvent()` sets `e.Ts = time.Now().Unix()` when `Ts == 0`
- Fix: use `SystemTime::now()` as fallback in `handle_event_packet()` when `packet.timestamp` is `None`. Resolves #1528.

## Test result

```
PASS dsd-events (62.96s)
13,382 events captured on both sides, all matched.
```

## Test plan
- [x] `make test-correctness-case CASE=dsd-events` passes end-to-end
- [x] Deliberately breaking `aggregation_key` in ADP produces 6,617 failures — test catches real regressions
- [x] Baseline captures ~13,382 events via `/intake/` JSON (stock agent)
- [x] Comparison captures same count via `/api/v1/events_batch` protobuf (ADP)
- [x] `test-correctness-dsd-events` added to GitLab e2e and release pipelines
- [x] Pre-commit checks pass (fmt, clippy, license, docs)

Generated with Claude Code

### DIFF
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -457,20 +457,25 @@ async fn add_dsd_pipeline_to_blueprint(
     //               │                 │                          │ service checks           │ events
     //               │                 ▼                          ▼                          ▼
     //               │      ┌─────────────────────┐    ┌─────────────────────┐    ┌─────────────────────┐
-    //               │      │  DSD Prefix/Filter  │    │ DSD Service Checks  │    │     DSD Events      │
-    //               │      │     (transform)     │    │      (encoder)      │    │      (encoder)      │
+    //               │      │  DSD Prefix/Filter  │    │ DSD Service Checks  │    │   Events Enrich     │
+    //               │      │     (transform)     │    │      (encoder)      │    │     (transform)     │
     //               │      └─────────────────────┘    └─────────────────────┘    └─────────────────────┘
     //               │                 │                          │                          │
-    //               │                 ▼                          │                          └─ ─ ─ ┐
-    //               │      ┌─────────────────────┐               └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐   │
-    //               │      │     DSD Enrich      │                                             │   │
-    //               │      │ (chained transform) │        ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐    │   │
-    //               │      │┌───────────────────┐│        │        Metrics Pipeline       │    │   │
-    //               │      ││    DSD Mapper     ││ ─ ─ ─▶ │  (aggregate, enrich, encode)  │    │   │
-    //               │      │└───────────────────┘│        └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘    │   │
-    //               │      └─────────────────────┘                       │                     │   │
-    //               │                                                    │                     │   │
-    //               ▼                                                    ▼                     ▼   ▼
+    //               │                 ▼                          │                          ▼
+    //               │      ┌─────────────────────┐               │               ┌─────────────────────┐
+    //               │      │     DSD Enrich      │               │               │     DSD Events      │
+    //               │      │ (chained transform) │               │               │      (encoder)      │
+    //               │      │┌───────────────────┐│               │               └─────────────────────┘
+    //               │      ││    DSD Mapper     ││               │                          │
+    //               │      │└───────────────────┘│               │                          │
+    //               │      └─────────────────────┘               │                          │
+    //               │                 │                          │                          │
+    //               │                 │        ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐           │
+    //               │                 └ ─ ─ ─▶ │        Metrics Pipeline       │           │
+    //               │                          │  (aggregate, enrich, encode)  │           │
+    //               │                          └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘           │
+    //               │                                       │                               │
+    //               ▼                                       ▼                               ▼
     //    ┌─────────────────────┐    ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐
     //    │      DSD Stats      │    │                           Forwarder                             │
     //    │    (destination)    │    │                       (Datadog Platform)                        │
@@ -487,6 +492,10 @@ async fn add_dsd_pipeline_to_blueprint(
         .error_context("Failed to configure metric tag filterlist transform.")?;
     let dsd_agg_config =
         AggregateConfiguration::from_configuration(config).error_context("Failed to configure aggregate transform.")?;
+    let events_enrich_config = ChainedConfiguration::default().with_transform_builder(
+        "host_enrichment",
+        HostEnrichmentConfiguration::from_environment_provider(env_provider.clone()),
+    );
     let dd_events_config = DatadogEventsConfiguration::from_configuration(config)
         .map(BufferedIncrementalConfiguration::from_encoder_builder)
         .error_context("Failed to configure Datadog Events encoder.")?;
@@ -501,6 +510,7 @@ async fn add_dsd_pipeline_to_blueprint(
         .add_transform("dsd_enrich", dsd_enrich_config)?
         .add_transform("dsd_tag_filterlist", dsd_tag_filterlist_config)?
         .add_transform("dsd_agg", dsd_agg_config)?
+        .add_transform("events_enrich", events_enrich_config)?
         .add_encoder("dd_events_encode", dd_events_config)?
         .add_encoder("dd_service_checks_encode", dd_service_checks_config)?
         .add_destination("dsd_stats_out", dsd_stats_config)?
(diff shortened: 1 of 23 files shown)


bin/agent-data-plane/src/cli/run.rs
```diff
@@ -457,20 +435,25 @@ async fn add_dsd_pipeline_to_blueprint(
...
+    //               │      ││    DSD Mapper     ││               │                          │
+    //               │      │└───────────────────┘│               │                          │
+    //               │      └─────────────────────┘               │                          │
+    //               │                 │                          │                          │
+    //               │                 │        ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐           │
+    //               │                 └ ─ ─ ─▶ │        Metrics Pipeline       │           │
+    //               │                          │  (aggregate, enrich, encode)  │           │
+    //               │                          └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘           │
+    //               │                                       │                               │
+    //               ▼                                       ▼                               ▼
```
thieman, line 478: This could probably be a Mermaid document but it's kind of neat watching it update it in-place like this

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1178,13 +1178,29 @@ fn handle_event_packet(
...
+    // When no d: field is present, backfill the current time — matching the stock Datadog Agent's
+    // behavior in pkg/aggregator/aggregator.go (addEvent), which sets e.Ts = time.Now().Unix()
+    // for any event with Ts == 0.
+    let timestamp = packet
+        .timestamp
+        .or_else(|| SystemTime::now().duration_since(UNIX_EPOCH).ok().map(|d| d.as_secs()));
+
     let eventd = EventD::new(packet.title, packet.text)
-        .with_timestamp(packet.timestamp)
+        .with_timestamp(timestamp)
```
thieman, line 1189: Fix for timestamp defaulting. Is it ok to just jam this in the dogstatsd source or should this be handled as an explicit transform in the events pipeline? Or something else?

tobz: In the source makes the most sense to me.

bin/correctness/panoramic/src/correctness/analysis/events/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        // Use i64::MAX as the sentinel rather than 0. ADP currently stores 0 for events without
+        // a d: field (bug #1528), so using 0 would silently mask the mismatch. i64::MAX is
+        // guaranteed never to appear as a real lading timestamp (u32 range tops out at ~4.3B,
+        // well below i64::MAX) so it unambiguously marks a normalized fill-in value.
+        for event in baseline_events.iter_mut().chain(comparison_events.iter_mut()) {
+            if (event.timestamp - now).abs() < one_hour {
+                event.timestamp = i64::MAX;
+            }
+        }
```
thieman, line 40: Highlighting wobbly timestamp comparisons for events that don't have explicit timestamps on them from lading

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1178,13 +1178,29 @@ fn handle_event_packet(
...
+        // When no source type is provided, default to "api" — the same default the stock Datadog
+        // Agent applies when serializing DogStatsD events to the intake JSON format. The agent
+        // groups events by source type name and uses "api" as the key for events without an
+        // explicit `s:` field. See: pkg/serializer/internal/metrics/events.go (writeItem).
+        .with_source_type_name(Some(
+            packet
+                .source_type_name
+                .map(|s| s.into())
+                .unwrap_or_else(|| "api".into()),
+        ))
```
thieman, line 1203: Highlighting fix for source type defaulting

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -487,6 +492,10 @@ async fn add_dsd_pipeline_to_blueprint(
         .error_context("Failed to configure metric tag filterlist transform.")?;
     let dsd_agg_config =
         AggregateConfiguration::from_configuration(config).error_context("Failed to configure aggregate transform.")?;
+    let events_enrich_config = ChainedConfiguration::default().with_transform_builder(
+        "host_enrichment",
+        HostEnrichmentConfiguration::from_environment_provider(env_provider.clone()),
+    );
```
thieman, line 498: Extends and reuses the `HostEnrichment` transform to provide `hostname` defaulting when not explicitly provided

bin/correctness/panoramic/src/correctness/analysis/events/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+        let now = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .map(|d| d.as_secs() as i64)
+            .unwrap_or(0);
+        let one_hour = 3600i64;
+
+        // Use i64::MAX as the sentinel rather than 0. ADP currently stores 0 for events without
+        // a d: field (bug #1528), so using 0 would silently mask the mismatch. i64::MAX is
+        // guaranteed never to appear as a real lading timestamp (u32 range tops out at ~4.3B,
+        // well below i64::MAX) so it unambiguously marks a normalized fill-in value.
```
tobz: Well, not as of this PR, no? We're applying a default timestamp in the DogStatsD source now.

Given that, it seems like our comparison should actually move down to the analysis section where check that the baseline/comparison timestamps are within some number of seconds within each other, like 15 seconds or whatever.

This means we can't just depend on straightforward `Eq`-based equivalence, so we'd probably want to copy/pasta the `treediff` stuff over from the traces analysis code, which would let us supply a custom field comparator just for timestamps.

tobz (changes requested):

bin/correctness/panoramic/src/correctness/analysis/events/mod.rs
thieman: Comment updated as yes we are defaulting now and this PR fixes 1528 as-is

With the approach here (translating near-now timestamps to a specific sentinel) we can continue to use `Eq` in the analysis which seemed simpler and less error-prone than trying to match up unsortable events pairwise.

tobz: Hmm, the unsortable events bit is a good point.

Maybe let's split the difference and just tighten the bound here? Anything within 5 minutes of the current wall-clock time should be sufficient to achieve the same outcome.

tobz (approved):
LGTM with a non-blocking nit/conclusion on my previous comment.

PR #1531, opened 2026-04-29
feat(dogstatsd): implemented dogstatsd debug logging

### LINKED ISSUE #1356: Implement DogStatsD debug logging (`dogstatsd_logging_enabled`, `dogstatsd_log_file`).

lucastemb:
## Summary

This adds DogStatsD debug logging currently present in the Agent to ADP. When enabled, one debug line is written per metric sample to a dedicated rotating log file (it contains metric name, tags, sample count, and last-seen timestamp). To support this 5 config options were added:
 
To activate the feature, both **must be enabled** (set to true):
- `dogstatsd_metrics_stats_enable`
- `dogstatsd_logging_enabled`

Log file settings: 
- `dogstatsd_log_file`
- `dogstatsd_log_file_max_size`
- `dogstatsd_log_file_max_rolls`

_The writer runs on a dedicated background thread with a bounded buffer. Lines are dropped (as to never block metric ingestion) if the buffer fills._ 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

I tested this PR predominantly through unit tests to cover rotation and error paths. I also tested locally by running ADP and sending synthetic metrics and seeing if the debug log file would contain the appropriate content. 
Screenshot 2026-04-29 at 2 19 53 PM

## References

- Closes: https://github.com/DataDog/saluki/issues/1356

### DIFF
--- a/lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
+++ b/lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
@@ -218,7 +218,7 @@ impl DogStatsDDebugLog {
         )
         .map_err(|e| {
             generic_error!(
-                "Failed to write DogStatsD debug log line to DogStatsD debug log file '{}': {}",
+                "Failed to write to DogStatsD debug log file '{}': {}",
                 self.log_file.display(),
                 e
             )
--- a/lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
+++ b/lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
@@ -125,11 +125,22 @@ struct ContextNoOrigin {
 
 impl DogStatsDDebugLogConfiguration {
     /// Creates a new `DogStatsDDebugLogConfiguration` from the given configuration.
-    pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
+    ///
+    /// If `dogstatsd_log_file` is empty, `default_log_file_path` is used.
+    pub fn from_configuration(
+        config: &GenericConfiguration, default_log_file_path: PathBuf,
+    ) -> Result<Self, GenericError> {
         let mut cfg: Self = config.as_typed()?;
 
         if cfg.log_file.as_os_str().is_empty() {
-            cfg.log_file = Self::default_log_file_path();
+            cfg.log_file = default_log_file_path;
+        }
+
+        if cfg.log_file.to_str().is_none() {
+            return Err(generic_error!(
+                "dogstatsd_log_file must be valid UTF-8, got '{}'",
+                cfg.log_file.display()
+            ));
         }
 
         cfg.configuration = Some(config.clone());
@@ -155,11 +166,6 @@ impl DogStatsDDebugLogConfiguration {
     /// Returns the number of rotated debug log files to keep.
     pub const fn log_file_max_rolls(&self) -> usize {
         self.log_file_max_rolls
-    }
-
-    /// Returns the default DogStatsD debug log file path for the current platform.
-    pub fn default_log_file_path() -> PathBuf {
-        default_dogstatsd_log_file_path()
     }
 }
 
@@ -228,13 +234,6 @@ impl DogStatsDDebugLog {
     fn ensure_writer(&mut self) -> Result<(), GenericError> {
         if self.writer.is_some() {
             return Ok(());
-        }
-
-        if self.log_file.to_str().is_none() {
-            return Err(generic_error!(
-                "dogstatsd_log_file must be valid UTF-8, got '{}'",
-                self.log_file.display()
-            ));
         }
 
         let appender = RollingFileAppenderBase::new(
@@ -335,27 +334,6 @@ fn format_timestamp(timestamp: u64) -> String {
         .unwrap_or_else(|| timestamp.to_string())
 }
 
-#[cfg(target_os = "macos")]
-fn default_dogstatsd_log_file_path() -> PathBuf {
-    PathBuf::from("/opt/datadog-agent/logs/dogstatsd_info/dogstatsd-stats.log")
-}
-
-#[cfg(target_os = "windows")]
-fn default_dogstatsd_log_file_path() -> PathBuf {
-    std::env::var_os("ProgramData")
-        .map(PathBuf::from)
-        .unwrap_or_else(|| PathBuf::from(r"c:\programdata"))
-        .join("datadog")
-        .join("logs")
-        .join("dogstatsd_info")
-        .join("dogstatsd-stats.log")
-}
-
-#[cfg(not(any(target_os = "macos", target_os = "windows")))]
-fn default_dogstatsd_log_file_path() -> PathBuf {
-    PathBuf::from("/var/log/datadog/dogstatsd_info/dogstatsd-stats.log")
-}
-
 #[cfg(test)]
 mod tests {
     use std::{
@@ -373,11 +351,15 @@ mod tests {
     use crate::config_registry::structs;
     use crate::config_registry::test_support::run_config_smoke_tests;
 
+    fn test_default_log_file_path() -> PathBuf {
+        PathBuf::from("/tmp/default-dogstatsd-stats.log")
+    }
+
     async fn deser_config(raw_json: &str) -> DogStatsDDebugLogConfiguration {
         let value = serde_json::from_str(raw_json).expect("test config should be valid JSON");
         let (config, _) = saluki_config::ConfigurationLoader::for_tests(Some(value), None, false).await;
 
-        DogStatsDDebugLogConfiguration::from_configuration(&config)
+        DogStatsDDebugLogConfiguration::from_configuration(&config, test_default_log_file_path())
             .expect("DogStatsDDebugLogConfiguration should deserialize")
     }
 
@@ -388,10 +370,7 @@ mod tests {
         assert!(config.enabled());
         assert!(!config.metrics_stats_enabled);
         assert!(config.logging_enabled);
-        assert_eq!(
-            config.log_file(),
-            DogStatsDDebugLogConfiguration::default_log_file_path()
-        );
+        assert_eq!(config.log_file(), test_default_log_file_path());
         assert_eq!(config.log_file_max_size(), ByteSize::mb(10));
         assert_eq!(config.log_file_max_rolls(), 3);
     }
@@ -445,14 +424,14 @@ mod tests {
         let value = json!({ "dogstatsd_log_file_max_rolls": -1 });
         let (config, _) = saluki_config::ConfigurationLoader::for_tests(Some(value), None, false).await;
 
-        let result = DogStatsDDebugLogConfiguration::from_configuration(&config);
+        let result = DogStatsDDebugLogConfiguration::from_configuration(&config, test_default_log_file_path());
         assert!(result.is_err());
     }
 
     #[tokio::test]
     async fn smoke_test() {
         run_config_smoke_tests(structs::DOGSTATSD_DEBUG_LOG_CONFIGURATION, &[], json!({}), |cfg| {
-            DogStatsDDebugLogConfiguration::from_configuration(&cfg)
+            DogStatsDDebugLogConfiguration::from_configuration(&cfg, test_default_log_file_path())
                 .expect("DogStatsDDebugLogConfiguration should deserialize")
         })
         .await
@@ -472,7 +451,7 @@ mod tests {
         )
         .await;
 
-        DogStatsDDebugLogConfiguration::from_configuration(&config)
+        DogStatsDDebugLogConfiguration::from_configuration(&config, test_default_log_file_path())
             .expect("DogStatsDDebugLogConfiguration should deserialize")
     }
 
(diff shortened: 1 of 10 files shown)


lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+}
+
+/// Configuration for the DogStatsD debug log destination.
+#[derive(Clone, Debug, Deserialize)]
+#[cfg_attr(test, derive(PartialEq, serde::Serialize))]
+pub struct DogStatsDDebugLogConfiguration {
+    /// Whether DogStatsD metric-level statistics are enabled.
+    ///
+    /// This defaults to `false`. The debug log destination is wired when `dogstatsd_logging_enabled` is `true`, but
+    /// it drops metrics until this runtime flag becomes `true`.
```
tobz: Small tweak to order of things to match our normal comment structure:

```suggestion
    /// Whether DogStatsD metric-level statistics are enabled.
    ///
    /// The debug log destination is always present when `dogstatsd_logging_enabled` is `true`,
    /// but it drops metrics until this runtime flag becomes `true`, and only during that period.
    ///
    /// Defaults to `false`.
```

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+                        tracing::warn!(
+                            key = DOGSTATSD_METRICS_STATS_ENABLE_KEY,
+                            "Dropped dynamic configuration update events while watching DogStatsD metrics stats."
+                        );
+                    }
+                    Err(tokio::sync::broadcast::error::RecvError::Closed) => break,
+                }
+            }
+        });
+    }
```
tobz: Skip all of this, and the `RuntimeBool` stuff, and just use `FieldUpdateWatcher`.

You can look at `tag_filterlist` for how we store a copy of `GenericConfiguration` to use when building the component, and how we use `GenericConfiguration::watch_for_updates` to get a watcher that just gives us update events for the specified key, and `select`-ing over that in the component's core run loop.

With that, then it's just as simple as `metrics_stats_enabled` being a `bool`.

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+
+    let appender = RollingFileAppenderBase::new(
+        log_file,
+        RollingConditionBase::new().max_size(log_file_max_size.as_u64()),
+        log_file_max_rolls,
+    )
+    .map_err(|e| generic_error!("Failed to open dogstatsd_log_file '{}': {}", log_file.display(), e))?;
+
+    let (writer, guard) = NonBlockingBuilder::default()
+        .thread_name("dogstatsd-debug-log-writer")
```
tobz: Make this `dsd-dbg-writer`, since otherwise, it will get chopped down to just `dogstatsd-debug` on Linux. 😅

(Thread names on Linux are limited to 15 characters.)

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+            log_file.display()
+        ));
+    }
+
+    let appender = RollingFileAppenderBase::new(
+        log_file,
+        RollingConditionBase::new().max_size(log_file_max_size.as_u64()),
+        log_file_max_rolls,
+    )
+    .map_err(|e| generic_error!("Failed to open dogstatsd_log_file '{}': {}", log_file.display(), e))?;
```
tobz:
```suggestion
    .map_err(|e| generic_error!("Failed to open DogStatsD debug log file '{}': {}", log_file.display(), e))?;
```

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+
+impl MemoryBounds for DogStatsDDebugLogConfiguration {
+    fn specify_bounds(&self, builder: &mut MemoryBoundsBuilder) {
+        builder
+            .minimum()
+            .with_single_value::<DogStatsDDebugLog>("component struct");
+    }
+}
+
+fn build_debug_log_writer(
```
tobz: Probably just move this into `ensure_writer`? We don't call it from anywhere else.

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+    }
+
+    fn write_metric(&mut self, metric: &Metric) -> Result<(), GenericError> {
+        self.ensure_writer()?;
+
+        let context = metric.context();
+        let metric_context = ContextNoOrigin {
+            name: context.name().clone(),
+            tags: context.tags().clone(),
+        };
```
tobz, line 209: So the thing here is that we want to be order-oblivious to the tags: `foo{a:1, b:2}` should be considered the same as `foo{b:2, a:1}` but normal hashing/equivalence won't hold, so this code would actually have entries for the above metric seen with the tags in those different orders.

I don't have a great answer for getting around this right now, and we can leave it as-is because the tag order changing from the same application should almost never happen, but I didn't want to forget about following up on this eventually.

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+        .join("datadog")
+        .join("logs")
+        .join("dogstatsd_info")
+        .join("dogstatsd-stats.log")
+}
+
+#[cfg(not(any(target_os = "macos", target_os = "windows")))]
+fn default_dogstatsd_log_file_path() -> PathBuf {
+    PathBuf::from("/var/log/datadog/dogstatsd_info/dogstatsd-stats.log")
+}
```
tobz: We should be using `PlatformSettings` instead, and adding a new method -- `get_default_dogstatsd_log_file_path` -- that returns the appropriate path.

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+
+        let writer = self.writer.as_mut().expect("writer should be initialized");
+        writeln!(
+            writer.writer,
+            "Metric Name: {} | Tags: {{{}}} | Count: {} | Last Seen: {}",
+            context.name(),
+            format_tags(context.tags()),
+            sample.count,
+            format_timestamp(sample.last_seen)
+        )
```
tobz, line 224: We'll almost certainly want to follow up after this PR to do this write incrementally so that we can, for example, write out the tags incrementally without having to allocate an intermediate `String` to hold them.

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+
+fn build_debug_log_writer(
+    log_file: &Path, log_file_max_size: ByteSize, log_file_max_rolls: usize,
+) -> Result<DebugLogWriter, GenericError> {
+    if log_file.to_str().is_none() {
+        return Err(generic_error!(
+            "dogstatsd_log_file must be valid UTF-8, got '{}'",
+            log_file.display()
+        ));
+    }
```
tobz: We should validate this in `DogStatsDDebugLogConfiguration::from_configuration`.

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+
+        loop {
+            select! {
+                _ = health.live() => continue,
+                maybe_events = context.events().next() => match maybe_events {
+                    Some(events) => {
+                        for event in events {
+                            if let Event::Metric(metric) = event {
+                                if let Err(error) = self.process_metric(&metric) {
+                                    tracing::warn!(error = %error, "Failed to write DogStatsD debug log line; continuing.");
```
tobz: Import this instead of using the fully-qualified name.

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+}
+
+/// DogStatsD destination that writes metric debug lines to a rotating file.
+struct DogStatsDDebugLog {
+    log_file: PathBuf,
+    log_file_max_size: ByteSize,
+    log_file_max_rolls: usize,
+    writer: Option<DebugLogWriter>,
+    metrics_stats_enabled: RuntimeBool,
+    stats: HashMap<ContextNoOrigin, MetricSample>,
```
tobz: `HashMap` -> `FastHashMap`

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+    /// it drops metrics until this runtime flag becomes `true`.
+    #[serde(rename = "dogstatsd_metrics_stats_enable", default)]
+    metrics_stats_enabled: bool,
+
+    #[serde(skip)]
+    metrics_stats_enabled_state: RuntimeBool,
+
+    /// Whether DogStatsD metric-level statistics should also be written to a log file.
+    ///
+    /// This defaults to `true`, matching the core Agent. This controls whether the destination is added to the
```
tobz: Same comment here about reordering so we mention the default as the last line.

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+
+    /// Maximum size of the active debug log file before rotation.
+    ///
+    /// This defaults to `10Mb`.
+    #[serde(rename = "dogstatsd_log_file_max_size", default = "default_log_file_max_size")]
+    log_file_max_size: ByteSize,
+
+    /// Number of rotated debug log files to keep.
+    ///
+    /// This defaults to `3`.
```
tobz:
```suggestion
    /// Defaults to `3`.
```

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -0,0 +1,720 @@
...
+
+    /// Path to the DogStatsD debug log file.
+    ///
+    /// This defaults to the platform-specific core Agent DogStatsD stats log path when the configured value is empty.
+    #[serde(rename = "dogstatsd_log_file", default)]
+    log_file: PathBuf,
+
+    /// Maximum size of the active debug log file before rotation.
+    ///
+    /// This defaults to `10Mb`.
```
tobz:
```suggestion
    /// Maximum size, in bytes, of the active debug log file before rotation.
    ///
    /// Defaults to `10Mb`.
```

tobz (changes requested):

lib/saluki-components/src/destinations/dsd_debug_log/mod.rs
```diff
@@ -275,7 +218,7 @@ impl DogStatsDDebugLog {
         )
         .map_err(|e| {
             generic_error!(
-                "Failed to write DogStatsD debug log line to dogstatsd_log_file '{}': {}",
+                "Failed to write DogStatsD debug log line to DogStatsD debug log file '{}': {}",
```
tobz: Sorry, I should have caught this before but the message is a little repetitive now.

Can we just make it:

```suggestion
                "Failed to write to DogStatsD debug log file '{}': {}",
```

lucastemb: Ha, no worries. I agree with you though.

tobz (approved):

PR #1532, opened 2026-04-29
feat(panoramic): add `dsd-service-checks` correctness test

thieman:
## Summary

- Adds a new `ServiceCheck` type to `stele` for representing simplified service check telemetry
- Adds `/api/v1/check_run` capture and `/service_checks/dump` endpoints to `datadog-intake`, replacing the previous stub that silently discarded payloads; handles DDA's habit of sending explicit `null` for optional fields (tags, message, host_name) rather than omitting them
- Adds a `ServiceChecks` analysis mode and `ServiceChecksAnalyzer` to `panoramic`, with near-now timestamp normalization using `u64::MAX` as sentinel (matching the events analyzer's `i64::MAX` approach)
- Adds `test/correctness/dsd-service-checks/` test config driving 100% `service_check` DSD payloads through both the stock agent and ADP for parity comparison
- No manual CI job needed — `generate-correctness-pipeline` auto-discovers the new test directory

**Bug fixes found while running the test locally and in CI:**
- `ServiceCheck` gains `set_hostname()` and `try_as_service_check_mut()`, mirroring `EventD`
- `HostEnrichment` transform now enriches service checks with the agent hostname when `h:` is absent, matching DDA behavior; separate `events_enrich` and `service_checks_enrich` pipeline stages handle each type independently
- Service checks without a `d:` timestamp now get the current time backfilled, matching DDA behavior and fixing a CI failure where baseline timestamps normalized to `u64::MAX` while comparison had `None`

## Test plan

- [x] `make test-correctness-case CASE=dsd-service-checks` passes locally (24,930 checks matched)
- [x] Deliberately breaking `enrich_service_check` causes the test to fail
- [ ] CI `test-correctness-dsd-service-checks` passes end-to-end

### DIFF
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -446,35 +446,37 @@ async fn add_dsd_pipeline_to_blueprint(
     // We're creating the "front half" of the DogStatsD pipeline, which deals solely with accepting DogStatsD payloads,
     // and enriching/processing them in DSD-specific ways, relevant to how the Datadog Agent is expected to behave.
     //
-    //                                        ┌─────────────────────┐
-    //                       metrics         │      DogStatsD      │   events + service checks
-    //          ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ │       (source)      │ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐
-    //          │                │            └─────────────────────┘                          │
-    //          │                ▼                                                             ▼
-    //          │     ┌─────────────────────┐                                    ┌─────────────────────┐
-    //          │     │  DSD Prefix/Filter  │                                    │  DSD Non-Metrics    │
-    //          │     │     (transform)     │                                    │  Enrich (chained)   │
-    //          │     └─────────────────────┘                                    │ ┌─────────────────┐ │
-    //          │                │                                               │ │ Host Enrichment │ │
-    //          │                ▼                                               │ └─────────────────┘ │
-    //          │     ┌─────────────────────┐                                    └─────────────────────┘
-    //          │     │     DSD Enrich      │                                         │           │
-    //          │     │ (chained transform) │                                         │           │
-    //          │     │┌───────────────────┐│                                         ▼           ▼
-    //          │     ││    DSD Mapper     ││                              ┌──────────────┐ ┌──────────────┐
-    //          │     │└───────────────────┘│                              │  DSD Events  │ │  DSD Service │
-    //          │     └─────────────────────┘                              │   (encoder)  │ │    Checks    │
-    //          │                │                                         └──────────────┘ │   (encoder)  │
-    //          │                │        ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐               └──────────────┘
-    //          │                └ ─ ─ ─▶ │        Metrics Pipeline       │                      │
-    //          │                         │  (aggregate, enrich, encode)  │                      │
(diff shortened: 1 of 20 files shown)


bin/correctness/panoramic/src/correctness/analysis/service_checks/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        // When `d:` is absent, some pipeline stages may backfill the current time. Any timestamp
+        // within 5 minutes of now is treated as a fill-in (probability of a lading value landing
+        // that close to now: ~0.0000070%) and normalized to None so both sides compare equal.
+        // Explicit `d:` timestamps are compared exactly.
+        let now = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .map(|d| d.as_secs())
+            .unwrap_or(0);
+        let five_minutes = 300u64;
```
thieman, line 29: Reused the timestamp hack from event tests

bin/correctness/panoramic/src/correctness/analysis/mod.rs
```diff
@@ -6,18 +6,22 @@ pub use self::collected::CollectedData;
 
 mod events;
 mod metrics;
+mod service_checks;
 mod traces;
 
 /// Types of analysis to perform on collected data
 #[derive(Clone, Deserialize)]
-#[serde(rename_all = "lowercase")]
+#[serde(rename_all = "snake_case")]
```
thieman, line 14: ServiceChecks is multiple words so this gives us `service_checks` instead of `servicechecks`

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -493,6 +493,10 @@ async fn add_dsd_pipeline_to_blueprint(
         "host_enrichment",
         HostEnrichmentConfiguration::from_environment_provider(env_provider.clone()),
     );
+    let service_checks_enrich_config = ChainedConfiguration::default().with_transform_builder(
```
thieman: Sharing hostname enrichment similar to what we did for events

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -493,6 +493,10 @@ async fn add_dsd_pipeline_to_blueprint(
...
+        "host_enrichment",
+        HostEnrichmentConfiguration::from_environment_provider(env_provider.clone()),
+    );
```
tobz: Can we just reuse the existing chained enrichment transform and call it like `add_missing_host` or whatever? Point both `dsd_in.events` and `dsd_in.service_checks` to it.

tobz (changes requested):
One small change then good 2 go, I think.

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -522,11 +515,10 @@ async fn add_dsd_pipeline_to_blueprint(
...
         .connect_component("metrics_enrich", ["dsd_agg"])?
-        // Events.
-        .connect_component("events_enrich", ["dsd_in.events"])?
-        .connect_component("dd_events_encode", ["events_enrich"])?
-        .connect_component("service_checks_enrich", ["dsd_in.service_checks"])?
-        .connect_component("dd_service_checks_encode", ["service_checks_enrich"])?
+        // Events and service checks share a single enrichment transform for host backfill.
+        .connect_component("dsd_non_metrics_enrich", ["dsd_in.events", "dsd_in.service_checks"])?
+        .connect_component("dd_events_encode", ["dsd_non_metrics_enrich"])?
+        .connect_component("dd_service_checks_encode", ["dsd_non_metrics_enrich"])?
```
tobz: Not that I'm saying this because I want you to change it back, but I just had the realization that this will mean we actually send both events and service checks to each of the event and service check encoders... which they'll properly filter out, because they have to -- the API contract demands they be able to do that -- but just wasn't even thinking about that when I originally commented.

tobz (approved):

PR #1533, opened 2026-04-29
feat(dogstatsd): simplified filter strings for log level to first-par…

### LINKED ISSUE #1380: Design and document ADP's log verbosity controls, and warn on unused `logging_frequency`.

lucastemb:
…ty dependencies

## Summary

ADP's log_level previously passed any value directly to tracing_subscriber::EnvFilter, which meant setting debug or trace would flood output with logs from third-party crates (hyper, tokio, tonic). This PR makes plain log levels (trace/debug/info/warn/error) apply only to ADP and Saluki-owned targets, matching the core Agent's behavior. Advanced EnvFilter directive strings (e.g. warn,agent_data_plane=debug,hyper=warn) are still supported and passed through unchanged for users who need fine-grained control.

Additionally, logging_frequency is intentionally left unused by ADP — successful forwarder operations are already logged
below info level, so there's nothing to throttle. ADP now emits a one-time warning if logging_frequency is detected in
config, directing users to log_level instead.
## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

This PR was tested mainly with dedicated unit tests. 

## References

- Closes https://github.com/DataDog/saluki/issues/1380

### DIFF
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -44,9 +44,7 @@ use crate::{
         ottl_transform_processor::OttlTransformConfiguration, tag_filterlist::TagFilterlistConfiguration,
     },
     internal::{
-        create_internal_supervisor,
-        logging::{warn_if_logging_frequency_configured, LoggingConfigurationTranslator},
-        remote_agent::RemoteAgentBootstrap,
+        create_internal_supervisor, logging::LoggingConfigurationTranslator, remote_agent::RemoteAgentBootstrap,
     },
 };
 use crate::{config::DataPlaneConfiguration, env_provider::ADPEnvironmentProvider};
@@ -134,7 +132,6 @@ pub async fn handle_run_command(
                     "Failed to translate logging configuration from Agent; continuing with bootstrap logging settings."
                 ),
             }
-            warn_if_logging_frequency_configured(&dynamic_config);
 
             // Reload our data plane configuration based on the dynamic configuration.
             let dynamic_dp_config = DataPlaneConfiguration::from_configuration(&dynamic_config)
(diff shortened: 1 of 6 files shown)


bin/agent-data-plane/src/cli/run.rs
```diff
@@ -129,6 +134,7 @@ pub async fn handle_run_command(
                     "Failed to translate logging configuration from Agent; continuing with bootstrap logging settings."
                 ),
             }
+            warn_if_logging_frequency_configured(&dynamic_config);
```
tobz: From our internal conversation, let's remove any changes here related to warning around `logging_frequency` being set.

tobz (changes requested):

tobz (approved):

PR #1541, opened 2026-04-30
feat(panoramic): add kind-based correctness test runtime

thieman:
## Human Summary

Adds the ability to run `kind` (kubernetes-in-docker) correctness tests. The actual goal here is to add multiple tests covering origin detection which has a lot of Kubernetes-specific behavior we'd like to cover. This PR does not add origin detection tests yet, but it does add a `dsd-plain-kind` test which is just the existing `dsd-plain` correctness test but modified to run under kind. This test will be deleted when we add the origin detection tests.

Example run including the kind test is provided immediately below. You'll need to [install kind](https://kind.sigs.k8s.io/docs/user/quick-start/#installing-from-release-binaries) first:

`cargo run --profile release --package panoramic -- run -d test/correctness -d test/integration/cases -t basic-startup,dsd-plain,dsd-plain-kind`

## Summary

- Adds a `runtime: kubernetes_in_docker` option to correctness test configs that runs test groups as multi-container Kubernetes pods inside a kind cluster, unlocking origin detection testing scenarios that are impossible in the plain Docker framework (real pod UIDs, containerd container IDs, K8s labels, External Data injection via `DD_EXTERNAL_ENV`)
- Introduces `dsd-plain-kind` as the initial kind-based test — verifies the existing dsd-plain workload passes end-to-end through the kind path as a baseline before origin-detection-specific tests are layered on top
- CI integration follows the same dynamic pipeline approach as existing Docker tests: the pipeline generator emits kind jobs automatically when `runtime: kubernetes_in_docker` is detected, using a new `.test-correctness-kind-definition` mixin that extends the existing Docker mixin with a longer timeout
- `runtime` is now a required field in all correctness configs (no default); all existing tests explicitly declare their runtime
- Rebased onto main after PR #1552 (dynamic dispatch with Test trait and Runner) and latest main (May 7)

## What changed

**New runtime path (`k8s.rs`):**
- Integrates with PR 1552's `Test` trait architecture — `Config::run(tctx: TestContext)` dispatches to the kind path when `runtime: kubernetes_in_docker`
- Each test group (baseline + comparison) runs as a multi-container pod in its own namespace, labelled `created-by=panoramic-kind` for orphan cleanup
- `datadog-intake`, `target` (agent), and `millstone` share a pod with an emptyDir at `/airlock` for the UDS socket
- Config files injected via ConfigMap with `subPath` mounts so the agent's `/etc/datadog-agent/` remains writable (needed for `auth_token` creation)
- Millstone wrapped in a socket-wait shell command so it doesn't send before the agent is ready
- After the pod reaches Running, background tasks stream each container's logs to `<log-dir>/<baseline|comparison>/<container>.log`; ANSI escape codes stripped via `crate::utils::strip_ansi_codes` (no duplicate implementations)
- Data collected via kube-rs port-forward to datadog-intake port 2049; forward cancelled via `CancellationToken` after collection
- `wait_for_millstone_exit` bounded by `MILLSTONE_EXIT_TIMEOUT` (300s)
- Both baseline and comparison errors reported when both groups fail simultaneously
- Malformed env vars (missing `=`) in target config emit `warn!` instead of being silently dropped

**ANSI stripping (both runtimes):**
- `airlock/src/driver.rs` strips ANSI codes from Docker container log output
- `k8s.rs` and `kind.rs` share `strip_ansi_codes` from `crate::utils`; `airlock` keeps its own copy as a separate crate
- `kind create cluster` output is captured and emitted through tracing at debug level; raw emoji/ANSI output suppressed

**Kind cluster lifecycle (`kind.rs`, `main.rs`, `runner.rs`, `test.rs`):**
- Kind cluster setup only runs when the selected test set includes at least one `kubernetes_in_docker` test; running `-t dsd-plain` never touches kind
- Kind cluster setup runs as a background task — Docker-runtime tests start immediately without waiting
- Kind tests wait for the cluster-ready signal **before acquiring a concurrency slot**, gated on `test.runtime() == "kubernetes_in_docker"` so Docker tests are completely unaffected; each task holds its own cloned `watch::Receiver` with independent "last seen" state — no mutex needed
- The wait loop uses `tokio::select!` on the cancellation token so Ctrl-C during cluster setup doesn't deadlock kind test futures; both `run_parallel` and `run_fail_fast` perform this wait
- `check_kind_installed` verifies the exit code in addition to binary presence
- A warning is emitted when kind setup fails after cluster creation, alerting users to a potentially dangling cluster
- Kind setup emits `TestEvent::StatusLine` messages visible in both TUI and logging modes
- panoramic manages the full kind cluster lifecycle: creates if absent (reuses if present), pulls images only if not already in the local daemon (pull failure is fatal), loads images in parallel, deletes cluster after tests
- `--no-delete-kind-cluster` keeps the cluster alive between runs (useful locally)
- `--kind-cluster-name` overrides the default (`saluki-correctness`)
- Flush wait changed from 32s to 30s (`FLUSH_WAIT: Duration`)

**CI (`.gitlab/correctness-mixins.yml`, `generate-correctness-pipeline.sh`):**
- `.test-correctness-kind-definition` extends `.test-correctness-definition` with a longer timeout — all cluster/image management is inside panoramic
- Pipeline generator emits kind jobs via the kind mixin; mixin selection driven by `test.runtime()` from the `Test` trait

**kind pre-installed in `SALUKI_BUILD_CI_IMAGE`:**
- `.ci/install-kind.sh` installs kind in the same `RUN` layer as `install-docker-cli.sh`, with per-arch SHA256 checksums hardcoded in the script

**Explicitness:**
- `Runtime` enum has no default; all correctness configs have an explicit `runtime:` field (including `dsd-tag-filterlist` from PR 1552)

**Local dev:**
- `cargo run --profile release --package panoramic -- run -d test/correctness -t dsd-plain-kind --no-tui --no-delete-kind-cluster`
- `make clean-kind` / `make clean-correctness`

**Dependencies:** kube 0.93 (client + rustls-tls + ws), k8s-openapi 0.22, tokio-util compat; RUSTSEC-2025-0134 ignored (rustls-pemfile, transitive via kube)

## Test plan

- [x] `dsd-plain-kind` passes locally
- [x] Running `-t dsd-plain` (Docker test) does not trigger kind cluster setup
- [x] Docker and kind tests run concurrently — Docker tests start immediately, kind tests wait without holding concurrency slots
- [x] Ctrl-C during cluster setup doesn't deadlock kind test futures
- [x] Kind setup progress visible in both TUI and logging modes via `TestEvent::StatusLine`
- [x] Container logs are plain text without ANSI codes (both runtimes)
- [x] All correctness tests discovered with explicit runtime fields (including  from latest main)
- [x] Branch rebased cleanly onto main post-PR-1552
- [ ] `SALUKI_BUILD_CI_IMAGE` rebuilt with kind (trigger `generate-build-ci-image` via Run Pipeline on this branch)
- [ ] CI pipeline generates `test-correctness-dsd-plain-kind` job
- [ ] Existing Docker correctness tests unaffected

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/bin/correctness/airlock/src/driver.rs
+++ b/bin/correctness/airlock/src/driver.rs
@@ -982,7 +982,7 @@ impl Driver {
                 match log_result {
                     Ok(log) => match log {
                         LogOutput::StdErr { message } => {
-                            if let Err(e) = stderr_file.write_all(&message[..]).await {
+                            if let Err(e) = stderr_file.write_all(&strip_ansi_codes(&message)).await {
                                 error!(error = %e, "Failed to write log line to standard error log file.");
                                 break;
                             }
@@ -992,7 +992,7 @@ impl Driver {
                             }
                         }
                         LogOutput::StdOut { message } => {
-                            if let Err(e) = stdout_file.write_all(&message[..]).await {
+                            if let Err(e) = stdout_file.write_all(&strip_ansi_codes(&message)).await {
                                 error!(error = %e, "Failed to write log line to standard output log file.");
                                 break;
                             }
@@ -1022,6 +1022,25 @@ impl Driver {
 
         Ok(())
     }
+}
+
+/// Removes ANSI escape sequences (`ESC[...letter`) from a byte slice.
+fn strip_ansi_codes(input: &[u8]) -> Vec<u8> {
+    let mut out = Vec::with_capacity(input.len());
+    let mut i = 0;
+    while i < input.len() {
+        if input[i] == 0x1b && input.get(i + 1) == Some(&b'[') {
+            i += 2;
+            while i < input.len() && !input[i].is_ascii_alphabetic() {
+                i += 1;
+            }
+            i += 1;
+        } else {
+            out.push(input[i]);
+            i += 1;
+        }
+    }
+    out
 }
 
 fn get_alpine_container_image() -> String {
--- a/bin/correctness/panoramic/src/utils.rs
+++ b/bin/correctness/panoramic/src/utils.rs
@@ -1,0 +1,18 @@
+/// Removes ANSI escape sequences (`ESC[...letter`) from a byte slice.
+pub(crate) fn strip_ansi_codes(input: &[u8]) -> Vec<u8> {
+    let mut out = Vec::with_capacity(input.len());
+    let mut i = 0;
+    while i < input.len() {
+        if input[i] == 0x1b && input.get(i + 1) == Some(&b'[') {
+            i += 2;
+            while i < input.len() && !input[i].is_ascii_alphabetic() {
+                i += 1;
+            }
+            i += 1;
+        } else {
+            out.push(input[i]);
+            i += 1;
+        }
+    }
+    out
+}
--- a/.ci/images/build/Dockerfile
+++ b/.ci/images/build/Dockerfile
@@ -29,15 +29,13 @@ RUN --mount=type=bind,from=ci-scripts,target=/scripts \
     /scripts/clean-temporary-caches.sh
 
 # Install the Docker CLI directly, as well as the CI-specific credential helper.
+# Also install kind (Kubernetes in Docker), used for kind-based correctness tests.
 #
 # We intentionally avoid basing our image off of the `docker` image because it's fat and bloated. ¯\_(ツ)_/¯
 RUN --mount=type=bind,from=ci-scripts,target=/scripts \
     /scripts/install-docker-cli.sh && \
+    /scripts/install-kind.sh && \
     /scripts/clean-temporary-caches.sh
-
-# Install kind (Kubernetes in Docker), used for kind-based correctness tests.
-RUN --mount=type=bind,from=ci-scripts,target=/scripts \
-    /scripts/install-kind.sh
 
 # Install Go, which we need to build AWS-LC in FIPS-compliant mode.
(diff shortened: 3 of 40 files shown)


bin/correctness/airlock/src/driver.rs
```diff
@@ -1024,6 +1024,25 @@ impl Driver {
     }
 }
 
+/// Removes ANSI escape sequences (`ESC[...letter`) from a byte slice.
+fn strip_ansi_codes(input: &[u8]) -> Vec<u8> {
```
thieman, line 1028: Driveby, cleans up the log artifacts produced by the correctness tests (not just kind, this was affecting all of them)

bin/correctness/panoramic/src/utils.rs
```diff
@@ -0,0 +1,18 @@
+/// Removes ANSI escape sequences (`ESC[...letter`) from a byte slice.
+pub(crate) fn strip_ansi_codes(input: &[u8]) -> Vec<u8> {
```
thieman, line 2: There are two copies of this function, the alternative would be to share it from `airlock`. Didn't seem like it made a ton of sense to do that since it's not directly in line with airlock's purpose, but can do that if you'd rather.

tobz (approved):

PR #1548, opened 2026-05-01
fix(agent-data-plane): correct JSON parsing for /config response, body           truncation, and scrubber quoting

rahulkaukuntla:
## Summary

Three related bugs that all manifested when running `target/devel/agent-data-plane config` (after having run `DD_DATA_PLANE_USE_NEW_CONFIG_STREAM_ENDPOINT=true make run-adp`) against a live agent that is using config streaming:

**1. Wrong parser for `/config` response (`cli/config.rs`)**
The privileged `/config` endpoint (`ConfigAPIHandler`) returns JSON, but the CLI was deserializing it through `serde_yaml`. YAML is a superset of JSON so small payloads worked accidentally, but the output was re-serialized as YAML which loses structure (e.g. collapses object keys). Fixed by parsing with  `serde_json` and pretty-printing the result.

**2. Response body silently truncated for large payloads (`cli/utils.rs`)** `collect_body` was calling `Buf::chunk().to_vec()` on the aggregated body. `chunk()` returns only the *first contiguous slice* of a `Buf` (typically ~16  KiB), so responses larger than one internal chunk were silently truncated — exactly the case for a full agent config dump. Fixed by calling `Collected::to_bytes()`, which merges all frames before converting.

**3. Scrubber strips closing `"` from JSON string values (`scrubber.rs`)** 
The `password_replacer` regex did not capture the trailing `"` delimiter, so `"password":"secret"` became `"password":"********` — missing the closing quote — which made the scrubbed JSON unparseable. Fixed by adding `(")` as capture group `$4` and including `$4` in the replacement, so the delimiter is  preserved. Added unit tests to assert scrubbed output is still valid JSON.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

- New unit tests in `scrubber.rs` covering JSON password scrubbing, API key 
scrubbing on large payloads, and YAML double-quoted values
- New integration tests in `lib/saluki-config/tests/config_duplicate_keys.rs` 
for both new `GenericConfiguration` methods 
- Full `saluki-common` test suites pass (`cargo test -p
saluki-common -p saluki-config`)
- Clippy clean with `-D warnings` across all affected crates


rahulkaukuntla:
@codex review

lib/saluki-common/src/scrubber.rs
```diff
@@ -114,9 +114,11 @@ impl Default for Scrubber {
             repl_func: None,
         };
 
+        // Capture the optional closing `"` as $4 so the replacement preserves it for JSON values without breaking
+        // unquoted values (plain text / YAML). Without $4, `"password":"secret"` → `"password":"********` (invalid JSON).
         let password_replacer = Replacer {
-            regex: Some(Regex::new(r#"(?i)(\"?(?:pass(?:word)?|pswd|pwd)\"?)((?:=| = |: )\"?)([0-9A-Za-z#!$%&'()*+,\-./:;<=>?@\[\\\]^_{|}~]+)"#).unwrap()),
-            repl: Some(b"$1$2********".to_vec()),
+            regex: Some(Regex::new(r#"(?i)(\"?(?:pass(?:word)?|pswd|pwd)\"?)((?:=| = |: )\"?)([0-9A-Za-z#!$%&'()*+,\-./:;<=>?@\[\\\]^_{|}~]+)(\"?)"#).unwrap()),
```
lucastemb: After running it through Claude and Codex, it looks like the Regex doesn't handle compact JSON password fields (i.e. `{"password":"secret"}`). Compact JSON password values may remain unscrubbed.

bin/agent-data-plane/src/cli/config.rs
```diff
@@ -31,8 +31,18 @@ pub async fn handle_config_command(bootstrap_config: &GenericConfiguration) {
...
+        Ok(v) => v,
+        Err(e) => {
+            error!(
+                "Failed to parse configuration response as JSON after scrubbing (malformed payload or scrubber bug): {:#}",
+                e
+            );
+            std::process::exit(1);
+        }
+    };
+    let formatted = serde_json::to_string_pretty(&config_value).unwrap_or_default();
```
lucastemb: **Nit:** This should probably fail "loudly" with an error upon failure; It currently silently fails by printing an empty config. Consider using `.expect(...)` or match the existing pattern in the file of:
```rust
Ok(x) => x,
Err(e) => {
    error!("Failed to format configuration response as JSON {:#}", e);
    std::process::exit(1);
}
```

lucastemb (changes requested):
Solid work! Just two _small_ comments. 👍

tobz (commented):
All three of the bug fixes mentioned make sense here. What I'm not seeing is the changes to `saluki-config`.

The stated purpose makes sense, but we don't actually _want_ callers to have to drop down to raw maps, messing around with them... that defeats the purpose of the configuration object being _effectively_ read-only and unidirectional: delivered from Core Agent and used as-is by remote agents.

Further, I don't actually see us using it here which makes me skeptical of its value. I think we should remove this portion of the PR for right now and, if anything, open a new PR with it by itself, showing how it's intended to be used, so we can talk about the design and whether or not we want to move forward with something like that.

bin/agent-data-plane/src/cli/config.rs
rahulkaukuntla: I went with matching the existing pattern: [05794cd](https://github.com/DataDog/saluki/pull/1548/commits/05794cd9899096e5642e349d1bcf59acf7e64b1e)

lib/saluki-common/src/scrubber.rs
rahulkaukuntla: I updated the regex to `|:[ ]?` to match compact JSON ("password":"secret"), and added a test case covering it.

rahulkaukuntla:
@tobz you are correct as usual 😅. I removed the saluki-config API's from this PR (in [2e4e7c6](https://github.com/DataDog/saluki/pull/1548/commits/2e4e7c6a1bd6debd3b5d0d46981a41cc85600da1)) and just stuck to fixing the 3 bugs. I'm not sure why the `dd-gitlab/check-deny` job is now failing, and I'm pretty sure that the issue is extraneous to this PR.

lucastemb (approved):
Not sure what to do about the failing `cargo deny` test, so I would defer to @tobz for a decision on that, but your changes LGTM. :shipit:

tobz:
We can bypass for now since the `check-deny` failure isn't related to this PR and will need to be fixed separately.

PR #1550, opened 2026-05-01
chore(deps): fix cargo deny failures (hickory-proto advisories + stale ignore)

jszwedko:
## Summary

Restores a passing `cargo deny` check on `main` by adding targeted ignores for two `hickory-proto` 0.25.x advisories (RUSTSEC-2026-0118, RUSTSEC-2026-0119) and removing the now-resolved stale ignore for RUSTSEC-2026-0097. No code upgrade is possible because `hyper-hickory 0.8` — our only consumer of hickory — pins to `hickory ^0.25` and no newer release of `hyper-hickory` exists. Both advisories are low-risk in this build: RUSTSEC-2026-0118 requires DNSSEC features (`dnssec-ring`/`dnssec-aws-lc-rs`) that are not enabled, and RUSTSEC-2026-0119 requires an attacker to control multi-record DNS messages being encoded, which does not apply to a client-side resolver sending simple outgoing queries.

## Test plan

- [ ] `make check-deny` passes
- [ ] CI `check-deny` job passes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -721,7 +721,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.48.0",
 ]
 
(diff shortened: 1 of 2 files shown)


jszwedko:
Closing since we'll introduce our own wrapper.

Issue #1536, opened 2026-04-30
Add workloads and output assertions to integration tests.


webern:
TODO: write this up better, but...

Basically we have to modes of testing right now:
- correctness: sends randomized payloads through ADP and the Agent both and compares what comes out the other side.
- integration: runs ADP and allows asserting for liveness, exit health and log patterns.

What we need, the ability to send sculpted inputs through ADP and write specific assertions on the output.

Directional proposal:

- Enhance integration tests to add `millstone` + `datadog-intake`
- Add ability to specify inputs and assert on outputs
- Try to keep them fast-ish (< ~30 seconds)
- Potentially run the same integration test against the Agent (for now) which crosses over slightly with correctness testing, but does so with sculpted inputs.

PR #1552, opened 2026-05-01
chore(panoramic): dynamic dispatch with test trait and runner

webern:
## Summary

Test execution was organized around two known test types. This PR introduces a `Test` trait and central `Runner` so that new types can be added without modifying the dispatch or scheduling code.

This PR replaces the enum dispatch with a `Test` trait and a `Runner` struct that owns scheduling, timeout, and cancellation uniformly. Discovery returns `Vec<Box<dyn Test>>`. Each test receives a `TestContext` at execution time carrying a cancellation token and log directory. The `Runner` creates a per-test cancel token, enforces the timeout with a grace period, and propagates program-level cancellation to running tests.

`TestResult` no longer carries filesystem paths - the `Runner` computes log directories and passes them through events. This is a minor change, but `TestResult` carried `log_dir` to a lot of places where it wasn't needed.

The `--no-logs` flag was removed. I didn't see it being used anywhere in the codebase and couldn't think of a great usecase. Meanwhile it was a bit unwieldy to wire it through to where it needed to be.

These changes make it straightforward to add new test types or create test cases dynamically. They just need to implement the trait, be added to the runner, and they will run and report like any other test.

| Change                                         | Reason / Benefit                                       |
|------------------------------------------------|--------------------------------------------------------|
| `Test` trait with `run(tctx)` signature        | New test types implement one trait, no enum arms       |
| `Runner` struct owns scheduling and timeout    | Uniform timeout/cancel for all test types              |
| `TestContext` passed to `run()`                | No stored runtime state                                |
| Per-test `CancellationToken` created by Runner | Runner controls shutdown; tests respond cooperatively  |
| Global cancel propagates to running tests      | Users can cancel via TUI or ctrl-c                     |
| `log_dir` removed from `TestResult`            | TestResult is a bit more pure without it               |
| `--no-logs` CLI flag removed                   | I can put it back if needed, but it was a little wonky |
| `TestRunner` -> `IntegrationRunner`            | Disambiguates from the generic `Runner` scheduler      |
| `TestRunner` -> `CorrectnessRunner`            | Disambiguates from the generic `Runner` scheduler      |
| `DiscoveredTest` enum deleted                  | Replaced by trait objects; removes match arms          |
| `RunArgs` builder for runner configuration     | Reduces parameters for the run_tests function call     |

## Change Type

## How did you test this PR?

- [x] CI
- [ ] Run locally on main
- [ ] Run locally on branch (check for differences)

Using:

```bash
docker ps -q --filter name=airlock- | xargs docker rm -f
docker network prune -f

make \
  build-datadog-intake-image \
  build-millstone-image \
  build-datadog-agent-image-release \
  build-datadog-agent-image \
  build-adp-image && \
cargo run --release --bin panoramic -- \
  run -d test/integration/cases \
      -d test/correctness \
      --no-tui
```

## References

Related, precursor to #1536

### DIFF
--- a/bin/correctness/panoramic/src/correctness/runner.rs
+++ b/bin/correctness/panoramic/src/correctness/runner.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::HashMap,
     future::Future,
-    path::{Path, PathBuf},
+    path::PathBuf,
     pin::Pin,
     task::{ready, Context, Poll},
     time::{Duration, Instant},
@@ -26,19 +26,18 @@ use crate::correctness::{
 use crate::{
     assertions::AssertionResult,
     reporter::{PhaseTiming, TestResult},
+    test::TestContext,
 };
 
 /// Run a single correctness test and return a panoramic `TestResult`.
-pub async fn run_correctness_test(
-    name: String, config: Config, log_dir: Option<PathBuf>, mounts_dir: PathBuf,
-) -> TestResult {
+pub async fn run_correctness_test(name: String, config: Config, tctx: TestContext) -> TestResult {
     let started = Instant::now();
 
     // Phase 1: spawn containers
     let spawn_start = Instant::now();
-    let test_runner = match TestRunner::from_config(&config, log_dir.clone(), &mounts_dir).await {
+    let test_runner = match CorrectnessRunner::from_config(&config, tctx).await {
         Ok(r) => r,
-        Err(e) => return make_error_result(name, started, "spawn_containers", e, log_dir),
+        Err(e) => return make_error_result(name, started, "spawn_containers", e),
     };
     let spawn_duration = spawn_start.elapsed();
 
@@ -46,7 +45,7 @@ pub async fn run_correctness_test(
     let collect_start = Instant::now();
     let (baseline_data, comparison_data) = match test_runner.run().await {
         Ok(data) => data,
-        Err(e) => return make_error_result(name, started, "collect_data", e, log_dir),
+        Err(e) => return make_error_result(name, started, "collect_data", e),
     };
     let collect_duration = collect_start.elapsed();
 
@@ -93,7 +92,6 @@ pub async fn run_correctness_test(
             }],
             error: None,
             phase_timings,
-            log_dir,
             assertion_details: vec![],
         },
         Err((e, details)) => {
@@ -111,16 +109,13 @@ pub async fn run_correctness_test(
                 }],
                 error: Some(summary),
                 phase_timings,
-                log_dir,
                 assertion_details: vec![details],
             }
         }
     }
 }
 
-fn make_error_result(
-    name: String, started: Instant, phase: &str, e: GenericError, log_dir: Option<PathBuf>,
-) -> TestResult {
+fn make_error_result(name: String, started: Instant, phase: &str, e: GenericError) -> TestResult {
     TestResult {
         name,
         passed: false,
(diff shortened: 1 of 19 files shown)


bin/correctness/panoramic/src/correctness/runner.rs
```diff
@@ -285,9 +281,9 @@ impl TestRunner {
             .unwrap_or_shutdown("collect_data", maybe_baseline_data.await, maybe_comparison_data.await)
             .await?;
 
-        // We've gotten our data back, so signal to any remaining containers that they can shutdown now.
```
webern, line 288: restore this comment

bin/correctness/panoramic/src/main.rs
```diff
@@ -122,65 +122,48 @@ async fn run_tests(mut cmd: cli::RunCommand, use_tui: bool) -> ExitCode {
...
-    ));
+    let runner_handle = tokio::spawn(async move { registry.run_tests(args).await });
 
     // Spawn the appropriate consumer based on mode.
     let all_passed = if use_tui {
-        run_with_tui_consumer(rx, cancel_token, log_dir, runner_handle).await
+        run_with_tui_consumer(rx, cancel_all, Some(log_dir), runner_handle).await
     } else {
-        run_with_logging_consumer(rx, &cmd, log_dir, runner_handle).await
+        run_with_logging_consumer(rx, &cmd, Some(log_dir), runner_handle).await
```
thieman, line 177: cancel doesn't go into this path?

webern: No, I ran out of time trying to get this done, but I don't think we already had a signal capture for ctrl-c in the program and I... ran out of time. I think we just need to trap ctrl-c and wire it up.

bin/correctness/panoramic/src/runner.rs
```diff
@@ -23,210 +23,287 @@ use tokio::sync::{mpsc, RwLock, Semaphore};
...
-            mounts_dir,
-            cancel_token,
-        )
-        .await
-    };
+/// A sender that the `Runner` can use to send `TestEvent`s.
+pub(crate) type EventSender = mpsc::UnboundedSender<TestEvent>;
+
+/// The amount of time a test has to clean up after cancellation or timing out.
+const GRACE_TIME: Duration = Duration::from_secs(5);
```
thieman: This will likely be pretty tight for kind tests that tear down the cluster after themselves, maybe give it 30s just to be accommodating?

bin/correctness/panoramic/src/runner.rs
```diff
@@ -23,210 +23,287 @@ use tokio::sync::{mpsc, RwLock, Semaphore};
...
 
-        let failed = !result.passed;
-        let _ = event_tx.send(TestEvent::TestCompleted { result: result.clone() });
-        results.push(result);
+/// Owns the set of registered tests and orchestrates their execution.
+///
+/// Handles test filtering, parallelism, timeout enforcement, cancellation propagation,
+/// log directory creation, and event dispatch. Individual test logic lives in the `Test`
+/// implementations.
+pub(crate) struct Runner {
```
thieman, line 100: Will this also hold options passed from the CLI? An example I was thinking of adding was a `--no-delete-kind-cluster` flag that would need to eventually get passed down to the correctness test

webern: My idea for that is `test.rs`:

```
    /// A directory from which files should be mounted into one or more of the domain-specific containers used in this
    /// test.
    // TODO: this is a hack introduced to support the PANORAMIC_DYNAMIC feature. Consider generalizing if needed.
    // For example: this could become runtime_config: HashMap<String, String> for shuttling domain specific items from
    // runtime to a test.
    mounts_dir: PathBuf,
```

Instead of mounts_dir there we could have runtime_config: HashMap<String, String>

And you could pass stuff that way.

But... I'm not sure `--no-delete-kind-cluster` at the CLI sounds great.

What I actually think is that there should probably be a whole new `Test` type for Kind cluster tests.

It bears thinking through design tradeoffs.

The issue with something like `--no-delete-kind-cluster` at the CLI is that now we are pushing extremely specific stuff into the CLI whereas panoramic might be capable of running tests that don't use containers or can't run on kind. So I think we should consider how we design things.

thieman: > But... I'm not sure --no-delete-kind-cluster at the CLI sounds great.

It's not really a property of the test itself, it's a runtime option for how you'd like to actually execute the tests. In this case the idea would be "I'm futzing with this test locally and will need to run it a bunch, so let's skip starting and killing a full kind cluster each time." The CLI seems like the natural place for those, I shouldn't need to futz with YAML for a temporary change I don't want to commit.

> What I actually think is that there should probably be a whole new Test type for Kind cluster tests.

In my current PR the config for it is limited to a single `runtime` field on the correctness test def, if it gets more complex than that could certainly consider splitting. But if they're 99% equivalent, seems like they can share.

thieman (approved):

bin/correctness/panoramic/src/runner.rs
webern: I would think of that in terms of test lifecycle being something like this:

- test resource set-up
- test execution
- test resource teardown
 
And what you're asking panoramic to do is `--skip-teardown` which could be applicable to other test types as well.

Edit: in fact, `--skip-teardown` can be important sometimes if you need to get in there and see what went wrong.

bin/correctness/panoramic/src/runner.rs
webern: Seems like a mistake, I thought I had 30 there. Oops, will change

bin/correctness/panoramic/src/runner.rs
thieman: Skipping the kind cluster teardown specifically is more granular than that, skipping teardown entirely would also skip cleaning up any of the test-specific namespaces, pods, containers, etc. Not saying we shouldn't support `--skip-teardown`, we should, but I think a general mechanism for passing CLI flags down to Test implementations will be necessary in either case.

webern: I think my ideal, when it comes to kubernetes, would be to create the cluster and tear it down without necessarily even caring about namespaces, etc. I can't remember if `kind` is one of those tools that needs a cluster to create your cluster... but in general, it's much cleaner if you can create and discard clusters.

thieman: That's the default, but if you're intending to run the test repeatedly on your Mac then that's going to add ~2 minutes to each test run that you'd probably rather not add. It's purely a convenience option.

bin/correctness/panoramic/src/main.rs
webern: Added this. It was a tiny thing, I was just feeling rushed on Friday to get the PR open.

bin/correctness/panoramic/src/runner.rs
webern: Fixed.

tobz (approved):

PR #1558, opened 2026-05-01
feat(dogstatsd): Add post aggregation histogram filter component

aqian01:
## Summary

DogStatsD histograms are expanded after aggregation into metrics like <name>.max, <name>.avg, <name>.count, and <name>.95percentile.

This filter runs after aggregation so we can drop those generated histogram metrics, since they do not exist yet when the original DogStatsD sample is received.

This PR adds an ADP-only DogStatsD post-aggregate filter transform that runs immediately after `dsd_agg`. 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests

### DIFF
--- a/bin/agent-data-plane/src/components/dogstatsd_post_aggregate_filter/mod.rs
+++ b/bin/agent-data-plane/src/components/dogstatsd_post_aggregate_filter/mod.rs
@@ -1,9 +1,6 @@
 //! DogStatsD post-aggregate metric filter transform.
 //!
 //! Drops post-aggregation scalar metrics whose generated histogram aggregate names match the metric filterlist.
-
-mod telemetry;
-
 use async_trait::async_trait;
 use memory_accounting::{MemoryBounds, MemoryBoundsBuilder};
 use saluki_config::GenericConfiguration;
@@ -25,6 +22,8 @@ use serde::Deserialize;
 use stringtheory::MetaString;
 use tokio::select;
 use tracing::{debug, error};
+
+mod telemetry;
 
 use self::telemetry::Telemetry;
 
@@ -89,7 +88,7 @@ impl DogStatsDPostAggregateFilterConfiguration {
     ///
     /// If the configuration cannot be deserialized, an error is returned.
     pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
-        let mut typed_config: DogStatsDPostAggregateFilterConfiguration = config.as_typed()?;
+        let mut typed_config: Self = config.as_typed()?;
         typed_config.configuration = Some(config.clone());
         Ok(typed_config)
     }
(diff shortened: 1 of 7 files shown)


bin/agent-data-plane/src/components/dogstatsd_post_aggregate_filter/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+}
+
+impl DogStatsDPostAggregateFilterConfiguration {
+    /// Creates a new `DogStatsDPostAggregateFilterConfiguration` from the given configuration.
+    ///
+    /// # Errors
+    ///
+    /// If the configuration cannot be deserialized, an error is returned.
+    pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
+        let mut typed_config: DogStatsDPostAggregateFilterConfiguration = config.as_typed()?;
```
tobz: Can just do `Self` as the type to keep things a little shorter. 😅

bin/agent-data-plane/src/components/dogstatsd_post_aggregate_filter/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+struct EffectiveFilterlist {
+    metric_filterlist: Vec<String>,
+    metric_filterlist_match_prefix: bool,
+    metric_blocklist: Vec<String>,
+    metric_blocklist_match_prefix: bool,
+}
```
tobz, line 191: Just saying this out loud so I don't forget it, but it's not a change we need to/should make here: since this code is now duplicated between the DSD Prefix Filter transform _and_ the Tag Filterlist transform _and_ this transform... it could potentially make sense for us to pull the DSD Prefix Filter transform into the `agent-data-plane` crate and stick these helper types in a common module where they can all be depended on.

bin/agent-data-plane/src/components/dogstatsd_post_aggregate_filter/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+    #[test]
+    fn sketch_metrics_are_not_filtered() {
+        let filter = noop_filter(vec!["request.duration.max"], false, vec![], false);
+
+        let names = filter_metric_names(
+            &filter,
+            vec![
+                Metric::distribution("request.duration.max", [1.0, 2.0, 3.0]),
+                Metric::histogram("request.duration.max", [1.0, 2.0, 3.0]),
+                Metric::gauge("request.duration.max", 1.0),
```
tobz: These should all have different prefixes, since otherwise the test isn't actually able to disambiguate _which_ metrics are being filtered out.

aqian01:
There is also a rounding change to match Core Agent behavior. The Agent parses histogram percentiles with `int(i*100 + 0.5)`, so values like `0.299` map to `30percentile`, not `29percentile`.

Agent reference: https://github.com/DataDog/datadog-agent/blob/12213fe95538f47d98d73bd945a87b3e24189285/pkg/metrics/histogram.go#L53-L70

bin/agent-data-plane/src/components/dogstatsd_post_aggregate_filter/mod.rs
```diff
@@ -0,0 +1,695 @@
+//! DogStatsD post-aggregate metric filter transform.
+//!
+//! Drops post-aggregation scalar metrics whose generated histogram aggregate names match the metric filterlist.
+
+mod telemetry;
```
tobz: Move this down, right above the `use self::telemetry::Telemetry;` line.

tobz (changes requested):
Mostly nits, but one change related to one of the unit tests.

tobz (approved):

PR #1559, opened 2026-05-01
chore(dev): unify correctness tools into the same container image

### LINKED ISSUE #1547: Improve local rebuild times of `millstone` and `datadog-intake` images.
`millstone` and `datadog-intake` should receive a similar fix as `agent-data-plane` such that they don't rebuild the world each time you rebuild them locally. See #1479 - these probably need something quite similar.

Edit: this could probably all be done in the same Dockerfile if it isn't already 🤔

tobz:
## Summary

This PR unifies the container images for `datadog-intake` and `millstone` into a single new container image, `correctness-tools`, in order to reduce build time and to allow for aggressive reuse of cached compilation artifacts.

We've applied the same optimizations to this new Dockerfile as we did to the ADP one: we utilize Docker's support for `cache` mounts to reuse the Cargo registry, and incremental/intermediate build artifacts, between builds. This allows near-native build performance -- change a single file, don't rebuild the entire world -- across image builds. Additionally, since we're building both tools in the same container, the heavy overlap in dependencies lets us build the second binary much faster than before, even excluding the build caching itself.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Making sure the new Make target runs locally, and that CI passes cleanly.

## References

Closes #1547


tobz:
* **#1559** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1559?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1559?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

webern (approved):
Nice. Actually when I opened the issue I was thinking the ADP build, millstone and datadog-intake could all still be independent container images, but that they could be built in the same Dockerfile using `buildx` target, 

e.g.

`FROM ubuntu-or-whatever as rust-build-setup`

Set's up everything needed to build our software

`FROM rust-build-setup as build-adp`
`FROM rust-build-setup as build-millstone`
`FROM rust-build-setup as build-datadog-intake`

Each of those builds one thing

`FROM build-adp as adp`
`FROM build-millsone as millstone`
`FROM build-datadog-intake as datadog-intake`

Each of those copies one binary

Each is using perhaps the same build cache

And you `docker build --target adp` to get the final ADP container.

Anyway, your solution get's most of the benefit I think.... and perhaps the ADP build does a lot that millstone and datadog-intake don't need for their builds.

tobz:
> Nice. Actually when I opened the issue I was thinking the ADP build, millstone and datadog-intake could all still be independent container images, but that they could be built in the same Dockerfile using `buildx` target,
> 
> e.g.
> 
> `FROM ubuntu-or-whatever as rust-build-setup`
> 
> Set's up everything needed to build our software
> 
> `FROM rust-build-setup as build-adp` `FROM rust-build-setup as build-millstone` `FROM rust-build-setup as build-datadog-intake`
> 
> Each of those builds one thing
> 
> `FROM build-adp as adp` `FROM build-millsone as millstone` `FROM build-datadog-intake as datadog-intake`
> 
> Each of those copies one binary
> 
> Each is using perhaps the same build cache
> 
> And you `docker build --target adp` to get the final ADP container.
> 
> Anyway, your solution get's most of the benefit I think.... and perhaps the ADP build does a lot that millstone and datadog-intake don't need for their builds.

@webern Good point about better reuse of build cache across _all_ of those binary targets and build tooling unification. Could definitely be a follow up.

PR #1560, opened 2026-05-01
feat(dogstatsd): add support for `skip_ssl_validation`

### LINKED ISSUE #1371: Investigate whether `skip_ssl_validation` needs to be supported or documented.
The core agent exposes [`skip_ssl_validation`][1] (default `false`). When enabled, it sets [`InsecureSkipVerify: true`][2] on the Go TLS config, disabling certificate verification for all outbound HTTPS connections. This is a debug feature for environments with self-signed certs or corporate proxies doing TLS interception.

ADP's TLS builder has a [`danger_accept_invalid_certs`][s1] method that achieves the same thing (installs an [`AcceptAllServerCertVerifier`][s2]), but it is only used in [test code][s3]. There is no config key reading `skip_ssl_validation` and no production code path that enables it.

This is something we would prefer not to add to ADP. Investigate whether customers are using this setting in practice. If usage is negligible, it may be fine to leave as-is. If customers depend on it, decide whether to implement or document that ADP does not support it.

[1]: https://github.com/DataDog/datadog-agent/blob/141f12062790/pkg/config/setup/common_settings.go#L1087
[2]: https://github.com/DataDog/datadog-agent/blob/141f12062790/pkg/util/http/transport.go#L82
[s1]: https://github.com/DataDog/saluki/blob/e9373ef98013/lib/saluki-tls/src/lib.rs#L106-L107
[s2]: https://github.com/DataDog/saluki/blob/e9373ef98013/lib/saluki-tls/src/lib.rs#L123-L130
[s3]: https://github.com/DataDog/saluki/blob/e9373ef98013/lib/saluki-components/src/common/datadog/proxy.rs#L742

lucastemb:
## Summary

This PR aims to add support for `skip_ssl_validation` for customers and environments with self-signed certs or corporate proxies doing TLS interception.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Mainly through unit tests

## References

- Closes: https://github.com/DataDog/saluki/issues/1371


lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -599,3 +643,168 @@ impl<T: Retryable> PendingTransactions<T> {
...
+    use tokio_rustls::TlsAcceptor;
+
+    use super::*;
+
+    fn forwarder_config_from_value(value: serde_json::Value) -> ForwarderConfiguration {
+        serde_json::from_value(value).expect("ForwarderConfiguration should deserialize")
+    }
+
+    fn init_tls_crypto_provider() {
+        static INIT: OnceLock<()> = OnceLock::new();
```
tobz, line 674: Can we add a TODO comment here to eventually clean this up? Process-wide side effects like this -- setting environment variables, initializing the crypto provider, etc -- are fragile to do directly in unit tests and usually indicate that we might have a need for a different / better pattern.

```suggestion
        // TODO: Figure out a better pattern for testing that doesn't involve initializing
        // the process-wide crypto provider.
        static INIT: OnceLock<()> = OnceLock::new();
```

tobz (approved):
Just one comment I want to make sure we leave behind but other than that, makes sense to me. 👍🏻

PR #1565, opened 2026-05-04
feat(dogstatsd): add units for timed metrics

atanzu:
## Summary
Send unit information for aggregate metrics calculated from timed metrics.

Currently there is no way to distinguish between abstract histograms and timed metrics which are represented as histograms. To disambiguate, we send the unit information with the aggregate metrics calculated from the histogram values.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?
* added autotests to validate the new code.
* did an end-to-end test with a personal account to confirm that timed metrics have units: 
Screenshot 2026-05-04 at 11 34 49

## References

- Closes: AGTMETRICS-433


tobz (changes requested):
We should use `MetaString` instead of `Arc<str>` because otherwise we're always allocating for a string that can be inlined.

tobz (approved):
LGTM. 👍🏻

Issue #1567, opened 2026-05-04
OLTP tests contend for DogStatsD port (was Running correctness tests reliably fails in my environment).


webern:
From this commit on main [507d91](https://github.com/DataDog/saluki/commit/507d913e4d1245250311a3ab992c18b390f5ad21), which happens to exactly match with v0.1.36 of ADP. The following commands reliably fail. On Friday fellow team members were unable to repro, but I'm not sure if they followed these exact repro steps. Also the list of tests that fail seems consistent.

Note: I experienced the same failures on Friday with https://github.com/DataDog/saluki/commit/86ffd2e799047ea7da7e5fd97c5c2469f09a7a12

Note that I did the following first:

```bash
docker rm -f $(docker ps -aq)
docker rmi -f $(docker images -aq)
docker system prune -af
```

I am using lima, so I completely nuked and rebuilt the Lima VM

```bash
limactl delete --force docker
limactl create --name=docker ~/some/path/lima/config/docker.yaml
limactl start docker
```

Then I built the containers (this succeeded):

```bash
echo "--------------------------------------------------------------------------";
echo "make build-datadog-intake-image"
make build-datadog-intake-image
echo "--------------------------------------------------------------------------";
echo "make build-millstone-image"
make build-millstone-image
echo "--------------------------------------------------------------------------";
echo "make build-datadog-agent-image-release"
make build-datadog-agent-image-release
echo "--------------------------------------------------------------------------";
echo "make build-datadog-agent-image "
make build-datadog-agent-image 
echo "--------------------------------------------------------------------------";
echo "make build-adp-image"
make build-adp-image
echo "--------------------------------------------------------------------------";
```

Then I ran the tests like this:

```bash
cargo run --release --bin panoramic -- \
  run -d test/integration/cases \
      -d test/correctness \
      --no-tui
```

Which resulted in these failures:

```text
2026-05-04T10:05:44.275385Z ERROR Baseline group runner completed step 'collect_data' successfully, but comparison group runner encountered an error.
FAIL otlp-traces-probabilistic (132.68s)
  Error: Failed to drive millstone to completion; process exited with non-zero exit code (1). Error message: <no error message provided>
  Phase timings:
    collect_data (132.68s)
  Logs: /var/folders/dc/w_0dw8vs1n5cx5snqzx3k9yh0000gp/T/panoramic-20260504-115823/correctness/otlp-traces-probabilistic
[...]

PR #1570, opened 2026-05-04
chore(dogstatsd): do not run dogstatsd twice in oltp tests

webern:
## Summary

OTLP correctness tests never disabled DogStatsD, which was fine when it defaulted to off. After the default flipped to on, ADP races the Core Agent for UDP port 8125 and crash-loops if it loses. Disable DogStatsD explicitly in OTLP test configs.

## Change Type
- [x] Bug fix

## How did you test this PR?

The tests were reliably failing for me before this change and passed after the change.

## References

#1567


webern:
Apparently this is the wrong approach. Closing.

jszwedko (approved):

thieman (approved):

PR #1575, opened 2026-05-04
fix(dogstatsd): Move the DSD prefix/filter component in front of the DSD enrich component

aqian01:
## Summary

Move the DogStatsD mapper before dsd_prefix_filter.

[Issue](https://datadoghq.atlassian.net/browse/DADP-67)

This makes ADP match the Core Agent order: map the metric name first, then apply statsd_metric_namespace and metric blocklist/filterlist rules. This fixes cases where blocklist entries for mapped metric names were not applied.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

The new correctness test sends the same DogStatsD traffic through Core Agent and ADP, then compares the emitted metrics.

It generates two raw metric names:

`mapper.blocklist.raw`
`mapper.blocklist.keep`
The DogStatsD mapper rewrites them as:

`mapper.blocklist.raw` -> `mapper.blocklist.mapped`
`mapper.blocklist.keep` -> `mapper.blocklist.allowed`
The config blocklists only the mapped name:
```
statsd_metric_blocklist:
  - mapper.blocklist.mapped
  ```
So the expected result is:

`mapper.blocklist.mapped` is dropped after remapping
`mapper.blocklist.allowed` is still emitted

If ADP applies the blocklist before the mapper, it checks `mapper.blocklist.raw`, misses the blocklist, then remaps it to `mapper.blocklist.mapped` and emits it. The correctness test catches that because Core Agent drops it but ADP would leak it.

### DIFF
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -464,17 +464,17 @@ async fn add_dsd_pipeline_to_blueprint(
     //               │                 │                          │ service checks           │ events
     //               │                 ▼                          ▼                          ▼
     //               │      ┌─────────────────────┐    ┌─────────────────────┐    ┌─────────────────────┐
-    //               │      │  DSD Prefix/Filter  │    │  Service Checks     │    │   Events Enrich     │
-    //               │      │     (transform)     │    │  Enrich (chained)   │    │    (chained)        │
-    //               │      └─────────────────────┘    └─────────────────────┘    └─────────────────────┘
-    //               │                 │                          │                          │
-    //               │                 ▼                          ▼                          ▼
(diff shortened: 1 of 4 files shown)


tobz (approved):
I think this makes sense to me if that's how we're doing it in the Core Agent.

Saying it out loud to try and not forget, but this is something that would be good to have a correctness test for... but I think we currently lack the ability to have `lading` spit out metrics with certain prefixes, which makes this infeasible to currently test.

Not a blocker to merging.

aqian01:
@tobz yup im working on adding the correctness test right now

tobz:
> @tobz yup im working on adding the correctness test right now

Ohhh, nice. I forgot about the support for specific metrics that got added to help with ASTA testing. 🙌🏻

PR #1591, opened 2026-05-05
enhancement(app): drive memory limiting behavior through configurable modes

tobz:
## Summary

This PR adds support for configuring memory limiting behavior, and by extension, the validation of memory bounds, through a set of configurable memory "modes."

Presently, memory limiting behavior is always on: if we detect a memory limit (either specified through configuration, or extracted from the ambient cgroups limits), we validate the bounds of the configured topology and either proceed or exit if we detect that we cannot firmly adhere to the configured bounds. Technically correct, but not particularly flexible... especially during the early stage of introducing ADP to the world.

In this PR, we address this by introducing the concept of a memory "mode": a way to specify how memory limiting and bounds validation should work. At the core, we expose three modes:

- disabled (do nothing, go forth and allocate freely)
- permissive (evaluate bounds and _warn_ if we may exceed them)
- strict (evaluate bounds and _exit_ if we may exceed them)

We default to **disabled** to ensure that users have an equivalent experience when using ADP as they're used to with the Datadog Agent. Permissive and strict are designed to allow for gradually opting in to memory limiting behavior: allowing things like the global memory limiter to actively influence component behavior, dialing in configuration settings to explore the resulting memory bounds, eventually culminating in (hopefully) switching to strict mode to lock in their tuning.

The global memory limiter is also left as-is, in terms of being configurable but defaulting to enabled. However, in the "disabled" mode, we simply set the memory limiter's grant to an (essentially) unbounded value. This was simply the expedient design choice but I'm thinking I may forcefully disable it when in the "disabled" memory mode just to be a little cleaner.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Existing tests, and five new integration tests that exercise the different memory modes.

## References

DADP-2

### DIFF
--- a/lib/memory-accounting/src/grant.rs
+++ b/lib/memory-accounting/src/grant.rs
@@ -53,15 +53,6 @@ impl MemoryGrant {
         Self::with_slop_factor(effective_limit_bytes, 0.0)
     }
 
-    /// Creates a new memory grant with the maximum possible effective limit.
-    ///
-    /// Useful for callers that want a valid grant whose effective limit is large enough that any reasonable memory
-    /// usage will fall well below it (e.g. when memory limiting should be effectively disabled but a real limiter
-    /// instance is still desired).
-    pub fn unbounded() -> Self {
-        Self::effective(MAX_GRANT_BYTES).expect("MAX_GRANT_BYTES is a valid grant size")
-    }
-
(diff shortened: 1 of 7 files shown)


lucastemb (approved):
I know you mentioned wanting to go over the doc comments, but the code looks good to me 👍

Issue #1540, opened 2026-04-30
`additional_endpoints` API keys not refreshed dynamically at runtime


jszwedko:
The ADP forwarder supports dynamic API key refresh for the primary endpoint — \`ResolvedEndpoint\` re-reads from the live config on each call via \`endpoint.api_key()\`. However, \`additional_endpoints\` are built as static \`ResolvedEndpoint\` instances without a config reference, so their API keys cannot be updated without a restart.

This means if a user defines \`additional_endpoints\` with secrets (e.g., \`ENC[]\` blocks) and the Core Agent performs a secret refresh, ADP silently ignores the updated values.

## Impact

Any user relying on runtime secret rotation for \`additional_endpoints\` will have stale credentials until restart.

## Proposed solution

Mirror Core Agent's approach in [\`domain_resolver.go:120-156\`](https://github.com/DataDog/datadog-agent/blob/23a2b178760d412ab88277107e5b95d2bc95fe57/comp/forwarder/defaultforwarder/resolver/domain_resolver.go#L120-L156) — give \`additional_endpoints\` resolved endpoints a reference to the live config so API keys are re-queried on each use.

## References

[Slack thread](https://dd.slack.com/archives/C070PQKLLP5/p1777500498095819)

PR #1595, opened 2026-05-06
feat(io): Retry transactions if secrets are enabled

atanzu:
## Summary
Retry transient 403 Forbidden responses in the Datadog forwarder when secrets management is active, so requests in flight during an API key refresh aren't dropped.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?
 - Unit tests for `StandardHttpClassifier` predicate management (`with_/set_/remove_status_code_predicate`) and `DefaultHttpRetryPolicy::with_backoff_and_classifier`.
  - Unit tests for `is_secrets_in_use` across the four `secret_*` config keys.
  - Unit test that mutates the underlying `GenericConfiguration` and confirms the same retry-policy instance flips its 403 decision.
  - End-to-end integration tests in `lib/saluki-components/src/common/datadog/io.rs` using a test TCP server, asserting 403 is not retried without secrets and is retried when secrets are in use.

## References

- AGTMETRICS-504

### DIFF
--- a/lib/saluki-components/src/common/datadog/io.rs
+++ b/lib/saluki-components/src/common/datadog/io.rs
@@ -83,7 +83,8 @@ where
 /// Transaction forwarder for Datadog endpoints.
 pub struct TransactionForwarder<B> {
     context: ComponentContext,
-    config: ForwarderConfiguration,
+    config: ForwarderConfiguration,            // static snapshot of forwarder settings
+    live_config: Option<GenericConfiguration>, // runtime-mutable configuration
     telemetry: ComponentTelemetry,
     metrics_builder: MetricsBuilder,
     client: HttpClient,
@@ -141,13 +142,13 @@ where
 {
     /// Creates a new `TransactionForwarder` instance from the given configuration.
     pub fn from_config<F>(
-        context: ComponentContext, config: ForwarderConfiguration, configuration: Option<GenericConfiguration>,
+        context: ComponentContext, config: ForwarderConfiguration, live_config: Option<GenericConfiguration>,
         endpoint_name: F, telemetry: ComponentTelemetry, metrics_builder: MetricsBuilder,
     ) -> Result<Self, GenericError>
     where
         F: Fn(&Uri) -> Option<MetaString> + Send + Sync + 'static,
     {
-        let endpoints = config.endpoint().build_resolved_endpoints(configuration)?;
+        let endpoints = config.endpoint().build_resolved_endpoints(live_config.clone())?;
         let mut client_builder = HttpClient::builder()
             .with_request_timeout(config.request_timeout())
             .with_bytes_sent_counter(telemetry.bytes_sent().clone())
@@ -167,6 +168,7 @@ where
         Ok(Self {
             context,
             config,
+            live_config,
             telemetry,
             metrics_builder,
             client,
@@ -186,6 +188,7 @@ where
         let Self {
             context,
             config,
+            live_config,
             telemetry,
             metrics_builder,
             client,
@@ -200,6 +203,7 @@ where
                 io_shutdown_tx,
                 context,
                 config,
+                live_config,
                 client,
                 telemetry,
                 metrics_builder,
@@ -214,10 +218,12 @@ where
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 async fn run_io_loop<B>(
     mut transactions_rx: mpsc::Receiver<Transaction<B>>, io_shutdown_tx: oneshot::Sender<()>,
-    context: ComponentContext, config: ForwarderConfiguration, service: HttpClient, telemetry: ComponentTelemetry,
-    metrics_builder: MetricsBuilder, resolved_endpoints: Vec<ResolvedEndpoint>,
+    context: ComponentContext, config: ForwarderConfiguration, live_config: Option<GenericConfiguration>,
+    service: HttpClient, telemetry: ComponentTelemetry, metrics_builder: MetricsBuilder,
+    resolved_endpoints: Vec<ResolvedEndpoint>,
 ) where
     B: Body + Buf + Clone + Send + Sync + 'static,
     B::Data: Send,
@@ -243,6 +249,7 @@ async fn run_io_loop<B>(
                 task_barrier,
                 context.clone(),
                 config.clone(),
+                live_config.clone(),
                 service.clone(),
                 telemetry.clone(),
                 txnq_telemetry,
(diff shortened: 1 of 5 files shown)


lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -311,7 +319,7 @@ async fn run_endpoint_io_loop<B>(
         .map_request(with_version_info())
         .concurrency_limit(config.endpoint_concurrency())
         .layer(RetryCircuitBreakerLayer::new(
-            config.retry().to_default_http_retry_policy(),
+            config.retry().to_default_http_retry_policy(live_config.clone()),
```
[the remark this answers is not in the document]
atanzu: Correct, I will follow-up on that as a part of #1540

lib/saluki-io/src/net/util/retry/classifier/http.rs
```diff
@@ -11,24 +19,207 @@ use super::RetryClassifier;
...
+                match status {
+                    // There's some status codes that likely indicate a fundamental misconfiguration or bug on the
+                    // client side which won't be resolved by retrying the request.
+                    StatusCode::BAD_REQUEST
+                    | StatusCode::UNAUTHORIZED
+                    | StatusCode::FORBIDDEN
+                    | StatusCode::PAYLOAD_TOO_LARGE => false,
+
+                    // For all other status codes, we'll only retry if they're in the client/server error range.
+                    status => status.is_client_error() || status.is_server_error(),
```
webern, line 86: Just a thought. The predicate idea could be more open to extensibility if it is just a functor that takes the response and returns a bool.

```rust
pub type HttpRetryPredicate<B> = Arc<dyn Fn(&http::Response<B>) -> bool + Send + Sync>;

fn default_should_retry<B>(resp: &http::Response<B>) -> bool {
    match resp.status() {
        StatusCode::BAD_REQUEST
        | StatusCode::UNAUTHORIZED
        | StatusCode::FORBIDDEN
        | StatusCode::PAYLOAD_TOO_LARGE => false,
        _ => resp.status().is_client_error() || resp.status().is_server_error(),
    }
}

pub struct StandardHttpClassifier<B> {
    predicates: Vec<HttpRetryPredicate<B>>,
}

impl<B: 'static> StandardHttpClassifier<B> {
    /// Starts with the default status-code predicate installed.
    pub fn new() -> Self {
        Self { predicates: vec![Arc::new(default_should_retry)] }
    }

    /// Starts with no predicates; all responses are non-retriable until one is added.
    pub fn empty() -> Self {
        Self { predicates: Vec::new() }
    }

    /// Adds a predicate. OR semantics: retry if any predicate returns `true`.
    pub fn with_predicate(mut self, predicate: HttpRetryPredicate<B>) -> Self {
        self.predicates.push(predicate);
        self
    }
}

impl<B, Error> RetryClassifier<http::Response<B>, Error> for StandardHttpClassifier<B> {
    fn should_retry(&self, response: &Result<http::Response<B>, Error>) -> bool {
        match response {
            Ok(resp) => self.predicates.iter().any(|p| p(resp)),
            Err(_) => true,
        }
    }
}
```

The secrets predicate in `retry.rs` becomes:

```rust
let secrets_gate: HttpRetryPredicate<_> = Arc::new(move |resp| {
    resp.status() == StatusCode::FORBIDDEN && is_secrets_in_use(&config)
});
let classifier = StandardHttpClassifier::new().with_predicate(secrets_gate);
```

This removes the `HashMap<StatusCode, predicate>` keying and the `set_/remove_status_code_predicate`
surface - predicates own their own logic and have the full response available.

lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -186,6 +188,7 @@ where
         let Self {
```
webern, line 188: Not a request to fix in this PR, but this destructure is where the `too_many_arguments` warnings (suppressed a few lines down) seem to originate. All these fields are owned by `self`, destructured here, and then passed individually through `run_io_loop` and `run_endpoint_io_loop`.

Perhaps `self` could be left in-tact here and then passed through as a single argument. Not sure.

```suggestion
        // TODO: do not destructure self as a way to fix the #[allow(clippy::too_many_arguments)] annotations
        let Self {
```

lib/saluki-components/src/common/datadog/retry.rs
```diff
@@ -177,26 +181,90 @@ impl RetryConfiguration {
...
+///
+/// This is computed dynamically from `config` so that runtime configuration updates (delivered over the dynamic
+/// configuration stream from the Core Agent) are reflected on every call. The gate is enabled when any of the
+/// following are true:
+///
+/// - `secret_refresh_on_api_key_failure_interval` is greater than zero
+/// - `secret_refresh_interval` is greater than zero
+/// - `secret_backend_type` is set to a non-empty string
+/// - `secret_backend_command` is set to a non-empty string
+pub fn is_secrets_in_use(config: &GenericConfiguration) -> bool {
```
webern, line 219: The four keys this reads (`secret_backend_command`, `secret_backend_type`, `secret_refresh_interval`, `secret_refresh_on_api_key_failure_interval`) describe whether secrets management is configured, not whether a refresh is currently in progress. I believe these are set at startup and not expected to change at runtime (right?), so the dynamic evaluation on every 403 may not be necessary and they could be read just once.

If that's right, a `bool` computed at construction time in `from_config` would behave identically:

```rust
let retry_403_on_secrets = live_config.as_ref().map_or(false, is_secrets_in_use);
```

That `bool` could then be captured directly in the predicate closure, and `Option<GenericConfiguration>` wouldn't need to be threaded through `TransactionForwarder`, `run_io_loop`, and `run_endpoint_io_loop`.

lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -141,13 +142,13 @@ where
 {
     /// Creates a new `TransactionForwarder` instance from the given configuration.
     pub fn from_config<F>(
-        context: ComponentContext, config: ForwarderConfiguration, configuration: Option<GenericConfiguration>,
+        context: ComponentContext, config: ForwarderConfiguration, live_config: Option<GenericConfiguration>,
```
webern, line 145: If the `is_secrets_in_use` question (`retry.rs:219`) resolves to a `bool`, this parameter goes away and the threading through `run_io_loop` / `run_endpoint_io_loop` goes with it.

webern (commented):
There might be a way to simplify if my comments are correct...

PR #1601, opened 2026-05-06
enhancement(dogstatsd): wire `dogstatsd_eol_required` through codec config

### LINKED ISSUE #1339: Wire `dogstatsd_eol_required` into DogStatsD codec configuration.
The DogStatsD codec supports requiring end-of-line delimiters, but this may not be wired into configuration. The core agent exposes `dogstatsd_eol_required` (disabled by default).

Verify whether the config path exists. If not, wire the existing codec support to the config key.

lucastemb:
## Summary

The DogStatsD codec supports requiring end-of-line delimiters, but was not wired into configuration. This PR aims to achieve parity with the Core Agent.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

My changes were tested mainly using unit tests and a manual smoke check.  

I started ADP with `dogstatsd_eol_required: ["udp"]`, sent a UDP DogStatsD packet without a trailing newline, and verified telemetry incremented `dsd_in error_type="framing"`.

```bash
lucas.tembras %  curl -s http://127.0.0.1:5102/metrics | rg 'component_errors_total|component_events_received_total'
# TYPE adp__component_events_received_total counter
adp__component_events_received_total{component_id="dd_metrics_encode",component_type="encoder"} 0
adp__component_events_received_total{component_id="events_enrich",component_type="transform"} 0
adp__component_events_received_total{component_id="dsd_agg",component_type="transform"} 0
adp__component_events_received_total{component_id="internal_metrics_out",component_type="destination"} 19992
adp__component_events_received_total{component_id="dsd_tag_filterlist",component_type="transform"} 0
adp__component_events_received_total{component_id="dsd_prefix_filter",component_type="transform"} 0
adp__component_events_received_total{component_id="dsd_post_agg_filter",component_type="transform"} 0
adp__component_events_received_total{component_id="dsd_in",component_type="source",message_type="metrics",listener_type="udp"} 0
adp__component_events_received_total{component_id="dd_service_checks_encode",component_type="encoder"} 0
adp__component_events_received_total{component_id="dsd_enrich",component_type="transform"} 0
adp__component_events_received_total{component_id="service_checks_enrich",component_type="transform"} 0
adp__component_events_received_total{component_id="dsd_in",component_type="source",message_type="events",listener_type="udp"} 0
adp__component_events_received_total{component_id="dd_events_encode",component_type="encoder"} 0
adp__component_events_received_total{component_id="metrics_enrich",component_type="transform"} 0
adp__component_events_received_total{component_id="dd_out",component_type="forwarder"} 0
adp__component_events_received_total{component_id="dsd_in",component_type="source",message_type="service_checks",listener_type="udp"} 0
adp__component_events_received_total{component_id="dsd_debug_log_out",component_type="destination"} 0
adp__component_events_received_total{component_id="dsd_stats_out",component_type="destination"} 0
# TYPE adp__component_errors_total counter
adp__component_errors_total{component_id="dsd_in",component_type="source",listener_type="udp",error_type="decode",message_type="metrics"} 1
adp__component_errors_total{component_id="dsd_in",component_type="source",listener_type="udp",error_type="decode",message_type="events"} 0
adp__component_errors_total{component_id="dd_events_encode",component_type="encoder",error_type="http_send"} 0
adp__component_errors_total{component_id="dd_service_checks_encode",component_type="encoder",error_type="http_send"} 0
adp__component_errors_total{component_id="dsd_in",component_type="source",listener_type="udp",error_type="decode",message_type="service_checks"} 0
adp__component_errors_total{component_id="dsd_in",component_type="source",listener_type="udp",error_type="framing"} 1
adp__component_errors_total{component_id="dd_metrics_encode",component_type="encoder",error_type="http_send"} 0
adp__component_errors_total{component_id="dd_out",component_type="forwarder",error_type="http_send"} 0
```

## References

- Closes: https://github.com/DataDog/saluki/issues/1339

### DIFF
--- a/lib/saluki-components/src/sources/dogstatsd/framer.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/framer.rs
@@ -19,8 +19,8 @@ impl Framer for DsdFramer {
     }
 }
 
-pub fn get_framer(listen_address: &ListenAddress) -> DsdFramer {
-    let newline_framer = NewlineFramer::default().required_on_eof(false);
+pub fn get_framer(listen_address: &ListenAddress, eol_required: bool) -> DsdFramer {
+    let newline_framer = NewlineFramer::default().required_on_eof(eol_required);
 
     match listen_address {
         ListenAddress::Tcp(_) => DsdFramer::Stream(NestedFramer::new(newline_framer, LengthDelimitedFramer)),
@@ -29,5 +29,72 @@ pub fn get_framer(listen_address: &ListenAddress) -> DsdFramer {
         ListenAddress::Unixgram(_) => DsdFramer::NonStream(newline_framer),
         #[cfg(unix)]
         ListenAddress::Unix(_) => DsdFramer::Stream(NestedFramer::new(newline_framer, LengthDelimitedFramer)),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::{
+        collections::VecDeque,
+        net::{Ipv4Addr, SocketAddr, SocketAddrV4},
+    };
+
+    use saluki_io::{
+        deser::framing::{Framer, FramingError},
+        net::ListenAddress,
+    };
+
+    use super::get_framer;
+
+    fn udp_address() -> ListenAddress {
+        ListenAddress::Udp(SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, 8125)))
+    }
+
+    fn missing_delimiter_err(len: usize) -> FramingError {
+        FramingError::InvalidFrame {
+            frame_len: len,
+            reason: "reached EOF without finding newline delimiter",
+        }
+    }
+
+    #[test]
+    fn udp_missing_newline_is_accepted_by_default() {
+        let payload = b"test.metric:1|c";
+        let mut buf = VecDeque::from(payload.to_vec());
+        let mut framer = get_framer(&udp_address(), false);
+
+        let frame = framer
+            .next_frame(&mut buf, true)
+            .expect("framing should not fail")
+            .expect("frame should be available at EOF");
+
+        assert_eq!(&frame[..], payload);
+    }
+
+    #[test]
+    fn udp_missing_newline_is_rejected_when_required() {
+        let payload = b"test.metric:1|c";
+        let mut buf = VecDeque::from(payload.to_vec());
+        let mut framer = get_framer(&udp_address(), true);
+
+        assert_eq!(
+            framer.next_frame(&mut buf, true),
+            Err(missing_delimiter_err(payload.len()))
+        );
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn uds_stream_missing_newline_is_rejected_when_required() {
+        let payload = b"test.metric:1|c";
+        let mut buf = VecDeque::new();
+        buf.extend((payload.len() as u32).to_le_bytes());
+        buf.extend(payload);
+        let mut framer = get_framer(&ListenAddress::Unix("/tmp/dsd-stream.sock".into()), true);
+
+        assert_eq!(
+            framer.next_frame(&mut buf, true),
+            Err(missing_delimiter_err(payload.len()))
+        );
     }
 }
--- a/lib/saluki-components/src/sources/dogstatsd/framer.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/framer.rs
@@ -53,7 +53,7 @@ mod tests {
     fn missing_delimiter_err(len: usize) -> FramingError {
         FramingError::InvalidFrame {
             frame_len: len,
-            reason: "reached EOF without finding newline delimiter",
+            reason: "reached EOF (end of UDP frame) without finding newline delimiter, which is required by dogstatsd_eol_required.",
         }
     }
 
--- a/lib/saluki-components/src/sources/dogstatsd/framer.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/framer.rs
@@ -50,13 +50,6 @@ mod tests {
         ListenAddress::Udp(SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, 8125)))
     }
 
-    fn missing_delimiter_err(len: usize) -> FramingError {
-        FramingError::InvalidFrame {
-            frame_len: len,
-            reason: "reached EOF (end of UDP frame) without finding newline delimiter, which is required by dogstatsd_eol_required.",
-        }
-    }
-
     #[test]
     fn udp_missing_newline_is_accepted_by_default() {
         let payload = b"test.metric:1|c";
@@ -77,10 +70,10 @@ mod tests {
         let mut buf = VecDeque::from(payload.to_vec());
         let mut framer = get_framer(&udp_address(), true);
 
-        assert_eq!(
+        assert!(matches!(
             framer.next_frame(&mut buf, true),
-            Err(missing_delimiter_err(payload.len()))
-        );
+            Err(FramingError::InvalidFrame { .. })
+        ));
     }
 
     #[cfg(unix)]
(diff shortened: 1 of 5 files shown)


lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -982,10 +1039,14 @@ async fn drive_stream(
...
-                    let mut frames = io_buffer.framed(&mut framer, reached_eof);
                     'frame: loop {
-                        match frames.next() {
-                            Some(Ok(frame)) => {
+                        match next_frame_or_discard_connectionless(
+                            &mut framer,
+                            &mut *io_buffer,
+                            reached_eof,
+                            stream.is_connectionless(),
+                        ) {
```
tobz: It feels like we've changed all of this just to support being able to unit test it.

`IoBufferManager`, which uses `ObjectPool`, already ensures that for connectionless usages, every time we call `get_buffer_mut`, we're getting back a fresh buffer and that the old buffer is cleared before anyone else acquires it.

When we require EOL on EOF in the framer, it would be returning an error here, which takes us down to the check we do to see if the stream is connectionless, and if so, we `continue 'read`, which causes us to loop around, ultimately calling `get_buffer_mut` which gives our fresh buffer, and so on.

It seems like simply keeping the `eol_required` plumbing gets us the intended behavior, even if it means we can't unit test it. Everything else here that, put together, gives us our desired behavior... is already tested.

lib/saluki-components/src/sources/dogstatsd/framer.rs
```diff
@@ -31,3 +31,70 @@ pub fn get_framer(listen_address: &ListenAddress) -> DsdFramer {
...
+    use super::get_framer;
+
+    fn udp_address() -> ListenAddress {
+        ListenAddress::Udp(SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, 8125)))
+    }
+
+    fn missing_delimiter_err(len: usize) -> FramingError {
+        FramingError::InvalidFrame {
+            frame_len: len,
+            reason: "reached EOF without finding newline delimiter",
```
webern:
```suggestion
            reason: "reached EOF (end of the UDP frame) without finding newline delimiter which is required by dogstatsd_eol_required.",
```

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -407,6 +422,38 @@ pub struct DogStatsDConfiguration {
...
+impl EolRequired {
+    fn from_config_values(values: &[String]) -> Self {
+        let mut eol_required = Self::default();
+
+        for value in values {
+            match value.as_str() {
+                "udp" => eol_required.udp = true,
+                "uds" => eol_required.uds = true,
+                "named_pipe" => {}
+                _ => warn!(value = value.as_str(), "Invalid dogstatsd_eol_required value."),
```
webern:
```suggestion
                _ => warn!(value = value.as_str(), "Invalid dogstatsd_eol_required value. Expected 'true' or 'false'"),
```

webern (commented):
Just a couple of UX nits. LGTM while deferring to tobs on the request to simplify.

lib/saluki-components/src/sources/dogstatsd/mod.rs
lucastemb: I cleaned up some of the bloat you're pointing out here and simplified. But, for UDP and Unixgram, `stream.is_connectionless` returns `true` meaning `should_retain` is `true`. So, when we call `get_buffer_mut`, we end up collapsing and returning the same buffer.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -407,6 +422,41 @@ pub struct DogStatsDConfiguration {
...
+        let mut eol_required = Self::default();
+
+        for value in values {
+            match value.as_str() {
+                "udp" => eol_required.udp = true,
+                "uds" => eol_required.uds = true,
+                "named_pipe" => {}
+                _ => warn!(
+                    value = value.as_str(),
+                    "Invalid dogstatsd_eol_required value. Expected 'true' or 'false'"
```
tobz: This isn't actually a boolean field, though. We should be listing out the possible values: `'udp', 'uds', or 'named_pipe'.`

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -407,6 +422,41 @@ pub struct DogStatsDConfiguration {
...
+    fn from_config_values(values: &[String]) -> Self {
+        let mut eol_required = Self::default();
+
+        for value in values {
+            match value.as_str() {
+                "udp" => eol_required.udp = true,
+                "uds" => eol_required.uds = true,
+                "named_pipe" => {}
+                _ => warn!(
+                    value = value.as_str(),
```
tobz: Should just be able to do `value` instead of `value = value.as_str()`.. and if that doesn't work, then `%value` to use the `Display` implementation of it.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -982,10 +1042,9 @@ async fn drive_stream(
                         bytes_read
                     );
 
-                    let mut frames = io_buffer.framed(&mut framer, reached_eof);
                     'frame: loop {
-                        match frames.next() {
-                            Some(Ok(frame)) => {
+                        match framer.next_frame(&mut *io_buffer, reached_eof) {
+                            Ok(Some(frame)) => {
```
tobz: Go back to the original code, using `io_buffer.framed(...)`.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1019,6 +1078,7 @@ async fn drive_stream(
                                 }
 
                                 if stream.is_connectionless() {
+                                    io_buffer.clear();
```
tobz, line 1082: We don't need to do this.

It already happens through the `IoBufferManager` when a new buffer is acquired at the top of the `'read`, thus releasing the old buffer and clearing it before returning it to the pool.

tobz (changes requested):

lib/saluki-components/src/sources/dogstatsd/framer.rs
```diff
@@ -31,3 +31,70 @@ pub fn get_framer(listen_address: &ListenAddress) -> DsdFramer {
...
+    use super::get_framer;
+
+    fn udp_address() -> ListenAddress {
+        ListenAddress::Udp(SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::LOCALHOST, 8125)))
+    }
+
+    fn missing_delimiter_err(len: usize) -> FramingError {
+        FramingError::InvalidFrame {
+            frame_len: len,
+            reason: "reached EOF (end of UDP frame) without finding newline delimiter, which is required by dogstatsd_eol_required.",
```
tobz: Tests are also failing because this message doesn't actually match the error that the newline framer returns... so we should simply use `matches!(FramingError::InvalidFrame { .. }, ..)` instead for our test assertions.

The error message as it exists in the newline framer is as specific as we can make it: we don't know if it's being used for UDP or TCP or UDS, and we also don't know if it's being used for DogStatsD or something else.

lib/saluki-components/src/sources/dogstatsd/mod.rs
tobz: After chatting about this with Lucas, I realized I'm wrong: I was conflating the retain behavior as working to replace the buffer for connectionless streams on every call to `get_buffer_mut`, which is where my insistence was coming.

We should leave this change as-is. 👍🏻

tobz (approved):

PR #1602, opened 2026-05-07
chore(env): create `datadog-agent-commons` crate for Agent-specific shared code

tobz:
## Summary

This PR introduces a new crate, `datadog-agent-commons`, as part of the start of trying to separate Datadog Agent-specific code from `saluki`-specific crates and allow for better sharing of helpers/types/etc between the various Datadog-specific components.

The bulk of this change is related to two major moves: putting the Agent-specific `saluki-env` code into `agent-data-plane`, and putting IPC-related helpers into `datadog-agent-commons`. I structured the change this way because the environment provider we use in ADP is inherently specific to it: nothing else is going to reuse (at least not yet) so I wanted to isolate it there... while things like the Agent IPC code, and platform helpers for getting specific paths and so on... it's inherently shared/commons-y even if _maybe_ we eventually decide it's also worth moving into `agent-data-plane`.

There's still a ton of Agent-specific code, namely component implementations, that live in `saluki-components`. For now, those are staying put, as it makes the configuration testing story ("config registry") simpler to wrap our heads around with configuration living in a single place. I'll be evaluating moving them out -- along with ensuring we don't lose configuration testing coverage -- in a follow-up PR.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Existing unit tests and a _few_ new ones around IPC auth settings.

## References

DADP-2


webern (approved):
Looks like you did not yet move the config registry which is helpful with my diff pending. I'll try to open a PR today.

PR #1607, opened 2026-05-07
feat(correctness): add kind-based origin detection tests and fix |card: frame drop

thieman:
## Human Summary

Adds the `dsd-origin-detection-client` correctness test which runs under Kubernetes-in-Docker. This exercises some of the k8s origin detection surface. Subsequent PRs will add coverage on additional configuration options related to origin detection.

There are a few other major changes in this PR:

- Kind correctness tests now run a single millstone instance that emits to both the baseline and comparison target containers. This was necessary for each target to get equivalent inputs with origin detection enabled.
- Updates lading to pull in [DataDog/lading#1871](https://github.com/DataDog/lading/pull/1871). This allows us to provide a pool of values for the DogStatsD extensions (`c:`, `e:`, `card:`) which you can see in `millstone.yaml` for the new test.
- `millstone.yaml` now resolves `{{CONTAINER_ID}}` and `{{POD_UID}}` which are used to pass down real values from the running millstone container. This is necessary so that origin detection resolves correctly within the agent based on the incoming metrics payloads.
- Fixes a correctness issue where ADP would drop the entire frame when a metric had an unsupported value in the `card:` field.

**TODOs for subsequent PRs:**

- More origin detection tests covering more of the configuration options. Ideally I'd like to make this parameterized somehow.
- Standalone PR swapping the Docker correctness tests to use a single millstone instance for consistency.

## Summary

- Replaces `dsd-plain-kind` (a kind test that disabled origin detection) with `dsd-origin-detection-kind`, which exercises PID-based origin detection in a real Kubernetes cluster via the containerd socket
- Adds `dsd-origin-detection-client`, which exercises all three DogStatsD origin detection protocol extension fields (`|c:`, `|e:`, `|card:`) including valid, unresolvable, wrong-case, and unrecognized values
- Switches all kind tests to a **shared millstone topology**: two agent pods + one shared millstone pod, both runs parallel with identical inputs
- Bumps lading-payload to `3eaedaca` (PR #1871) which adds `container_ids`, `external_data`, and `cardinality` pool fields to the DogStatsD payload generator
- Fixes three bugs in `|card:` handling: ADP was case-sensitive while the Go agent is case-insensitive (`strings.ToLower`); non-UTF-8 bytes previously invoked UB via `from_utf8_unchecked`; millstone's `wait` script masked a failed child exit code

## Key changes

**Shared millstone topology (`k8s.rs`):** Each agent pod mounts a unique HostPath directory as its airlock volume (`type: DirectoryOrCreate`). A shared millstone pod mounts both socket directories and runs two millstone processes in parallel with the same seed and config, giving both agents bitwise-identical inputs. Origin data (pod UID, container ID) is gathered from the millstone pod so both configs resolve the same entity.

**`|card:` parse fixes (`helpers.rs`, `origin.rs`):**
- Frame drop on unrecognized values: replaced `alt(known_values)` with `rest` + graceful `try_from(...).ok()`
- UB on non-UTF-8 bytes: replaced `from_utf8_unchecked` with `simdutf8::basic::from_utf8` (consistent with the rest of the codec)
- Case sensitivity mismatch: `OriginTagCardinality::try_from` now matches lowercase first (zero-allocation), then falls through to `eq_ignore_ascii_case` (also allocation-free) to match the Go agent's `strings.ToLower` behavior. Discovered by the correctness test: `|card:HIGH` was treated as high cardinality by baseline DSD but ignored by ADP, causing systematic value divergence.

**dsd-origin-detection-kind:** PID-based origin detection via containerd socket. Passes.

**dsd-origin-detection-client:** `dogstatsd_origin_detection_client: true`. millstone cardinality pool includes all four canonical values, the `"orch"` shorthand (`ShortOrchestratorCardinalityString` in the Go agent), wrong-case variants (`"HIGH"`, `"ORCH"`), and an unrecognized value. Both tests pass.

## Test plan

- [x] `make test-correctness-case CASE=dsd-origin-detection-kind` — PASS
- [x] `make test-correctness-case CASE=dsd-origin-detection-client` — PASS


bin/correctness/millstone/src/config.rs
```diff
@@ -56,7 +56,7 @@ impl TryFrom<String> for TargetAddress {
 pub enum Payload {
     /// DogStatsD-encoded metrics.
     #[serde(rename = "dogstatsd")]
-    DogStatsD(lading_payload::dogstatsd::Config),
+    DogStatsD(Box<lading_payload::dogstatsd::Config>),
```
thieman, line 59: [Claude Sonnet 4.6] Boxed because the lading-payload bump (adding `container_ids`, `external_data`, `cardinality` as `Vec<String>`) grew `dogstatsd::Config` to ≥256 bytes while the other two variants stay ~32 bytes. Without the box, every `Payload` instance allocates 256 bytes regardless of which variant is stored — a `Payload::OpenTelemetryMetrics` value wastes 224 bytes. Boxing the `DogStatsD` variant drops it to 8 bytes (a pointer), equalizing all variants. This also satisfies Clippy's `large_enum_variant` lint which is promoted to a hard error via `#![deny(warnings)]` in this binary.

Cargo.toml
```diff
@@ -182,7 +182,7 @@ tower-http = { version = "0.6", default-features = false }
 bollard = { version = "0.20", default-features = false }
 home = { version = "0.5", default-features = false }
 reqwest = { version = "0.13", default-features = false }
-lading-payload = { git = "https://github.com/DataDog/lading", rev = "01ef8d70d6967be86fb7ef558b5d6cb3b6f13dae" }
+lading-payload = { git = "https://github.com/DataDog/lading", rev = "3eaedacabff0f3fc9947b019c020c5d020adf808" }
```
thieman, line 185: Bump to pick up https://github.com/DataDog/lading/pull/1871

test/correctness/dsd-origin-detection-client/millstone.yaml
```diff
@@ -0,0 +1,69 @@
...
+      external_data:
+        - "pu-{{POD_UID}},cn-millstone,it-false"
+        - "this-is-not-valid-external-data"
+      # Per-metric cardinality override pool for the |card: extension field.
+      cardinality:
+        - "low"
+        - "orchestrator"
+        - "high"
+        - "none"
+        - "this-is-not-a-valid-cardinality"
```
thieman, line 78: Passing `this-is-not-a-valid-cardinality` for cardinality triggers the frame drop bug that is addressed in the `helpers.rs` changes

lib/saluki-io/src/deser/codec/dogstatsd/helpers.rs
```diff
@@ -170,30 +169,16 @@ pub fn external_data(input: &[u8]) -> IResult<&[u8], &str> {
...
+        // SAFETY: We know the bytes in `b` can only be comprised of UTF-8 characters, because our tags are all based on valid
+        // UTF-8 strings, which ensures that it's valid to interpret the bytes directly as UTF-8.
+        unsafe { std::str::from_utf8_unchecked(b) }
+    })
     .parse(input)?;
 
-    OriginTagCardinality::try_from(raw_cardinality)
-        .map(|cardinality| (remaining, Some(cardinality)))
-        .map_err(|_| nom::Err::Error(Error::new(input, ErrorKind::Verify)))
+    Ok((remaining, OriginTagCardinality::try_from(raw_cardinality).ok()))
```
thieman: Fix for dropping the frame when a metric has an invalid `card:` extension value

lib/saluki-io/src/deser/codec/dogstatsd/helpers.rs
```diff
@@ -170,30 +169,16 @@ pub fn external_data(input: &[u8]) -> IResult<&[u8], &str> {
...
-            // SAFETY: We know the bytes in `b` can only be comprised of UTF-8 characters, because our tags are all based on valid
-            // UTF-8 strings, which ensures that it's valid to interpret the bytes directly as UTF-8.
-            unsafe { std::str::from_utf8_unchecked(b) }
-        },
-    )
+    let (remaining, raw_cardinality) = map(all_consuming(preceded(tag(CARDINALITY_PREFIX), rest)), |b| {
+        // SAFETY: The DogStatsD frame is validated as UTF-8 upstream before reaching this parser,
+        // so any bytes remaining after the prefix are valid UTF-8.
+        unsafe { std::str::from_utf8_unchecked(b) }
+    })
```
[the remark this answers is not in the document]
thieman: Fixed in dbc82a8ce0 — replaced `from_utf8_unchecked` with `std::str::from_utf8(...).ok()` so non-UTF-8 bytes return `None` (same as an unrecognized value) rather than invoking UB. Added a unit test `cardinality_non_utf8_bytes_returns_none` that directly exercises this with `\xff\xfe` bytes — it would have panicked in debug or produced garbage in release with the old implementation.

bin/correctness/panoramic/src/correctness/k8s.rs
```diff
@@ -457,27 +625,106 @@ fn build_pod(cfg: PodConfig<'_>) -> Pod {
...
+    } = cfg;
+
+    let wait_cmd = format!(
+        "until [ -f /etc/millstone/baseline.toml ] && \
+               [ -f /etc/millstone/comparison.toml ] && \
+               [ -S /baseline-airlock/metrics.sock ] && \
+               [ -S /comparison-airlock/metrics.sock ]; do sleep 1; done; \
+         exec sh -c '{bin} /etc/millstone/baseline.toml & \
+                      {bin} /etc/millstone/comparison.toml & \
+                      wait'",
```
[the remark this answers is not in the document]
thieman: Fixed in dbc82a8ce0 — the wait script now captures both PIDs and combines exit codes: `wait $P1; R1=$?; wait $P2; R2=$?; exit $((R1 | R2))`. Either failure propagates as a non-zero container exit, which `wait_for_millstone_exit` will correctly surface as an error.

lib/saluki-io/src/deser/codec/dogstatsd/helpers.rs
```diff
@@ -170,30 +169,71 @@ pub fn external_data(input: &[u8]) -> IResult<&[u8], &str> {
...
-
-    OriginTagCardinality::try_from(raw_cardinality)
-        .map(|cardinality| (remaining, Some(cardinality)))
-        .map_err(|_| nom::Err::Error(Error::new(input, ErrorKind::Verify)))
+    let (remaining, raw_bytes) = all_consuming(preceded(tag(CARDINALITY_PREFIX), rest)).parse(input)?;
+
+    // Use checked UTF-8 conversion: non-UTF-8 bytes in the cardinality value are treated the same
+    // as an unrecognized string — return None so the frame continues processing rather than
+    // invoking undefined behavior or hard-failing.
+    let cardinality = std::str::from_utf8(raw_bytes)
```
tobz: `simdutf8::basic::from_utf8(input)` instead.

tobz (changes requested):
One small thing.

lib/saluki-io/src/deser/codec/dogstatsd/helpers.rs
thieman: Fixed in https://github.com/DataDog/saluki/pull/1607/changes/76fa0d77755913c75fc83c75ceeeeee7dab6f80b

tobz (approved):

PR #1613, opened 2026-05-08
chore(o11y): split internal telemetry across INFO/DEBUG/TRACE and set default level to INFO

### LINKED ISSUE #1534: Histograms in internal telemetry can accumulate forever if actively filtered.
## Context

When filtering is active for internal metrics, we bail out early during iteration over each set of metrics by type: if the metric at hand is being filtered, don't process it. For counters and gauges, this is fine because it simply means they never get emitted downstream. For histograms, however, this creates a scenario where the memory used to hold the histogram samples grows over time because histograms must be "cleared".

We do clear histograms when there are no consumers attached to the metrics stream, but we don't do the same thing if a given histogram is being filtered out. We should update the filter logic to clear the histogram before proceeding to the next one to ensure they don't accumulate endlessly.

https://github.com/DataDog/saluki/blob/ef4bdace87acb39fb5bf547fb063089d1228ded0/lib/saluki-core/src/observability/metrics.rs#L262-L282

tobz:
## Summary

This PR introduces a first pass at starting to classify Saluki's internal telemetry at more appropriate levels to allow for effectively filtering out useful, always-on telemetry from more low level/granular telemetry that is primarily useful for debugging.

Saluki, based on the underlying metrics library it uses, has a concept of metric "levels": levels are synonymous with the levels you see in logging (INFO, DEBUG, etc), and are designed to allow for filtering out expensive/high-cardinality metrics that might not always be necessary. Most of the internal telemetry emitted by Saluki was already classified at different levels, but the leveling wasn't consistent across the board, and we had a default level filter of TRACE, meaning we always emitted everything anyways.

In this PR, we've done a few things:

- added the ability to configure a default metric filter level, and set that default to INFO
- reclassified the level of many internal metrics to reduce the volume when filtering at the INFO level
- fixed a bug with histogram sample accumulation when a histogram is actively filtered (#1534)

### Level reclassification

We took a simple but principled approach during reclassification:

- metrics that are ultimately used for RAR/COAT should be INFO so we never skip emitting them
- component telemetry (almost everything that looks like `component_*_total`) should be INFO as this provides general performance/health indicators for the topology
- almost all other things should be DEBUG
- anything in the DEBUG category which is very high-cardinality (some runtime metrics, task poll latency, etc) becomes TRACE just to give a little bit of separation

One additional change is that we removed a remapped metrics -- `            "dogstatsd.channel_latency_secs` -- ads it would have required us to make a high-cardinality histogram sit at the INFO level, and we don't actually need to mirror that metric because it was always sort of an approximation of the original metric to begin with.

### Default INFO level

With the metric filter level now exposed in configuration (`metrics_level`, or `DD_METRICS_LEVEL`), we can now trivially update the default filter level to INFO as it can be overridden from the default in scenarios where a user always wants DEBUG or TRACE metrics flowing. Additionally, we already have existing support for dynamically adjusting the metrics level via `agent-data-plane debug set-metric-level`.

### Required follow-up

This PR constrains the _emission_ of internal telemetry, but one thing it doesn't yet change is how the internal observability topology holds on to it. We currently use a Prometheus destination for exposing our internal telemetry outside of the process, and a "reflector" for holding it prior to sending it over RAR. Both of these mechanisms will accumulate whatever metrics they receive without any additional filtering or clearing out of metrics once the metrics filter level changes.

Essentially, this means that once the metrics filter level is changed to something more verbose, that "level" will be exposed until the process is restarted even if the more-verbose metrics aren't actually being updated.

We'll need to make some subsequent changes to the aforementioned mechanisms so that they can participate and observe the changes to the metric levels (somehow), allowing them to clear out metrics which fall outside of the updated metrics filter level to stop them from being exposed.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

- [x] Existing unit tests.
- [x] Ran ADP (standalone mode) and observed only INFO level metrics in internal telemetry output.
- [x] Ran ADP (standalone mode) with `metrics_level: trace` and observed all metrics in internal telemetry output.

## References

DADP-65

Fixes #1534

### DIFF
--- a/.github/workflows/check-pr-title.yml
+++ b/.github/workflows/check-pr-title.yml
@@ -44,6 +44,7 @@ jobs:
             blackhole
             buffered incremental
             chained
+            checks ipc
             ci
             common
             components
@@ -53,6 +54,7 @@ jobs:
             core
             correctness
             datadog
+            datadog-agent-commons
             datadog-intake
             datadog-protos
             ddsketch
@@ -61,14 +63,12 @@ jobs:
             docs
             dogstatsd
             dogstatsd mapper
-            dogstatsd prefix filter
             dsd debug log
             dsd stats
             env
             error
             heartbeat
             host enrichment
-            host tags
             integration
             internal metrics
             io
@@ -77,6 +77,7 @@ jobs:
             metric router
             metrics
             millstone
+            o11y
             otlp
             otlp-protos
             ottl
--- a/lib/saluki-core/src/observability/metrics.rs
+++ b/lib/saluki-core/src/observability/metrics.rs
@@ -51,12 +51,12 @@ pub struct FilterHandle {
 impl FilterHandle {
     /// Overrides the current metrics filter level.
     pub fn override_filter(&self, level: Level) {
-        *self.state.current_level.lock().unwrap() = Some(level);
+        *self.state.current_level.lock().unwrap() = level;
     }
 
-    /// Resets the metrics filter level to the default (INFO).
+    /// Resets the metrics filter level to the default that was configured when the metrics subsystem was initialized.
     pub fn reset_filter(&self) {
-        *self.state.current_level.lock().unwrap() = None;
+        *self.state.current_level.lock().unwrap() = self.state.default_level;
     }
 }
 
@@ -65,7 +65,8 @@ struct State {
     level_map: FastConcurrentHashMap<Key, Level>,
     flush_tx: broadcast::Sender<SharedEvents>,
     metrics_prefix: String,
-    current_level: Mutex<Option<Level>>,
+    default_level: Level,
+    current_level: Mutex<Level>,
 }
 
 impl State {
@@ -86,7 +87,7 @@ struct MetricsRecorder {
 }
 
 impl MetricsRecorder {
-    fn new(metrics_prefix: String) -> Self {
+    fn new(metrics_prefix: String, default_level: Level) -> Self {
         let (flush_tx, _) = broadcast::channel(2);
         Self {
             state: Arc::new(State {
@@ -94,7 +95,8 @@ impl MetricsRecorder {
                 level_map: FastConcurrentHashMap::default(),
                 flush_tx,
                 metrics_prefix,
-                current_level: Mutex::new(None),
+                default_level,
+                current_level: Mutex::new(default_level),
             }),
         }
     }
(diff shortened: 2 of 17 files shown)


.github/workflows/check-pr-title.yml
```diff
@@ -61,14 +63,12 @@ jobs:
...
             dogstatsd
             dogstatsd mapper
-            dogstatsd prefix filter
             dsd debug log
             dsd stats
             env
             error
             heartbeat
             host enrichment
-            host tags
```
tobz, line 71: These were just missed updates to the list of title scopes.

lib/saluki-core/src/observability/metrics.rs
```diff
@@ -86,15 +87,16 @@ struct MetricsRecorder {
...
         let (flush_tx, _) = broadcast::channel(2);
         Self {
             state: Arc::new(State {
                 registry: Registry::new(AtomicStorage),
                 level_map: FastConcurrentHashMap::default(),
                 flush_tx,
                 metrics_prefix,
-                current_level: Mutex::new(None),
+                default_level,
+                current_level: Mutex::new(default_level),
```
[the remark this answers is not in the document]
tobz: This is a good point.

Updated the code in 3c050b05f2 to clear counters out when filtered, _and_ when we have no downstream listeners.

lib/saluki-core/src/observability/metrics.rs
```diff
@@ -220,67 +222,67 @@ async fn flush_metrics(flush_interval: Duration) {
...
 
             let context = context_resolver.resolve_from_key(key);
-            let value = counter.swap(0, Ordering::Relaxed) as f64;
-
             let metric = Metric::counter(context, value);
             metrics.push(Event::Metric(metric));
         }
 
         for (key, gauge) in gauges {
+            let value = f64::from_bits(gauge.load(Ordering::Relaxed));
```
thieman, line 266: Unlike the swap above this won't actually modify the `gauge`, right? If so then should this stay below the `continue` on filtering?

thieman (approved):

lib/saluki-core/src/observability/metrics.rs
tobz: True. It's a no-op change.

I just did it so that things are laid out consistently in the three different loops, just to try and reduce the cognitive burden of seeing it ordered differently and being like "wait, why?".

PR #1618, opened 2026-05-11
chore(correctness): switch Docker runtime to single shared millstone container

thieman:
## Human Summary

Follows up on #1607, which switched the kind correctness tests to use a single shared millstone pod. This PR does the same for the Docker runtime for consistency, and while doing so fixes several pre-existing issues in the kind runtime as well.

## Summary

**`$GROUP` placeholder in `millstone.yaml`:**
- All `millstone.yaml` files now use `$GROUP` as a placeholder in the `target:` URL (e.g. `unixgram:///$GROUP-airlock/metrics.sock`, `grpc://$GROUP:4317/...`, `udp://$GROUP:8125`). At test runtime the shared millstone container uses `sed` to substitute `baseline` or `comparison` to produce two per-target configs, then runs both millstone processes in parallel.

**Docker runtime (`runner.rs`, `airlock/driver.rs`):**
- Removes millstone from the baseline and comparison `GroupRunner`s; each group now only contains `[datadog-intake, target]`
- Adds a shared millstone `GroupRunner` with its own isolation group, connected to both agent networks
- Each agent container gets a network alias (`baseline` or `comparison`) via `NetworkingConfig` at creation time so the shared millstone can address each agent unambiguously for both DSD (Unix socket) and TCP/gRPC/UDP tests
- Splits `DataCollector` into `AgentGroupCollector` (intake port) and `MillstoneWaiter` (millstone handle)
- Adds `DriverConfig::with_volume_mount()`, `DriverConfig::with_network_alias()`, `DriverConfig::with_network()`, and `DriverConfig::from_image()` (pub) to airlock
- The shared millstone container bind-mounts the original `millstone.yaml` (same mechanism as before) and derives the two per-target configs at startup with `sed`
- Fixes Docker volumes and networks leaking on test failure paths — `cleanup_groups` is now called on every exit path, not just the success path

**Kind runtime (`k8s.rs`):**
- Fixes the same TCP/gRPC/UDP target bug that existed in the kind runtime (was hardcoded to socket paths regardless of test type). Now uses `$GROUP` substitution: `baseline`/`comparison` for socket tests, agent pod cluster IPs for TCP/gRPC/UDP tests
- The millstone pod startup wait condition is now conditional: only waits for socket files on DSD tests (`is_socket_target`)
- For TCP/gRPC/UDP kind tests, agent pods are started first and their IPs fetched before the millstone pod is created; the millstone pod then includes `nc -z host port` readiness checks in its startup wait condition so millstone does not attempt a connection before the agent has bound its port
- Port is extracted from the `$GROUP` placeholder URL via `serde_yaml` — parses the template as YAML and reads the `target` field

**New test cases from main (rebased):**
- `dsd-uds-stream` (Unix stream socket) and `dsd-udp` (UDP) updated to use `$GROUP` placeholder

## Key changes

**Target rewriting (`$GROUP`):**

Every `millstone.yaml` uses `$GROUP` in the target URL. At test runtime:
- Docker: `sed 's/\$GROUP/baseline/g'` substitutes the network alias
- Kind/DSD: `template.replace("$GROUP", "baseline")` substitutes the group name (for airlock volume path)
- Kind/TCP: `template.replace("$GROUP", pod_ip)` substitutes the agent pod's cluster IP

**Docker millstone container startup:**
```sh
sed 's/\$GROUP/baseline/g' /etc/millstone/config.toml > /tmp/millstone-baseline.toml &&
sed 's/\$GROUP/comparison/g' /etc/millstone/config.toml > /tmp/millstone-comparison.toml &&
exec sh -c 'millstone /tmp/millstone-baseline.toml & P1=$!; millstone /tmp/millstone-comparison.toml & P2=$!; wait $P1; R1=$?; wait $P2; R2=$?; exit $((R1|R2))'
```

**Kind port-readiness (non-socket targets):**
1. Start agent pods (Phase 1a)
2. Fetch pod IPs; build `nc -z ip port` checks using port extracted via `serde_yaml` (Phase 1b)
3. Start millstone pod with those checks baked into its wait condition (Phase 1c)

## Test plan

- [x] `dsd-plain` — PASS
- [x] `dsd-added-tags` — PASS
- [x] `dsd-events` — PASS
- [x] `otlp-metrics` — PASS
- [x] `otlp-traces` — PASS
- [x] `otlp-traces-kind` (temporary test case, removed after validation) — PASS


bin/correctness/panoramic/src/correctness/runner.rs
```diff
@@ -29,6 +29,15 @@ use crate::{
...
 };
 
+/// Directory mounted inside the shared millstone container that holds the per-target config files.
+const MILLSTONE_CONFIG_DIR_INTERNAL: &str = "/etc/millstone";
+
+/// How long to wait after millstone exits before querying datadog-intake for data.
+///
+/// This gives the agents time to flush any remaining aggregated metrics after millstone stops
+/// sending. The value is slightly longer than a full aggregation bucket width.
+const FLUSH_WAIT: Duration = Duration::from_secs(32);
```
thieman, line 41: This is the same as before, just moved around and hoisted to a constant

bin/correctness/panoramic/src/correctness/runner.rs
```diff
@@ -137,10 +162,40 @@ pub(crate) fn make_error_result(name: String, started: Instant, phase: &str, e:
...
 }
 
+/// Rewrites the `target:` field in the millstone config YAML for a specific test group.
+///
+/// Two substitutions are applied to the target value:
+///
+/// - Unix socket targets (`unixgram:///airlock/...`): `/airlock/` is replaced with
+///   `/{group}-airlock/`, directing millstone at the group's volume mount.
+/// - TCP/gRPC targets (`grpc://target:...`): `://target` is replaced with `://{group}`,
+///   directing millstone at the group's network alias.
```
thieman: I'm adding correctness tests on additional supported transports in a subsequent PR which might need to expand on this substitution piece. This is obviously pretty janky / magical, might make more sense to support a $GROUP substitution or something in the millstone YAML. Will take as a follow-up.

bin/correctness/panoramic/src/correctness/k8s.rs
```diff
@@ -497,6 +543,21 @@ fn is_socket_target(template: &str) -> bool {
...
 
+/// Extracts the port number from the millstone config template.
+///
+/// Parses the template as YAML, reads the `target` string, and extracts the port from the
+/// `scheme://$GROUP:<port>/...` URL. Returns `None` for Unix socket targets (which need no port
+/// check) or if parsing fails. Used to add a `nc -z` port-readiness check to the millstone
+/// pod's startup script for TCP/gRPC/UDP targets.
+fn extract_target_port(template: &str) -> Option<u16> {
+    let doc: serde_yaml::Value = serde_yaml::from_str(template).ok()?;
+    let target = doc.get("target")?.as_str()?;
```
thieman, line 554: This is a bit janky but importing the actual Millstone config schema would require giving it a `lib.rs` etc. Can certainly do that but this was already a lot more code than I was expecting and this should break pretty loudly if it ever drifts since we run correctness on every PR.

bin/correctness/panoramic/src/correctness/k8s.rs
```diff
@@ -655,19 +682,30 @@ fn build_millstone_pod(cfg: MillstonePodConfig<'_>) -> Pod {
...
+    // Build the readiness condition. Config files are always awaited since the harness writes
+    // them after the pod reaches Running. Socket files are only awaited for DSD tests.
+    let ready_cond = if use_socket_wait {
+        "[ -f /etc/millstone/baseline.toml ] && \
+               [ -f /etc/millstone/comparison.toml ] && \
+               [ -S /baseline-airlock/metrics.sock ] && \
+               [ -S /comparison-airlock/metrics.sock ]"
+    } else {
+        "[ -f /etc/millstone/baseline.toml ] && \
+               [ -f /etc/millstone/comparison.toml ]"
```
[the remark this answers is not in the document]
thieman: Addressed in https://github.com/DataDog/saluki/pull/1618/changes/2bbe5fb2f472eb621299d4172cad4e81e6a97c94

bin/correctness/panoramic/src/correctness/runner.rs
```diff
@@ -139,8 +150,10 @@ pub(crate) fn make_error_result(name: String, started: Instant, phase: &str, e:
 
 /// Manages the state and program flow of running a *correctness* test.
 ///
-/// In a correctness test, two isolated groups of containers are created. One containing the Agent alone, and the other
-/// containing ADP and the Agent working together. These are called 'baseline' and 'comparison', respectively.
+/// In a correctness test, two isolated groups of containers are created. One containing the Agent
+/// alone (baseline), and the other containing ADP and the Agent working together (comparison). A
+/// single shared millstone container runs two parallel millstone processes — one targeting each
```
webern, line 168: Does it have to be two processes? I don't know how, but could the same bytes be, hmmm, copied to two sockets?

test/correctness/dsd-events/millstone.yaml
```diff
@@ -1,5 +1,5 @@
 seed: [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131]
-target: "unixgram:///airlock/metrics.sock"
+target: "unixgram:///$GROUP-airlock/metrics.sock"
```
webern, line 2: Sigh.

webern (approved):

bin/correctness/panoramic/src/correctness/k8s.rs
webern: I dream of:

```bash
find . -type f -name "*.yaml" -delete
```

bin/correctness/panoramic/src/correctness/runner.rs
thieman: That's probably doable for UDP sockets with something like `tee`, not sure about gRPC. It's comparatively a lot of complexity imo compared to just running two processes, which lets us take advantage of the fact that Lading is deterministic

webern (approved):

PR #1620, opened 2026-05-11
feat(correctness): add origin detection config matrix

thieman:
## Human Summary

Adds a new `correctness_matrix` Panoramic test type which allows us to define multiple variants of a given correctness test. Each variant can provide specific environment variables to override, but the rest of the test configuration is shared. This allows us to test multiple configurations for a given test with minimal duplication.

Test variants are individually resolvable in Panoramic, so it's possible to run one variant without running all variants of a given test. Variants are also individually discoverable, so each variant gets its own CI task.

This PR additionally introduces 6 new variants covering origin detection settings and folds the previous 2 we already had into a single matrix test.

## What

1. **Env var smoke test coverage for all origin detection config settings.** The six DogStatsD annotations that control origin detection (`dogstatsd_origin_detection`, `origin_detection_unified`, `dogstatsd_entity_id_precedence`, `dogstatsd_origin_detection_client`, `dogstatsd_origin_optout_enabled`, `dogstatsd_tag_cardinality`) had no `env_var_override` set, so the existing `DogStatsDConfiguration` smoke test never verified they could be loaded via `DD_*` env vars. This PR adds the overrides and the smoke test now exercises all six.

2. **A `correctness_matrix` test type** that expands a single test directory into multiple independent named correctness tests — one per variant. Each variant appends `additional_env_vars` to both baseline and comparison targets, allowing a matrix of agent configurations without duplicating the full test layout.

3. **A new matrix test** (`dsd-origin-detection-matrix`) with 8 variants providing full non-default coverage of every origin detection setting:
   - `legacy` — UDS PID-based detection, all defaults
   - `legacy-entity-id-precedence` — UDS PID suppressed when client pod UID present (`dogstatsd_entity_id_precedence=true`)
   - `legacy-high-cardinality` — legacy mode with `dogstatsd_tag_cardinality=high`
   - `legacy-optout-disabled` — `dogstatsd_origin_optout_enabled=false`; metrics with `|card:none` are enriched rather than skipped
   - `legacy-with-client-fields` — legacy + `|c:`/`|e:`/`|card:` protocol extension parsing
   - `unified` — priority waterfall, first-match-wins (`origin_detection_unified=true`)
   - `unified-with-client-fields` — unified + client field parsing
   - `unified-high-cardinality` — unified with `dogstatsd_tag_cardinality=high`

4. **Removal of `dsd-origin-detection-kind` and `dsd-origin-detection-client`**, which are now fully covered by the `legacy` and `legacy-with-client-fields` matrix variants respectively.

5. **Fix to `tooling/generate-correctness-pipeline.sh`** to sanitize matrix variant names (which contain `/`) into valid GitLab job name identifiers, while preserving the original name for `CORRECTNESS_TEST_CASE`.

## Why

The origin detection config settings had no env var test coverage. Without it, we have no guarantee that `DD_DOGSTATSD_ORIGIN_DETECTION=true` etc. actually work, which matters since that is how the correctness matrix variants control each agent's configuration.

The matrix mechanism removes the need to duplicate `config.yaml`, `datadog-base.yaml`, and `millstone.yaml` across N near-identical directories whenever we want to test multiple related configurations.

## Testing

- All existing unit tests pass (`make test`, 1225/1225)
- `DogStatsDConfiguration` smoke test now exercises env var paths for all six origin detection settings
- All 8 matrix variants pass locally against a kind cluster
- `panoramic list -d test/correctness` confirms all 8 variants expand with correct names and the two deleted tests are gone
- CI pipeline generator produces valid GitLab job names for all variants


lib/saluki-components/src/config_registry/datadog/dogstatsd.rs
```diff
@@ -162,7 +162,7 @@ crate::declare_annotations! {
         schema: &schema::DOGSTATSD_ORIGIN_DETECTION,
         support_level: SupportLevel::Full,
         additional_yaml_paths: &[],
-        env_var_override: None,
+        env_var_override: Some(&["DD_DOGSTATSD_ORIGIN_DETECTION"]),
```
webern, line 165: I don't think these are necessary (and shouldn't be there) for `DD_{{the-rest}}` when `{{the-rest}}` matches `yaml-path.to_upper()`. I'm not exactly sure how dots are handled (but I think they are). Anyway, the whole config apparatus is equating yaml paths with DD_* env vars if I understand it correctly.

lucastemb (approved):

bin/correctness/panoramic/src/config.rs
```diff
@@ -404,11 +409,194 @@ impl IntegrationConfig {
...
+}
+
 /// Discover all test cases across one or more directories.
 ///
 /// Each `config.yaml` found in a direct subdirectory must have a top-level `type` field set to
-/// either `"integration"` or `"correctness"`. Files with a missing or unknown `type` are skipped
-/// with a warning. Multiple test types may coexist freely within the same directory.
+/// `"integration"`, `"correctness"`, or `"correctness_matrix"`. Files with a missing or unknown
+/// `type` are skipped with a warning. Multiple test types may coexist freely within the same
+/// directory.
```
webern: It's a lie that I previously missed when I introduced the panic at line 624

```suggestion
/// `type` cause a panic. Multiple test types may coexist freely within the same directory.
```

webern (commented):
Cool.

I think we are going to need to consolidate on a single config for testing instead of having a switch case on different config types. I don't suggest doing that now. Just noticing.

lib/saluki-components/src/config_registry/datadog/dogstatsd.rs
thieman: My thinking was that adding them here _makes sure_ that we are reading in those env vars for those config values, at test time. Ideally there would be a matching test on the DDA side. Without this, it's possible that these new correctness tests wouldn't be exercising any functionality at all, because the env vars aren't being read.

webern (approved):

thieman:
Benchmarks passed on the second-to-last commit and the last commit is literally just a comment change, so... sending to queue

tobz:
Rerunning the benchmark job because that ain't making sense to me.

PR #1621, opened 2026-05-11
feat(correctness): add DogStatsD UDP and UDS stream transport correctness tests

thieman:
## Summary

Adds correctness test coverage for two DogStatsD transports that were previously untested. All nine existing DSD correctness tests use only UDS datagram (`unixgram://`). This PR covers the remaining transports supported by both Saluki and the stock Datadog Agent.

## Key changes

**New test cases:**
- `test/correctness/dsd-udp` — metrics over UDP (port 8125, `dogstatsd_non_local_traffic: true` so the millstone container can reach the agent across Docker bridge networks)
- `test/correctness/dsd-uds-stream` — metrics over UDS stream socket (`SOCK_STREAM`), which exercises the nested length-prefix + newline framer — a completely different code path from the datagram-based tests

**Millstone changes** required to support network-addressed targets:
- `config/target`: `Tcp` and `Udp` variants now store `host:port` as `String` rather than `SocketAddr`. `SocketAddr::from_str` only accepts IP literals; panoramic rewrites the millstone target to Docker network aliases (e.g. `baseline:8125`) which must be resolved via Docker DNS at connection time.
- `driver/config`: new `send_delay_us` field (default `0`). Linux UDP socket receive buffers default to ~208 KiB (~25 packets at 8 KiB each). Without pacing, millstone overflows the buffer and causes asymmetric packet loss between agents. `500 µs` between sends keeps peak buffer occupancy low and delivers all 10,000 payloads in ~5 seconds — well within a single 10-second aggregation bucket, preventing cross-bucket histogram skew.

**TCP note:** `dogstatsd_tcp_port` is a Saluki-only feature. The stock Agent does not register this config key and has no TCP listener implementation (`comp/dogstatsd/listeners/` has no `tcp.go`). It cannot be covered by the baseline-vs-comparison correctness framework.

No changes to panoramic's target rewriting logic were needed — `unix://` already benefits from the `/airlock/` path substitution used by `unixgram://`, and `udp://` already benefits from the `://target` host substitution used by `grpc://`. Millstone is unconditionally joined to both agent networks, so UDP hostname resolution works without further changes.

## Test plan

Both tests pass locally:
- `dsd-uds-stream`: PASS
- `dsd-udp`: PASS (confirmed twice with the `send_delay_us: 500` pacing)

### DIFF
--- a/bin/correctness/millstone/src/config.rs
+++ b/bin/correctness/millstone/src/config.rs
@@ -1,5 +1,4 @@
 use std::{
-    net::SocketAddr,
     num::NonZeroUsize,
     path::{Path, PathBuf},
 };
@@ -11,10 +10,16 @@ use serde::Deserialize;
 #[serde(try_from = "String")]
 pub enum TargetAddress {
     /// TCP socket.
-    Tcp(SocketAddr),
+    ///
+    /// Stored as a `host:port` string so that Docker network aliases (e.g. `baseline:8125`)
+    /// are resolved at connection time rather than at config-parse time.
+    Tcp(String),
 
     /// UDP socket.
-    Udp(SocketAddr),
+    ///
+    /// Stored as a `host:port` string so that Docker network aliases (e.g. `baseline:8125`)
+    /// are resolved at connection time rather than at config-parse time.
+    Udp(String),
 
     /// Unix Domain Socket in SOCK_DGRAM mode.
     UnixDatagram(PathBuf),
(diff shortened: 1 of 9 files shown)


bin/correctness/millstone/src/config.rs
```diff
@@ -131,6 +146,22 @@ pub struct Config {
...
+    /// receiver has a chance to drain it.
+    ///
+    /// For example, a value of `500` (500 µs) limits the send rate to ~2,000 payloads/s, which is well within
+    /// what a DogStatsD agent can drain. At 8 KiB per payload, that is ~16 MB/s — low enough that the 208 KiB
+    /// buffer never accumulates more than a handful of packets at any moment, and all 10,000 payloads are
+    /// delivered in ~5 seconds — well within a single 10-second aggregation bucket.
+    ///
+    /// Defaults to `0` (no delay).
+    #[serde(default)]
+    pub send_delay_us: u64,
```
thieman, line 163: I'm a bit worried about this flaking but I'm not sure how else we might account for it, without any delay at all I was seeing significant dropped payloads on both sides

tobz (approved):

bin/correctness/millstone/src/config.rs
tobz: I agree although I think pacing is a reasonable approach so long as we document it as such.

PR #1624, opened 2026-05-11
enhancement(dogstatsd): allow autoscaling UDP stream handlers on available parallelism

### LINKED ISSUE #1593: Autoscale the number of UDP workers for DogStatsD source.
## Context

Currently, the DogStatsD source will use a _single_ stream handler for UDP, consistent with how we spawn a single handler for UDS datagram: there's no "connections" to naturally spawn off, so we do one task.

However, a trick available to us for UDP is to use `SO_REUSEPORT` such that we can run multiple stream handlers, all bound to the same UDP listening address, leaving it to the OS to load balance UDP packets across them based on the 4-tuple. This could be as simple as spawning N stream handlers for every 6-8 vCPUs we detect: something simple to add a _little_ bit of concurrency when we know there's multiple cores that can actually be utilized.

This would let us gain a little bit of extra processing capacity when operating with multiple cores, avoiding saturation of a single task which risks dropped packets due to socket buffers filling up when the stream handler task yields to the runtime and can't otherwise keep up with the incoming rate of packets.

tobz:
## Summary

This PR adds the ability to "autoscale" the UDP stream handlers in the DogStatsD source based on available parallelism.

Currently, the UDP listener for the DogStatsD source emits a single stream, in the first `accept` call, and then pends forever after that. It does the same thing for UDS datagram listeners. The reason is simple: these are connectionless transports, so there's really only ever one "stream" we can yield back unlike TCP or UDS streams.

However, on Linux platforms, we have a trick available to us for UDP: `SO_REUSEPORT`. This socket option allows binding N UDP sockets to the same address/port in a way that the kernel will load balance incoming datagrams across them, letting us exceed the processing capacity of a single task.

This PR introduces support for yielding multiple UDP streams from `Listener` and using that in the DogStatsD source through a new configuration setting, `dogstatsd_autoscale_udp_listeners`. When enabled, the DogStatsD source will configure the UDP listener to emit N UDP streams, where N is calculated from the available parallelism: for every 8 vCPUs, we yield an additional UDP stream up to a maximum of 4. This gives us a small amount of additional parallelism without trying to track too closely to vCPU, such that the smaller count of UDP stream handlers can hopefully maximize their on-CPU time.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Performance

## How did you test this PR?

- [x] New unit tests for the `saluki-io` changes.
- [x] New integration test to ensure that things don't break on Linux with UDP listener autoscaling enabled.

## References

DADP-2

Closes #1593

### DIFF
--- a/lib/saluki-components/src/config_registry/datadog/dogstatsd.rs
+++ b/lib/saluki-components/src/config_registry/datadog/dogstatsd.rs
@@ -11,6 +11,13 @@ static DOGSTATSD_ALLOW_CONTEXT_HEAP_ALLOCS_SCHEMA: SchemaEntry = SchemaEntry {
     default: Some("true"),
 };
 
+static DOGSTATSD_AUTOSCALE_UDP_LISTENERS_SCHEMA: SchemaEntry = SchemaEntry {
+    yaml_path: "dogstatsd_autoscale_udp_listeners",
+    env_vars: &[],
+    value_type: ValueType::Bool,
+    default: Some("false"),
+};
+
 static DOGSTATSD_BUFFER_COUNT_SCHEMA: SchemaEntry = SchemaEntry {
     schema: Schema::Saluki,
     yaml_path: "dogstatsd_buffer_count",
@@ -318,6 +325,17 @@ crate::declare_annotations! {
     /// `dogstatsd_allow_context_heap_allocs` — allow heap allocations when interner is full.
     DOGSTATSD_ALLOW_CONTEXT_HEAP_ALLOCS = SalukiAnnotation {
         schema: &DOGSTATSD_ALLOW_CONTEXT_HEAP_ALLOCS_SCHEMA,
+        support_level: SupportLevel::Full,
+        additional_yaml_paths: &[],
+        env_var_override: None,
+        used_by: &[structs::DOGSTATSD_CONFIGURATION],
+        value_type_override: None,
+        test_json: None,
+    };
+
+    /// `dogstatsd_autoscale_udp_listeners` — bind multiple UDP sockets with SO_REUSEPORT for kernel load balancing.
+    DOGSTATSD_AUTOSCALE_UDP_LISTENERS = SalukiAnnotation {
+        schema: &DOGSTATSD_AUTOSCALE_UDP_LISTENERS_SCHEMA,
         support_level: SupportLevel::Full,
         additional_yaml_paths: &[],
         env_var_override: None,
--- a/lib/saluki-components/src/config_registry/datadog/dogstatsd.rs
+++ b/lib/saluki-components/src/config_registry/datadog/dogstatsd.rs
@@ -12,6 +12,7 @@ static DOGSTATSD_ALLOW_CONTEXT_HEAP_ALLOCS_SCHEMA: SchemaEntry = SchemaEntry {
 };
 
 static DOGSTATSD_AUTOSCALE_UDP_LISTENERS_SCHEMA: SchemaEntry = SchemaEntry {
+    schema: Schema::Saluki,
     yaml_path: "dogstatsd_autoscale_udp_listeners",
     env_vars: &[],
     value_type: ValueType::Bool,
--- a/lib/saluki-components/src/sources/dogstatsd/mod.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/mod.rs
@@ -1599,7 +1599,8 @@ mod tests {
     }
 
     #[test]
-    fn autoscale_udp_listeners_from_config() {
+    #[cfg(target_os = "linux")]
+    fn autoscale_udp_listeners_from_config_linux() {
         let config = deser_config(r#"{"dogstatsd_autoscale_udp_listeners": true}"#);
         assert!(config.autoscale_udp_listeners);
 
@@ -1611,6 +1612,15 @@ mod tests {
             (1..=4).contains(&n),
             "expected 1..=4 streams from vCPU formula, got {n}"
         );
+    }
+
+    #[test]
+    #[cfg(not(target_os = "linux"))]
+    fn autoscale_udp_listeners_from_config_non_linux() {
+        let config = deser_config(r#"{"dogstatsd_autoscale_udp_listeners": true}"#);
+        assert!(config.autoscale_udp_listeners);
+
+        assert_eq!(None, config.udp_streams_to_yield());
     }
 
     #[test]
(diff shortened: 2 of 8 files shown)


tobz:
* **#1624** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1624?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1624?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/saluki-io/src/net/listener.rs
```diff
@@ -272,6 +300,47 @@ where
...
+#[cfg(target_os = "linux")]
+async fn bind_udp_sockets(
+    addr: SocketAddr, maybe_socket_count: Option<NonZeroUsize>,
+) -> io::Result<VecDeque<TokioUdpSocket>> {
+    use socket2::{Domain, Protocol, SockAddr, Socket, Type};
+
+    fn bind_one(addr: SocketAddr) -> io::Result<TokioUdpSocket> {
+        let socket = Socket::new(Domain::for_address(addr), Type::DGRAM, Some(Protocol::UDP))?;
+        socket.set_reuse_address(true)?;
+        socket.set_reuse_port(true)?;
```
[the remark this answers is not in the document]
tobz: This is fair, and I addressed it in 3d5ba43.

lib/saluki-components/src/config_registry/datadog/dogstatsd.rs
```diff
@@ -358,6 +365,17 @@ crate::declare_annotations! {
         test_json: None,
     };
 
+    /// `dogstatsd_autoscale_udp_listeners` — bind multiple UDP sockets with SO_REUSEPORT for kernel load balancing.
+    DOGSTATSD_AUTOSCALE_UDP_LISTENERS = SalukiAnnotation {
```
webern, line 338: Is it necessary to make it configurable? To have it on? Would it cause problems.

This is just coming from configuration fatigue 😅

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -494,6 +512,27 @@ impl DogStatsDConfiguration {
         EolRequired::from_config_values(&self.eol_required)
     }
 
+    /// Returns the number of UDP stream handlers to spawn, derived from `dogstatsd_autoscale_udp_listeners` and
+    /// the number of available vCPUs.
+    ///
+    /// Returns `None` when autoscaling is disabled, which keeps the legacy single-socket behavior. The platform
+    /// gate for `SO_REUSEPORT` lives inside the listener — this method intentionally stays platform-agnostic.
```
webern, line 519: 👍 for a platform agnostic function

lib/saluki-io/src/net/listener.rs
```diff
@@ -93,10 +93,12 @@ enum ListenerInner {
 /// continually "accepted". Instead, `Listener` will emit a single `Stream` that can be used to send and receive data
 /// from multiple remote peers.
 ///
-/// ## Missing
+/// ## UDP autoscaling
 ///
-/// - Ability to configure `Listener` to emit multiple streams for connectionless address families, allowing for load
-///   balancing. (Only possible for UDP via SO_REUSEPORT, as UDS does not support SO_REUSEPORT.)
+/// On Linux, UDP listeners can be configured to bind multiple sockets to the same address using `SO_REUSEPORT`,
```
webern, line 101: Should work for unix/darwin to, right?

webern (approved):
Cool. Is there a known performance gain or is it more theoretical?

lib/saluki-io/src/net/listener.rs
```diff
@@ -272,6 +299,55 @@ where
...
+#[cfg(target_os = "linux")]
+async fn bind_udp_sockets(
+    addr: SocketAddr, maybe_socket_count: Option<NonZeroUsize>,
+) -> io::Result<VecDeque<TokioUdpSocket>> {
+    use socket2::{Domain, Protocol, SockAddr, Socket, Type};
+
+    fn bind_one(addr: SocketAddr) -> io::Result<TokioUdpSocket> {
+        let socket = Socket::new(Domain::for_address(addr), Type::DGRAM, Some(Protocol::UDP))?;
+        socket.set_reuse_address(true)?;
+        socket.set_reuse_port(true)?;
```
thieman, line 327: Would it be possible to try this flow with set_reuse_port and fall back to the standard approach if it fails? That seems like it'd be better for targeting than target_os, since IIUC the practical target set here is "linux 3.9+ and darwin?" I'm not sure how much <3.9 is out in the wild but RHEL 7 is only on 3.10 and we have quite a few of those

lib/saluki-components/src/config_registry/datadog/dogstatsd.rs
tobz: It would _probably_ be fine, but...

My main hesitation is that, internally, we have _very little_ UDP-based DSD traffic... and it's almost the complete opposite for customers. Whatever insight we derive about its behavior when dogfooding will likely be irrelevant when running in a customer environment.

tobz:
> Cool. Is there a known performance gain or is it more theoretical?

@webern There's definitely an upper bound to how fast we can accept datagrams -- whether UDP or UDS -- with a single task. For connectionless scenarios, you run the risk of dropping data if you can't keep up since there's no backpressure signal to the clients.

The Datadog Agent tries to handle this by separating receive from decode, but it pays the price in terms of having to allocate and shuttle large pre-decode buffers between the receive and decode ("pipeline") tasks.. whereas we want to keep all of our benefits of how we do it now (zero-copy decode before `Event` conversion, etc) while scaling the number of tasks that we have available to do so.

lib/saluki-io/src/net/listener.rs
tobz: `SO_REUSEPORT`/`SO_REUSEADDR` _are_ available on Darwin, but they don't provide any load balancing like they do on Linux.

lib/saluki-io/src/net/listener.rs
tobz: "Yes" to having it try to fallibly set the option as a sideband way to detect support, "no" to Darwin, since Darwin doesn't actually do any load balancing between the sockets.

PR #1626, opened 2026-05-11
feat(dogstatsd): Add dogstatsd telemetry compatibility

aqian01:
## Summary

Implements [DADP-45](https://datadoghq.atlassian.net/wiki/spaces/DADP/pages/6685394989/DADP-45+DogStatsD+pipeline+telemetry+compatibility+for+ADP) DogStatsD telemetry compatibility for ADP.

- Enables ADP telemetry by default on :5102.
- Adds explicit Prometheus scrape routes:
  - `/metrics` for raw ADP telemetry
  - `/compat/metrics` for go_expvar-compatible telemetry
  - `/` for backward-compatible raw telemetry
  - unknown paths return 404
- Adds compat remapping rules for DogStatsD, UDP/UDS listener, forwarder, and retry queue metrics.
- Adds DogStatsD origin-detection error telemetry.
- Adds HTTP forwarder error classification for DNS, connection, TLS, write, send, and HTTP status errors.
- Adds retry queue gauges:
  - `network_http_retry_queue_size`
  - `network_http_retry_queue_bytes_per_sec`
- Updates retry queue bytes/sec to match Agent-style incoming payload rate semantics.

this is a related [PR](https://github.com/DataDog/datadog-agent/compare/andrewq/adp_openmetrics_compat_config?expand=1) for adding the example config that is mentioned [here](https://datadoghq.atlassian.net/wiki/spaces/DADP/pages/6685394989/DADP-45+DogStatsD+pipeline+telemetry+compatibility+for+ADP#4.-Customer-integration-config)
## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

unit tests

Manual testing:
Build the core agent
```
❯ dda inv agent.build
```
Run the core agent
```
DD_DATA_PLANE_FORCE_ENABLE=true \
DD_DATA_PLANE_ENABLED=true \
./bin/agent/agent run -c dev/dist/datadog.yaml
```

Build ADP
```
make build-adp
```
Run ADP
```
DD_API_KEY=apikey \
DD_ADP_OTLP_ENABLED=true \
DD_OTLP_CONFIG='{}' \
DD_METRIC_TAG_FILTERLIST='[]' \
DD_AUTH_TOKEN_FILE_PATH=/Users/andrew.qian/Documents/Code/agent-related/datadog-agent/dev/dist/auth_token \
RUST_LOG=info \
make run-adp

```

Note I am using the fish terminal so these commands may not work for you.
1.
verify ADP is actually running and listening on localhost:5102.
The raw telemetry endpoint works:
```
curl -s -o /tmp/adp.metrics -w '%{http_code}\n' http://localhost:5102/metrics
curl -s -o /tmp/adp.compat.metrics -w '%{http_code}\n' http://localhost:5102/compat/metrics

200
200
```

2.
Verified /compat/metrics exposes DogStatsD UDP compat metrics and increments dogstatsd_metric_packets, dogstatsd_udp_packets, and dogstatsd_udp_bytes after sending a DogStatsD UDP metric through ADP.

Commands run:

`printf 'manual.compat.counter:1|c\n' | nc -u -w0 127.0.0.1 9191`
`sleep 15`
`curl -sf http://localhost:5102/compat/metrics >/tmp/adp.compat.metrics`

`grep '^dogstatsd_metric_packets' /tmp/adp.compat.metrics`
`grep '^dogstatsd_udp_packets' /tmp/adp.compat.metrics`
`grep '^dogstatsd_udp_bytes' /tmp/adp.compat.metrics`
Observed output:

`dogstatsd_metric_packets 1`
`dogstatsd_udp_packets 1`
`dogstatsd_udp_bytes 26`

3.
Verified `/compat/metrics` exposes ADP forwarder compat metrics with the `source="agent-data-plane"` label and increments `forwarder_transactions_success` after sending DogStatsD UDP traffic through ADP.

Commands run:

`for i in (seq 1 20); printf 'manual.compat.counter:1|c\n' | nc -u -w0 127.0.0.1 9191; end`

`sleep 30`

`curl -sf http://localhost:5102/compat/metrics >/tmp/adp.compat.metrics`

`grep '^forwarder_transactions' /tmp/adp.compat.metrics`

Observed output included:

`forwarder_transactions_success{source="agent-data-plane"} 11`

and zero-valued error/drop/retry metrics with the same `source="agent-data-plane"` label.

4.
Verified `/compat/metrics` exposes and increments DogStatsD event and service check compat metrics after sending event and service check packets through ADP over UDP.

Commands run:

`printf '_e{11,11}:compat test|hello world\n' | nc -u -w0 127.0.0.1 9191`

`printf '_sc|manual.compat.check|0\n' | nc -u -w0 127.0.0.1 9191`

`sleep 15`

`curl -sf http://localhost:5102/compat/metrics >/tmp/adp.compat.metrics`

`grep '^dogstatsd_event_packets' /tmp/adp.compat.metrics`

`grep '^dogstatsd_service_check_packets' /tmp/adp.compat.metrics`

`grep '^dogstatsd_udp_packets' /tmp/adp.compat.metrics`

`grep '^dogstatsd_udp_bytes' /tmp/adp.compat.metrics`

Observed output:

`dogstatsd_event_packets 1`

`dogstatsd_service_check_packets 1`

`dogstatsd_udp_packets 23`

`dogstatsd_udp_bytes 606`

5.
Verified `/compat/metrics` exposes and increments DogStatsD UDS compat metrics after sending DogStatsD packets through ADP’s Unix datagram and stream sockets.

Commands run:

`python3 -c "import socket; s=socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM); s.sendto(b'manual.compat.uds_dgram_counter:1|c\n', '/tmp/adp-dogstatsd-dgram.sock')"`

`printf 'manual.compat.uds_stream_counter:1|c\n' | nc -U -w0 /tmp/adp-dogstatsd-stream.sock`

`sleep 15`

`curl -sf http://localhost:5102/compat/metrics >/tmp/adp.compat.metrics`

`grep '^dogstatsd_uds_packets' /tmp/adp.compat.metrics`

`grep '^dogstatsd_uds_bytes' /tmp/adp.compat.metrics`

Observed output:

`dogstatsd_uds_packets 3`

`dogstatsd_uds_bytes 110`

6.

Verified `/compat/metrics` exposes and increments DogStatsD parse error compat metrics after sending malformed DogStatsD UDP packets through ADP.

Commands run:

`printf 'not-a-valid-dogstatsd-packet\n' | nc -u -w0 127.0.0.1 9191`

`printf 'broken.metric:abc|c\n' | nc -u -w0 127.0.0.1 9191`

`printf '_e{bad,bad}:bad|event\n' | nc -u -w0 127.0.0.1 9191`

`sleep 15`

`curl -sf http://localhost:5102/compat/metrics >/tmp/adp.compat.metrics`

`grep '^dogstatsd_metric_parse_errors' /tmp/adp.compat.metrics`

`grep '^dogstatsd_event_parse_errors' /tmp/adp.compat.metrics`

`grep '^dogstatsd_udp_packet_reading_errors' /tmp/adp.compat.metrics`

Observed output:

`dogstatsd_metric_parse_errors 2`

`dogstatsd_event_parse_errors 1`

`dogstatsd_udp_packet_reading_errors 0`

7.

Verified the customer OpenMetrics ingestion path using the ADP compat config example.

Setup commands:

`cd /Users/andrew.qian/Documents/Code/agent-related/datadog-agent`

`mkdir -p dev/dist/conf.d/openmetrics.d`

`cp cmd/agent/dist/conf.d/openmetrics.d/adp_compat.yaml.example dev/dist/conf.d/openmetrics.d/adp_compat.yaml`

Check command:

`DD_CONFD_PATH=$PWD/dev/dist/conf.d /opt/datadog-agent/bin/agent/agent check openmetrics -c $PWD/dev/dist`

Observed result:

`Configuration Source: file:/Users/andrew.qian/Documents/Code/agent-related/datadog-agent/dev/dist/conf.d/openmetrics.d/adp_compat.yaml[0]`

`Instance ID: openmetrics:datadog.agent:e69c74259b7d0dcc [OK]`

`Metric Samples: Last Run: 23, Total: 23`

Observed submitted metrics included:

`datadog.agent.forwarder.transactions.retry_queue_size`

`datadog.agent.retry_queue_duration.bytes_per_sec`

Both carried:

`endpoint:http://localhost:5102/compat/metrics`

`source:agent-data-plane`

Observed service check:

`datadog.agent.openmetrics.health` with status `OK`

[DADP-45]: https://datadoghq.atlassian.net/browse/DADP-45?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -642,6 +745,26 @@ impl<T: Retryable> PendingTransactions<T> {
...
+        self.telemetry.record_retry_queue_bytes_per_sec(bytes_per_sec);
+    }
+}
+
+fn current_unix_time_secs() -> u64 {
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .expect("system time should be after Unix epoch")
+        .as_secs()
 }
```
tobz: This already exists: `saluki_common::time::get_unix_timestamp`.

lib/saluki-io/src/net/client/http/telemetry.rs
```diff
@@ -16,6 +16,74 @@ use tower::{Layer, Service};
...
+
+    fn increment_http_error(&self) {
+        self.http_errors.increment(1);
+    }
+}
+
+fn register_scoped_error(builder: &MetricsBuilder, error_type: &'static str, error_scope: &'static str) -> Counter {
+    builder.register_counter_with_tags(
+        "network_http_requests_errors_total",
+        [("error_type", error_type), ("error_scope", error_scope)],
```
tobz, line 83: Where does `error_scope` come from? Not seeing it anywhere in the Agent codebase.

lib/saluki-io/src/net/unix/linux.rs
```diff
@@ -49,25 +49,31 @@ where
...
+                Some(creds) => ProcessIdentity::Credentials(ProcessCredentials {
+                    pid: creds.pid,
+                    uid: creds.uid,
+                    gid: creds.gid,
+                }),
+                None => ProcessIdentity::Error(ProcessCredentialsError::InvalidCredentials),
+            }
         }
     } else {
-        None
```
tobz, line 67: This seems wrong?

In the Agent, we don't emit any origin detection errors when there's no ancillary/out-of-band data. Probably should be `ProcessIdentity::Unavailable`?

lib/saluki-io/src/net/addr.rs
```diff
@@ -219,22 +278,45 @@ pub enum ConnectionAddress {
...
             #[cfg(unix)]
-            Self::ProcessLike(maybe_creds) => match maybe_creds {
-                None => write!(f, "<unbound>"),
-                Some(creds) => write!(f, "<pid={} uid={} gid={}>", creds.pid, creds.uid, creds.gid),
+            Self::ProcessLike(identity) => match identity {
+                ProcessIdentity::Credentials(creds) => {
+                    write!(f, "<pid={} uid={} gid={}>", creds.pid, creds.uid, creds.gid)
+                }
+                ProcessIdentity::Error(error) => write!(f, "<origin-detection-error: {}>", error),
+                ProcessIdentity::Unavailable => write!(f, "<unbound>"),
```
tobz: This should be `<no-origin>`.

lib/saluki-io/src/net/addr.rs
```diff
@@ -219,22 +278,45 @@ pub enum ConnectionAddress {
...
             Self::SocketLike(addr) => write!(f, "{}", addr),
             #[cfg(unix)]
-            Self::ProcessLike(maybe_creds) => match maybe_creds {
-                None => write!(f, "<unbound>"),
-                Some(creds) => write!(f, "<pid={} uid={} gid={}>", creds.pid, creds.uid, creds.gid),
+            Self::ProcessLike(identity) => match identity {
+                ProcessIdentity::Credentials(creds) => {
+                    write!(f, "<pid={} uid={} gid={}>", creds.pid, creds.uid, creds.gid)
+                }
+                ProcessIdentity::Error(error) => write!(f, "<origin-detection-error: {}>", error),
```
tobz: This is gonna generate really verbose debug/trace logs when we include the connection address. We should add a method to `ProcessCredentialsError` that just returns a simplified identifier for each variant, like `no-oob-data` or `zero-pid`, etc.

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -170,13 +173,27 @@ impl hyper::rt::Write for HttpsCapableConnection {
...
-        this.inner.poll_flush(cx)
+        match this.inner.poll_flush(cx) {
+            Poll::Ready(Err(error)) => {
+                if let Some(error_telemetry) = this.error_telemetry.as_ref() {
+                    error_telemetry.increment_wrote_request_error();
+                }
+                Poll::Ready(Err(error))
+            }
+            other => other,
+        }
```
tobz, line 196: Note to self, kind of: I think this is probably lower-level than we want to be here. We'd really want something at the request-level because, from what I can tell, the Go equivalent in `httptrace.WroteRequestInfo` isn't delivered on a per-write/flush _call_ basis, but in terms of the overall operation.

Probably fine for now because errors here should leave to the request failing and so the methods wouldn't be called again for the same buffer, etc... but it sort of hides the _intent_ of the metric to have it here, IMO.

bin/agent-data-plane/src/state/metrics/mod.rs
```diff
@@ -339,6 +343,20 @@ pub fn render_rar_telemetry(
...
+) -> String {
+    render_telemetry(state, rules, renderer)
+}
+
+/// Renders ADP telemetry using Core Agent `go_expvar`-compatible metric names.
+pub fn render_compat_telemetry(
+    state: &AggregatedMetricsState, rules: &[RemapperRule], renderer: &mut PrometheusRenderer,
+) -> String {
+    render_telemetry(state, rules, renderer)
+}
```
tobz: These are identical... 😅

tobz (changes requested):

lib/saluki-io/src/net/client/http/telemetry.rs
aqian01: It's ADP-only, not from the Agent. We add it to our internal error counter so the compat remapper can tell transaction-level errors apart from phase-level ones (DNS/TLS/connect/write) when mapping into forwarder_transactions_errors. It's only used for matching, so it doesn't show up on /compat/metrics.

bin/agent-data-plane/src/internal/observability.rs
```diff
@@ -50,8 +60,19 @@ impl Supervisable for InternalTelemetryWorker {
...
+        let metrics_state = get_shared_metrics_state().await;
+        let compat_rules = get_compat_remappings();
+        let compat_renderer = Mutex::new(PrometheusRenderer::new());
+        let compat_provider: Arc<dyn PrometheusPayloadProvider> = Arc::new(move || {
+            let state = metrics_state.state();
+            let mut renderer = compat_renderer
+                .lock()
+                .expect("compat renderer mutex should not be poisoned");
+            render_telemetry(state, &compat_rules, &mut renderer)
+        });
```
tobz, line 72: Note to self: part of me thinks that we should just ditch the Prometheus destination at some point and just directly expose API endpoints that use the internal metrics reflector and Prometheus rendering code since it would let us avoid having to do hacks like this to support multiple paths on the telemetry endpoint... and we could avoid needing the separate telemetry endpoint in the first place.

tobz (approved):
Overall, this looks good enough to merge so we can get telemetry where it needs to be.

We're definitely going to end up refactoring the telemetry endpoint stuff in the near future, and I'd like to try and consolidate some of the HTTP telemetry stuff... but those aren't blockers here.

PR #1635, opened 2026-05-13
chore(app): dynamically publish config API via supervised worker

tobz:
## Summary

This PR wraps up the config API handler into a supervised worker to take advantage of the dynamic route publishing capabilities of the (un)privileged API builders.

Really, it's more unification than anything else since we were able to directly construct `ConfigAPIHandler` before this PR as well, but... 🤷🏻 

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Non-functional (chore, refactoring, docs)
- [ ] Performance

## How did you test this PR?

- [x] Built and ran ADP locally, ensuring the `/config` endpoint still worked.
- [x] Added a new assertion to the `privileged-api-endpoints` integration test to assert the `/config` exists.

## References

DADP-2

PR #1636, opened 2026-05-13
chore(agent-data-plane): update environment provider to be supervisor-driven

tobz:
## Summary

This PR updates the ADP environment provider, as well as some of the underlying shared pieces from `saluki-env`, to be supervisor-based.

Environment providers are generally composed of data stores which are dynamically updated in some way, either by querying the system directly for information or receiving it from external processes/systems. These updates happen asynchronously, and so environment providers (really, their underlying subproviders like workload and autodiscovery) will naturally spawn asynchronous tasks to handle this work. We want to bring these tasks under supervision.

We've made a number of small changes to support this:

- The core primitives for workload metadata collection -- `MetadataAggregator` and `MetadataCollectorWorker` -- can now be added to a supervision tree directly. They play nicely with restarts.
- `RemoteAgentWorkloadProvider` now returns its own supervisor that drives the aforementioned metadata workers, and it runs a dedicated worker for dynamically publishing the API routes related to querying the workload metadata stores (tags, External Data, etc).
- `RemoteAgentAutodiscoveryProvider` will also return its own supervisor that drives the underlying workers that handle forwarding AD events to subscribers
- `ADPEnvironmentProvider` returns an optional supervisor (depending on if standalone mode is enabled) that is a rollup of the new supervisors returned by the underlying workload and AD providers

Essentially, we've turned all the spots where we previously spawned asynchronous tasks into supervised workers, and rolled them up into nested supervisors.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Existing unit, integration, and correctness tests.

## References

DADP-2


tobz:
* **#1661** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1661?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1636** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1636?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1636?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1635** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1635?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1633** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1633?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1632** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1632?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -286,9 +296,9 @@ pub async fn handle_run_command(
 
     // Signal the internal supervisor to shutdown (if still running) and drive it to completion.
     // If the supervisor already exited (i.e., the select! above matched its branch), both the send
-    // and await resolve immediately — the send is a no-op and the future is already complete.
+    // and await resolve immediately — the send is a no-op and the JoinHandle is already ready.
     let _ = internal_shutdown_tx.send(());
```
aqian01, line 300: Does the early return on errors on line 221-222 prevent the shutdown from being run? 

```
let built_topology = blueprint.build().await?;
let mut running_topology = built_topology.spawn(&health_registry, memory_limiter).await?;
```

aqian01 (commented):
For this code in 
`lib/saluki-env/src/workload/collectors/cgroups.rs` (line 80)
```
poller_handle = Some(tokio::task::spawn_blocking(move || cgroups_manager.poll(operations_tx)));
```
When the collector (now supervised) is dropped, I don't think it drops the poller it spawns?

bin/agent-data-plane/src/cli/run.rs
tobz: When we're at the point of waiting for everything to become healthy in the internal supervisor, we haven't spawned the primary topology yet so there's nothing to shutdown.

tobz:
> When the collector (now supervised) is dropped, I don't think it drops the poller it spawns?

@aqian01 Good catch. Switched to using `AbortOnDropHandle` in [409835e](https://github.com/DataDog/saluki/pull/1636/commits/409835e6633572034091828d28c26872e20dee14) to hold the poller task handle to ensure it gets aborted when the collector future is dropped.

aqian01 (approved):

PR #1638, opened 2026-05-13
feat(agent-data-plane): emit point telemetry

thieman:
## Summary

Emit ADP-side point telemetry for metric payloads so Core Agent can consume `point.sent` and `point.dropped` via Remote Agent Registry.

Key changes:
- Track metric data point counts through payload and transaction metadata.
- Count point sends on successful Datadog forwarder delivery.
- Count point drops for permanent HTTP/send failures and retry queue drops.
- Emit the ADP point telemetry instruments as gauges while preserving cumulative semantics.
- Expose ADP point telemetry through RAR remapping as `point.sent` / `point.dropped` with the `domain` tag. RAR handles remote-agent attribution.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

- `make fmt`
- `cargo check --workspace`
- `cargo check -p saluki-components --lib`
- `cargo check -p saluki-components --tests`
- `cargo check -p saluki-io --lib`
- `cargo test -p agent-data-plane state::metrics::tests::test_render_rar_telemetry --bin agent-data-plane`
- `cargo test -p saluki-components common::datadog::request_builder --lib`
- `cargo test -p saluki-components common::datadog::request_builder::tests::split_oversized_request_tracks_data_points_written --lib`
- `cargo test -p saluki-components common::datadog::transaction::tests::basic_transaction_ser_deser_roundtrip --lib`
- `cargo test -p saluki-components encoders::datadog::metrics::tests::input_data_point_count_tracks_metric_values --lib`
- `cargo test -p saluki-io net::util::retry::queue --lib`
- Pre-commit checks run by `git commit`:
  - Rust formatting
  - Cargo.toml formatting
  - Clippy lints
  - third-party license freshness
  - dependency advisories/license/source checks
  - API documentation build

## References

- DADP-71


bin/agent-data-plane/src/state/metrics/rules/transaction.rs
```diff
@@ -12,6 +12,12 @@ pub fn get_transaction_remappings() -> Vec<RemapperRule> {
             "transactions.success_bytes",
         )
         .with_original_tags(["domain", "endpoint"]),
+        RemapperRule::by_name("adp.component_data_points_sent_total", "point.sent")
+            .with_original_tags(["domain"])
+            .with_additional_tags(["remote_agent:agent-data-plane"]),
```
tobz: We shouldn't have to add a `remote_agent` tag: RAR does that for us.

lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -479,16 +473,35 @@ fn generate_retry_queue_id(context: ComponentContext, endpoint: &ResolvedEndpoin
...
+        "{}://{}",
+        endpoint.scheme(),
+        endpoint.host_str().expect("resolved endpoint must have a host")
+    );
+    if let Some(port) = endpoint.port() {
+        domain.push(':');
+        domain.push_str(&port.to_string());
+    }
+    domain
+}
```
tobz: How is this any different from `endpoint.endpoint().to_string()`?

lib/saluki-components/src/common/datadog/request_builder.rs
```diff
@@ -605,9 +622,14 @@ where
             }));
         }
 
+        let data_points_written = inputs
+            .iter()
+            .map(|input| self.encoder.input_data_point_count(input))
+            .sum();
+
```
tobz: Seems like this should be calling `self.encoded_data_point_count()`?

lib/saluki-components/src/common/datadog/telemetry.rs
```diff
@@ -31,6 +33,22 @@ pub struct ComponentTelemetry {
     http_errors_by_code: Arc<Mutex<HashMap<StatusCode, Counter>>>,
 }
 
+struct DataPointGauge {
+    gauge: Gauge,
+    value: u64,
+}
```
tobz: This is just a gauge with an extra step. What's the deal here?

lib/saluki-io/src/net/util/retry/queue/mod.rs
```diff
@@ -60,12 +68,14 @@ impl PushResult {
     pub fn merge(&mut self, other: Self) {
         self.items_dropped += other.items_dropped;
         self.events_dropped += other.events_dropped;
+        self.data_points_dropped += other.data_points_dropped;
     }
 
     /// Tracks a single dropped item.
-    pub fn track_dropped_item(&mut self, event_count: u64) {
+    pub fn track_dropped_item(&mut self, event_count: u64, data_point_count: u64) {
```
tobz: Random thought, not a blocker: feels like we ought to actually be passing in `&dyn EventContainer` here so it can just read the event count and data point count directly, avoiding the need to manually extract them before calling this method, leaving a gap to use the wrong values.

tobz (changes requested):

lib/saluki-components/src/common/datadog/telemetry.rs
thieman: Claude originally implemented this as a counter and when I told it to change it to a gauge it did it in the dumbest way possible, will fix. Haven't gotten aroud to polishing this one yet, trying to get the trio working in tandem locally so I know it actually works end-to-end

lib/saluki-components/src/common/datadog/io.rs
thieman: I believe the endpoint actually includes for example `/api/v2/series` but I'll check

thieman: Nope turns out `endpoint.endpoint()` is pretty close to what we want, this now calls `endpoint.endpoint().origin().ascii_serialization()` which includes the port iff it's not the default for the scheme. That seems like what we'd want but I'll check if that matches once I get everything integrated

lib/saluki-components/src/common/datadog/io.rs
rayz, line 272: Do we need to account this as a permanently dropped transaction and increment `point.dropped`? Right now we only log below

thieman: [GPT 5.5] Agreed — this is a permanent drop for that endpoint because the endpoint I/O task receiver is gone, so the transaction will never be enqueued or retried there. Fixed in e462a4c82e by tracking it via `track_permanently_failed_transaction(transaction.metadata(), None, endpoint_domain)` when the fanout send fails.

lib/saluki-components/src/common/datadog/io.rs
thieman: This matched in the correctness test so I think we're good here

tobz (approved):

rayz (approved):

Issue #1642, opened 2026-05-13
Support `use_v2_api.*` settings.


webern:
Title says it all. Add support for these config settings.

tobz:
Closed via #1646.

PR #1646, opened 2026-05-13
enhancement(datadog): add support for series V1 API

### LINKED ISSUE #1641: Correctly categorize `use_v2_api.*` keys. DADP-72 Finding 2
We do not support `use_v2_api.*` settings but these are incorrectly listed in `ignored_keys.yaml`.

In fact, this is an important enough configuration setting that we should
* Error and Exit (`Severity::High`)

Also we should update the "confkey" ledger and doc. Me and my Claude know what this means even if no one else does.

tobz:
## Summary

This PR adds support for sending series metrics to the legacy series V1 API for Datadog destinations.

Prior to this, we only supported the series V2 API, the newer, Protocol Buffers-based API which superseded the old V1 API. Well... we want to allow using ADP in more configurations, and some customers still use the series V1 API, so here we are. :)

This PR implements support for the series V1 API, which is kind of self-evident, but we've made the following changes:

- updated the existing Datadog Metrics encoder to support both series API versions (mutually exclusive) based on the configuration of `use_v2_api.series`
- updated the configuration registry to account for `use_v2_api.series`, and update our supportability status for it
- added a few unit tests, and a new correctness test, to assert that we're doing series V1 correctly

I also made three small changes to the correctness tooling outside of adding series V1 support:

- `make test-correctness-case` runs with `--no-tui` now so that you can actually see all of the failing assertions (if any); no need for the TUI if you're just running a single case, really
- added detection of diagnose-related requests for the two series API endpoints in `datadog-intake`: DDA will send dummy payloads when trying to hit those endpoints during diagnose/connectivity check runs... so now we directly detect them and log them as such to avoid having spurious failure messages in the `datadog-intake` logs
- added normalization of the various ways we propagate the "host" value in different metrics output formats such that it makes it into the `stele` normalized metric as a tag, such that we can ensure we're properly comparing whatever the generated host value is, whether it's a tag or specialized field

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

- [x] Existing unit/correctness tests.
- [x] Added a new correctness test (`dsd-v1-api-series`) that runs in series V1 mode.

## References

DADP-72

Closes #1641, #1642

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -567,7 +567,7 @@ test-correctness: ## Runs the complete correctness suite (all test cases in para
 test-correctness-case: build-panoramic
 test-correctness-case: ## Runs a single correctness test case by name (usage: make test-correctness-case CASE=dsd-plain)
 	@echo "[*] Running '$(CASE)' correctness test case..."
-	@target/release/panoramic run -d $(shell pwd)/test/correctness -t $(CASE)
+	@target/release/panoramic run -d $(shell pwd)/test/correctness -t $(CASE) --no-tui
 
 .PHONY: build-panoramic
 build-panoramic: check-rust-build-tools
--- a/lib/saluki-components/src/encoders/datadog/metrics/mod.rs
+++ b/lib/saluki-components/src/encoders/datadog/metrics/mod.rs
@@ -171,7 +171,7 @@ pub struct DatadogMetricsConfiguration {
     /// the V2 endpoint (`/api/beta/sketches`) regardless of this setting.
     ///
     /// Defaults to `true`.
-    #[serde(default = "default_use_v2_api_series", rename = "use_v2_api_series")]
+    #[serde(default = "default_use_v2_api_series")]
     use_v2_api_series: bool,
 
     /// Additional tags to apply to all forwarded metrics.
--- a/lib/saluki-components/src/encoders/datadog/metrics/mod.rs
+++ b/lib/saluki-components/src/encoders/datadog/metrics/mod.rs
@@ -1,4 +1,4 @@
-use std::{num::NonZeroU64, time::Duration};
+use std::{fmt, num::NonZeroU64, time::Duration};
 
 use async_trait::async_trait;
 use datadog_protos::metrics as proto;
@@ -26,6 +26,7 @@ use saluki_error::{generic_error, ErrorContext as _, GenericError};
 use saluki_io::compression::CompressionScheme;
 use saluki_metrics::MetricsBuilder;
 use serde::Deserialize;
+use serde_json::{Map as JsonMap, Number as JsonNumber, Value as JsonValue};
 use tokio::{select, sync::mpsc, time::sleep};
 use tracing::{debug, error, warn};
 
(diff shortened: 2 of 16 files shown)


test/correctness/dsd-v1-api-series/config.yaml
```diff
@@ -0,0 +1,25 @@
...
+  additional_env_vars:
+    - DD_API_KEY=correctness-test
+comparison:
+  image: saluki-images/datadog-agent:testing-release
+  files:
+    - datadog.yaml:/etc/datadog-agent/datadog.yaml
+  additional_env_vars:
+    - DD_API_KEY=correctness-test
+    - DD_DATA_PLANE_ENABLED=true
+    - DD_DATA_PLANE_STANDALONE_MODE=true
```
webern: Isn't it better if most of our tests are not in standalone mode? Not for any particular reason just... because?

lib/saluki-components/src/encoders/datadog/metrics/mod.rs
```diff
@@ -147,6 +164,16 @@ pub struct DatadogMetricsConfiguration {
...
     zstd_compressor_level: i32,
 
+    /// Whether to use the V2 API for series metrics.
+    ///
+    /// When `true` (the default), series metrics are sent to the V2 protobuf endpoint (`/api/v2/series`). When
+    /// `false`, series metrics are sent to the legacy V1 JSON endpoint (`/api/v1/series`). Sketch metrics always use
+    /// the V2 endpoint (`/api/beta/sketches`) regardless of this setting.
+    ///
+    /// Defaults to `true`.
+    #[serde(default = "default_use_v2_api_series", rename = "use_v2_api_series")]
```
webern: That rename matches the field name, btw.

lib/saluki-components/src/encoders/datadog/metrics/mod.rs
```diff
@@ -714,6 +826,86 @@ fn encode_series_metric(
...
+        // V1 emits each point as a [timestamp, value] tuple — not a nested object.
+        let value_json = JsonNumber::from_f64(value)
+            .map(JsonValue::Number)
+            .unwrap_or_else(|| JsonValue::from(0));
+        points.push(JsonValue::Array(vec![JsonValue::from(timestamp), value_json]));
+    }
+    obj.insert("points".into(), JsonValue::Array(points));
+
+    // Walk the deduplicated tag set once, extracting the first `device:<value>` tag into the device JSON field while
+    // dropping `dd.internal.resource` (which is a V2-protobuf-only concept with no V1 representation).
```
webern, line 862: These sorts of things are matching Agent code?

Makefile
```diff
@@ -567,7 +567,7 @@ test-correctness: ## Runs the complete correctness suite (all test cases in para
 test-correctness-case: build-panoramic
 test-correctness-case: ## Runs a single correctness test case by name (usage: make test-correctness-case CASE=dsd-plain)
 	@echo "[*] Running '$(CASE)' correctness test case..."
-	@target/release/panoramic run -d $(shell pwd)/test/correctness -t $(CASE)
+	@target/release/panoramic run -d $(shell pwd)/test/correctness -t $(CASE) --no-tui
```
webern, line 570: 👍

webern (approved):

test/correctness/dsd-v1-api-series/config.yaml
tobz: I think this is just Claude having cargo cult'd the other tests, but you're right: we should get away from this in correctness tests.

I'll update the test definition.

lib/saluki-components/src/encoders/datadog/metrics/mod.rs
tobz: 🧐 

Yep, silly Claude. Good catch. Will fix.

lib/saluki-components/src/encoders/datadog/metrics/mod.rs
tobz: Yep.

test/correctness/dsd-v1-api-series/config.yaml
tobz: Updated in 35d387e1555.

lib/saluki-components/src/encoders/datadog/metrics/mod.rs
tobz: Fixed in https://github.com/DataDog/saluki/commit/915a0ce91141c8e18580985f0c33e2eee1464b9e.

PR #1647, opened 2026-05-13
feat(dogstatsd): added capture functionality 

lucastemb:
## Summary

Adds DogStatsD capture support to ADP, modeled after the Go Agent `dogstatsd-capture` flow.

This PR adds:

- the ability to start DogStatsD traffic captures through the control-plane API and CLI
- persistence of captured UDS packets to Datadog capture files
- replay tagger state in the capture trailer

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests 

## References

Related to the Go Agent DogStatsD capture/replay implementation used as the parity reference.


lib/saluki-components/src/sources/dogstatsd/replay/writer.rs
```diff
@@ -0,0 +1,819 @@
...
+            if !state.accepting {
+                return false;
+            }
+
+            state.traffic_tx.clone()
+        };
+
+        match traffic_tx {
+            Some(traffic_tx) => traffic_tx.send(record).is_ok(),
+            None => false,
```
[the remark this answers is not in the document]
lucastemb: Blocking behavior matches Go Agent capture semantics. Furthermore, switching this to `try_send` would make ADP captures best-effort, which is less than ideal for a feature that is supposed to serve as a debugging tool.

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -104,21 +127,35 @@ struct StatsResponse<'a> {
...
             }
         }
+        DogstatsdSubcommand::Capture(config) => {
+            let mut api_client = match DataPlaneAPIClient::from_config(bootstrap_config) {
+                Ok(client) => client,
+                Err(e) => {
+                    error!("Failed to create data plane API client: {:#}", e);
+                    std::process::exit(1);
+                }
+            };
```
tobz: It looks like maybe this was cleaned up to switch back to the regular data plane API client, but now we have a duplicate instantiation. We should just revert this so that we instantiate the client once at the top of `handle_dogstatsd_command`.

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -153,6 +190,21 @@ async fn handle_dogstatsd_stats(api_client: &mut DataPlaneAPIClient, cmd: StatsC
     Ok(())
 }
 
+async fn handle_dogstatsd_capture(
+    api_client: &mut DataPlaneAPIClient, cmd: CaptureCommand,
+) -> Result<(), GenericError> {
+    println!("Starting a dogstatsd traffic capture session...\n");
```
tobz:
```suggestion
    info!("Starting a DogStatsD traffic capture session...");
```

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -153,6 +190,21 @@ async fn handle_dogstatsd_stats(api_client: &mut DataPlaneAPIClient, cmd: StatsC
...
+
+    let capture_duration = cmd.capture_duration.to_string();
+    let capture_path = api_client
+        .dogstatsd_capture(&capture_duration, cmd.capture_path.as_deref(), cmd.compressed)
+        .await?;
+
+    println!("Capture started, capture file being written to: {capture_path}");
```
tobz:
```suggestion
    info!("Capture started. Data will be written to '{capture_path}'.");
```

bin/agent-data-plane/src/cli/utils.rs
```diff
@@ -170,6 +184,38 @@ impl DataPlaneAPIClient {
...
+        })
+        .error_context("Failed to serialize DogStatsD capture request.")?;
+        let req = Request::post(uri)
+            .header(CONTENT_TYPE, "application/json")
+            .body(body)
+            .expect("valid request");
+        let response_body = self
+            .client
+            .send(req)
+            .and_then(process_capture_response_body)
```
tobz: This seems almost identical to `process_response_body`... I would just use that instead.

lib/datadog-agent-commons/src/ipc/client/mod.rs
```diff
@@ -23,7 +23,7 @@ use tracing::warn;
 use crate::ipc::{config::RemoteAgentClientConfiguration, session::SessionId, tls::build_ipc_client_ipc_tls_config};
 
 mod bearer_auth;
-use self::bearer_auth::BearerAuthInterceptor;
+pub use self::bearer_auth::BearerAuthInterceptor;
```
tobz: Why?

lib/saluki-components/src/sources/dogstatsd/replay/writer.rs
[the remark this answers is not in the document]
tobz: I'm going to unresolve this because I think Codex's mention here is important.

Serializing every single turn of every UDS stream handler loop behind a rendezvous channel is not good. It's not good because it totally screws up the asynchronous runtime by blocking executor threads. I understand that's how the Datadog Agent does it, but Go has different semantics for how goroutines are scheduled which don't apply here.

I agree that fallible sends aren't what we want, so I think we need to update the code to set a _minimum_ capture depth size (since DDA always defaults to zero) of something like 1024. This at least gives the stream handler tasks a fighting chance of not hitting a full channel when doing their send at moderate DSD volumes.

lib/saluki-components/src/sources/dogstatsd/replay/writer.rs
```diff
@@ -0,0 +1,819 @@
...
+            "DogStatsD capture path is not configured and no explicit path was provided."
+        ));
+    }
+
+    ensure_target_dir_exists(target_dir)?;
+
+    let timestamp = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map_err(|e| generic_error!("Failed to compute capture file timestamp: {}", e))?
+        .as_secs();
```
tobz: `saluki_common::time::get_unix_timestamp()`.

lib/saluki-config/src/duration_string.rs
```diff
@@ -24,7 +24,7 @@ use snafu::Snafu;
 /// Accepted inputs:
 ///
 /// - Strings with Go time-unit suffixes: `"30s"`, `"1h30m"`, `"250ms"`, `"2h45m30s"`, `"1.5h"`. Valid suffixes: `ns`,
-///   `us`, `µs`, `ms`, `s`, `m`, `h`.
+///   `us`, `µs`, `μs`, `ms`, `s`, `m`, `h`.
```
tobz, line 27: Duplicate?

lib/saluki-io/src/net/unix/linux.rs
```diff
@@ -49,12 +49,14 @@ where
...
     let control_len = msg_hdr.control_len();
 
-    let process_identity = if control_len > 0 {
+    let (process_identity, ancillary) = if control_len > 0 {
         unsafe {
             ancillary_data.set_len(control_len);
-            let messages = ancillary_data.messages();
 
-            match messages
+            let ancillary = ancillary_data.bytes().to_vec();
```
tobz: Nah, we can't allocate a random extra buffer for every single UDS read.

After looking at the Agent-side implementation of this, it never even uses the ancillary data during replay... it's just totally ignored.

We should unwind all of these changes in `saluki-io` and just always set the ancillary data field in the capture record to an empty `Vec`. We're going to redo all of that stuff ASAP anyways, before implementing capture... but right now, let's avoid all this unnecessary work around the ancillary data.

tobz (changes requested):

lib/saluki-config/src/duration_string.rs
lucastemb: They are both distinct characters. (µ != μ). They look identical, but one is supposed to be the micro sign while the other is supposed to be mu. Go also explicitly accepts both.

tobz (approved):

PR #1652, opened 2026-05-14
fix(agent-data-plane): change default listen ports to avoid conflicts

### LINKED ISSUE #1651: ADP listen ports can interfere with non-standard Agent port configurations.
Seen in a real config:

You can, for example, set

`expvar_port: 5100`

In which case the Core Agent binds that port. Then ADP tries to bind it as its `data_plane.api_listen_address`, which defaults to 5100, and exists because is can't bind the port.

Edits:
- An idea that surfaced is to simply change ADP's defaults for these:
    - Port 5100 (default): `data_plane.api_listen_address`
    - Port 5101 (default): `data_plane.secure_api_listen_address`
    - Port 5102 (default): `data_plane.telemetry_listen_addr`
- Based on a search, 5100 is coming up a lot, so it might be worth changing.

## Resolved

- We will change our defaults to 5400/5401/5402, respectively.
- We also need to check the helm charts to see if there is any interaction there.

webern:
## Summary

ADP's default listen ports (5100/5101/5102) can collide with non-standard Agent port configurations such as `expvar_port: 5100`. Change the defaults to 5400/5401/5402 to avoid interference:

- Port 5400: `data_plane.api_listen_address`
- Port 5401: `data_plane.secure_api_listen_address`
- Port 5402: `data_plane.telemetry_listen_addr`

Updates all integration tests, SMP experiments, Makefile run targets, and documentation to match.

## Change Type
- [x] Bug fix

## How did you test this PR?

CI

## References

Closes #1651

Note:
- This seems necessary: https://github.com/DataDog/datadog-operator/pull/3014
- Changes to helm charts and deployment charts do not seem pressing since it's already a configuration in those (the "configured default" will just not match the binary's default)


thieman:
Worth noting this is not intended as some sort of actual fix to the problem, but based on current customer configs this reduces (or eliminates?) the incidence rate of active conflicts

thieman (approved):

tobz:
Note to passersby: we're holding on this until we cut the 1.0.0 release.

tobz:
Closing this for now as we have a pathway towards reducing the ports we need to expose, and a workaround if any initial customers have a pressing need to change the ports used by ADP.

PR #1654, opened 2026-05-14
fix(tls): support min_tls_version for forwarding

### LINKED ISSUE #1370: Support `min_tls_version` to enforce minimum TLS version for outbound connections.
The core agent exposes [`min_tls_version`][1] (default `"tlsv1.2"`). It accepts `tlsv1.0`, `tlsv1.1`, `tlsv1.2`, `tlsv1.3` and [applies the setting][2] to the TLS config on all outbound HTTPS connections.

ADP builds its rustls [`ClientConfig`][s1] with no protocol version override — rustls defaults to TLS 1.2 and TLS 1.3, which matches the core agent's default behavior. There is no config key to change this. A customer setting `min_tls_version: tlsv1.3` to enforce TLS 1.3-only would have no effect on ADP.

[1]: https://github.com/DataDog/datadog-agent/blob/141f12062790/pkg/config/setup/common_settings.go#L1219
[2]: https://github.com/DataDog/datadog-agent/blob/141f12062790/pkg/util/http/transport.go#L85
[s1]: https://github.com/DataDog/saluki/blob/e9373ef98013/lib/saluki-tls/src/lib.rs#L118-L148

thieman:
## Summary

Fixes #1370.

Adds core Agent-compatible `min_tls_version` support for ADP outbound Datadog HTTPS forwarding.

## Key changes

- Add a documented client-side minimum TLS version model in `saluki-tls`.
- Parse `min_tls_version` from ADP configuration with default `tlsv1.2` and accepted values `tlsv1.0`, `tlsv1.1`, `tlsv1.2`, and `tlsv1.3` case-insensitively.
- Apply the configured TLS floor to shared Datadog `TransactionForwarder` HTTPS clients.
- Preserve existing `skip_ssl_validation`, proxy, timeout, connection-age, telemetry, and FIPS behavior.
- Move `min_tls_version` out of unsupported config registry/docs and into supported/transparent config parity surfaces.

## Test plan

- `cargo test -p saluki-tls`
- `cargo test -p saluki-components common::datadog::config`
- `cargo test -p saluki-components common::datadog::io`
- `cargo test -p saluki-components config_registry`
- `cargo check -p saluki-tls`
- `cargo check -p saluki-io`
- `cargo check -p saluki-components`
- `make fmt`

### DIFF
--- a/lib/saluki-components/src/config_registry/datadog/forwarder.rs
+++ b/lib/saluki-components/src/config_registry/datadog/forwarder.rs
@@ -95,6 +95,17 @@ crate::declare_annotations! {
         test_json: None,
     };
 
+    /// `min_tls_version` — minimum TLS version for outbound HTTPS forwarding.
+    MIN_TLS_VERSION = SalukiAnnotation {
+        schema: &schema::MIN_TLS_VERSION,
+        support_level: SupportLevel::Full,
+        additional_yaml_paths: &[],
+        env_var_override: None,
+        used_by: &[structs::FORWARDER_CONFIGURATION],
+        value_type_override: None,
+        test_json: Some(r#""tlsv1.3""#),
+    };
+
     /// `skip_ssl_validation` — disables TLS certificate validation for Datadog intake forwarding.
     SKIP_SSL_VALIDATION = SalukiAnnotation {
         schema: &schema::SKIP_SSL_VALIDATION,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4313,6 +4313,7 @@ dependencies = [
 name = "saluki-tls"
 version = "0.1.0"
 dependencies = [
+ "facet",
  "rustls",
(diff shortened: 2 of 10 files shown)


lib/saluki-components/src/config_registry/datadog/forwarder.rs
```diff
@@ -95,6 +95,17 @@ crate::declare_annotations! {
         test_json: None,
     };
 
+    /// `min_tls_version` — minimum TLS version for outbound HTTPS forwarding.
+    MIN_TLS_VERSION = SalukiAnnotation {
+        schema: &schema::MIN_TLS_VERSION,
+        support_level: SupportLevel::Full,
```
tobz, line 101: Hmm, are we actually at full support if we don't allow going down past TLS 1.2? 🤔

PR #1671, opened 2026-05-18
fix(env): use mpsc for autodiscovery events

### LINKED ISSUE #1605: Change autodiscovery channel to MPSC to avoid loss.
## Context

In #1602, a Codex code review comment pointed out that we were changing the size of a broadcast channel used in an autodiscovery provider implementation, which being smaller, would risk losing events when there was a large burst of events.

While this is true, the bigger problem is that we can lose those events in the first place. This should really be a bounded MPSC channel to avoid that scenario ever occurring.

omribz156:
## Summary
Changes autodiscovery event delivery from a bounded broadcast channel to a bounded MPSC channel. Sends are now awaited, so a burst of autodiscovery events applies backpressure instead of letting a lagging broadcast receiver miss events.

This also updates the provider receiver type, hands out the single receiver through the existing `subscribe()` API, and adds a local-provider regression test that uses a capacity-1 channel with two config events.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?
- `cargo fmt --check`
- `git diff --check`
- `cargo metadata --no-deps --format-version 1`

I attempted `cargo test -p saluki-env autodiscovery::providers::local --lib`, but this Windows workspace is missing the MSVC linker / Windows SDK libraries (`link.exe`, `kernel32.lib`, etc.), so the command fails while compiling dependency build scripts before it reaches Saluki code.

This was implemented with Codex assistance, with the final diff reviewed against the issue scope and the existing autodiscovery provider tests.

## References

- Closes: https://github.com/DataDog/saluki/issues/1605

### DIFF
--- a/bin/agent-data-plane/src/internal/env/autodiscovery.rs
+++ b/bin/agent-data-plane/src/internal/env/autodiscovery.rs
@@ -23,9 +23,10 @@ use tracing::{debug, warn};
 /// autodiscovery API operates in a streaming fashion, which the provider uses to then broadcast updates to subscribers.
 #[derive(Clone)]
 pub struct RemoteAgentAutodiscoveryProvider {
-    sender: Sender<AutodiscoveryEvent>,
-    receiver: Arc<Mutex<Option<Receiver<AutodiscoveryEvent>>>>,
+    subscribers: AutodiscoverySubscribers,
 }
+
+type AutodiscoverySubscribers = Arc<Mutex<Vec<Sender<AutodiscoveryEvent>>>>;
 
 impl RemoteAgentAutodiscoveryProvider {
     /// Creates a new `RemoteAgentAutodiscoveryProvider` based on the given configuration, along with a [`Supervisor`] that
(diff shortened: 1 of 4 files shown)


tobz (changes requested):
While this PR is superficially correct, the implementation misses an important aspect: only one subscriber can exist now, and only one _ever_, not even just "at most one."

What we need to do instead is actually keep track of the separate subscriptions, creating an MPSC channel for every subscription and holding on to the senders. Every time we go to send, we'd use the result of the send -- did it succeed or is the receiver gone? -- and use that to influence whether or not we remove the attached sender from the list before forwarding the next event.

omribz156:
Thanks, fixed in the latest push.

Changed both autodiscovery providers to keep a list of per-subscriber `mpsc::Sender`s. Each `subscribe()` now creates its own bounded channel, and event delivery fans out to all current subscribers while pruning senders whose receivers have gone away. I also added local-provider coverage for fan-out plus closed-receiver pruning.

Verified with:
- `cargo fmt --check`
- `cargo metadata --no-deps --format-version 1`
- `git diff --check`

I also tried `cargo test -p saluki-env autodiscovery::providers::local --lib`, but this Windows environment fails earlier in `process-memory` with `cannot find type StatSource in this scope`, before the focused autodiscovery tests can run.

This follow-up was Codex-assisted, with the final diff reviewed before pushing.

tobz (approved):

tobz:
Looks like all that's left is fixing the formatting lint and then this should be good to go.

jszwedko:
@omribz156 I pushed a commit to fix the formatting, but your earlier commits are unsigned which is blocking merge. Are you able to amend the commits to sign and force-push?

omribz156:
Thanks, done. I rewrote the branch without changing the diff and force-pushed signed commits.

Verified on GitHub that all three PR commits now show valid signatures:
- `4e34b88072`
- `21918bc79a`
- `f0ba958b6d`

jszwedko:
(merged manually since check-licenses was the only failure; that was fixed on main, and I had trouble updating the branch here)

PR #1677, opened 2026-05-18
docs(dogstatsd): additional audit of dogstatsd options in the Core Agent

jszwedko:
## Summary

Prompted by the fact that we missed the `observability_pipelines_worker` and `use_api_v2.series` options in the initial audit of dogstatsd config options, I did an additional extensive audit of dogstatsd options that impact the Core Agent (by asking Claude to walk the full path from receiving dogstatsd packets to forwarding them) and identified additional ones I believe we need to investigate and possibly add support for.


tobz (approved):

tobz:
Actually, is this now breaking tests because of the switch to some settings as unsupported? 🤔

lib/saluki-components/src/config_registry/datadog/unsupported.rs
```diff
@@ -410,4 +414,340 @@ crate::declare_annotations! {
...
+        env_var_override: None,
+        used_by: &[],
+        value_type_override: None,
+        test_json: None,
+    };
+    /// `dogstatsd_experimental_http.enabled` - experimental HTTP/H2C DSD listener toggle.
+    DOGSTATSD_EXPERIMENTAL_HTTP_ENABLED = SalukiAnnotation {
+        schema: &schema::DOGSTATSD_EXPERIMENTAL_HTTP_ENABLED,
+        // Not implemented. #1682
+        support_level: SupportLevel::Incompatible(Severity::High),
```
tobz, line 445: This is experimental... this has gotta be low severity at _worst_.

lib/saluki-components/src/config_registry/datadog/unsupported.rs
```diff
@@ -410,4 +414,340 @@ crate::declare_annotations! {
...
+        env_var_override: None,
+        used_by: &[],
+        value_type_override: None,
+        test_json: None,
+    };
+    /// `dogstatsd_experimental_http.listen_address` - experimental HTTP DSD listener bind address.
+    DOGSTATSD_EXPERIMENTAL_HTTP_LISTEN_ADDRESS = SalukiAnnotation {
+        schema: &schema::DOGSTATSD_EXPERIMENTAL_HTTP_LISTEN_ADDRESS,
+        // Not implemented. #1682
+        support_level: SupportLevel::Incompatible(Severity::High),
```
tobz, line 456: Same comment as above.

lib/saluki-components/src/config_registry/datadog/unsupported.rs
```diff
@@ -410,4 +414,340 @@ crate::declare_annotations! {
...
+        env_var_override: None,
+        used_by: &[],
+        value_type_override: None,
+        test_json: None,
+    };
+    /// `enable_json_stream_shared_compressor_buffers` - shared JSON stream compressor buffer allocation.
+    ENABLE_JSON_STREAM_SHARED_COMPRESSOR_BUFFERS = SalukiAnnotation {
+        schema: &schema::ENABLE_JSON_STREAM_SHARED_COMPRESSOR_BUFFERS,
+        // Not implemented. #1686
+        support_level: SupportLevel::Incompatible(Severity::Low),
```
tobz, line 467: Irrelevant to ADP.

lib/saluki-components/src/config_registry/datadog/unsupported.rs
```diff
@@ -410,4 +414,340 @@ crate::declare_annotations! {
...
+        env_var_override: None,
+        used_by: &[],
+        value_type_override: None,
+        test_json: None,
+    };
+    /// `forwarder_requeue_buffer_size` - forwarder in-memory requeue buffer size.
+    FORWARDER_REQUEUE_BUFFER_SIZE = SalukiAnnotation {
+        schema: &schema::FORWARDER_REQUEUE_BUFFER_SIZE,
+        // Not implemented. #1680
+        support_level: SupportLevel::Incompatible(Severity::Low),
```
tobz, line 489: Irrelevant to ADP.

lib/saluki-components/src/config_registry/datadog/unsupported.rs
```diff
@@ -410,4 +414,340 @@ crate::declare_annotations! {
...
+        env_var_override: None,
+        used_by: &[],
+        value_type_override: None,
+        test_json: None,
+    };
+    /// `heroku_dyno` - Heroku dyno name override for agent telemetry.
+    HEROKU_DYNO = SalukiAnnotation {
+        schema: &schema::HEROKU_DYNO,
+        // Not implemented. #1685
+        support_level: SupportLevel::Incompatible(Severity::High),
```
tobz, line 511: Not seeing how this is high severity since we're not initially targeting Heroku support? Medium at best.

lib/saluki-components/src/config_registry/datadog/unsupported.rs
```diff
@@ -410,4 +414,340 @@ crate::declare_annotations! {
...
+        env_var_override: None,
+        used_by: &[],
+        value_type_override: None,
+        test_json: None,
+    };
+    /// `telemetry.dogstatsd.aggregator_channel_latency_buckets` - DSD-to-aggregator channel latency histogram buckets.
+    TELEMETRY_DOGSTATSD_AGGREGATOR_CHANNEL_LATENCY_BUCKETS = SalukiAnnotation {
+        schema: &schema::TELEMETRY_DOGSTATSD_AGGREGATOR_CHANNEL_LATENCY_BUCKETS,
+        // Not implemented. #1679
+        support_level: SupportLevel::Incompatible(Severity::Low),
```
tobz, line 599: Irrelevant to ADP.

lib/saluki-components/src/config_registry/datadog/unsupported.rs
```diff
@@ -410,4 +414,340 @@ crate::declare_annotations! {
...
+        env_var_override: None,
+        used_by: &[],
+        value_type_override: None,
+        test_json: None,
+    };
+    /// `telemetry.dogstatsd.listeners_channel_latency_buckets` - DSD listener channel latency histogram buckets.
+    TELEMETRY_DOGSTATSD_LISTENERS_CHANNEL_LATENCY_BUCKETS = SalukiAnnotation {
+        schema: &schema::TELEMETRY_DOGSTATSD_LISTENERS_CHANNEL_LATENCY_BUCKETS,
+        // Not implemented. #1679
+        support_level: SupportLevel::Incompatible(Severity::Low),
```
tobz, line 610: Irrelevant to ADP.

lib/saluki-components/src/config_registry/datadog/unsupported.rs
```diff
@@ -410,4 +414,340 @@ crate::declare_annotations! {
...
+        env_var_override: None,
+        used_by: &[],
+        value_type_override: None,
+        test_json: None,
+    };
+    /// `telemetry.dogstatsd.listeners_latency_buckets` - DSD listener processing latency histogram buckets.
+    TELEMETRY_DOGSTATSD_LISTENERS_LATENCY_BUCKETS = SalukiAnnotation {
+        schema: &schema::TELEMETRY_DOGSTATSD_LISTENERS_LATENCY_BUCKETS,
+        // Not implemented. #1679
+        support_level: SupportLevel::Incompatible(Severity::Low),
```
tobz, line 621: Irrelevant to ADP.

tobz (changes requested):

lib/saluki-components/src/config_registry/datadog/unsupported.rs
jszwedko: It is experimental, but if the user has it enabled it and is using it then they will start losing metrics if they enable ADP. That is the test I've been using for assigning severity: what is the behavioral change? Is there the chance things would completely break? High. Is the behavior different, but functionally the same (e.g. performance knobs)? Medium. Is there negligible difference to behavior? Low.

Note, no customers should have this enabled in practice, but in the off-chance they do we wouldn't want them to be running ADP.

lib/saluki-components/src/config_registry/datadog/unsupported.rs
jszwedko: Same as my earlier comment, if this happens to be set then ADP's lack of support would result in incorrect metrics, but, again, we don't _expect_ this to be set so users shouldn't hit this one in practice.

lib/saluki-components/build.rs
```diff
@@ -167,7 +167,10 @@ fn yaml_value_to_json_str(value: &serde_yaml::Value) -> Option<String> {
             let escaped = s.replace('\\', "\\\\").replace('"', "\\\"");
             Some(format!("\"{}\"", escaped))
         }
-        serde_yaml::Value::Sequence(seq) if seq.is_empty() => Some("[]".to_string()),
+        serde_yaml::Value::Sequence(seq) => {
```
jszwedko, line 170: Needed to handle the default of `["container.memory.usage","container.cpu.usage"]` for `autoscaling.failover.metrics`.

lib/saluki-components/src/config_registry/datadog/unsupported.rs
jszwedko: Makes sense, I expect that many of these options will not be applicable to ADP, but I'd like to investigate them further (via the linked issues) and document for each one why it isn't supported.

docs/agent-data-plane/configuration/dogstatsd.md
```diff
@@ -1,14 +1,14 @@
 # Configuring DogStatsD on Agent Data Plane
```
jszwedko, line 1: Make sure to expand this file.

lib/saluki-components/src/config_registry/datadog/unsupported.rs
tobz: Hmmm, that's fair.

lib/saluki-components/src/config_registry/datadog/unsupported.rs
```diff
@@ -410,4 +410,325 @@ crate::declare_annotations! {
...
+        env_var_override: None,
+        used_by: &[],
+        value_type_override: None,
+        test_json: None,
+    };
+    /// `enable_payloads.json_to_v1_intake` - legacy /api/v1/intake JSON payload delivery.
+    ENABLE_PAYLOADS_JSON_TO_V1_INTAKE = SalukiAnnotation {
+        schema: &schema::ENABLE_PAYLOADS_JSON_TO_V1_INTAKE,
+        // Not implemented. #1686
+        support_level: SupportLevel::Incompatible(Severity::High),
```
tobz: This one also strikes me as not high, but low: the Agent only uses it for [sending process metadata](https://github.com/DataDog/datadog-agent/blob/a1e28a8e0de788f6414754f85ae38d7d880ce145/pkg/serializer/serializer.go#L375).

lib/saluki-components/src/config_registry/datadog/unsupported.rs
tobz: Fairrrr.

lib/saluki-components/src/config_registry/datadog/unsupported.rs
```diff
@@ -410,4 +410,314 @@ crate::declare_annotations! {
...
+        env_var_override: None,
+        used_by: &[],
+        value_type_override: None,
+        test_json: None,
+    };
+    /// `multi_region_failover.enabled` - multi-region failover mode.
+    MULTI_REGION_FAILOVER_ENABLED = SalukiAnnotation {
+        schema: &schema::MULTI_REGION_FAILOVER_ENABLED,
+        // Not implemented. #1678
+        support_level: SupportLevel::Incompatible(Severity::High),
```
webern, line 555: This was flagged as breaking internal deployments. But... since we already silently ignore this, and because the PR represents good data that we should merge, I'd suggest switching this `Medium`, merging, and then following up on the deployment changes needed.

```suggestion
        // TODO: follow up with necessary deployment changes
        support_level: SupportLevel::Incompatible(Severity::Medium),
```

webern (changes requested):
I'd suggesting switching that one that we flagged in conversation to `Medium` in order to merge this rather than leaving it hanging.

lib/saluki-components/src/config_registry/datadog/unsupported.rs
jszwedko: Ended up making the internal deployment changes first so I think we can use this as-is.

tobz (approved):

lib/saluki-components/src/config_registry/datadog/unsupported.rs
jszwedko: Aaah, I misunderstood what this option does. I think ADP should just ignore this one as only applicable to the Core Agent since ADP won't ever be sending these payloads (right?). Thanks for flagging.

webern (approved):

PR #1655, opened 2026-05-14
feat(dogstatsd): add support for `observability_pipelines_worker.metrics` / `vector.metrics` config keys

aqian01:
## Summary

[Source](https://github.com/DataDog/saluki/issues/1586)

Adds ADP support for routing metric series and sketches to Observability Pipelines Worker using:

- `observability_pipelines_worker.metrics.enabled`
- `observability_pipelines_worker.metrics.url`

legacy aliases:
- `vector.metrics.enabled`
- `vector.metrics.url`

When enabled with a valid URL, metric payloads use OPW as the primary endpoint. Non-metric payloads continue using the normal Datadog primary endpoint. additional_endpoints continue receiving dual-shipped payloads.

Changes:

- Added OPW metrics config parsing with vector.metrics.* fallback.
- Added per-path routing for series and sketches in the Datadog forwarder.
- Kept dynamic API-key refresh attached to OPW-routed traffic.
- Registered OPW metrics keys as supported config.
- Kept OPW logs/traces keys explicitly unsupported.
- Updated DogStatsD config docs and known config inventory.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Non-functional (chore, refactoring, docs)
- [ ] Performance

## How did you test this PR?

Unit tests

Manual Testing
Started a local OPW listener:
`nc -l 127.0.0.1 8080`
Started the dev core Agent with ADP enabled:
```
DD_DATA_PLANE_FORCE_ENABLE=true \
DD_DATA_PLANE_ENABLED=true \
DD_DATA_PLANE_DOGSTATSD_ENABLED=true \
./bin/agent/agent run -c dev/dist/datadog.yaml
```
Started ADP with OPW metrics routing enabled:
```
DD_API_KEY=<api_key> \
DD_CMD_PORT=5001 \
DD_DATA_PLANE_ENABLED=true \
DD_DATA_PLANE_DOGSTATSD_ENABLED=true \
DD_DOGSTATSD_PORT=9191 \
DD_DOGSTATSD_SOCKET=/tmp/adp-dogstatsd-dgram.sock \
DD_DOGSTATSD_STREAM_SOCKET=/tmp/adp-dogstatsd-stream.sock \
DD_DATA_PLANE_TELEMETRY_ENABLED=true \
DD_DATA_PLANE_TELEMETRY_LISTEN_ADDR=tcp://127.0.0.1:5102 \
[...]

PR #1690, opened 2026-05-18
fix(config): move OPW/Vector logs+traces keys to ignored_keys

jszwedko:
## Summary

OPW logs and traces routing is entirely out of scope for ADP, so the `observability_pipelines_worker.{logs,traces}.{enabled,url}` and `vector.{logs,traces}.{enabled,url}` keys have no effect on ADP behavior. Moving them from `unsupported.rs` (which surfaces `Incompatible` warnings to operators) to `ignored_keys.yaml` (silently ignored) avoids misleading noise for users who have these keys set.

These annotations were added in https://github.com/DataDog/saluki/pull/1655 but I think this was a mistake.

## Test plan
- [ ] `cargo check --workspace` passes
- [ ] Verify config registry emits no warning for `observability_pipelines_worker.logs.enabled` when set

### DIFF
--- a/lib/saluki-components/etc/ignored_keys.yaml
+++ b/lib/saluki-components/etc/ignored_keys.yaml
@@ -2484,6 +2484,14 @@
   reason: initial bulk
 - name: notable_events.enabled
   reason: initial bulk
+- name: observability_pipelines_worker.logs.enabled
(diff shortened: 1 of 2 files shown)


tobz (approved):
Hmm, yeah, fair. 👍🏻

PR #1698, opened 2026-05-19
feat(io): Add `min_tls_version` support to ADP forwarding

aqian01:
## Summary
[Source](https://github.com/DataDog/saluki/issues/1370)

- Added explicit TLS minimum-version control to Saluki’s TLS stack so ADP can honor `min_tls_version` instead of always using rustls defaults.
- Wired the setting through the HTTP client surface and Datadog forwarder config.
- Updated config registry metadata and DogStatsD docs to reflect partial compatibility: TLS 1.2/1.3 are enforced, while TLS 1.0/1.1 are accepted and clamped to TLS 1.2.

## Testing
- Added unit tests for forwarder config parsing, TLS version mapping, and TLS 1.3-only behavior against a TLS 1.2-only test server.
- Verified with workspace checks and targeted `nextest` runs across `saluki-tls`, `saluki-io`, and `saluki-components`.
- Ran formatting and diff checks successfully.

## Manual QA
- Started a local HTTPS server restricted to TLS 1.2 only.
- Ran ADP with `min_tls_version: tlsv1.2`, `skip_ssl_validation: true`, and a local `dd_url`; sent a DogStatsD metric; confirmed the server received `POST /api/v2/series` with `tls=TLSv1.2`.
- Ran ADP with the same setup but `min_tls_version: tlsv1.3`; sent a DogStatsD metric; confirmed ADP failed with `ServerTlsVersionIsDisabledByOurConfig` and the server logged TLS protocol-version handshake failures.

These checks prove the setting is applied to real Datadog forwarder HTTPS traffic: TLS 1.2 is allowed when configured, and the same TLS 1.2 endpoint is rejected when ADP is configured as TLS 1.3-only.

### DIFF
--- a/lib/saluki-components/src/common/datadog/config.rs
+++ b/lib/saluki-components/src/common/datadog/config.rs
@@ -188,12 +188,18 @@ pub struct ForwarderConfiguration {
     /// Saluki clamps them to TLS 1.2 because rustls does not support older protocol versions.
     #[serde(default = "default_min_tls_version")]
     min_tls_version: String,
+
+    /// Parsed minimum TLS protocol version for Datadog intake forwarding.
+    #[serde(skip)]
+    #[facet(opaque)]
+    parsed_min_tls_version: TlsMinimumVersion,
 }
 
 impl ForwarderConfiguration {
     /// Creates a new `ForwarderConfiguration` from the given configuration.
     pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
         let mut forwarder_config = config.as_typed::<Self>()?;
+        forwarder_config.parsed_min_tls_version = min_tls_version_from_config_value(&forwarder_config.min_tls_version);
 
         // Handle fixing up the forwarder storage path if it's empty.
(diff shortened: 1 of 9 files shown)


thieman (approved):

lib/saluki-components/src/common/datadog/config.rs
```diff
@@ -28,6 +29,39 @@ const fn default_forwarder_connection_reset_interval() -> u64 {
...
+}
+
+fn min_tls_version_from_config_value(value: &str) -> TlsMinimumVersion {
+    let trimmed = value.trim();
+    match trimmed.to_lowercase().as_str() {
+        MIN_TLS_VERSION_TLS10 | MIN_TLS_VERSION_TLS11 => {
+            warn!(
+                config_key = "min_tls_version",
+                value = trimmed,
+                "Configured TLS minimum version is lower than rustls supports; using tlsv1.2."
```
jszwedko, line 48: Nit: we shouldn't expose implementation details in logs consumed by customers.

```suggestion
                "Configured TLS minimum version is lower than supported; using tlsv1.2."
```

PR #1699, opened 2026-05-19
chore(dogstatsd mapper): add support for `dogstatsd_mapper_cache_size` to cache mapper results

### LINKED ISSUE #1687: Investigate DogStatsD server internals: `dogstatsd_host_socket_path` and `dogstatsd_mapper_cache_size`.
## Problem

Two DogStatsD server config keys need formal investigation:

**`dogstatsd_host_socket_path`** — the default host-mounted UDS socket directory injected by the Kubernetes Cluster Agent admission controller into pods. In the core agent this key is read only by the admission controller mutator to set the socket path in pod specs; it is not read by the DogStatsD server itself. ADP likely does not need this key, but this should be confirmed and documented.

**`dogstatsd_mapper_cache_size`** — the LRU cache capacity for DogStatsD mapper regex/wildcard match results. The core agent caches mapper results per metric name to avoid re-running pattern matching on every sample. ADP's mapper currently re-applies all profile rules on every event with no result cache. This is a meaningful behavioral difference for workloads with high-cardinality metric names that hit mapper rules.

## What needs to happen

For `dogstatsd_host_socket_path`: confirm ADP never reads this key. If confirmed, document as Not Planned with explanation.

For `dogstatsd_mapper_cache_size`: implement a mapper result cache in ADP's DogStatsD mapper transform so that repeated metric names skip full regex evaluation, and expose `dogstatsd_mapper_cache_size` to configure its capacity. If a cache is impractical in ADP's streaming architecture, document the divergence.

tobz:
## Summary

This PR adds support for `dogstatsd_mapper_cache_size` in the DogStatsD Mapper transform, allowing mapper results to be cached. As mappings are fixed based on a given metric name, their transformations can be cached to reduce most of the overhead of calculating the new metric name and the set of tags derived from the old name.

Our implementation has one notable divergence from the Core Agent: when `dogstatsd_mapper_cache_size` is set to `0`, we disable caching but allow the mapper to continue processing. In contrast, the Datadog Agent will silently disabling mapping. I believe the Datadog Agent's behavior is actually a bug and I'm treating it as such here, leaving the filing of an issue on the Agent side as a follow-up rather than a pre-requisite before implementing support for caching in the first place.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

- Added five new unit tests around caching of mapper results.
- Existing unit, integration, and correctness tests.

## References

DADP-18

Closes #1687

### DIFF
--- a/docs/agent-data-plane/configuration/dogstatsd.md
+++ b/docs/agent-data-plane/configuration/dogstatsd.md
@@ -72,6 +72,7 @@ default values.
 
 | Config Key                          | Description                      | Agent Behavior                                 | ADP Behavior                                                   |
 | ----------------------------------- | -------------------------------- | ---------------------------------------------- | -------------------------------------------------------------- |
+| `dogstatsd_mapper_cache_size`       | Mapper result LRU cache size     | `0` disables mapping; positive sizes the LRU   | `0` disables the cache only; mapping still runs ([#1687])      |
 | `dogstatsd_metrics_stats_enable`    | Enable per-metric debug stats    | Config toggle                                  | Gates debug log; stats API on-demand ([#1352], [#1356])        |
 | `dogstatsd_stats_enable`            | Enable internal stats endpoint   | Config toggle                                  | On-demand via API ([#1352])                                    |
 | `dogstatsd_stats_buffer`            | Internal stats buffer size       | Configurable                                   | On-demand via API ([#1352])                                    |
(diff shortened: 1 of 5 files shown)


lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
```diff
@@ -46,6 +54,19 @@ pub struct DogStatsDMapperConfiguration {
...
+    /// When enabled, mapped metrics will be cached by name to avoid repeat evaluation of the configured mapper rules.
+    ///
+    /// When set to `0`, the cache is disabled.
+    ///
+    /// Defaults to `1000`.
+    #[serde(
+        rename = "dogstatsd_mapper_cache_size",
+        default = "default_dogstatsd_mapper_cache_size"
+    )]
+    cache_size: usize,
```
thieman, line 68: Do we know what happens if someone passes a negative int here for some reason? If it underflows, might be better to error somewhere?

tobz: It would error during deserialization, because such an input cannot be correctly deserialized to an unsigned integer.

docs/agent-data-plane/configuration/dogstatsd.md
```diff
@@ -193,6 +194,21 @@ OpenMetrics check pointed at the agent's own telemetry endpoint. ADP has a separ
...
+
+ADP and the core agent both cache mapper results to skip regex evaluation on repeat metric names.
+With the default value of `1000`, and with any positive integer, behavior matches the core agent:
+results are cached in an LRU keyed by the original metric name, including a negative-cache entry for
+names that match no profile.
+
+The two implementations diverge when this setting is `0`. In the core agent, `0` is rejected by the
+underlying LRU library, which causes the entire mapper to be silently disabled — mapping profiles
+configured by `dogstatsd_mapper_profiles` are not applied. In ADP, `0` disables the result cache
+only; mapping profiles still run, so each metric pays the regex evaluation cost without amortization.
```
jszwedko, line 207: Interesting. I almost wonder if we should change the Core Agent behavior to match this.

tobz: We _absolutely_ should. It's clearly a correctness issue.

thieman (approved):

PR #1707, opened 2026-05-20
fix(datadog): honor forwarder http protocol

### LINKED ISSUE #1361: Allow using HTTP/1.1 via `forwarder_http_protocol`.
The core agent exposes [`forwarder_http_protocol`][1] (default `"auto"`). `"auto"` [negotiates HTTP/2 via ALPN][2] with HTTP/1.1 fallback; `"http1"` [forces HTTP/1.1 only][3].

ADP's HTTP client [always enables both HTTP/1.1 and HTTP/2 via ALPN][4] — equivalent to `"auto"`. There is no way to force HTTP/1.1 only. A customer setting `forwarder_http_protocol: http1` will get no effect.

Likely simple to implement — configure the hyper-rustls connector to advertise only `http/1.1` in ALPN instead of calling `enable_all_versions()` when set to `"http1"`. Might be worth checking whether anyone actually uses this setting first.

[1]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/pkg/config/setup/common_settings.go#L1544
[2]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/comp/forwarder/defaultforwarder/domain_forwarder.go#L278-L294
[3]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/pkg/util/http/transport.go#L234-L241
[4]: https://github.com/DataDog/saluki/blob/dbb968c2db21/lib/saluki-io/src/net/client/http/conn.rs#L369

thieman:
## Summary
- Honor `forwarder_http_protocol` in the Datadog forwarder configuration.
- Add HTTP protocol selection to the shared HTTP client so `http1` uses hyper-rustls HTTP/1-only mode.
- Promote `forwarder_http_protocol` from unsupported to fully supported in the config registry.

## Key changes
- `auto` remains the default and now advertises both `h2` and `http/1.1` through ALPN because `hyper-rustls` now enables its `http1` feature.
- `http1` uses `hyper_rustls::HttpsConnectorBuilder::enable_http1()`, which leaves TLS ALPN empty per hyper-rustls HTTP/1-only behavior.
- Invalid values such as `http2` are rejected during forwarder config deserialization.

## Test plan
- [x] `make fmt`
- [x] `make test`
- [x] `cargo nextest run -p saluki-io net::client::http::conn::tests::auto_protocol_advertises_h2_and_http1_alpn net::client::http::conn::tests::http1_protocol_leaves_alpn_empty --no-fail-fast`
- [x] `cargo test -p saluki-io net::client::http::conn::tests:: -- --nocapture`
- [x] `cargo test -p saluki-components common::datadog::config::tests::forwarder_http_protocol -- --nocapture`
- [x] `cargo test -p saluki-components config_registry::datadog::registry_tests:: -- --nocapture`
- [x] `cargo test -p saluki-components config_registry::classifier::tests:: -- --nocapture`

Fixes #1361.

### DIFF
--- a/lib/saluki-components/src/common/datadog/config.rs
+++ b/lib/saluki-components/src/common/datadog/config.rs
@@ -3,7 +3,7 @@ use std::time::Duration;
 use facet::Facet;
 use saluki_config::GenericConfiguration;
 use saluki_error::GenericError;
-use saluki_io::net::client::http::TlsMinimumVersion;
+use saluki_io::net::client::http::{HttpProtocol, TlsMinimumVersion};
 use serde::Deserialize;
 use tracing::warn;
 
@@ -64,7 +64,8 @@ fn min_tls_version_from_config_value(value: &str) -> TlsMinimumVersion {
 
 /// HTTP protocol selection for the Datadog forwarder.
 #[repr(u8)]
-#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Facet)]
+#[derive(Clone, Copy, Debug, Default, Deserialize, Eq, PartialEq, Facet)]
+#[serde(rename_all = "lowercase")]
 #[cfg_attr(test, derive(serde::Serialize))]
 pub enum ForwarderHttpProtocol {
     /// Automatically negotiate HTTP/2 with HTTP/1.1 fallback.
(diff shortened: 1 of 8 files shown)


lib/saluki-components/src/common/datadog/config.rs
```diff
@@ -28,6 +28,44 @@ const fn default_forwarder_connection_reset_interval() -> u64 {
     0
 }
 
+/// HTTP protocol selection for the Datadog forwarder.
+#[repr(u8)]
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Facet)]
+#[cfg_attr(test, derive(serde::Serialize))]
+pub enum ForwarderHttpProtocol {
```
tobz, line 70: You should be able to skip the custom `Deserialize` implementation, I believe, by doing something like:

```rust
#[derive(Deserialize)]
#[serde(untagged, rename_all = "lowercase")]
```

lib/saluki-components/src/common/datadog/config.rs
```diff
@@ -229,6 +273,11 @@ impl ForwarderConfiguration {
         self.endpoint_buffer_size
     }
 
+    /// Returns the HTTP protocol selection for outgoing forwarder requests.
+    pub fn http_protocol(&self) -> saluki_io::net::client::http::HttpProtocol {
```
tobz: Import this type instead of fully qualifying it.

tobz (approved):
Approved with non-blocking nits.

lib/saluki-components/src/common/datadog/config.rs
thieman: [Claude Sonnet 4] Fixed by deriving `Deserialize` with `#[serde(rename_all = "lowercase")]` and removing the custom implementation.

lib/saluki-components/src/common/datadog/config.rs
thieman: [Claude Sonnet 4] Fixed by importing `HttpProtocol` alongside `TlsMinimumVersion` and using the imported type in the accessor and conversion impl.

PR #1708, opened 2026-05-20
feat(agent-data-plane): version printing subcommand

webern:
## Summary

```
agent-data-plane version now prints the version as defined by saluki_metadata.
```

This could be handy during an incident response. Although I prefer `--version` over the Go style subcommand, it seems Argh is pretty minimal (compared to, e.g. Clap) and it was just less weird to add it as another subcommand.

## Change Type
- [x] New feature

## How did you test this PR?

```
$ make build-adp
(...)
```

```
$ ./target/devel/agent-data-plane version
v1.2.0-351262c79d
```

```
./target/devel/agent-data-plane version --json
{
  "full_name": "Agent Data Plane",
  "short_name": "data-plane",
  "identifier": "adp",
  "git_hash": "351262c79d",
  "version": "1.2.0",
  "build_time": "0000-00-00T00:00:00-00:00",
  "dev_build": true,
  "target_arch": "aarch64-apple-darwin"
}
```

```
$ ./target/devel/agent-data-plane version --help
Usage: agent-data-plane version [--json]

Prints the agent-data-plane version.

Options:
  --json            print the complete saluki_metadata version information as a
                    JSON object
  --help, help      display usage information
```
## References

N/A

### DIFF
--- a/bin/agent-data-plane/src/cli/version.rs
+++ b/bin/agent-data-plane/src/cli/version.rs
@@ -3,9 +3,21 @@ use argh::FromArgs;
 /// Prints the agent-data-plane version.
 #[derive(FromArgs, Debug)]
 #[argh(subcommand, name = "version")]
-pub struct VersionCommand {}
+pub struct VersionCommand {
+    /// print the complete saluki_metadata version information as a JSON object
+    #[argh(switch)]
+    pub json: bool,
+}
 
 /// Prints the agent-data-plane version.
-pub async fn handle_version_command() {
-    println!(env!("CARGO_PKG_VERSION"));
+pub async fn handle_version_command(json: bool) {
+    let app_data = saluki_metadata::get_app_details();
+    if json {
+        println!(
+            "{}",
+            serde_json::to_string_pretty(app_data).expect("Unable to serialize version information.")
+        )
+    } else {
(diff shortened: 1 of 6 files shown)


bin/agent-data-plane/src/cli/version.rs
```diff
@@ -0,0 +1,11 @@
+use argh::FromArgs;
+
+/// Prints the agent-data-plane version.
+#[derive(FromArgs, Debug)]
+#[argh(subcommand, name = "version")]
+pub struct VersionCommand {}
+
+/// Prints the agent-data-plane version.
+pub async fn handle_version_command() {
+    println!(env!("CARGO_PKG_VERSION"));
```
tobz: Ideally we'd be using the full version string based on `saluki_metadata`, following the pattern from `cli/run.rs`.

Version number-wise, it should be the same as `CARGO_PKG_VERSION`... but we get commit SHA, build date, etc, as well.

tobz (changes requested):

bin/agent-data-plane/src/cli/version.rs
webern: Ah, I didn't see that. Thanks. I created a quick look at version-sha, and a `--json` flag for printing the full object.

tobz (approved):

PR #1711, opened 2026-05-20
feat(dogstatsd): added replay functionality

lucastemb:
## What

Adds DogStatsD replay support across the staged replay layers:

- parses DogStatsD capture files and tagger state
- exposes replay trigger and stop control endpoints
- sends replay packets back into ADP over UDS with replay credentials
- marks replay-injected packets on receive and resolves tags from captured tagger state

## Why

Replay should preserve the origin/tag attribution from the capture file instead of resolving replayed traffic against the current live workload state.

## Validation

- `cargo check -p saluki-components -p saluki-context -p saluki-io --tests`

## Notes

Opened as draft for human review.


bin/agent-data-plane/src/cli/run.rs
```diff
@@ -380,12 +390,22 @@ fn check_and_warn_config(config: &GenericConfiguration) -> Result<(), GenericErr
     Ok(())
 }
 
+#[derive(Default)]
+struct TopologyControlSurfaces {
+    dogstatsd: Option<DogStatsDControlSurface>,
+}
```
tobz, line 386: I actually really like this concept of grouping together the necessary types to provide a "control surface" over a given pipeline.

One thing we should do as an immediate follow-up PR, I think, is actually unify this such that we also include the DSD Stats stuff in here as well. I think that would mean moving things around so we basically pass back just the API handler, and `create_topology` actually creates the component configuration and all of that... but it seems doable.

That would also, I believe, let us remove the need for a separate `DogStatsDControlPlaneConfiguration` type.

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -438,15 +458,19 @@ async fn create_topology(
     }
 
     if dp_config.dogstatsd().enabled() {
-        dsd_capture_api_handler =
-            Some(add_dsd_pipeline_to_blueprint(&mut blueprint, config, env_provider, dsd_stats_config).await?);
+        let (capture_api_handler, replay_api_handler) =
+            add_dsd_pipeline_to_blueprint(&mut blueprint, config, env_provider, dsd_stats_config).await?;
```
tobz: I would say that we should make `add_dsd_pipeline_to_blueprint` actually return `DogStatsdControlSurface` so we don't have to care up here about the fact it's two separate pieces, etc.

lib/saluki-context/src/origin.rs
```diff
@@ -128,6 +128,13 @@ pub struct RawOrigin<'a> {
     ///
     /// See [`ExternalData`] for more information.
     external_data: Option<RawExternalData<'a>>,
+
+    /// Whether this origin describes a packet injected by a replay session.
+    ///
+    /// Set to `true` when the receive-side packet handler observes the replay credentials marker on incoming
+    /// `SCM_CREDENTIALS`. Downstream tag resolution branches on this flag to consult the captured tagger snapshot
+    /// instead of the live workload provider.
+    is_replay: bool,
```
tobz: I don't love this implementation detail of DSD replay leaking out through `RawOrigin<'a>`, and I think we could do this in a simpler way with just process ID.

Right now, we're using a spoofed GID in `SCM_CREDENTIALS` to signal that the packet is a replay packet... but we could _actually_ just use a "tagged" PID. On Linux, the maximum PID value possible is right around 4,195,000... but we carry that value in `u32`, which has an upper bound of _~4,000,000,000_. This means that the upper bit of `u32` will never be used for a real PID.

If we use just the PID, we can "tag" the real PID by setting the upper bit to one before sending it (`let spoofed_pid = real_pid | 1 << 32`), and looking for that during origin detection (`let is_replay = pid & (1 << 32) != 0`) instead of doing `origin.is_replay()`.

tobz (changes requested):

lib/saluki-context/src/origin.rs
lucastemb: Done. I like that idea. For future reference, is the general philosophy to keep shared types as generic and universally applicable as possible? That is to say, should we avoid adding niche or feature-specific flags that might muddy or detract from a structs intended purpose?

lib/saluki-io/src/net/unix/replay_send.rs
```diff
@@ -0,0 +1,127 @@
...
+/// The captured PID is packed into `uid` and the replay marker GID into `gid`, matching the protocol the receive-side
+/// packet handler watches for.
+pub async fn send_replay_packet(
+    socket: &UnixDatagram, payload: &[u8], captured_pid: i32, replay_gid: u32,
+) -> io::Result<usize> {
+    let our_pid = std::process::id() as libc::pid_t;
+    let creds = libc::ucred {
+        pid: our_pid,
+        uid: captured_pid as u32,
+        gid: replay_gid,
```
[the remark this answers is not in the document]
lucastemb: Doesn't apply. Should be run by a user or operator with sudo privileges anyways.

lib/saluki-io/src/net/unix/non_linux.rs
```diff
@@ -49,3 +49,13 @@ where
...
+
+/// Stub: replay packet sending is Linux-only because it depends on `SCM_CREDENTIALS`.
+pub async fn send_replay_packet(
+    _socket: &tokio::net::UnixDatagram, _payload: &[u8], _captured_pid: i32, _replay_gid: u32,
+) -> io::Result<usize> {
+    Err(io::Error::new(
+        io::ErrorKind::Unsupported,
+        "DogStatsD replay is only supported on Linux",
+    ))
+}
```
tobz: I wouldn't even have this method exposed... only expose the Linux implementation, and then `#[cfg(...)]` gate it at the callsite in the CLI subcommand.

lib/saluki-io/src/net/unix/replay_send.rs
```diff
@@ -0,0 +1,127 @@
+//! Linux-only helpers for sending DogStatsD replay traffic.
```
tobz: Can we move this code to live in `linux.rs` and just call it something like `uds_sendmsg_with_creds`? We'd also want the doc comment for it to mirror what we have for functions like `unix_recvmsg`, except `Sends` instead of `Receives`, and mention that it uses the provided creds, the permissions needed, yadda yadda.

Also, we should have it take `ProcessCredentials` instead of the individual PID/GID values.

Basically, this is DogStatsD-flavored right now, and we want to make it generic.

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -1,14 +1,21 @@
...
 use std::collections::{HashMap, HashSet};
-use std::time::Duration;
+use std::path::{Path, PathBuf};
+use std::time::{Duration, Instant};
 
 use argh::{FromArgValue, FromArgs};
 use comfy_table::{presets::ASCII_FULL_CONDENSED, Cell, ContentArrangement, Row, Table};
+use saluki_components::sources::{
+    TimestampResolution, TrafficCaptureReader, DEFAULT_REPLAY_LOOPS, REPLAY_CREDENTIALS_GID,
+};
```
tobz: This is definitely a code smell to me, but also it feels non-blocking. I'd say we should follow up by seeing if we can reorganize some of the DSD replay code into `datadog-agent-commons`.

tobz (approved):

Issue #1476, opened 2026-04-24
Investigate `statsd_forward_host` / `statsd_forward_port` — raw packet forwarding gap


jszwedko:
The core agent's DogStatsD server supports raw packet forwarding: when both [`statsd_forward_host`][1] and [`statsd_forward_port`][1] are set, every received DogStatsD packet is forwarded verbatim (UDP) to the configured host:port, acting as a transparent proxy to another StatsD/DogStatsD server.

ADP has no equivalent. The config inventory marks both keys as [`MISSING`][s1] and the configuration docs note they are ["Not planned"][s2].

However, usage data shows **~253,000 customer agents** have `statsd_forward_host` and `statsd_forward_port` configured. This is a significant customer dependency that warrants investigation before ADP is broadly adopted.

Determine whether ADP needs to implement packet forwarding, or whether customers using this feature can be migrated to an alternative approach. If forwarding will remain unsupported, document the gap clearly so customers know what to expect.

[1]: https://github.com/DataDog/datadog-agent/blob/main/comp/dogstatsd/server/server.go
[s1]: https://github.com/DataDog/saluki/blob/main/docs/agent-data-plane/configuration/dogstatsd/known-configs.json
[s2]: https://github.com/DataDog/saluki/blob/main/docs/agent-data-plane/configuration/dogstatsd.md

jszwedko:
Closed by https://github.com/DataDog/saluki/pull/1712

PR #1712, opened 2026-05-20
feat(dogstatsd): Add DogStatsD packet forwarding support

aqian01:
## Summary
[Source](https://github.com/DataDog/saluki/issues/1476)

- Add DogStatsD forwarding support for `statsd_forward_host` and `statsd_forward_port`.
- Forward each decoded/framed DogStatsD message over UDP to the configured forwarding target before parsing, filtering, mapping, or aggregation.
- Support forwarding for messages received over UDP, UDS datagram, UDS stream, and TCP while preserving normal ingestion behavior.
- Keep forwarding best effort: setup and send failures are logged/tracked through telemetry and do not stop DogStatsD ingestion.
- Align transient forwarding backpressure by awaiting enqueue of forwarded copies instead of dropping them when the forwarding queue is temporarily full.
- Use a fixed forwarding queue capacity and support IPv4 and IPv6 forwarding targets without modeling the core Agent packet-buffer/flush pipeline.
- Promote `statsd_forward_host` and `statsd_forward_port` to supported DogStatsD config keys.
- Keep `dogstatsd_packet_buffer_size` and `dogstatsd_packet_buffer_flush_timeout` marked not applicable for ADP because ADP decodes DogStatsD inline.
- Update DogStatsD docs/config inventory, integration coverage, and correctness coverage for forwarding behavior.

## Design Notes

This PR intentionally forwards decoded DogStatsD messages rather than preserving core Agent UDP packet-buffer grouping or UDS/TCP stream outer-frame grouping.

For example, a stream frame containing:

```text
metric.one:1|c
metric.two:2|c
```

is forwarded as two UDP messages:

```text
metric.one:1|c
```

and:

```text
metric.two:2|c
```

This follows review feedback to avoid emulating the core Agent ingest -> packet-buffer -> worker pipeline in ADP. ADP still forwards the same DogStatsD messages, but it does not attempt byte-for-byte parity of forwarded UDP packet boundaries.

## Testing

- `make fmt`
- `cargo test -p saluki-components sources::dogstatsd::tests --lib -- --nocapture`
- `cargo test -p saluki-components sources::dogstatsd::framer::tests --lib -- --nocapture`
- `cargo test -p saluki-components packet_forwarder --lib -- --nocapture`
- `cargo test -p saluki-io deser::framing::tests --lib -- --nocapture`
- `cargo test -p panoramic dogstatsd_forwarding --bin panoramic -- --nocapture`
- `cargo check --workspace`
- `git diff --check`

## Coverage

- Unit coverage for forwarding one decoded message payload, IPv6 forwarding targets, queue backpressure, send-failure telemetry, and `statsd_forward_port` enable/disable behavior.
- Integration coverage for UDP malformed payloads, UDP valid metrics, TCP stream messages, UDS datagram messages, and UDS stream messages.
- Correctness coverage compares forwarded DogStatsD messages rather than exact UDP packet grouping, so core Agent and ADP can differ in packet boundaries while preserving forwarded message content.


lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -271,6 +302,50 @@ pub struct DogStatsDConfiguration {
...
+    statsd_forward_host: Option<MetaString>,
+
+    /// The port to forward raw DogStatsD packets to over UDP.
+    ///
+    /// Forwarding is enabled only when this value is within the UDP port range and `statsd_forward_host` is non-empty.
+    /// If this value is negative or greater than 65535, forwarding is disabled without failing DogStatsD ingestion.
+    ///
+    /// Defaults to 0.
+    #[serde(rename = "statsd_forward_port", default = "default_statsd_forward_port")]
+    statsd_forward_port: i64,
```
tobz: We should make this `u16`. That's the only thing a valid port can be.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -981,6 +1092,298 @@ impl Metrics {
...
+impl PacketForwarder {
+    fn from_config(config: &DogStatsDConfiguration) -> Option<Self> {
+        if config.has_invalid_statsd_forward_port() {
+            warn!(
+                port = config.statsd_forward_port,
+                min = MIN_STATSD_FORWARD_PORT,
+                max = MAX_STATSD_FORWARD_PORT,
+                "Invalid statsd forward port. Packet forwarding disabled."
+            );
+            return None;
```
tobz: Is this how it works in the Core Agent? If the port is invalid, it just ignores it and packet forwarding is disabled?

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -981,6 +1092,298 @@ impl Metrics {
...
+        let forwarder = self.clone();
+        spawn_traced_named("dogstatsd-packet-forwarder-setup", async move {
+            forwarder.connect().await;
+        });
+    }
+
+    async fn connect(&self) {
+        let host = &self.target_host;
+        let port = self.target_port;
+        let targets = match timeout(FORWARDER_RESOLVE_TIMEOUT, resolve_forward_targets(host, port)).await {
```
tobz: We don't actually need to do this separately. `tokio::net::UdpSocket::connect` already takes the address parameter as anything that implements `ToSocketAddrs`, which includes `(&str, u16)`.

Using that instead (still coupled with the timeout, though) lets us get rid of a _ton_ of the code here.

tobz (changes requested):
I left some specific feedback, but overall, I want to simplify this design: just forward the individual metrics packets (don't differentiate between the outer UDS frame and inner newline-delimited frames, and remove the code that captures/holds on to outer frames in `NestedFramer`) and avoid all of the "packet buffer" stuff.

My reasoning here is:

- we're still sending all of the same metrics in the end, but we skip needing to handle nested vs unnested frames
- the packet buffer infrastructure on the Agent side is really a consequence of the ingest -> worker model, so modeling it here isn't something we really _need_ to do
- additionally, modeling the packet buffer infrastructure means time-based non-determinism (due to the flushing support) which is harder to test for correctness than if all we do if blast out the individual metric packets in real-time

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -981,6 +1040,168 @@ impl Metrics {
...
+        while let Some(packet) = packets_rx.recv().await {
+            match self.socket.send(&packet.payload).await {
+                Ok(bytes_sent) => packet.record_success(bytes_sent),
+                Err(e) => {
+                    packet.record_error();
+                    debug!(target = %self.target, error = %e, "Failed to forward DogStatsD packet.");
+                }
+            }
+        }
+    }
```
tobz: A thought I had about this was that forwarding the unframed[^1] payload means we wouldn't be including a newline delimiter which _would_ be a difference if the original metrics were coming in with that delimiter. That said, we're also sending them one-by-one so we don't have any issues in terms of incorrectly smooshing two or more payloads together and the remote side being unable to tease them apart... just that if they _require_ a newline delimiter, and the original message had one... we would be forwarding it _without_ the delimiter here.

That said, the original StatsD "specification"[^2] never dictated needing a newline delimiter for anything _other_ than splitting multiple metrics up in a single payload, so I think we're fine here. Mostly just musing out loud. 😅 

[^1]: When a DSD message has the length prefix, or newline delimiter, it's "framed"... so it becomes "unframed" when we trim all of that off.
[^2]: https://github.com/b/statsd_spec

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -981,6 +1040,168 @@ impl Metrics {
...
+    packet_forwarding_errors: Counter,
+}
+
+impl ForwardPacket {
+    fn from_payload(payload: Bytes, metrics: &Metrics) -> Self {
+        Self {
+            payload,
+            packets_forwarded: metrics.packets_forwarded().clone(),
+            bytes_forwarded: metrics.bytes_forwarded().clone(),
+            packet_forwarding_errors: metrics.packet_forwarding_errors().clone(),
```
tobz: Just make `Metrics` clone and pass a clone of it to `PacketForwarder` when we create it initially.

It's not that it isn't cheap enough to clone these for every packet, but it _is_ wasteful to do so, and it means we're moving around more memory than we need to.

tobz (changes requested):
A few nits here and there.

One last major thing I think we should do is move all of the forwarder code -- except for the config settings themselves which obviously need to stay -- into a new `forwarder` module... just since `mod.rs` is already very long as is.

I think this might end up requiring us to also move the `Metrics` struct into its own module so we can reference it from the forwarder code, which is also good for cleaning up `mod.rs`.

tobz (approved):

PR #1713, opened 2026-05-21
feat(io): add composable HTTP retry predicates

atanzu:
## Summary
Replace the fixed standard HTTP classifier with a predicate-backed classifier that preserves the default retry behavior while allowing callers to add narrowly scoped retry predicates.

Expose `DefaultHttpRetryPolicy::with_backoff_and_classifier` so callers can pass a customized `StandardHttpClassifier` without reimplementing the standard retry lifecycle.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Add new unit-tests to validate the classifier behavior.

## References

- AGTMETRICS-504

This PR is a part of the series which addesses #1540.

### DIFF
--- a/lib/saluki-components/src/common/datadog/retry.rs
+++ b/lib/saluki-components/src/common/datadog/retry.rs
@@ -177,7 +177,7 @@ impl RetryConfiguration {
     }
 
     /// Creates a new [`DefaultHttpRetryPolicy`] based on the forwarder configuration.
-    pub fn to_default_http_retry_policy(&self) -> DefaultHttpRetryPolicy {
+    pub fn to_default_http_retry_policy<B: 'static>(&self) -> DefaultHttpRetryPolicy<B> {
         let retry_backoff = ExponentialBackoff::with_jitter(
             Duration::from_secs_f64(self.backoff_base),
             Duration::from_secs_f64(self.backoff_max),
--- a/lib/saluki-io/src/net/util/retry/classifier/http.rs
+++ b/lib/saluki-io/src/net/util/retry/classifier/http.rs
@@ -1,6 +1,29 @@
-use http::StatusCode;
+use std::sync::Arc;
+
+use http::Response;
 
 use super::RetryClassifier;
+
+/// A predicate that decides whether a response should be treated as retriable.
+///
+/// The predicate receives the response and returns `true` if the response should be retried.
+pub type HttpRetryPredicate<B = ()> = Arc<dyn Fn(&Response<B>) -> bool + Send + Sync>;
+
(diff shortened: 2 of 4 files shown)


webern (commented):
This is good, and I like that you broke out the interface change into its own PR first.

I still think this interface is better and more extensible:

```rust
pub type HttpRetryPredicate<B> = Arc<dyn Fn(&http::Response<B>) -> bool + Send + Sync>;
```

` pub fn with_predicate(mut self, predicate: HttpRetryPredicate<B>)`

But if I'm wrong or missing something let me know!

atanzu:
> This is good, and I like that you broke out the interface change into its own PR first.
> 
> I still think this interface is better and more extensible:
> 
> ```rust
> pub type HttpRetryPredicate<B> = Arc<dyn Fn(&http::Response<B>) -> bool + Send + Sync>;
> ```
> 
> ` pub fn with_predicate(mut self, predicate: HttpRetryPredicate<B>)`
> 
> But if I'm wrong or missing something let me know!

@webern I thought of that again and yes, IO is a Saluki framework component which means we need to extend it in a flexible manner, welcoming future use-cases. I've updated my PR to use the full response like you've suggested.

lib/saluki-components/src/common/datadog/retry.rs
```diff
@@ -177,7 +177,7 @@ impl RetryConfiguration {
     }
 
     /// Creates a new [`DefaultHttpRetryPolicy`] based on the forwarder configuration.
-    pub fn to_default_http_retry_policy(&self) -> DefaultHttpRetryPolicy {
+    pub fn to_default_http_retry_policy<B: 'static>(&self) -> DefaultHttpRetryPolicy<B> {
```
webern, line 180: Is
```
`static
```

required to get it to compile?

```suggestion
    pub fn to_default_http_retry_policy<B>(&self) -> DefaultHttpRetryPolicy<B> {
```

webern (approved):
Looks good!

lib/saluki-components/src/common/datadog/retry.rs
atanzu: Unfortunately yes:
```
error[E0310]: the parameter type `B` may not live long enough
   --> lib/saluki-components/src/common/datadog/retry.rs:188:9
    |
188 |         DefaultHttpRetryPolicy::with_backoff(retry_backoff)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |         |
    |         the parameter type `B` must be valid for the static lifetime...
    |         ...so that the type `B` will meet its required lifetime bounds
```

PR #1714, opened 2026-05-21
chore(docs): fix remaining docs errors + add scaffolding to run docs lints in CI

tobz:
## Summary

This PR fixes the remaining errors detected by Vale, and adds a new CI job to check/enforce that there are no error-level docs lints in PRs.

Crucially, we've done the following:

- changed `make check-docs` to filter to errors, since otherwise we were still at around ~1000 warnings
- changed `make check-docs` to filter out some paths that didn't make sense, like generated/vendored code
- fixed all the remaining errors
- added a new `check-docs` job to the `test` stage in CI (disabled at the moment)

There's still some tweaks to make to our style guide and agents ruleset to hopefully drive agents towards looping on fixing issues that `make check-docs` uncovers, or avoid them in the first place... but I'm optimistic it will manage just fine in the interim.

As well, the new `check-docs` CI job is set to manual mode only, as we need to update the build CI image to include the `vale` binary. I plan to do that as a follow-up PR.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

- [x] Ensured `make check-docs` and `make check-all` passed locally.

## References

DADP-2

### DIFF
--- a/bin/correctness/stele/src/traces/stats.rs
+++ b/bin/correctness/stele/src/traces/stats.rs
@@ -431,9 +431,9 @@ impl From<&proto::ClientGroupedStats> for AggregationKey {
 
 /// Aggregator for client statistics.
 ///
-/// Client statistics are aggregated by a number of fields that generate correspond to a specific span: service, name,
-/// and operation. Additional fields are used to further group the statistics, such as response codes and tags. This is
-/// referred to as the "aggregation key".
+/// Client statistics are represented by generating an "aggregation key" for a given span, and then grouping statistics
+/// by that key. The aggregation key includes fields like service, span name, span operation, response code, and more.
+/// This is meant to group like spans together,
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
 pub struct ClientStatisticsAggregator {
     groups: FastHashMap<AggregationKey, BucketedClientStatistics>,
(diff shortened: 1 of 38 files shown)


bin/correctness/stele/src/traces/stats.rs
```diff
@@ -476,9 +476,6 @@ impl ClientStatisticsAggregator {
     }
 
     /// Returns a reference to the aggregated statistics groups.
-    ///
-    /// Groups are split by "aggregation key", which is a combination of select fields in each client stats payload,
-    /// roughly corresponding to a specific span: name, operation, kind, tags, and so on.
```
jszwedko, line 481: Seems like we are losing some detail here 🤔

jszwedko (approved):

bin/correctness/stele/src/traces/stats.rs
tobz: Yes, but, it's already explained in the doc comments for `ClientStatisticsAggregator` itself.

Just avoiding repetition here since it's just an opportunity to have contradicting/out-of-sync information.

jszwedko: Aaah, I see, missed that.

PR #1715, opened 2026-05-21
feat(dogstatsd): support serializer payload limit configs

rayz:
## Summary

Adds ADP support for Datadog Agent serializer payload size limit configs and wires each setting to the matching payload path.

| Config setting | Affects in ADP | Agent behavior matched |
| --- | --- | --- |
| `serializer_max_payload_size` | V1 series, sketches, events, service checks | Generic serializer compressed limit: [JSON series/service checks](https://github.com/DataDog/datadog-agent/blob/main/pkg/serializer/internal/stream/json_payload_builder.go#L118), [events](https://github.com/DataDog/datadog-agent/blob/main/pkg/serializer/internal/metrics/events.go#L131), [sketches](https://github.com/DataDog/datadog-agent/blob/main/pkg/serializer/internal/metrics/sketch_series_list.go#L121) |
| `serializer_max_uncompressed_payload_size` | V1 series, sketches, events, service checks | Generic serializer uncompressed limit: [JSON series/service checks](https://github.com/DataDog/datadog-agent/blob/main/pkg/serializer/internal/stream/json_payload_builder.go#L119), [events](https://github.com/DataDog/datadog-agent/blob/main/pkg/serializer/internal/metrics/events.go#L132), [sketches](https://github.com/DataDog/datadog-agent/blob/main/pkg/serializer/internal/metrics/sketch_series_list.go#L122) |
| `serializer_max_series_payload_size` | V2 series | Series compressed limit: [V2 series builder](https://github.com/DataDog/datadog-agent/blob/main/pkg/serializer/internal/metrics/iterable_series.go#L189) |
| `serializer_max_series_uncompressed_payload_size` | V2 series | Series uncompressed limit: [V2 series builder](https://github.com/DataDog/datadog-agent/blob/main/pkg/serializer/internal/metrics/iterable_series.go#L190) |

This also moves the supported settings from unsupported to supported config registry entries and updates the DogStatsD config compatibility docs.

Configured values are clamped to endpoint/API bounds. This is stricter than the current Agent behavior, which passes configured values through directly.

`serializer_max_series_points_per_payload` is intentionally not added here because the Agent-side removal is pending.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?
Existing Unit Tests + CI


lib/saluki-components/src/encoders/datadog/metrics/mod.rs
```diff
@@ -227,6 +263,10 @@ impl EncoderBuilder for DatadogMetricsConfiguration {
         let mut series_rb = RequestBuilder::new(series_encoder, compression_scheme, RB_BUFFER_CHUNK_SIZE).await?;
         series_rb.with_max_inputs_per_payload(self.max_metrics_per_payload);
 
+        if series_endpoint == MetricsEndpoint::SeriesV2 {
+            series_rb.with_len_limits(self.max_series_uncompressed_payload_size, self.max_series_payload_size)?;
+        }
```
tobz: So this is the main thing that strikes me as insufficient: we should be using the API-enforced limits as an upper bound (`min(user_payload_size_limit, api_payload_size_limit)`)`.

We should never let users send payloads that risk being rejected due to exceeding size limits.

tobz: Also, to that end, I would change the doc comments for the config fields to say that every endpoint has an upper bound that is enforced by the API, and whatever value is specified here (as in: the user-supplied config value) will be clamped to that limit.

.. and then I'd probably also uniformly apply it to the V1 series payloads as well unless those are actually hard-coded in the Agent?

tobz (changes requested):

lib/saluki-components/src/encoders/datadog/metrics/mod.rs
rayz: v1 uses `serializer_max_payload_size` and`serializer_max_uncompressed_payload_size` for series, sketches, events and service checks. updated to support this ^ as well as the upperbound clamp in [505fa77](https://github.com/DataDog/saluki/pull/1715/commits/505fa7780caa6a447351745e815fc731a9b60570)

 pr description also includes a table of what each new config setting affects.

tobz (approved):

tobz (approved):

PR #1717, opened 2026-05-21
feat(config): retry 403 when secrets are active

atanzu:
## Summary
Thread the live GenericConfiguration into the forwarder retry policy so 403 Forbidden can be retried only when secrets management is active.

The policy re-evaluates the secrets gate on each 403, so runtime configuration updates are reflected without rebuilding endpoint services.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Added test cases to cover the new functionality.

### DIFF
--- a/lib/saluki-components/src/common/datadog/io.rs
+++ b/lib/saluki-components/src/common/datadog/io.rs
@@ -95,6 +95,7 @@ where
 pub struct TransactionForwarder<B> {
     context: ComponentContext,
     config: ForwarderConfiguration,
+    live_config: Option<GenericConfiguration>,
     telemetry: ComponentTelemetry,
     metrics_builder: MetricsBuilder,
     client: HttpClient,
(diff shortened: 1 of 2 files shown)


lib/saluki-config/src/lib.rs
```diff
@@ -497,6 +509,11 @@ fn build_figment_from_sources(sources: &[ProviderSource]) -> Figment {
     })
 }
 
+fn compute_secrets_in_use(figment: &Figment) -> bool {
```
tobz: This is only relevant to Datadog Agent-specific code, so this needs to move out of `saluki-config` and into the classifier predicate we're generating.

tobz (changes requested):

lib/saluki-config/src/lib.rs
atanzu: Agree, this is not a framework-level feature.
However, the straightforward implementation lacks the ability to do quick checks, because on each 403 we check vs the config against the pre-computed bool like it was in rev1. Maybe we could somehow utilize the notification mechanism to still be able to do so. Which might _probably_ not worth it because soon we'll enable key updates for additional endpoints, so we anyway go to the config before making every API call.

tobz (approved):

Issue #1580, opened 2026-05-05
Some integration tests are timing out locally when run in parallel.


webern:
The dogstatsd-bind-host integration test consistently fails due to a slow container_start phase that consumes nearly the entire test timeout.

Observed failure:

```
FAIL dogstatsd-bind-host (62.31s)

Assertions:
  - cancelled (5.84s)
    Test was cancelled.

Phase timings:
  container_start (53.87s)
  dynamic_vars    (284.35ms)
  assertions      (5.84s)
```

```
  Run 1: dogstatsd-default-bind failed, dogstatsd-non-local-overrides-bind-host passed.
  Run 2: dogstatsd-default-bind passed, dogstatsd-non-local-overrides-bind-host failed.

  driver_config_build (95.71us)
  container_start      (52.63s)
  assertions           ( 7.37s)   <- cancelled
  write_logs           ( 4.00ms)
  cleanup              ( 2.32s)
```

- The panoramic dynamic mechanism seems like a potential culprit. Otherwise I'm not sure why containers are taking so long to start just for these tests.
- Possibly the new cancellation timeout mechanism is too tight, or includes more of the docker setup time than it did before, but that doesn't make sense. I don't know why these are taking so long.

## Logs

[panoramic-20260505-120114.zip](https://github.com/user-attachments/files/27395713/panoramic-20260505-120114.zip)
[panoramic-20260505-120559.zip](https://github.com/user-attachments/files/27395711/panoramic-20260505-120559.zip)
[panoramic-20260505-120820.zip](https://github.com/user-attachments/files/27395712/panoramic-20260505-120820.zip)

## Edits/Updates

It looks like the issue is that the 60s timeout used to only cover the assertions phase of an integration test, and the change I made in eb250ff, which causes the timeout to be applied at the top level (i.e. including all phases of the test) makes 60s a bit too tight.  Here are the timings from logs left over in my tempdir from before the change. Occasionally the total was more than 60 seconds. I think a timeout of 90 seconds seems reasonable from this table. But I admittedly do not understand why certain container start phases are so slow. Perhaps it is due to parallelization.

```

  ┌─────────────────┬─────────────────────────────────────────┬────────┬─────────────────┬────────────┐
  │       Run       │                  Test                   │ Total  │ container_start │ assertions │
  ├─────────────────┼─────────────────────────────────────────┼────────┼─────────────────┼────────────┤
[...]

Issue #1579, opened 2026-05-05
Test `dsd-mapper-blocklist` fails locally.


webern:
Tested at commit e8a875de79bc617bed49151c0830c111c5c6bf40 (current `main` as of 2026-05-05).

## Steps to reproduce

```
git clone git@github.com:DataDog/saluki.git
cd saluki
make build-correctness-tools-image
make build-datadog-agent-image
make test-correctness-case CASE=dsd-mapper-blocklist
```

## What happens

The `dsd-mapper-blocklist` correctness test fails. The test configures a mapper profile that renames `mapper.blocklist.raw` to `mapper.blocklist.mapped`, and a `statsd_metric_blocklist` entry for `mapper.blocklist.mapped`. DDA (baseline) correctly drops the renamed metric. ADP (comparison) emits it.

```
Analyzing 98 unfiltered metrics from baseline target, and 129 unfiltered metrics from comparison target.
Filtered 66 internal telemetry metric(s) from baseline, and 66 internal telemetry metric(s) from comparison.
ERROR Mismatch in unique metrics between baseline and comparison!
ERROR Metrics in baseline but not in comparison:
ERROR Metrics in comparison but not in baseline:
ERROR   - mapper.blocklist.mapped[env:worker, service:worker] (type: rate)
ERROR   - mapper.blocklist.mapped[version:pod-c, version:staging] (type: rate)
ERROR   - mapper.blocklist.mapped[env:staging, pod:v2] (type: rate)
ERROR   - mapper.blocklist.mapped[env:worker, service:prod] (type: rate)
ERROR   - mapper.blocklist.mapped[pod:pod-a, version:pod-a] (type: rate)
ERROR   - mapper.blocklist.mapped[service:prod, version:staging] (type: rate)
ERROR   - mapper.blocklist.mapped[env:pod-a, pod:worker] (type: rate)
ERROR   - mapper.blocklist.mapped[env:v1] (type: rate)
ERROR   - mapper.blocklist.mapped[env:api, version:api] (type: rate)
ERROR   - mapper.blocklist.mapped[env:v2, service:prod] (type: rate)
ERROR   - mapper.blocklist.mapped[env:api, pod:worker] (type: rate)
ERROR   - mapper.blocklist.mapped[env:pod-c, service:pod-b] (type: rate)
ERROR   - mapper.blocklist.mapped[pod:api, version:worker] (type: rate)
ERROR   - mapper.blocklist.mapped[env:prod, env:v2] (type: rate)
[...]

PR #1724, opened 2026-05-22
feat(correctness): fan out from one millstone process

webern:
## Summary

Reduce payload timing divergence by running a single millstone process that fans out to both `baseline` and `comparison` in correctness tests.

Before this change, the millstone container ran `sh -c '... & P1=$! ... & P2=$! ...'` to fork two
independent millstones - one per target which is one place where drift may be introduced. 

In follow up PRs I intend to extend the solution to include
- waiting for a ready signal from both consumers before starting
- respecting a wall-clock-time-based boundary pause to avoid the bucket boundary

Key changes:

- `bin/correctness/millstone`: `Config` accepts `targets:` as a named map (`baseline:` /
  `comparison:`) instead of a single `target:`; `TargetSender` fans out each generated payload to
  all configured sinks; driver and corpus collapsed to a single send loop; errors include the
  target name.
- `bin/correctness/panoramic`: Docker (`runner.rs`) and k8s (`k8s.rs`) paths each spawn one
  millstone container/pod per test instead of two. New shared helpers in `correctness/config.rs`
  (`resolve_group_placeholders`, `millstone_targets_all_sockets`, `millstone_first_network_port`)
  substitute the `$GROUP` placeholder per-target on the host. The resolved YAML is written under
  the per-test `log_dir` (deliberately not `mounts_dir`, which is overlaid into the agent
  containers).
- 19 `test/correctness/*/millstone.yaml` migrated mechanically to the `targets:` shape.

## Change Type

- [x] Enhancement

## How did you test this PR

Locally on macOS against rebuilt `correctness-tools` and `datadog-agent` images at this commit:

- Full suite (`-d test/integration/cases -d test/correctness`) parallel (`-p 4` default): **51/52
  pass**. Single failure: `dsd-origin-detection-matrix/unified-high-cardinality` — window-edge
  divergence, the residual window boundary failure class that subsequent PRs intend to target.
- Full suite sequential (`-p 1`): **52/52 pass**. This is slow! Parallelization is a good idea if we can fix it.

Compare to the pre-fanout parallel baseline of 50/52 with all four `dsd-origin-detection-matrix/*`
variants plus `dsd-service-checks` failing under value-divergence.

No leaked airlock resources after either run. I did see leaked airlock resources on an aborted run, and that is something I also intend to work on downstream.

## References

Not directly, but documenting my ongoing sensitivity to local flakes:
- #1579 
- #1580

### DIFF
--- a/bin/correctness/millstone/src/config.rs
+++ b/bin/correctness/millstone/src/config.rs
@@ -326,6 +326,10 @@ targets:
             Err(e) => e,
         };
         let msg = err.to_string();
-        assert!(msg.contains("baseline"), "error should mention duplicate key, got: {}", msg);
+        assert!(
+            msg.contains("baseline"),
+            "error should mention duplicate key, got: {}",
+            msg
+        );
     }
 }
--- a/bin/correctness/millstone/src/config.rs
+++ b/bin/correctness/millstone/src/config.rs
@@ -1,14 +1,11 @@
 use std::{
-    fmt,
     num::NonZeroUsize,
     path::{Path, PathBuf},
 };
 
+use indexmap::IndexMap;
 use saluki_error::{generic_error, ErrorContext as _, GenericError};
-use serde::{
-    de::{self, MapAccess, Visitor},
-    Deserialize, Deserializer,
-};
+use serde::Deserialize;
 
 #[derive(Deserialize)]
 #[serde(try_from = "String")]
@@ -146,9 +143,9 @@ pub struct Config {
     /// Each generated payload is sent to every target in turn (fan-out), in declared order, before
     /// the next payload is generated. Keys are free-form target names (used in logs and errors);
     /// values are address strings parsed the same way as a single target. At least one target is
-    /// required.
-    #[serde(deserialize_with = "deserialize_targets")]
-    pub targets: Vec<(String, TargetAddress)>,
+    /// required. `IndexMap` preserves the order keys appear in the YAML file and rejects
+    /// duplicate keys via `serde_yaml`.
+    pub targets: IndexMap<String, TargetAddress>,
 
     /// Output volume.
     ///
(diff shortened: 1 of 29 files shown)


bin/correctness/millstone/src/corpus.rs
```diff
@@ -50,10 +50,16 @@ fn get_finalized_corpus_blueprint(config: &Config) -> Result<CorpusBlueprint, Ge
...
+    // stream, the framing flag applies to the whole corpus.
     match &mut blueprint.payload {
         Payload::DogStatsD(dsd_config) => {
-            if let TargetAddress::Unix(_) = config.target {
+            let any_unix_stream = config
+                .targets
+                .iter()
+                .any(|(_, addr)| matches!(addr, TargetAddress::Unix(_)));
+            if any_unix_stream {
                 dsd_config.length_prefix_framed = true;
```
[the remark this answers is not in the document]
webern: Not applicable. We don't plan to mix configs like that.

bin/correctness/panoramic/src/correctness/runner.rs
```diff
@@ -247,47 +248,52 @@ impl CorrectnessRunner {
...
+        // Resolve $GROUP per target on the host. For Docker, the substitution value is simply
+        // the entry's key, which is also the network alias the corresponding agent advertises.
+        let template = std::fs::read_to_string(&self.millstone_config.config_path).with_error_context(|| {
+            format!(
+                "Failed to read millstone config template '{}'.",
+                self.millstone_config.config_path.display()
+            )
+        })?;
+        let resolved =
+            crate::correctness::config::resolve_group_placeholders(&template, |key| Some(key.to_string()))?;
```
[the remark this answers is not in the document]
webern: Yeah, this is lame, but it works for now.

bin/correctness/millstone/src/config.rs
```diff
@@ -186,6 +196,140 @@ impl Config {
...
     }
 }
+
+/// Deserializes a YAML mapping into an order-preserving `Vec<(String, TargetAddress)>`.
+///
+/// Send order matches declared order, so we cannot use `HashMap` or `BTreeMap`. Visiting the
+/// underlying `MapAccess` directly preserves the order produced by `serde_yaml`'s YAML parser
+/// without adding an `indexmap` dependency. Duplicate keys are rejected because they would
+/// silently shadow each other in the resulting map and produce a misleading send count.
+fn deserialize_targets<'de, D>(deserializer: D) -> Result<Vec<(String, TargetAddress)>, D::Error>
```
webern: Is this necessary? Why not use a `BTreeMap<String, TargetAddress> `? I think the tests should be eliminated as well. This is not worthy of a deserializer and tests.

bin/correctness/panoramic/src/correctness/k8s.rs
```diff
@@ -115,7 +115,7 @@ pub async fn run_k8s_correctness_test(name: String, config: Config, tctx: TestCo
 
     let run_start = Instant::now();
 
-    let use_socket_wait = is_socket_target(&millstone_template);
+    let use_socket_wait = crate::correctness::config::millstone_targets_all_sockets(&millstone_template);
```
webern:
```suggestion
    let use_socket_wait = millstone_targets_all_sockets(&millstone_template);
```

bin/correctness/panoramic/src/correctness/k8s.rs
```diff
@@ -163,7 +163,7 @@ pub async fn run_k8s_correctness_test(name: String, config: Config, tctx: TestCo
     // in parallel with the agent pods (which is fine, as the socket won't appear until the agent is
     // ready regardless).
     let tcp_readiness_checks = if !use_socket_wait {
-        if let Some(port) = extract_target_port(&millstone_template) {
+        if let Some(port) = crate::correctness::config::millstone_first_network_port(&millstone_template) {
```
webern:
```suggestion
        if let Some(port) = millstone_first_network_port(&millstone_template) {
```

bin/correctness/panoramic/src/correctness/k8s.rs
```diff
@@ -216,13 +219,18 @@ pub async fn run_k8s_correctness_test(name: String, config: Config, tctx: TestCo
...
-    // For TCP/gRPC targets $GROUP is the agent pod's cluster IP (the only routable address
-    // from the separate millstone pod).
-    let (baseline_group_value, comparison_group_value) = if is_socket_target(&config_content) {
-        ("baseline".to_string(), "comparison".to_string())
+    let config_with_origin = substitute_origin_placeholders(&millstone_template, &origin_data);
+
+    // Per-target $GROUP values. For socket transports the substitution value is the entry key
+    // (the airlock HostPath directory name); for TCP/gRPC the value is the corresponding agent
+    // pod's cluster IP, since pod hostnames are not routable from a separate pod.
+    let group_values: std::collections::HashMap<String, String> = if use_socket_wait {
```
webern: Use imports
```suggestion
    let group_values: HashMap<String, String> = if use_socket_wait {
```

bin/correctness/panoramic/src/correctness/k8s.rs
```diff
@@ -233,32 +241,32 @@ pub async fn run_k8s_correctness_test(name: String, config: Config, tctx: TestCo
             Ok(ips) => ips,
             Err(e) => return make_error_result(name, started, "get_pod_ips", cleanup(e).await),
         };
-        (baseline_pod_ip, comparison_pod_ip)
+        [
+            ("baseline".to_string(), baseline_pod_ip),
```
webern, line 247:
```suggestion
        [
            // TODO: find a more dynamic mechanism for this.
            ("baseline".to_string(), baseline_pod_ip),
```

bin/correctness/panoramic/src/correctness/k8s.rs
```diff
@@ -216,13 +219,18 @@ pub async fn run_k8s_correctness_test(name: String, config: Config, tctx: TestCo
...
-    let (baseline_group_value, comparison_group_value) = if is_socket_target(&config_content) {
-        ("baseline".to_string(), "comparison".to_string())
+    let config_with_origin = substitute_origin_placeholders(&millstone_template, &origin_data);
+
+    // Per-target $GROUP values. For socket transports the substitution value is the entry key
+    // (the airlock HostPath directory name); for TCP/gRPC the value is the corresponding agent
+    // pod's cluster IP, since pod hostnames are not routable from a separate pod.
+    let group_values: std::collections::HashMap<String, String> = if use_socket_wait {
+        [
+            ("baseline".to_string(), "baseline".to_string()),
```
webern, line 230:
```suggestion
            // TODO: find a more dynamic mechanism for this.
            ("baseline".to_string(), "baseline".to_string()),
```

bin/correctness/panoramic/src/correctness/config.rs
```diff
@@ -284,3 +284,216 @@ impl Config {
...
+/// differently for each target. This helper parses the template as YAML, expects a top-level
+/// `targets:` mapping, and for each entry replaces `$GROUP` in that entry's address using the
+/// caller-supplied lookup (typically `key -> "baseline" | "comparison"` for Docker, or
+/// `key -> pod cluster IP` for k8s TCP/gRPC). Other parts of the document are left untouched.
+///
+/// The lookup closure must return `Some` for every target key present in the template; missing
+/// keys are reported as an error rather than silently leaving `$GROUP` unresolved.
+///
+/// Returns the rewritten YAML as a `String` ready to be written to the millstone container.
+pub fn resolve_group_placeholders<F>(template: &str, mut lookup: F) -> Result<String, GenericError>
```
webern, line 301:
```suggestion
/// Returns the rewritten YAML as a `String` ready to be written to the millstone container.
pub fn resolve_group_placeholders<F>(template: &str, mut lookup: F) -> Result<String, GenericError>
// TODO: let's design a better way to configure our containers than this
```

bin/correctness/millstone/src/driver.rs
```diff
@@ -50,8 +50,9 @@ impl Driver {
         let mut borrowed_payloads = payloads.iter().map(|b| &b[..]).cycle();
 
         let mut payloads_sent = 0;
+        // Aggregate across all fan-out targets. `TargetSender::send` returns the sum of bytes
+        // written to every backend for this payload.
         let mut payload_bytes_sent = 0;
-        let mut partial_sends = 0;
```
webern, line 54: Why did we remove the accounting of `partial_sends`? We should probably wire this up. Realizing it needs to be done separately for each `TargetSender`, there must be some way to do the accounting for each separately.

bin/correctness/millstone/src/corpus.rs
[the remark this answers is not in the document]
webern: It's a little wonky, but we're using this for correctness and as far as I know our use case is to always use the same framing mode for `baseline` and `correctness`.

bin/correctness/millstone/src/config.rs
```diff
@@ -137,8 +138,14 @@ pub struct Config {
...
-    /// The target to send payloads to.
-    pub target: TargetAddress,
+    /// The targets to send payloads to.
+    ///
+    /// Each generated payload is sent to every target in turn (fan-out), in declared order, before
+    /// the next payload is generated. Keys are free-form target names (used in logs and errors);
+    /// values are address strings parsed the same way as a single target. At least one target is
+    /// required.
+    // TODO: the names of the target can only be: baseline and/or comparison
+    pub targets: IndexMap<String, TargetAddress>,
```
tobz, line 148: Any reason why not to just make it `baseline_target` and `comparison_target` in this PR? Or, alternatively, create a dead simple `TargetConfiguration` struct with `baseline: TargetAddress`, etc.

tobz: Or, maybe even simpler: just make it use the same target address definition for both baseline and comparison automatically?

The updated correctness test configurations all have identical target addresses, so it doesn't seem like we actually need to differentiate at all? Like just do it internally, implicitly.

tobz (commented):
Left a comment about trying to simplify things so that we're not adding more boilerplate.

webern:
Abandoned. Sorry!

PR #1732, opened 2026-05-25
feat(config): per-pipeline incompatibility gate

### LINKED ISSUE #1606: Do not warn or error on unrelated incompatible configs. DADP-68
I am solving #1549 for the naive case where ADP only cares about DogStatsD (since this is the first thing we are shipping to users). However, ADP can, and will, do many other things, and we should not warn or error on configs for a system that is not enabled. For example, if `dogstatsd_blah` is deemed an incompatible config for ADP, we should not warn or error on it unless DogStatsD is enabled in ADP.

This may already be modeled sufficiently by the config structs in which configs are found, but I suspect it's going to be a bit more complicated and ambiguous than that, so I want to handle it after #1549 is done.

Edit: Actually, by definition, the offending keys will *not* be in the structs that map them to subsystems, so something definitely needs to change about the data model.

[DADP-68](https://datadoghq.atlassian.net/browse/DADP-68)

webern:
## Summary

Implements a data enrichment in the Config Registry and uses it in agent-data-plane such that we
will not warn or exit on a configuration key that is not relevant to ADP's configured state. More
specifically, if a particular pipeline is not enabled, then an unsupported configuration key
affecting only that pipeline will not emit a warning or error.

#### Data model:

- `Pipeline` enum: `DogStatsD`, `Checks`, `Otlp`, `Traces`.
- `PipelineAffinity` enum: either a list of affected pipelines of `CrossCutting`

## Change Type

- [x] New feature

## How did you test this PR?

- unit tests locally and in CI
- New integration test `adp-config-check-no-warn`: mirrors `adp-config-check-warn`, sets
  `DD_DATA_PLANE_DOGSTATSD_ENABLED=false`, and asserts
  `log_not_contains "Unsupported configuration key"` for
  `DD_DOGSTATSD_TELEMETRY_ENABLED_LISTENER_ID`. Confirms end-to-end warning suppression. (Checking
  in CI)

## References

- Closes: #1606


bin/agent-data-plane/src/cli/run.rs
```diff
@@ -151,7 +153,13 @@ pub async fn handle_run_command(
...
     }
 
-    check_and_warn_config(&config).error_context("Incompatible configuration detected.")?;
+    let classifier = ConfigClassifier::new();
+    let active = active_pipelines(&dp_config);
+    let flattened = config
+        .flattened_keys()
+        .error_context("Unable to flatten configuration into a list of dot-separated keys.")?;
+    let active_keys = filter_by_active_pipelines(flattened, &classifier, &active);
+    check_and_warn_config(&classifier, &active_keys).error_context("Incompatible configuration detected.")?;
```
webern: I don't like that we are using the classifier twice here. My intent was that we would do it more like this

Pseudocode:

```
check_and_warn_config(&config).error_context("Incompatible configuration detected.")?;


inside check_and_warn_config...

for (key, val) in flattened {
    let classification = classifier.classify(key, val);
    if !relevant(classification, enabled_pipelines) {
        continuel
    }
}

```

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -305,9 +313,54 @@ pub async fn handle_run_command(
...
+            match c.support_level {
+                SupportLevel::Incompatible(_) => match c.pipeline_affinity {
+                    PipelineAffinity::CrossCutting => true,
+                    PipelineAffinity::Pipelines(ps) => ps.iter().any(|p| active.contains(p)),
+                },
+                _ => true,
+            }
+        })
+        .collect()
+}
```
webern: No, we are using the classifier twice. That's dumb.

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -318,17 +371,12 @@ pub async fn handle_run_command(
...
-    for (key, val) in config
-        .flattened_keys()
-        .error_context("Unable to flatten configuration into a list of dot-separated keys.")?
-    {
+    for (key, val) in keys {
         // Get the classification. The classifier returns None if the config key is invalid or not-applicable to ADP.
-        let Some(classification) = config_classifier.classify(&key, &val) else {
+        let Some(classification) = classifier.classify(key, val) else {
             continue;
         };
```
webern, line 370: Here we should check if the classification is "relevant" using the list of configured pipelines.

lib/saluki-components/src/config_registry/datadog/containerd.rs
```diff
@@ -11,6 +11,7 @@ crate::declare_annotations! {
         used_by: &[structs::CONTAINERD_CONFIGURATION],
         value_type_override: None,
         test_json: None,
+        pipeline_affinity: PipelineAffinity::CrossCutting,
```
webern, line 15: Almost everything is labeled as CrossCutting still. These need to be reevaluated and assigned properly.

lib/saluki-components/src/config_registry/datadog/mod.rs
```diff
@@ -79,6 +79,22 @@ mod registry_tests {
...
+    #[test]
+    fn pipelines_affinity_slice_is_non_empty() {
+        for annotation in ALL_ANNOTATIONS.iter() {
+            if let PipelineAffinity::Pipelines(ps) = annotation.pipeline_affinity {
+                assert!(
+                    !ps.is_empty(),
+                    "annotation '{}' has PipelineAffinity::Pipelines(&[]) - \
+                     use PipelineAffinity::CrossCutting for keys with no pipeline restriction, \
+                     or list at least one Pipeline. \
+                     See config_registry/mod.rs module docs for details.",
```
webern: Improve wording.

lib/saluki-components/src/config_registry/mod.rs
```diff
@@ -141,6 +141,52 @@ pub mod structs {
     pub const GET_TYPED: &str = "get_typed";
 }
 
+/// Logical ADP data pipeline a config key belongs to.
```
webern:
```suggestion
/// The ADP data pipeline a config key affects.
```

lib/saluki-components/src/config_registry/mod.rs
```diff
@@ -141,6 +141,52 @@ pub mod structs {
...
+///
+/// Only primary user-controlled toggles and Traces (a sub-signal of OTLP) are represented.
+/// Derived forwarding baselines (Metrics, Logs, Events, ServiceChecks) are omitted: they have no
+/// distinct config keys and are fully implied by the primary toggles.
```
webern: Rewrite

lib/saluki-components/src/config_registry/mod.rs
```diff
@@ -141,6 +141,52 @@ pub mod structs {
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum Pipeline {
+    /// DogStatsD metrics pipeline (source + aggregator + forwarder).
+    DogStatsD,
+    /// Agent checks runner IPC source.
```
webern: Double check accuracy. `IPC`?

lib/saluki-components/src/config_registry/mod.rs
```diff
@@ -141,6 +141,52 @@ pub mod structs {
...
+    Checks,
+    /// OTLP ingestion frontend - active whenever `otlp().enabled()` is true, including proxy mode.
```
webern: What is proxy mode? Accurate?

lib/saluki-components/src/config_registry/mod.rs
```diff
@@ -141,6 +141,52 @@ pub mod structs {
...
+/// the invariant test.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PipelineAffinity {
+    /// Warn only when at least one of the listed pipelines is active.
+    ///
+    /// The slice must be non-empty.
+    Pipelines(&'static [Pipeline]),
+    /// Warn regardless of which pipelines are active.
+    ///
+    /// Use for keys that genuinely affect ADP broadly: forwarder, TLS, logging. Add a comment on
```
webern: fowader? Is that universally used?

"logging" is ambiguous since there may one day be a logging pipeline. Say "ADP log level"

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -305,33 +307,62 @@ pub async fn handle_run_command(
...
+    if dp_config.traces_pipeline_required() {
+        s.insert(Pipeline::Traces);
+    }
+    s
+}
+
 /// Check the resolved configuration against the config registry for incompatibilities.
 ///
-/// Classifies each flattened key in `config` with the config registry `Classifier`. Returns an
+/// Classifies each flattened key in `keys` with the config registry `Classifier`. Returns an
```
webern:
```suggestion
/// Classifies each flattened key in `config` with the config registry `Classifier`. Returns an
```

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -380,6 +411,23 @@ fn check_and_warn_config(config: &GenericConfiguration) -> Result<(), GenericErr
     Ok(())
 }
 
+/// Returns true if at least one of the `active_pipelines` is affected based on `pipeline_affinity`.
```
tobz:
```suggestion
/// Returns `true` if at least one of the `active_pipelines` is affected based on `pipeline_affinity`.
```

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -305,33 +307,62 @@ pub async fn handle_run_command(
...
         .flattened_keys()
         .error_context("Unable to flatten configuration into a list of dot-separated keys.")?
     {
         // Get the classification. The classifier returns None if the config key is invalid or not-applicable to ADP.
-        let Some(classification) = config_classifier.classify(&key, &val) else {
+        let Some(classification) = classifier.classify(&key, &val) else {
             continue;
         };
 
+        // Ignore it if none of the affected_pipelines are active.
```
tobz:
```suggestion
// Ignore it if none of the affected pipelines are active.
```

tobz (approved):
Overall, mostly just a few nits but this all seems reasonable to me.

One "take it or leave it" thing: it might be worth it to add helper methods to `PipelineAffinity` for common groups, like `PipelineAffinity::metrics()` gets you all pipelines that generate/emit metrics, and that sort of thing. This could at least help reduce the verbosity of some of the boilerplate, and make it easier to keep things synchronized given how many files these annotations are spread across.

webern:
> Overall, mostly just a few nits but this all seems reasonable to me.
> 
> One "take it or leave it" thing: it might be worth it to add helper methods to `PipelineAffinity` for common groups, like `PipelineAffinity::metrics()` gets you all pipelines that generate/emit metrics, and that sort of thing. This could at least help reduce the verbosity of some of the boilerplate, and make it easier to keep things synchronized given how many files these annotations are spread across.

I decided not to right now given that some of our intended follow up work will make this into even more of a data-only concept (i.e. the overlay file and code gen). Also it's a little bit nuanced as to when traces should be in the list with checks and dogstatsd and when it shouldn't.

PR #1734, opened 2026-05-26
chore(agent-data-plane): drop dedicated internal telemetry endpoint and expose on unprivileged API

tobz:
## Summary

This PR moves our internal telemetry endpoint to be exposed from the unprivileged API instead of requiring a dedicated telemetry endpoint.

As part of originally exposing the internal telemetry from ADP, we did so by creating a dedicated "internal observability" topology, which culminated with a Prometheus destination that exposed a scrape endpoint. This necessitated listening on a dedicated port, however, because the destination could not reasonably tie into the existing APIs in terms of registering routes. It was the simplest approach at the time and worked well.

As we push more and more, however, we want to consolidate and unify things where possible, and we now have many additional primitives that get us pretty close to what we need to replace our existing set-up:

- piecemeal rendering of Prometheus exposition payloads (`prometheus-exposition` crate)
- aggregated metrics storage (`Reflector`, `AggregatedMetricsState`)
- registering routes dynamically on particular API endpoints (`DynamicAPIBuilder`)

This PR is the culmination of replacing the original approach with the latest and greatest, which has let us remove the "internal observability" topology entirely, along with the dedicated listener just for internal telemetry.

Number of headline changes:

- internal telemetry is always on/exposed: no more configuration setting for it
- the "internal observability" topology has been removed
- the existing telemetry _routes_ (`/metrics` and `/compat/metrics`) are now exposed on the unprivileged API endpoint
- we added a new worker for internal telemetry which uses the existing reflected internal telemetry state
- we moved most of the metrics remapping code from `bin/agent-data-plane` to `saluki-core` to allow it to be reused
- we generalized the processing logic -- take remapper rules, iterate the reflector state, render Prometheus output -- by creating a new helper, `TelemetryProcessor` (don't love the name, might still change it)
- we added support for histograms to the aggregated metrics processor which previously only handled counters and gauges

The overall idea is that we now have a _single_ store for all of the internal telemetry, and we create individual telemetry processors, each with whatever the relevant set of remapper rules is (RAR, compat telemetry, or pure internal telemetry), in order to generate the necessary output.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Existing unit, integration, and correctness tests.

## References

DADP-2


lib/saluki-core/src/observability/metrics/aggregated.rs
```diff
@@ -0,0 +1,468 @@
...
+                let mut merged = a.clone();
+                merged.merge(&b);
+                Self {
+                    timestamp: None,
+                    value: AggregatedMetricValue::Histogram(merged),
+                }
+            }
+            (_, other) => Self {
+                timestamp: None,
+                value: other,
```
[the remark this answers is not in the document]
tobz: I'll give points for technically being correct.

Practically, internal telemetry _can't_ flip from one type to another during runtime, so this wouldn't ever happen. However, it's easy enough to fix to make it _impossible_, so that's what I did in 2603b76a11.

lib/saluki-core/src/observability/metrics/aggregated.rs
```diff
@@ -0,0 +1,468 @@
...
+        })
+        .await
+        .clone()
+}
+
+/// Merges two aggregated metric values.
+///
+/// Counters are summed, gauges use last-write-wins (the incoming value is kept), histograms are merged per-bucket and
+/// by sum/count.
+pub fn merge_aggregated_values(
```
lucastemb: It doesn't seem like this function is being called anywhere.`pub fn merge()` should handle it.

lib/saluki-core/src/observability/metrics/histogram.rs
```diff
@@ -0,0 +1,127 @@
+//! Fixed-bucket histogram aggregation for Prometheus exposition.
+
+use std::sync::LazyLock;
+
+use crate::data_model::event::metric::Histogram;
+
+const TIME_HISTOGRAM_BUCKET_COUNT: usize = 30;
```
lucastemb: Nit: Is there any reason that these are not consolidated into one general constant `HISTOGRAM_BUCKET_COUNT` or something?

lib/saluki-core/src/observability/metrics/aggregated.rs
```diff
@@ -0,0 +1,468 @@
...
+    /// Searches the state for all counter metrics with a matching name and returns their aggregated value.
+    ///
+    /// This method allows rolling up counter metrics that share a common name. For example, a metric may be emitted
+    /// with the same name, but N different values for a specific tag, potentially as a way to break down the metric by
+    /// a certain facet. This method can allow re-aggregating the value of each different facet value into a single
+    /// value.
+    ///
+    /// If multiple metrics are found with a matching name, but they're not all counter metrics, or if no metrics are
+    /// found with a matching name, `0.0` is returned.
+    pub fn get_aggregated(&self, name: &str) -> f64 {
```
lucastemb: Doesn't seem like this wrapper is being used anywhere

lib/saluki-core/src/observability/metrics/aggregated.rs
```diff
@@ -0,0 +1,468 @@
...
+
+    /// Searches the state for a single metric with a matching name.
+    ///
+    /// The latest value for the metric is returned. Histograms are skipped.
+    ///
+    /// If no metric is found with a matching name, or if multiple metrics are found with a matching name, `None` is
+    /// returned.
+    pub fn find_single(&self, name: &str) -> Option<f64> {
+        self.find_single_with_tags(name, &[])
+    }
```
lucastemb: Same as `get_aggregated`, doesn't seem like this wrapper is being used anywhere

lucastemb (changes requested):
Overall the PR solid. Main problems were just a a duplicate function and some unused wrappers. Not sure if that's something that will be used later down the line, and I just don't have the foresight atm. 

An additional suggestion that you could take or leave as well is maybe making some of the functions a little more restrictive in terms of accessibility `pub -> pub(crate)`, but not sure how strict we want to be on that in ADP.

lib/saluki-core/src/observability/metrics/histogram.rs
tobz: Hmm, yeah, no good reason this can't be a single constant.

lib/saluki-core/src/observability/metrics/aggregated.rs
tobz: Removed in d0020af228.

lib/saluki-core/src/observability/metrics/aggregated.rs
tobz: Removed in d0020af228.

lib/saluki-core/src/observability/metrics/histogram.rs
tobz: Updated in d0020af228.

lib/saluki-core/src/observability/metrics/aggregated.rs
tobz: Yep, good catch.

Removed in d0020af228.

lucastemb (approved):
Looks good :shipit:

bin/agent-data-plane/src/config.rs
```diff
@@ -45,10 +43,6 @@ impl DataPlaneConfiguration {
             secure_api_listen_address: config
                 .try_get_typed("data_plane.secure_api_listen_address")?
                 .unwrap_or_else(|| ListenAddress::any_tcp(5101)),
-            telemetry_enabled: config.try_get_typed("data_plane.telemetry_enabled")?.unwrap_or(true),
-            telemetry_listen_addr: config
-                .try_get_typed("data_plane.telemetry_listen_addr")?
-                .unwrap_or_else(|| ListenAddress::any_tcp(5102)),
             checks: DataPlaneChecksConfiguration::from_configuration(config)?,
```
jszwedko, line 46: Just a reminder that we'll need to remove these from the Agent too, now, after https://github.com/DataDog/datadog-agent/pull/51390 is merged.

Issue #1718, opened 2026-05-21
Track macOS platform support gaps in ADP


thieman:
## Context

ADP targets Linux as its first-class platform. macOS builds compile and run, but several features degrade silently or are disabled entirely. This issue catalogs the known gaps with enough detail to act on them.

Related issues: #580 (CI build/test), #625 (flaky `process-memory` test on macOS).

---

## What already works on macOS

- **`PlatformSettings`** (`lib/datadog-agent-commons/src/platform/macos_impl.rs`) exists: default config dir (`/opt/datadog-agent/etc`), log dir, and syslog URI (`unixgram:///var/run/syslog`) are all defined.
- **`process-memory`** has a `darwin` module using Mach task APIs — compiles and works.
- **UDS (Unix domain sockets)** compile and run via the `#[cfg(unix)]` path.
- **Syslog** works via the `#[cfg(unix)]` path.
- **Containerd collector** is not Linux-gated; it populates `ContainerPid → Container` alias entries in the tag store on macOS when containerd is running.

---

## Gap 1: UDS origin detection — a platform limitation, not just a missing implementation

**Files:** `lib/saluki-io/src/net/unix/non_linux.rs`, `lib/saluki-io/src/net/addr.rs`, `lib/saluki-components/src/sources/dogstatsd/mod.rs`

### The platform constraint

On Linux, ADP uses `SO_PASSCRED` + `SCM_CREDENTIALS` ancillary data: each `recvmsg` delivers the sender's `{pid, uid, gid}` as out-of-band data in the `ucred` struct. The PID is what the origin detection pipeline needs — it's what gets set via `origin.set_process_id()` and is used to resolve a container ID.

On macOS, the analogous APIs are `getsockopt(LOCAL_PEERCRED)` (returns `xucred`) and `getpeereid()`. Both give uid and gid only — **neither provides a PID**. This is a macOS kernel design decision with no standard workaround. PID-based UDS origin detection is not possible on macOS.

### What the code does today

- `enable_uds_socket_credentials` is a no-op (returns `Ok(())` without setting any socket option).
- `uds_recvmsg` in `non_linux.rs` calls `recvmsg` via `socket2` but passes no ancillary data buffer, returning `ProcessCredentialsError::UnsupportedPlatform` in the `ConnectionAddress`.

### The real symptom: misleading error telemetry

[...]

PR #1735, opened 2026-05-26
feat(panoramic): add native macOS runtime for integration tests

thieman:
## Human Summary

Adds support for running integration tests (*not* correctness tests) under Mac native. All but 3 of the existing integration tests are updated to support Mac. In CI, this runs under Tart VMs for arm64 and on native runners for amd64. To support this, we have to do a bunch of gymnastics so that our Agent under test doesn't conflict with the system Agent on the native runners.

Runtime is selected via a new `--runtime` flag to panoramic which defaults to Mac native tests when running on a Mac.

## Summary

Adds the ability to run ADP integration tests as **native macOS processes** (no Docker, no virtualization), and wires the suite into CI on the existing `macos:sonoma-arm64` / `macos:sonoma-amd64` runners.

**24 of 27 integration tests pass on `native_macos`** in ~3 minutes. Both standalone (just-ADP) and converged (Agent + ADP) tests are supported. Both CI jobs are passing.

This is the bare-metal subset of #1721; a Tart-based local-dev wrapper will land separately in a follow-up.

## Why

ADP has no test coverage on macOS today. The 27-test integration suite all runs inside a Linux Docker container. Adding native macOS support is the first concrete step toward macOS as a supported platform (tracked in #1718): it builds ADP for macOS, exercises real macOS startup behavior, and gives us a place to land further coverage as the platform gaps from #1718 are closed.

## Architecture

```
┌─────────────────────────────────────────────────────────────┐
│ Layer 2: macOS environment (CI bare-metal runner)           │
│   `make test-integration-macos-ci` ← CI entry point         │
└─────────────────┬───────────────────────────────────────────┘
                  │
                  ▼
┌─────────────────────────────────────────────────────────────┐
│ Layer 1: new Rust code (panoramic + airlock)                │
│   airlock::native::NativeProcess                            │
│   panoramic::native_runner::NativeIntegrationRunner         │
│   runtimes: [docker, native_macos] in test configs          │
│   requires_core_agent: true  for converged tests            │
└─────────────────────────────────────────────────────────────┘
```

The native runner is a parallel implementation of the Docker `IntegrationRunner`. The genuinely-duplicated assertion-loop logic is factored into a shared `run_assertion_steps` helper; the spawn / log capture / cleanup architectures stay separate because they're legitimately different (bollard vs `tokio::process`).

The Makefile intentionally does **not** expose a "run on your own Mac" convenience target. A Datadog Agent running on a dev host will collide with the test ADP on port 8125 (DSD UDP); the right answer for local-dev isolation is the Tart wrapper in the follow-up PR. Until then, CI is the only supported path.

### Rust framework

- **`airlock::native::NativeProcess`** (new, ~240 lines) — `tokio::process` wrapper with kill-on-drop, always places spawned processes in their own process group so cleanup can signal the entire group (parent plus any forked helpers like `trace-agent` / `process-agent`). Background watcher observes the child's exit, fires a `CancellationToken`, and records the exit code in a shared `OnceLock` cell.
- **`panoramic::native_runner::NativeIntegrationRunner`** (new, ~365 lines) — parallel to the Docker integration runner. Creates a per-test temp config directory, optionally spawns the Datadog Core Agent, waits for the IPC handshake to be ready, then spawns ADP. Both processes feed into the same `LogBuffer` so existing log-based assertions work unchanged.
- **`run_assertion_steps`** (new helper in `assertions/mod.rs`) — extracted from the Docker `IntegrationRunner`'s implementation (the more complete one: fail-fast on first failure, step-index debug logging, consistent error reporting). Both runners delegate to it.
- **`AssertionContext`** gains an `is_native` flag and an `Option<ExitCodeCell>` so:
  - `file_contains` reads files via `tokio::fs` on native (vs. `docker exec cat` on Docker)
  - `adp_exits_with` (new assertion) abstracts the runtime difference for "did ADP exit with code N":
    - On `docker`: greps the captured log buffer for `agent-data-plane exited with code N` (the s6 supervisor's exit message)
    - On `native_macos`: reads the exit code from the per-process `OnceLock`
- **`log_contains` / `log_not_contains` fixed** to do a final post-exit buffer read instead of bailing immediately on container exit. Required for tests where ADP exits within seconds (e.g., `adp-disabled-exit`).

### Config schema

- New `runtimes: Vec<String>` field on `IntegrationConfig`. Defaults to `["docker"]`. Configs that declare multiple runtimes expand into one `Test` instance per runtime at discovery time, named `{base_name}/{runtime}`.
- New `requires_core_agent: bool` field. When true on the `native_macos` runtime, the native runner spawns the Core Agent alongside ADP. Informational on `docker` (s6 always runs both).
- New `panoramic run --runtime <name>` filter so `make` targets can scope to one runtime cleanly.

### Provisioning + CI

- **`make provision-macos-test-env`** — idempotently installs the Datadog Agent at `/opt/datadog-agent` (downloads the official DMG, runs `installer -pkg`, tolerates the postinstall non-zero exit since the binaries land before postinstall runs) and bootstraps the IPC cert + auth_token by running the Agent briefly. Pinned to Agent `7.78.0` via `MACOS_TEST_AGENT_VERSION` for reproducibility.
- **`make test-integration-macos-ci`** — CI entry point: builds binaries, runs `provision-macos-test-env`, runs the native suite.
- **`make build-adp-native`** — `cargo build --release --bin agent-data-plane` with the same `APP_*` env vars that `build-adp-base` uses, so the binary logs as `| DATAPLANE |` instead of `| UNKNOWN |`.
- **`.gitlab/e2e.yml`** — new `test-integration-macos-arm64` and `test-integration-macos-amd64` jobs (e2e stage) extending the existing bare-metal macOS runner tags. `before_script` does a defensive `sudo pkill` to clean up any orphan Agent processes from prior shared-runner jobs.

### Tests enabled

| Category | Tests | Status |
|---|---|---|
| Standalone (just ADP) | 14 of 17 | All passing |
| Converged (Agent + ADP) | 10 of 10 | All passing |
| Container-supervisor behavior | 0 of 1 | Deferred |
| Bind-host (PANORAMIC_DYNAMIC shell hooks) | 0 of 2 | Deferred |

**3 tests intentionally not enabled** (filtered out via the `runtimes` field):

- `adp-rar-disabled`: assertion is "container stable for 10s," which really tests "s6 keeps restarting ADP." No equivalent supervisor on native. Re-enable when the native runner grows restart-on-failure semantics or the test is rewritten to assert on retry behavior directly.
- `dogstatsd-bind-host`, `dogstatsd-bind-custom-hostname`: use `PANORAMIC_DYNAMIC_*` env shell hooks that run `hostname -i` and `echo ... >> /etc/hosts` — Linux-container-isms not portable to a macOS host (`hostname -i` isn't supported on macOS; writing `/etc/hosts` needs root and has real host side effects).

## Test plan

Verified locally on Apple Silicon (macOS 26.4.1), inside an ephemeral Tart macOS Sequoia VM with a freshly provisioned Datadog Agent (this matches the shape of a clean CI runner):

```
$ make test-integration-macos-ci
...
PASSED: 24 passed, 0 failed, 24 total (192.50s)
```

CI passes the same tests on both `macos:sonoma-arm64` and `macos:sonoma-amd64`.

**Regression check (Linux Docker path):**

- `panoramic list -d test/integration/cases` shows the original tests plus the 24 new `*/native_macos` variants.
- The `docker` variant of every previously-enabled test still discovers and dispatches to the existing `IntegrationRunner`.
- `cargo test -p airlock -p panoramic` passes.
- `make check-clippy`, `make check-fmt`, `make check-docs` clean (pre-commit enforced on every commit).

## Notable findings while building this

1. **ADP's bootstrap requires a `datadog.yaml`** to exist (defaulting to `/opt/datadog-agent/etc/datadog.yaml` on macOS). The native runner creates a per-test empty config so it works on hosts with or without an Agent install.
2. **The Agent's authoritative config (via the config stream) overrides ADP's env vars** because `ConfigurationLoader` follows "later sources win." If ADP set `DD_AUTH_TOKEN_FILE_PATH=<per-test>` but the Agent didn't, the Agent would advertise the platform default and ADP would honor the stream value for post-bootstrap IPC clients, failing TLS with `UnknownIssuer`. Fix: set `DD_AUTH_TOKEN_FILE_PATH` on the Agent too.
3. **Core Agent forks `trace-agent` and `process-agent`** that orphan onto launchd when the parent is SIGKILL'd. Native runner spawns the Agent with `process_group(0)` and signals the whole group on cleanup.
4. **The DMG installer's postinstall hook fails** with exit code 112, but the binaries land in `/opt/datadog-agent` before the postinstall runs. The provision target tolerates this non-zero exit.
5. **Bootstrap-via-sudo leaves files owned by root** mode 600. The provision target chowns to the current user only if the files aren't already readable, so sudo isn't re-prompted on already-set-up systems.

Related: #1718 (macOS platform gap tracking), #1721 (companion PR with optional Tart VM wrapper for local-dev isolation).


atanzu:
I have a concern about native runners clashing with the local setup. The thing with Docker is, it allowes us to create an isolated environment which 1/ Doesn't have conflicts with the host setup and 2/ Doesn't allow the code under test to do weird things.

For the latter, did you consider things like Launchd limits or Sandbox/Seatbelt?

For the former, technically we have `chroot` in MacOS (with certain limitations, but still).

.githooks/pre-commit
```diff
@@ -7,6 +7,7 @@ make check-fmt
 make check-clippy
 make check-licenses
 make check-deny
+make check-docs
```
thieman, line 10: Driveby but this should have been in here already

bin/correctness/airlock/src/unix.rs
```diff
@@ -0,0 +1,255 @@
...
+        #[cfg(unix)]
+        if let Some(pgid) = self.process_group {
+            // SAFETY: killpg with a valid pgid is a safe syscall; we ignore the return value.
+            unsafe {
+                libc::killpg(pgid, libc::SIGTERM);
+            }
+            tokio::time::sleep(Duration::from_millis(500)).await;
+            unsafe {
+                libc::killpg(pgid, libc::SIGKILL);
+            }
```
thieman, line 210: Highlighting these two `unsafe` blocks... seems fine to me

bin/correctness/panoramic/src/config.rs
```diff
@@ -111,14 +111,78 @@ pub struct IntegrationConfig {
...
 
+    /// Runtimes under which this test is eligible to run.
+    ///
+    /// Each value must be either `"docker"` (the default) or `"mac"`. The active
+    /// runtime for any given panoramic invocation is chosen at the CLI level (`--runtime`,
+    /// defaulting to the host's native runtime); a test discovers only when this list contains
+    /// that active runtime. Tests with multiple entries are portable across runtimes, but still
+    /// execute only once per invocation, in the active runtime.
+    #[serde(default = "default_integration_runtimes")]
+    pub runtimes: Vec<String>,
```
thieman, line 133: Highlighting this change, `panoramic` now takes in a `--runtime` flag which determines which set of tests get run. Defaults to native on Mac now.

bin/correctness/panoramic/src/config.rs
```diff
@@ -111,14 +111,78 @@ pub struct IntegrationConfig {
...
+    ///
+    /// When `true`, host-process runtimes (such as `mac`) spawn the Datadog Core Agent as a side process before
+    /// starting ADP, sharing a per-test config directory so they communicate over IPC the same
+    /// way they would in production. When `false` (the default), only ADP is spawned (standalone
+    /// mode).
+    ///
+    /// On the `docker` runtime this field is informational; the converged image always runs
+    /// both processes via s6.
+    #[serde(default)]
+    pub requires_core_agent: bool,
```
thieman: Considered getting rid of this and just always starting up the core agent but it added a couple of seconds per test. Can remove if you'd rather.

test/integration/cases/adp-cmd-port/config.yaml
```diff
@@ -4,26 +4,31 @@
...
+# non-default path end-to-end.
 
 type: integration
 name: "adp-cmd-port"
 description: "Verifies ADP connects to the correct port when cmd_port is set"
 timeout: 120s
+runtimes: [docker, mac]
+requires_core_agent: true
+
+env:
```
thieman, line 20: Hoisted `env` out of the `container` field since it applies to both containerized and native tests

.gitlab-ci.yml
```diff
@@ -14,6 +14,15 @@ stages:
...
 workflow:
+  # Auto-cancel any interruptible jobs from a previous pipeline when a new commit lands on the
+  # same ref. The default ('conservative') only cancels the pipeline if no non-interruptible
+  # job has started yet — since our Linux jobs are non-interruptible and start immediately,
+  # that mode effectively never cancels anything. 'interruptible' cancels just the jobs marked
+  # interruptible: true (currently the macOS unit + integration jobs, which run on scarce
+  # bare-metal runner capacity); non-interruptible jobs continue to completion as before.
+  # https://docs.gitlab.com/ci/yaml/#workflowauto_cancelon_new_commit
+  auto_cancel:
+    on_new_commit: interruptible
```
thieman, line 25: Made Mac jobs specifically interruptible and added this definition to the workflow so they'll get canceled when new commits get pushed. Our Mac capacity is constrained, we only have 3 concurrent AMD64 runners across the project.

bin/correctness/panoramic/src/test_env.rs
```diff
@@ -0,0 +1,63 @@
...
+///
+/// Naming convention: every default port that's 4 digits gets a `5` prepended (8125 -> 58125,
+/// 5001 -> 55001, etc.). The GUI is disabled outright since we don't exercise it.
+///
+/// Note on env-var nesting: saluki-config (and figment) split env-var names on `__` to map to
+/// nested config keys. Single-underscore env vars like `DD_DATA_PLANE_API_LISTEN_ADDRESS` map
+/// to the flat key `data_plane_api_listen_address` and are silently ignored; we use double
+/// underscores at every dot boundary for the deep ADP / OTLP keys below. The top-level Agent
+/// env vars (`DD_CMD_PORT` etc.) are explicitly queried by the Agent so they don't need it.
+pub fn port_isolation_env() -> HashMap<String, String> {
```
thieman, line 22: This thing provides a shared set of env vars which is intended to provide non-default values for anything like port assignments or filepaths which can conflict with another Agent running on the machine. This is necessary since the amd64 runners are non-virtualized and have their own Agents running on them. For consistency these variables are applied to all test runs, native or otherwise.

Obviously this is super annoying and jank, so I'm open to better ideas here.

thieman:
@atanzu this is currently doing the following:

1. arm64 jobs run on a shared runner pool where each job now goes into a Tart VM that is provisioned and torn down with each job. Should be no risk of contaminating the runner here.

2. amd64 jobs still run directly on native Mac runners. These are "dedicated" runners only used by Saluki, so if we do happen to screw them up somehow the blast radius is limited to our project.

Additionally, we are going out of our way to install the version of the Agent under test into a non-standard install directory and override any port assignments or filepaths under test so that we do not interfere with an Agent already running on the machine. This is applicable to the amd64 runners since they are running their own Agents for internal observability. That's probably not perfect (I'd imagine we may be polluting some logs?) but it's thorough enough that the tests can pass. I think this is probably the best we can do here without any virtualization support. Something like `chroot` might help on some axes (filepaths?) but not others (ports?). This was already a bear to get working so I'm inclined to leave it as-is since the blast radius is already limited.

bin/correctness/panoramic/src/assertions/file_contains.rs
```diff
@@ -95,7 +95,12 @@ impl Assertion for FileContainsAssertion {
                 };
             }
 
-            match read_file_in_container(&ctx.container_name, &self.path).await {
+            let read_result = if ctx.is_host_process {
+                read_file_local(&self.path).await
```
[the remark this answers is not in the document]
tobz: Fair thing to raise... I see we're sandboxing our logs and what not in their own directory, but are we cleaning that up before/after test runs?

(Maybe I just haven't gotten far enough through the PR yet.)

bin/correctness/panoramic/src/config.rs
tobz: I'd say we should always run the Core Agent. It's that sort of thing that gives me some extra confidence that our tests are reflecting real-world usage.

Makefile
```diff
@@ -569,6 +569,106 @@ list-integration-tests: build-panoramic
...
+
+# Version of the Datadog Agent installed by `provision-macos-test-env`. Pinned for
+# reproducibility; bump when the integration tests need newer Agent behavior.
+MACOS_TEST_AGENT_VERSION ?= 7.78.0
+MACOS_TEST_AGENT_DMG_DIR ?= /tmp/saluki-dda-dmg-cache
+MACOS_TEST_AGENT_DMG_URL ?= https://s3.amazonaws.com/dd-agent/datadog-agent-$(MACOS_TEST_AGENT_VERSION)-1.$(shell uname -m).dmg
+# Sandbox directory the Agent is installed into. Deliberately *not* /opt/datadog-agent: keeping
+# our install isolated from any pre-existing system install (which a CI runner or developer host
+# may already have at a different, conflicting version) avoids surprises in both directions.
+MACOS_TEST_AGENT_INSTALL_DIR ?= /tmp/saluki-dda/datadog-agent
```
tobz: These should live at the top with other variables so they're discoverable.

bin/correctness/panoramic/src/config.rs
thieman: Does that include if we're intentionally running in standalone mode, which a bunch of the integration tests are? I'm not sure whether having Core running will actually break those or not, I can check

bin/correctness/panoramic/src/assertions/file_contains.rs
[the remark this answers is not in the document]
thieman: [GPT-5] Addressed in https://github.com/DataDog/saluki/pull/1735/commits/afa587d7c57c0b1ef5d2cce6f13029c412d19541. The Unix runner now wraps the per-test `panoramic-unix-*` state directory in an RAII guard and removes only the framework-owned directory it creates. I intentionally did not clean paths from test YAML (for example explicit log-file assertion paths), since those are test-owned rather than runner-owned.

bin/correctness/panoramic/src/config.rs
thieman: [GPT-5] Addressed in https://github.com/DataDog/saluki/pull/1735/commits/a1f12f01919a41415cef6e459dc0680c79fe93dc. The mac host-process runner now always starts the Core Agent before ADP, matching the Docker/s6 fixture shape. ADP standalone vs converged behavior is controlled only by test config now. I also removed the now-vestigial `requires_core_agent` field and mirrored Docker cont-init listener collision avoidance for DogStatsD and OTLP.

Makefile
thieman: [GPT-5] Addressed in https://github.com/DataDog/saluki/pull/1735/commits/afa587d7c57c0b1ef5d2cce6f13029c412d19541. Moved the `MACOS_TEST_AGENT_*` settings up with the other top-level Makefile variables so they are easier to find.

tobz (approved):

PR #1739, opened 2026-05-27
fix(datadog): stabilize additional endpoint retry queue IDs

atanzu:
## Summary

Derive additional endpoint retry queue IDs from the raw configured URL plus key index instead of the normalized endpoint or API key value. This preserves retry queues across API key rotation and avoids collisions when two raw additional endpoint URLs normalize to the same Datadog endpoint.

Note: this change is not backwards compatible. After the update ADP won't be able to pick up queued transactions from the disk. 

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Add unit test to validate the naming mechanism.

## References

- AGTMETRICS-504
- #1540

### DIFF
--- a/lib/saluki-components/src/common/datadog/io.rs
+++ b/lib/saluki-components/src/common/datadog/io.rs
@@ -2,7 +2,6 @@
 //!
 //! # Missing
 //!
-//! - Account for dynamic API key updates when deriving endpoint queue IDs.
 //! - Avoid initializing the process-wide crypto provider from tests.
 
 use std::{collections::VecDeque, error::Error as _, path::PathBuf, sync::Arc, time::Duration};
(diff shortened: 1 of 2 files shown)


lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -513,18 +512,21 @@ async fn run_endpoint_io_loop<B>(
...
-
-    // We set our queue ID/name to be unique for the component/endpoint/API key combination, which ensures that two
-    // instances of the same destination cannot collide with each other if they're using the same endpoint/API key
-    // combination.
-    let hash = hash_single_stable((context.component_id(), endpoint.endpoint(), endpoint.cached_api_key()));
+    // For primary and OPW endpoints we omit the key entirely; a single primary endpoint per
+    // component instance is sufficient to guarantee uniqueness.
+    // Use the raw (pre-normalization) URL for additional endpoints rather than the normalized form.
+    // Different raw strings can normalize to the same host (e.g. "app.datadoghq.com" and
+    // "https://app.datadoghq.com"), so hashing over the normalized URL would cause collisions.
```
tobz: Can you drop these comment lines? We've already got them in the doc comment for `additional_endpoint_queue_key` itself.

tobz (commented):
I think the one test that's missing... I think? is that we're actually generating separate queue IDs for different keys within a single endpoint definition (the whole point of the API key index bit).

tobz (approved):

PR #1762, opened 2026-05-28
fix(tls): gracefully handle errors during loading of platform root certificates

tobz:
## Summary

This PR updates the platform root certificate loading behavior in `saluki-tls` to gracefully handle a number of scenarios that we would otherwise emit an error on.

Ultimately, our goal here is to roughly mirror the behavior of Go's crypto stack which strives to not return errors unless it failed to load any _valid_ certificates: loading no certificates at all isn't an error, but finding some to load and all of them being invalid _is_ an error.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Added a number of new integration tests that assert that we gracefully/silent handle the various error conditions: no certificates found at all, certificates found but all invalid, some invalid, none invalid, etc.

## References

DADP-2

### DIFF
--- a/lib/saluki-tls/src/lib.rs
+++ b/lib/saluki-tls/src/lib.rs
@@ -287,7 +287,9 @@ pub fn initialize_default_crypto_provider() -> Result<(), GenericError> {
 ///
 /// ## Errors
 ///
-/// If any error occurs during the locating or loading the platform's native certificate store, an error will be returned.
+/// If errors occur during certificate loading and no certificates were ultimately added to the store, an error is
+/// returned. Missing files or directories referenced by `SSL_CERT_FILE`/`SSL_CERT_DIR` are tolerated and treated as
+/// "no certificates available" rather than as a load failure.
 ///
 /// [c_rehash]: https://www.openssl.org/docs/manmaster/man1/c_rehash.html
 pub fn load_platform_root_certificates() -> Result<(), GenericError> {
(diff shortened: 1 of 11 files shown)


lib/saluki-tls/src/lib.rs
```diff
@@ -300,23 +302,44 @@ pub fn load_platform_root_certificates() -> Result<(), GenericError> {
         ));
     }
 
+    let root_cert_store = load_platform_root_certificates_inner()?;
+
+    // The reason it should be impossible is that we intentionally only set it _here_, and we do so after acquiring the
+    // mutex, and only then do we make sure that it hasn't been set before proceeding to try to set it.
+    DEFAULT_ROOT_CERT_STORE
```
webern, line 309: Isn't there an FnOnce pattern for this?

webern (approved):

lib/saluki-tls/src/lib.rs
tobz: Yeah, I could do this with `std::sync::Once`. I think, originally, I did it this way just to ensure we didn't _needlessly_ load the platform root certificates twice only to realize someone else beat us to the punch, but `std::sync::Once` can provide that invariant too... so I'll just switch to that.

tobz: Actually, not going to change it.

When trying this out locally, it involved a lot of contortions (and `OnceLock`, as well) to maintain the return signature and then also be able to clone the result in case of errors.

Going to leave it as-is for now.

lib/saluki-tls/tests/loads_valid_certificate.rs
```diff
@@ -0,0 +1,16 @@
+//! Integration test: a directory containing a valid PEM certificate produces a populated store.
+
+mod common;
+
+use tempfile::TempDir;
+
+#[test]
+fn loads_valid_certificate() {
+    let temp_dir = TempDir::new().expect("should create temp dir");
+    common::write_self_signed_cert(&temp_dir.path().join("ca.pem"));
```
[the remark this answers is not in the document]
tobz: This is incorrect. `SSL_CERT_DIR` can have any regular ol' `.crt` or `.pem` files in it... it's the `c_rehash` tool from OpenSSL itself that scans such a directory and creates _symlinks` to those certificate files in the form of `HHHHHHHH.D`, but `rustls-native-certs` isn't actually using that pattern to _find_ eligible certificates when `SSL_CERT_DIR` is specified.

PR #1771, opened 2026-05-29
feat(components): support `aggregator_tag_filter_cache_capacity` config key in tag filterlist

### LINKED ISSUE #1667: Support `aggregator_tag_filter_cache_capacity` for the tag filter context cache.
## Problem

The core agent exposes `aggregator_tag_filter_cache_capacity` (default 1000) to configure the size of the dedup cache used by the `metric_tag_filterlist` context resolver. ADP hardcodes this to 100,000 entries:

- **Core agent registration:** [`pkg/config/setup/common_settings.go:1588`](https://github.com/DataDog/datadog-agent/blob/main/pkg/config/setup/common_settings.go#L1588)
- **Core agent usage:** [`pkg/aggregator/context_resolver.go:110`](https://github.com/DataDog/datadog-agent/blob/main/pkg/aggregator/context_resolver.go#L110)
- **ADP hardcoded constant:** [`bin/agent-data-plane/src/components/tag_filterlist/mod.rs:40`](https://github.com/DataDog/saluki/blob/55bca143624da31542e49d0da9b33f15bccfe08b/bin/agent-data-plane/src/components/tag_filterlist/mod.rs#L40)

The 100x difference means operators cannot tune cache size for memory-constrained environments, and there is no parity with the core agent's default behavior.

## Fix

Read `aggregator_tag_filter_cache_capacity` from config (default 1000 to match the core agent) and use it to size the `CONTEXT_CACHE_CAPACITY` constant in `TagFilterlistConfiguration`. The config key should be read in `TagFilterlistConfiguration::from_configuration()` or the `TagFilterlistWorker` constructor.

jszwedko:
## Summary

Use the configured `aggregator_tag_filter_cache_capacity` rather than the hardcoded constant. https://github.com/DataDog/datadog-agent/pull/51522 configures the default to 100k when ADP is enabled (otherwise 1k for Core Agent) so there should be no behavioral change.

Closes #1667

### DIFF
--- a/bin/agent-data-plane/src/components/tag_filterlist/mod.rs
+++ b/bin/agent-data-plane/src/components/tag_filterlist/mod.rs
@@ -151,10 +151,10 @@ pub struct TagFilterlistConfiguration {
     #[serde(default, rename = "metric_tag_filterlist")]
     entries: Vec<MetricTagFilterEntry>,
 
-    /// Maximum number of entries in the per-context dedup cache used by the tag filter.
+    /// Maximum number of entries in the per-context deduplication cache used by the tag filter.
     ///
     /// Each cache entry tracks whether a given metric context has already had its tags filtered,
-    /// avoiding redundant work on repeated submissions of the same metric.
+    /// avoiding redundant work on repeated submissions of the same metric context.
     ///
     /// Defaults to 100,000. The core Agent defaults to 1,000 for its own aggregator; ADP uses a
     /// higher value to accommodate the greater metric throughput it handles. When `data_plane.enabled`
--- a/bin/agent-data-plane/src/components/tag_filterlist/mod.rs
+++ b/bin/agent-data-plane/src/components/tag_filterlist/mod.rs
@@ -37,9 +37,12 @@ use tracing::{debug, error, warn};
 
 use crate::components::dogstatsd_filterlist::METRIC_TAG_FILTERLIST_CONFIG_KEY;
 
-const CONTEXT_CACHE_CAPACITY: usize = 100_000;
 const CONTEXT_CACHE_TTI: Duration = Duration::from_secs(30);
 const CONTEXT_CACHE_EXPIRATION_INTERVAL: Duration = Duration::from_secs(1);
+
+fn default_context_cache_capacity() -> usize {
+    100_000
+}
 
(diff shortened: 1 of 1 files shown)


bin/agent-data-plane/src/components/tag_filterlist/mod.rs
```diff
@@ -148,6 +151,24 @@ pub struct TagFilterlistConfiguration {
...
 
+    /// Maximum number of entries in the per-context deduplication cache used by the tag filter.
+    ///
+    /// Each cache entry tracks whether a given metric context has already had its tags filtered,
+    /// avoiding redundant work on repeated submissions of the same metric context.
+    ///
+    /// Defaults to 100,000. The core Agent defaults to 1,000 for its own aggregator; ADP uses a
+    /// higher value to accommodate the greater metric throughput it handles. When `data_plane.enabled`
+    /// and `data_plane.dogstatsd.enabled` are both true, the Agent raises its own default to 100,000
+    /// via the config stream so that a single value in `datadog.yaml` controls both processes.
```
tobz: Can you move the `Defaults to 100,000` as the last (and separate) line in the doc comment?

I'm also perhaps a little dubious of mentioning the specifics here since, ultimately, the effective default is 100,000 either way.. but not a strong objection for that part specifically.

tobz (changes requested):
One veryyyy small thing.

bin/agent-data-plane/src/components/tag_filterlist/mod.rs
jszwedko: Fixed in 'docs(components): simplify doc comment on aggregator_tag_filter_cache_capacity' (https://github.com/DataDog/saluki/commit/ee5640ccba). Moved 'Defaults to 100,000.' to a standalone final line and stripped the agent/ADP implementation details.

jszwedko: Agreed, that is pretty wordy. I slimmed it down and moved the default to a separate line in [ee5640c](https://github.com/DataDog/saluki/pull/1771/commits/ee5640ccba7627cae45e1a736a2fcdc1e123c70a)

tobz (approved):

PR #1772, opened 2026-05-29
fix(aggregate): clamp window duration to minimum of one second

tobz:
## Summary

This PR fixes a bug where sub-second window durations would cause the Aggregate transform to panic during sample insertion.

Since we take the window duration as an actual `Duration` field, we allow for sub-second aggregation windows.. but we also use whole seconds when calculating which bucket to drop a sample into, which means that if our bucket window is sub-second, our whole second math breaks down quickly, leading to division by zero and ultimately panics.

In this PR, we've moved entirely to whole seconds for the window duration, _and_ backed that up with switching to `NonZeroU64`. This solves both the fractional second _and_ "less than one" problem in one fell swoop. 

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

## References

DADP-2

### DIFF
--- a/docs/agent-data-plane/configuration/dogstatsd.md
+++ b/docs/agent-data-plane/configuration/dogstatsd.md
@@ -273,7 +273,7 @@ The following settings are specific to ADP and have no equivalent in the core ag
 | `aggregate_flush_interval`                  | Aggregator flush period                    |         |
 | `aggregate_flush_open_windows`              | Flush open windows on stop                 |         |
 | `aggregate_passthrough_idle_flush_timeout`  | Passthrough buffer flush delay             |         |
-| `aggregate_window_duration_secs`            | Aggregation window size                    |         |
+| `aggregate_window_duration_seconds`            | Aggregation window size                    |         |
 | `connect_retry_attempts`                    | IPC client connect retries                 |         |
 | `connect_retry_backoff`                     | IPC client retry delay                     |         |
 | `counter_expiry_seconds`                    | Idle counter keep-alive duration           | 300     |
--- a/docs/agent-data-plane/configuration/dogstatsd.md
+++ b/docs/agent-data-plane/configuration/dogstatsd.md
@@ -273,7 +273,7 @@ The following settings are specific to ADP and have no equivalent in the core ag
 | `aggregate_flush_interval`                  | Aggregator flush period                    |         |
 | `aggregate_flush_open_windows`              | Flush open windows on stop                 |         |
 | `aggregate_passthrough_idle_flush_timeout`  | Passthrough buffer flush delay             |         |
-| `aggregate_window_duration`                 | Aggregation window size                    |         |
+| `aggregate_window_duration_secs`            | Aggregation window size                    |         |
 | `connect_retry_attempts`                    | IPC client connect retries                 |         |
 | `connect_retry_backoff`                     | IPC client retry delay                     |         |
 | `counter_expiry_seconds`                    | Idle counter keep-alive duration           | 300     |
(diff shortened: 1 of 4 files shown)


docs/agent-data-plane/configuration/dogstatsd.md
```diff
@@ -286,7 +273,7 @@ The following settings are specific to ADP and have no equivalent in the core ag
 | `aggregate_flush_interval`                  | Aggregator flush period                    |         |
 | `aggregate_flush_open_windows`              | Flush open windows on stop                 |         |
 | `aggregate_passthrough_idle_flush_timeout`  | Passthrough buffer flush delay             |         |
-| `aggregate_window_duration`                 | Aggregation window size                    |         |
+| `aggregate_window_duration_secs`            | Aggregation window size                    |         |
```
thieman: Nit: should this be `seconds` to align with `counter_expire_seconds` below or is this already a hodgepodge of different ways to do it?

tobz: Yeah, no, you're right... we should align on not condensing unit suffixes. Will fix.

lib/saluki-components/src/transforms/aggregate/mod.rs
```diff
@@ -36,8 +36,9 @@ use self::config::{HistogramConfiguration, HistogramStatistic};
 
 const PASSTHROUGH_IDLE_FLUSH_CHECK_INTERVAL: Duration = Duration::from_secs(2);
 
-const fn default_window_duration() -> Duration {
-    Duration::from_secs(10)
+const fn default_window_duration_secs() -> NonZeroU64 {
+    // SAFETY: It's clearly not zero.
+    unsafe { NonZeroU64::new_unchecked(1) }
```
thieman: Dumb question for my behalf, is the unsafe / fright-factor overhead worth not just using `new`? I'd imagine this isn't in a particularly hot path?

tobz: Nah, we could do `NonZeroU64::new(..).expect(...)` just as well.

thieman (approved):

lib/saluki-components/src/transforms/aggregate/mod.rs
tobz: Resolved in ce2ad3e214.

docs/agent-data-plane/configuration/dogstatsd.md
tobz: Fixed in c50b39645.

lib/saluki-components/src/transforms/aggregate/mod.rs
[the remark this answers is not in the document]
tobz: _Oof_. Good catch.

Fixed in 2bad6b8211.

docs/agent-data-plane/configuration/dogstatsd.md
```diff
@@ -286,7 +273,7 @@ The following settings are specific to ADP and have no equivalent in the core ag
 | `aggregate_flush_interval`                  | Aggregator flush period                    |         |
 | `aggregate_flush_open_windows`              | Flush open windows on stop                 |         |
 | `aggregate_passthrough_idle_flush_timeout`  | Passthrough buffer flush delay             |         |
-| `aggregate_window_duration`                 | Aggregation window size                    |         |
+| `aggregate_window_duration_seconds`            | Aggregation window size                    |         |
```
jszwedko, line 276: Nit, whitespace is misaligned now.

PR #1777, opened 2026-05-29
feat(datadog): support `forwarder_outdated_file_in_days` for stale retry file cleanup

### LINKED ISSUE #1360: Support `forwarder_outdated_file_in_days` to clean up stale retry queue files.
The core agent exposes [`forwarder_outdated_file_in_days`][1] (default 10). On startup, the forwarder [deletes `.retry` files][2] from the disk-backed transaction queue that exceed this age. This prevents unbounded disk growth from stale retry data after outages.

ADP has a disk-backed retry queue but manages cleanup by size and disk ratio only ([`forwarder_storage_max_size_in_bytes`][3], `forwarder_storage_max_disk_ratio`). There is no file-age-based cleanup. The disk queue is not enabled by default, so this should fast-follow disk queue enablement.

[1]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/pkg/config/setup/common_settings.go#L1534
[2]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/comp/forwarder/defaultforwarder/default_forwarder.go#L330
[3]: https://github.com/DataDog/saluki/blob/dbb968c2db21/lib/saluki-components/src/destinations/datadog/retry.rs

jszwedko:
## Summary

Adds support for `forwarder_outdated_file_in_days` (default 10), matching the core Agent's startup behavior. When disk persistence is enabled, ADP now scans `forwarder_storage_path` at startup (matching Agent behavior) and removes `retry-*.json` files whose filesystem mtime exceeds the configured age, preventing unbounded disk growth from stale retry data after extended outages. Set to `0` to disable cleanup. The config key is moved from the unsupported registry to the supported forwarder registry.

Closes #1360

## Test plan

- [x] `cargo check --workspace && cargo check --workspace --tests` passes
- [x] `cargo test -p saluki-components --lib outdated` passes (3 new tests)
- [x] `cargo test -p saluki-components --lib config_registry` passes
- [ ] With `forwarder_storage_max_size_in_bytes` set and `forwarder_outdated_file_in_days: 10`, old retry files are removed at startup; recent files and non-retry files are untouched
- [ ] With `forwarder_outdated_file_in_days: 0`, no files are removed


lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -392,6 +393,9 @@ async fn run_endpoint_io_loop<B>(
 
     // If the storage size is set, enable disk persistence for the retry queue.
     if config.retry().storage_max_size_bytes() > 0 {
+        // Remove stale retry files before opening the queue.
+        remove_outdated_retry_files(config.retry().storage_path(), config.retry().outdated_file_in_days()).await;
```
jszwedko: This matches the Agent's behavior of only checking at start-up.

lib/saluki-components/src/common/datadog/retry.rs
```diff
@@ -130,6 +134,20 @@ pub struct RetryConfiguration {
...
     storage_max_disk_ratio: f64,
+
+    /// Maximum age in days for retry files on disk before they are deleted at startup.
+    ///
+    /// When disk persistence is enabled, ADP removes any `retry-*.json` files in the
+    /// per-queue subdirectory of `forwarder_storage_path` that are older than this many days
+    /// each time it starts. This prevents unbounded disk growth from stale retry data left
+    /// behind after long outages.
+    ///
+    /// Setting this to `0` removes all retry files on startup. Defaults to 10.
```
jszwedko:
```suggestion
    /// Defaults to 10.
```

lib/saluki-io/src/net/util/retry/queue/mod.rs
```diff
@@ -434,6 +435,7 @@ mod tests {
                 u64::MAX,
                 1.0,
                 Arc::new(DiskUsageRetrieverImpl::new(root_path.clone())),
+                0,
```
jszwedko:
```suggestion
                10,
```

lib/saluki-io/src/net/util/retry/queue/persisted.rs
```diff
@@ -454,6 +459,61 @@ async fn create_directory_recursive(path: PathBuf) -> Result<(), GenericError> {
     .error_context("Failed to spawn directory creation blocking task.")?
 }
 
+/// Deletes files in `queue_path` whose filename-embedded creation timestamp is older than
+/// `max_age_days`. Does nothing if the directory does not exist.
+///
+/// Setting `max_age_days` to `0` deletes all retry files (cutoff = now), matching the behavior
+/// of the core Agent's `FileRemovalPolicy` with `outdatedFileDayCount = 0`.
```
jszwedko, line 500:
```suggestion
/// Deletes files in `queue_path` whose filename-embedded creation timestamp is older than
/// `max_age_days`. Does nothing if the directory does not exist.
```

lib/saluki-io/src/net/util/retry/queue/persisted.rs
```diff
@@ -518,6 +578,7 @@ mod tests {
             1024,
             0.8,
             DiskUsageRetrieverWrapper::new(Arc::new(DiskUsageRetrieverImpl::new(root_path.clone()))),
+            0,
```
jszwedko:
```suggestion
            10,
```

Let's use the default of 10 for the tests where we aren't explicitly testing the stale deletion behavior.

lib/saluki-components/src/common/datadog/retry.rs
jszwedko: Fixed in 'docs/test: apply review suggestions' (https://github.com/DataDog/saluki/commit/4588d13ff6).

lib/saluki-io/src/net/util/retry/queue/mod.rs
jszwedko: Fixed in 'docs/test: apply review suggestions' (https://github.com/DataDog/saluki/commit/4588d13ff6).

lib/saluki-io/src/net/util/retry/queue/persisted.rs
```diff
@@ -454,6 +459,58 @@ async fn create_directory_recursive(path: PathBuf) -> Result<(), GenericError> {
...
 }
 
+/// Deletes files in `queue_path` whose filename-embedded creation timestamp is older than
+/// `max_age_days`. Does nothing if the directory does not exist.
+async fn remove_outdated_retry_files(queue_path: &Path, max_age_days: u32) {
+    let mut dir = match tokio::fs::read_dir(queue_path).await {
+        Ok(d) => d,
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => return,
+        Err(e) => {
+            warn!(path = %queue_path.display(), error = %e, "Failed to open retry queue directory for age-based cleanup.");
```
webern: I don't love the idea of swallowing errors deep in a call stack. I think it's better to have that be explicitly done by callers.

lib/saluki-io/src/net/util/retry/queue/persisted.rs
```diff
@@ -111,13 +111,18 @@ where
...
         // Make sure the directory exists first.
         create_directory_recursive(root_path.clone())
             .await
             .with_error_context(|| format!("Failed to create retry directory '{}'.", root_path.display()))?;
 
+        // Remove stale retry files before loading state. This must run after directory creation
+        // but before refresh_entry_state so the queue doesn't load files that are about to be
+        // removed.
+        remove_outdated_retry_files(&root_path, max_age_days).await;
+
```
webern: Is this only supposed to happen on creation of the store? Or does it need to be called at some interval during its lifetime as well?

I think this function is doing more that what `from_root_path` implies (which would be construction only). It's also becoming a nasty function signature. At the leat, to function doc comment should be updated to mention cleanup of stale files.

Possibly a better signature would be

```rust
#[derive(Debug)]
pub struct PersistedQueueArgs {
    pub root_path: PathBuf,
    pub total_on_disk_bytes: u64,
    pub max_on_disk_bytes: u64,
    pub storage_max_disk_ratio: f64,
    pub max_age_days: u32,
}
```

```rust
pub async fn create(args: PersistedQueueArgs) -> Result<Self, GenericError> {
    // instantiate and create-the-dir-if-not-exists
}

pub async fn remove_stale_files(&self) -> Result {
///...
}
```

At instantiation:

```rust
let queue_args = PersistedQueueArgs{
   ///...
}

let persisted_queue = PersistedQueue::create(args).await.context("Unable to create the persisted retry queue").?;
if let Err(e) = persisted_queue.remove_stale_files().await {
    warn!(error = e, "Unable to clean up stale retry queue files");
}
```

Sorry it's pseudo code, but the idea is just to make error handling more visible in the calling code.

webern (approved):
Opinionated and debatable nits. Looks good.

lib/saluki-io/src/net/util/retry/queue/persisted.rs
jszwedko: 👍 good call, I do like making this behavior more visible to the caller (and it also lets us potentially call it under other conditions). The Core Agent only runs this at start-up so I'm matching that behavior, but there is the potential we may want to call it on some interval in the future.

jszwedko: Updated in [85ebb26](https://github.com/DataDog/saluki/pull/1777/commits/85ebb260ae907f4bb07524f052ba85198f4034a1)

PR #1790, opened 2026-06-01
feat(dev): add Windows cross-build support

thieman:
## pls look toby

With this PR it's now possible to cross-compile ADP-for-Windows on a Mac. Nothing is actually runnable but I wanted to get this sorted before I started working on Windows CI.

I scoped this to only saluki and agent-data-plane, testing tools like `correctness` are not in scope and probably still don't build on Windows. I'm not planning on running `correctness` but if any of those have problems I'll get to them as part of the CI PR.

## Summary
- Add Windows platform settings with Core Agent-compatible ConfigRoot registry lookup and ProgramData fallback paths.
- Add local macOS Windows cross-build setup/build targets using `cargo-xwin` and Homebrew LLVM.
- Fix Windows cfg/build blockers across networking, workload helpers, correctness tools, and ADP so the workspace cross-builds for `x86_64-pc-windows-msvc`.

## Notes
- This does not add Windows runtime support for ADP service management, named pipes, or Windows-specific DogStatsD ingestion.
- Unix socket address types remain available for config/type handling, but Unix socket listener construction returns an unsupported-platform error on non-Unix platforms.

## Test plan
- [x] `make fmt`
- [x] `make build-windows-cross`
- [x] `PATH="/opt/homebrew/opt/llvm/bin:$PATH" cargo xwin check --target x86_64-pc-windows-msvc --workspace --tests`
- [x] pre-commit checks run by `git commit`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,7 @@ export CARGO_TOOL_VERSION_dd-rust-license-tool ?= 1.0.3
 export CARGO_TOOL_VERSION_cargo-deny ?= 0.18.9
 export CARGO_TOOL_VERSION_cargo-hack ?= 0.6.30
 export CARGO_TOOL_VERSION_cargo-nextest ?= 0.9.99
+export CARGO_TOOL_VERSION_cargo-xwin ?= 0.22.0
 export CARGO_TOOL_VERSION_cargo-autoinherit ?= 0.1.5
 export CARGO_TOOL_VERSION_cargo-sort ?= 1.0.9
 export CARGO_TOOL_VERSION_dummyhttp ?= 1.1.0
@@ -51,6 +52,11 @@ export CARGO_TOOL_VERSION_cargo-machete ?= 0.9.1
 export CARGO_TOOL_VERSION_rustfilt ?= 0.2.1
 export DDPROF_VERSION ?= 0.20.0
 export LADING_VERSION ?= sha-d608ffbce8f8c77b147d6750b3bb6d6948af239a
+
+# Windows cross-compilation settings. These targets currently assume a local macOS host.
+export WINDOWS_CROSS_TARGET ?= x86_64-pc-windows-msvc
+export WINDOWS_CROSS_LLVM_BIN ?= /opt/homebrew/opt/llvm/bin
+export WINDOWS_CROSS_CARGO_ARGS ?= --workspace
 
 # Version of source repositories (Git tag) for vendored Protocol Buffers definitions.
 export PROTOBUF_SRC_REPO_DD_AGENT ?= 7.73.x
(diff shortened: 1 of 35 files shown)


Makefile
```diff
@@ -208,6 +214,28 @@ build-correctness-tools-image: ## Builds the correctness tools suite (datadog-in
...
+	@rustup target add $(WINDOWS_CROSS_TARGET)
+	@echo "[*] Ensuring cargo-xwin@$(CARGO_TOOL_VERSION_cargo-xwin) is installed..."
+	@test -f "$(CARGO_BIN_DIR)/cargo-xwin" || cargo install cargo-xwin --version "$(CARGO_TOOL_VERSION_cargo-xwin)" --locked
+	@echo "[*] Ensuring Homebrew LLVM is installed..."
+	@brew list llvm >/dev/null 2>&1 || brew install llvm
+	@test -x "$(WINDOWS_CROSS_LLVM_BIN)/llvm-lib" || \
+		(echo "Missing llvm-lib at $(WINDOWS_CROSS_LLVM_BIN)/llvm-lib. Set WINDOWS_CROSS_LLVM_BIN or install Homebrew LLVM." && exit 1)
+
+.PHONY: build-windows-cross
+build-windows-cross: install-windows-cross-tools ## Builds ADP for Windows from a local macOS host (override WINDOWS_CROSS_CARGO_ARGS as needed)
```
thieman, line 235: Adding this for now as a way to locally test Windows support. May remove this once we have Windows CI working

lib/datadog-agent-commons/src/platform/windows_impl.rs
```diff
@@ -0,0 +1,113 @@
...
+    ptr,
+    sync::OnceLock,
+};
+
+use windows_sys::Win32::{
+    Foundation::ERROR_SUCCESS,
+    System::Registry::{RegGetValueW, HKEY, HKEY_LOCAL_MACHINE, RRF_RT_REG_SZ, RRF_SUBKEY_WOW6464KEY},
+};
+
+// TODO: Add Windows-specific tests for registry-backed path resolution once we have Windows CI coverage.
```
thieman, line 8: This file is vibed from the core agent impl but untested as we don't actually have tests running yet. Added a TODO to come back and add tests.

lib/datadog-agent-commons/src/platform/windows_impl.rs
```diff
@@ -0,0 +1,113 @@
...
+    os::windows::ffi::{OsStrExt as _, OsStringExt as _},
+    path::{Path, PathBuf},
+    ptr,
+    sync::OnceLock,
+};
+
+use windows_sys::Win32::{
+    Foundation::ERROR_SUCCESS,
+    System::Registry::{RegGetValueW, HKEY, HKEY_LOCAL_MACHINE, RRF_RT_REG_SZ, RRF_SUBKEY_WOW6464KEY},
+};
```
tobz: This is wild when the [`windows-registry`](https://crates.io/crates/windows-registry) crate exists and abstracts this all.

C'mon bot. 😤

lib/saluki-components/src/sources/dogstatsd/framer.rs
```diff
@@ -34,9 +34,7 @@ pub fn get_framer(listen_address: &ListenAddress, eol_required: bool) -> DsdFram
     match listen_address {
         ListenAddress::Tcp(_) => DsdFramer::Stream(NestedFramer::new(newline_framer, LengthDelimitedFramer)),
         ListenAddress::Udp(_) => DsdFramer::NonStream(newline_framer),
-        #[cfg(unix)]
```
tobz, line 37: Why are we dropping these cfg gates?

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -234,6 +234,7 @@ impl hyper::rt::Write for HttpsCapableConnection {
 #[derive(Clone)]
 struct InnerConnector {
     http: HickoryHttpConnector,
+    #[cfg(unix)]
     connect_timeout: Duration,
```
tobz, line 238: I feel like I'm missing why connection timeouts are only desired/used when we're building for *nix platforms...

lib/saluki-components/src/sources/dogstatsd/framer.rs
thieman: I asked it to try to minimize the gating when it could, this change gets rid of around ~30 existing gates. The tradeoff is that Unixgram/Unix can be represented in Windows but would fail at bind time down here https://github.com/DataDog/saluki/pull/1790/changes#diff-3f1b5f842346a88f141b84d5300fa21b42804b11097d5403bb358284a9f9440aR201-R206

Can revert this if you'd rather. Or alternatively, maybe we add a platform gate at the parser so it still fails at config load time in Windows, but we don't need to gate everywhere?

lib/saluki-io/src/net/client/http/conn.rs
thieman: The InnerConnector only applies connect_timeout to UDS. In the TCP flow the connect timeout is applied in the http connectors, not in this InnerConnector. https://github.com/DataDog/saluki/blob/9af5deb088170802f1174915ae62daec15d999cc/lib/saluki-io/src/net/client/http/conn.rs#L430-L451

So we need to add the gate or else we get a dead code error on non-Unix since InnerConnector doesn't consume the field at all

tobz: That would _indeed_ explain it. 👍🏻

lib/saluki-components/src/sources/dogstatsd/framer.rs
tobz: Hmm, fair, I can appreciate the overall reduction here since we're still gating things, as you pointed out.

Thinking about it a little more, I think I'm fine with it as is... if it turns out to really bother me, well, then... I can always open a PR. 🤷🏻

tobz (approved):

PR #1791, opened 2026-06-01
feat(agent-data-plane): added multi-region failover for metrics

### LINKED ISSUE #1678: Investigate multi-region failover support for `multi_region_failover.*` config keys.
## Problem

The core agent supports multi-region failover (MRF) for metrics via six config keys:

- `multi_region_failover.enabled` — activates a secondary metrics-forwarding endpoint
- `multi_region_failover.failover_metrics` — routes the metrics datatype to the failover endpoint
- `multi_region_failover.metric_allowlist` — restricts MRF to a named subset of metrics
- `multi_region_failover.api_key` — API key for the failover region
- `multi_region_failover.dd_url` — intake URL for the failover region (overrides `multi_region_failover.site`)
- `multi_region_failover.site` — Datadog site for the failover region

These are wired through the Go domain forwarder (`comp/forwarder/defaultforwarder`) and the serializer. ADP does not have an equivalent MRF subsystem.

## What needs to happen

Determine whether ADP should:
1. Implement MRF forwarding natively (send metrics to a secondary endpoint when enabled), or
2. Rely on the core agent to handle MRF routing after ADP delivers metrics to it, or
3. Document these as not applicable with a clear explanation.

If ADP needs to implement MRF, identify what forwarder-side changes are required to add a secondary endpoint, the credential/URL configuration, and the metric allowlist filtering.

lucastemb:
## Summary

Parses six config keys (`multi_region_failover.enabled`,  `multi_region_failover.failover_metrics`, `multi_region_failover.metric_allowlist`,`multi_region_failover.api_key`, `multi_region_failover.site`, `multi_region_failover.dd_url`) to build multi-region failover configuration.

Creates a new transform that ingests the configuration to determine what endpoint to send data to as a failover, and which metrics to forward (if a `metric_allowlist` is configured).

The architecture also opens the possibility to expand capabilities to logs and traces in a way that is compliant with existing ADP patterns.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests. 

## References

- Closes: https://github.com/DataDog/saluki/issues/1678


bin/agent-data-plane/src/cli/run.rs
```diff
@@ -548,6 +549,46 @@ async fn add_baseline_metrics_pipeline_to_blueprint(
...
+        .error_context("Failed to configure Multi-Region Failover Datadog forwarder.")?;
+    let mrf_forwarder_config = if mrf_config.is_metrics_active() {
+        match (mrf_config.metrics_endpoint_url(), mrf_config.api_key()) {
+            (Some(dd_url), Some(api_key)) => mrf_forwarder_base.with_endpoint_override(dd_url, api_key.to_string()),
+            _ => {
+                warn!(
+                    "Multi-Region Failover metrics forwarding is enabled, but endpoint or API key configuration is \
+                     missing. The MRF metrics gateway will drop all events."
+                );
+                mrf_forwarder_base
```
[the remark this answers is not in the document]
lucastemb: Valid concern. We wouldn't want to handle the same metrics twice. If it's misconfigured, we will log a warning and it will not wire the MRF metric branch stuff into the topology. :P

lib/saluki-components/src/forwarders/datadog/mod.rs
```diff
@@ -66,6 +66,7 @@ impl DatadogConfiguration {
         endpoint.set_dd_url(dd_url);
         endpoint.set_api_key(api_key);
         self.forwarder_config.clear_opw_metrics_endpoint();
+        self.configuration = None;
```
rayz: Is `multi_region_failover.api_key` refreshed anywhere at all? `with_endpoint_override()` drops the live config reference, so this looks static even if the config stream later updates the MRF API key.

lucastemb: To my understanding, I don't think that the API Key should be updated dynamically based on the following snippet from the Core Agent (https://github.com/DataDog/datadog-agent/blob/ec7bc68731a4e3561a4c5f5366a46c24b9e28368/cmd/agent/subcommands/run/command.go#L556-L573) which lists the config vars that could be changed at runtime. 

I could be wrong though. So feel free to clarify.

rayz: It can be updated, when secrets are refreshed. I'm not sure what the code you linked me is but thats not the list of the config vars that could be changed at runtime. (If that list was the list of the only keys that can be updated, you would see `api_key` on there which can definitely be updated

tobz (approved):

PR #1794, opened 2026-06-01
enhancement(core): Add efficient trace payload (aka v1) encoder

ajgajg1134:
## Summary

  Migrates the Datadog trace encoder in agent-data-plane from the "classic" TracerPayload wire format to the string-indexed idx trace payload format (the idx / ETP variant used by the newer trace intake). Instead of embedding string values inline on every span, chunk, and tracer-payload field, the encoder now interns strings into a per–tracer-payload string table and emits *_ref indices, materially shrinking trace payloads with repeated strings (service names, resource names, tags, attribute keys, etc.).

## What changed

Trace encoder (lib/saluki-components/.../encoders/datadog/traces/mod.rs)
 - Move the datadog trace encoder to emit the new "efficient trace payload" format (aka IDX format) which uses a deduplicated string table to allow for more compact payloads.
 - The new payload format also provides some new field improvements like 128-bit trace IDs, and moving things like origin and sampling mechanism to top level fields
 - Meta/Metrics/MetaStruct are gone now and replaced with a single typed Attributes map

## How did you test this PR?
  - New protobuf round-trip test for mixed classic/indexed AgentPayload.
  - Encoder unit tests updated to decode the idx wire format directly (handling string-table-after-chunks ordering) and continue to verify Error Tracking Standalone chunk-tag
  behavior.
  - stele parser extended for idx payloads, keeping correctness cases valid.
  - Wrote a small smoke test application to emit traces and observed them in the UI

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance


bin/correctness/datadog-intake/src/app/traces/state.rs
```diff
@@ -36,8 +36,14 @@ impl TracesState {
...
-    pub fn merge_agent_payload(&self, payload: AgentPayload) -> Result<(), GenericError> {
-        let new_spans = Span::get_spans_from_agent_payload(&payload);
+    ///
+    /// Both the classic `tracerPayloads` (field 5) and the idx `idxTracerPayloads` (field 11)
+    /// paths are decoded. `raw_body` must be the original protobuf-encoded bytes of the
+    /// `AgentPayload` and is used to decode field 11 directly, bypassing the incorrectly-typed
+    /// generated field in the `AgentPayload` struct.
+    pub fn merge_agent_payload(&self, payload: AgentPayload, raw_body: &[u8]) -> Result<(), GenericError> {
+        let mut new_spans = Span::get_spans_from_agent_payload(&payload);
+        new_spans.extend(Span::get_spans_from_idx_bytes(&payload, raw_body));
```
tobz: Are these always mutually exclusive depending on payload version?

lib/protos/datadog/src/lib.rs
```diff
@@ -57,6 +61,12 @@ pub mod traces {
     pub mod builders {
         pub use super::super::trace_piecemeal_include::datadog::trace::*;
     }
+
+    /// String-indexed (`idx`) trace payload types used by the v0.2 trace intake format.
```
tobz: v0.2? :suspect:

lib/protos/datadog/build.rs
```diff
@@ -107,6 +107,20 @@ fn main() {
         .customize_callback(SerdeCapableStructs)
         .run_from_script();
 
+    // Separate invocation for idx proto types to avoid filename collision with
```
tobz: Totally a take-it-or-leave-it thing, but it feels very opaque as an outsider to see "idx proto types" instead of just "v1"... like "v1" is immediately grokable/intuitive, but "idx" less so. Like we talk about it as the v1 protocol internally, not the "idx protocol."

lib/saluki-components/src/encoders/datadog/traces/mod.rs
```diff
@@ -62,6 +59,54 @@ static CONTENT_TYPE_PROTOBUF: HeaderValue = HeaderValue::from_static("applicatio
...
+const TAG_ETS_STANDALONE_ERROR_KEY: &str = "_dd.error_tracking_standalone.error";
+const TAG_ETS_STANDALONE_ERROR_VALUE: &str = "true";
+
+/// String interning table for the `idx` tracer payload format.
+///
+/// Index 0 is always the empty string. All other strings are assigned indices in insertion order.
+#[derive(Debug)]
+struct StringTable {
+    strings: Vec<MetaString>,
+    indices: FastHashMap<MetaString, u32>,
```
tobz: If you switch to `FastIndexMap`, you can drop `strings` and just use `self.indices.insert_full` to get back a tuple of `(index, <maybe previous value>)`.. and then for lookups, you'd do `self.indices.get_index_of` to check if it existed and, if so, get the index for it in one fell swoop.

Although as I say that, I realize all we'd _really_ need in that scenario is a set, not a full map.... so you would probably want to end up adding an equivalent alias for `indexmap::IndexSet` to `lib/saluki-common/src/collections/mod.rs` and then actually use `FastIndexSet` instead.

lib/protos/datadog/src/lib.rs
ajgajg1134: haha yeah, versions are... confusing, this new payload format is v1 from tracer -> trace agent, but the endpoint the trace-agent uses at intake is called `api/v0.2/traces`.... which internally they actually call v1... but since this is a new field it's the same endpoint version on the backend...and the non indexed version of a tracer payload is v0.7 for tracers.... it's a mess

lib/protos/datadog/build.rs
ajgajg1134: yeah this is kinda my bad. See my other comment on trace versioning. I've started to try and talk about the project as the "Efficient trace payload" aka ETP to avoid this "V1" naming collision that happens everywhere. I went with idx in the trace-agent code since the biggest difference is that strings are 'indexed'. But I'm fine with renaming this to whatever is clearest here :P

bin/correctness/datadog-intake/src/app/traces/state.rs
ajgajg1134: structurally no, but functionally at the moment yes. The trace-agent will only ever emit one or the other in a single agent payload. I'd have to double check if trace intake can support receiving both at the same time though

lib/saluki-components/src/encoders/datadog/traces/mod.rs
ajgajg1134: ooo good call. TIL about IndexSet/IndexMap. The intern function is so much simpler now too

lib/saluki-components/src/encoders/datadog/traces/mod.rs
```diff
@@ -81,29 +80,21 @@ impl Default for StringTable {
...
     fn intern(&mut self, s: &str) -> u32 {
-        if let Some(&idx) = self.indices.get(s) {
-            return idx;
-        }
-        let idx = self.strings.len() as u32;
-        let ms = MetaString::from(s);
-        self.strings.push(ms.clone());
-        self.indices.insert(ms, idx);
-        idx
+        let (idx, _) = self.indices.insert_full(MetaString::from(s));
```
tobz: This will reconstruct a `MetaString` for every call, even if it doesn't end up actually inserting it due to already being present, so we should prefer the split `get_index_of`/`insert_full` paradigm.

lib/saluki-components/src/encoders/datadog/traces/mod.rs
```diff
@@ -609,42 +531,36 @@ impl TraceEndpointEncoder {
...
-                }
-                Ok(())
-            })?;
-
             // Tracer payload metadata refs (skip default/empty values).
+            // Strings are interned on the fly; the string table is written at the end of this
+            // closure so it is complete by the time tp.strings() is called.
             if !trace.payload.container_id.is_empty() {
-                tp.container_id_ref(self.string_table.indices[trace.payload.container_id.as_ref()])?;
+                tp.container_id_ref(self.string_table.intern(trace.payload.container_id.as_ref()))?;
```
tobz: You _should_ be able to do `&trace.payload.container_id` instead of `trace.payload.container_id.as_ref()` (and the same for all the other instances of this pattern where you have a `MetaString`.) Just makes it _slightly_ more succinct.

(Rust will use "[deref coercion](https://doc.rust-lang.org/std/ops/trait.Deref.html#deref-coercion)" to take `&MetaString` and automatically coerce it to `&str` since `MetaString` implements `Deref<Target = str>`.)

lib/protos/datadog/build.rs
tobz: I think ETP or v1 make the most sense since they give something more tangible / intuitive to use when discussing.

lib/saluki-components/src/encoders/datadog/traces/mod.rs
tobz: Another thought, which I'll flesh out when I have a chance: making this generic over `stringtheory::CheapMetaString` to avoid reconstructing when we already are passing in something that is derived from `MetaString` vs something that only ever comes as `&str` by the time we want to intern it.

lib/protos/datadog/build.rs
```diff
@@ -37,11 +53,77 @@ impl CustomizeCallback for SerdeCapableStructs {
...
+    stage_proto(
+        format!("{VENDORED_TRACE_ROOT}/idx/span.proto"),
+        staged_trace_root.join("idx/idx_span.proto"),
+        &[],
+    );
+    stage_proto(
+        format!("{VENDORED_TRACE_ROOT}/idx/tracer_payload.proto"),
+        staged_trace_root.join("idx/idx_tracer_payload.proto"),
+        &[("datadog/trace/idx/span.proto", "datadog/trace/idx/idx_span.proto")],
+    );
```
tobz, line 103: Hmmm, I'm not entirely sure I'm following: _why_ are we moving the proto files around here?

lib/saluki-common/src/collections/mod.rs
```diff
@@ -29,3 +29,6 @@ pub type PrehashedHashSet<T> = std::collections::HashSet<T, NoopU64BuildHasher>;
 /// This is only suitable when using `u64` for the key type, or another type which only wraps over a `u64` value. See
 /// [`NoopU64Hasher`][crate::hash::NoopU64Hasher] for more details.
 pub type PrehashedHashMap<K, V> = std::collections::HashMap<K, V, NoopU64BuildHasher>;
+
+/// A hash set with stable insertion order based on `indexset` ([`IndexSet`][indexmap::IndexSet]) using [`FastHasher`][crate::hash::FastHasher].
+pub type FastIndexSet<K> = indexmap::IndexSet<K, FastBuildHasher>;
```
tobz: Can we just move this up to go after `FastIndexMap`?

lib/saluki-components/src/encoders/datadog/traces/mod.rs
```diff
@@ -451,49 +494,44 @@ impl TraceEndpointEncoder {
...
-        let hostname = resolve_hostname_from_payload(
-            trace.payload.hostname.as_ref(),
+        // Clone to avoid a lifetime conflict: hostname_str borrows from this copy, and
+        // self.string_table is mutably borrowed inside the encoding closure below.
+        let default_hostname_copy = self.default_hostname.as_ref().to_string();
+        let hostname_str: Option<&str> = resolve_hostname_from_payload(
+            &trace.payload.hostname,
             source.as_ref(),
-            Some(self.default_hostname.as_ref()),
+            Some(default_hostname_copy.as_str()),
```
tobz: From testing locally, I was able to use `Some(self.default_hostname.as_ref())` without issue.

Maybe the comment _was_ valid at some point, but seemingly no longer.

lib/saluki-components/src/encoders/datadog/traces/mod.rs
```diff
@@ -639,21 +727,15 @@ impl TraceEndpointEncoder {
...
-            }
+            // Write the string table after all refs so the table is complete.
+            // Protobuf allows fields in any order; decoders that do a full parse before
+            // resolving refs handle strings-after-refs correctly.
+            tp.strings(|sb| {
+                for s in self.string_table.indices.iter() {
+                    sb.add(s.as_ref())?;
+                }
+                Ok(())
+            })?;
```
tobz: You can shrink this a _little_ bit by using `add_many_mapped`:

```rust
tp.strings(|sb| sb.add_many_mapped(&self.string_table.indices, |s| &**s))?;
```

(This takes all the values from any value which can be coerced into an `Iterator`, and since we need a `&str` but only have `&MetaString`, we use `add_many_mapped`, instead of `add_many`, which then lets us deref it above [`&**s`] which turns it into `&str`.)

lib/saluki-components/src/encoders/datadog/traces/mod.rs
```diff
@@ -954,15 +1160,13 @@ mod tests {
...
     async fn ets_chunk_tag_absent_for_non_error_trace() {
         let mut encoder = make_encoder(true).await;
-        let trace = make_trace(); // no error
+        let trace = make_trace();
         let mut buf = Vec::new();
         encoder.encode(&trace, &mut buf).expect("encode should succeed");
-        let payload = AgentPayload::parse_from_bytes(&buf).expect("should parse AgentPayload");
-        let has_tag = payload
-            .tracerPayloads
+        let chunk_attrs = decode_etp_chunk_attributes(&buf);
```
tobz, line 1053: This is a general comment related to all the instances in the tests where we're now using manual implementations of protobuf decoding: was there some particular issue with using the code-generated version that made it infeasible?

It strikes me as fragile to manually handroll these decoder methods, on top of just adding a lot of boilerplate, when we could depend on the code-generated version instead.

lib/protos/datadog/build.rs
ajgajg1134: This was an area of confusion for me too. If I understood the errors I was seeing correctly, this is due to name collisions related to the `idx` proto package using the same names as the package above it. Like `TracerPayload` exists in both namespaces but with a different definition. My original approach was a totally separate custom proto generation for the IDX stuff but it meant that if you accidentally referred to the _old_ idx code it would be wrong 🤔 (This is an area of the rust code that I'm especially not familiar / comfortable with so I'm very open to any ideas or pairing on this problem)

lib/saluki-components/src/encoders/datadog/traces/mod.rs
ajgajg1134: Yeah that makes sense. I think the primary purpose of having the custom decoders is that they also do double duty in decoding the string table to make the assertions clearer, but I agree here I think I can refactor this to use the code-generated decoder and then use helper methods to do string dereferencing

lib/protos/datadog/build.rs
tobz: Alright, from looking at this more locally, it seems like there's just a bug in `protobuf-codegen` where it naively manipulates the input filename to get the resulting output module name, which then leads to the observed name collisions.

This doesn't happen with `piecemeal`- or `prost`-based codegen because they do, in fact, properly construct the module path from the absolute input file path.

This is fine for now: as mentioned in our chat, this area is ripe for consolidation (getting from three to two, or even one, codegen crate) so it's fine for now since it works and it's just a build time thing.

tobz (approved):

PR #1795, opened 2026-06-01
feat(components): read tag filterlist cache capacity from `data_plane.dogstatsd.aggregator_tag_filter_cache_capacity`

jszwedko:
## Summary

Follow up to https://github.com/DataDog/saluki/pull/1771. [A review comment on the associated datadog-agent PR](https://github.com/DataDog/datadog-agent/pull/51522#discussion_r3333012309) caused me to realize that these tag cache capacity settings actually need to be separate since the Core Agent can still run the tag aggregation feature even if ADP is enabled (and so needs its own cache limit).

This PR adds a new setting specifically for ADP.

Don't merge until DataDog/datadog-agent#51522 is merged.

## Test plan

- [x] All `tag_filterlist` unit tests pass
- [ ] With `data_plane.dogstatsd.aggregator_tag_filter_cache_capacity: 50000` in `datadog.yaml`, confirm ADP uses capacity 50,000
- [ ] Without override, confirm ADP uses default capacity 100,000

### DIFF
--- a/lib/saluki-components/src/config_registry/datadog/unsupported.rs
+++ b/lib/saluki-components/src/config_registry/datadog/unsupported.rs
@@ -325,17 +325,6 @@ crate::declare_annotations! {
         test_json: None,
         pipeline_affinity: PipelineAffinity::Pipelines(&[Pipeline::DogStatsD]),
     };
-    /// `aggregator_tag_filter_cache_capacity` - not read by ADP; use `data_plane.dogstatsd.aggregator_tag_filter_cache_capacity` instead.
-    AGGREGATOR_TAG_FILTER_CACHE_CAPACITY = SalukiAnnotation {
-        schema: &schema::AGGREGATOR_TAG_FILTER_CACHE_CAPACITY,
-        support_level: SupportLevel::Incompatible(Severity::Low),
-        additional_yaml_paths: &[],
-        env_var_override: None,
-        used_by: &[],
-        value_type_override: None,
-        test_json: None,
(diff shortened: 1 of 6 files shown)


lib/saluki-components/src/config_registry/datadog/unsupported.rs
```diff
@@ -325,6 +325,17 @@ crate::declare_annotations! {
         test_json: None,
         pipeline_affinity: PipelineAffinity::Pipelines(&[Pipeline::DogStatsD]),
     };
+    /// `aggregator_tag_filter_cache_capacity` - not read by ADP; use `data_plane.dogstatsd.aggregator_tag_filter_cache_capacity` instead.
+    AGGREGATOR_TAG_FILTER_CACHE_CAPACITY = SalukiAnnotation {
+        schema: &schema::AGGREGATOR_TAG_FILTER_CACHE_CAPACITY,
+        support_level: SupportLevel::Incompatible(Severity::Low),
```
tobz: This issues a debug log only, right?

jszwedko: It does, but come to think of it, we shouldn't actually we shouldn't even look at this key since it may be set in addition to `data_plane.dogstatsd.aggregator_tag_filter_cache_capacity`. It should be ignored instead.

jszwedko: Moved to ignored in [e579a97](https://github.com/DataDog/saluki/pull/1795/commits/e579a97e81be0216219c4bbfaea490a945fd486e)

tobz (approved):

PR #1796, opened 2026-06-02
enhancement(core): simpler method for connecting sequential components in `TopologyBlueprint`

tobz:
## Summary

This PR introduces a new method, `connect_components_in_order`, to `TopologyBlueprint`, which allows specifying a list of component IDs to be connected together sequentially.

In many cases, we create a number of components and then connect them together one after the other, so we might create A, B, C, and D... and then connect them in the shape of `A -> B -> C -> D`. With the typical approach of using `TopologyBlueprint::connect_component`, this means three method calls where source/destination are inverted, which doesn't read super great:

```rust
blueprint.connect_component("b", ["a"])?
    .connect_component("c", ["b"])?
    .connect_component("d", ["c"])?;
```

With `connect_components_in_order`, we can instead pass a list of the components, in the intended connection order, which matches the way we might naturally describe how the components are connected:

```rust
blueprint.connect_components_in_order(["a", "b", "c", "d"])?;
```

This new method only handles connecting together components in a single output fashion: you can't specify multiple outputs to attach to the next component in the list, and you can't refer to named outputs, because each element in the list needs to be usable as both a component output ID _and_ component input ID, which is only satisfiable when referring to a default output. Even with this caveat, though, it's still useful, as evidenced by the number of existing `connect_component` calls we were able to replace in this PR with the new method.

We'll likely have a follow-up PR to change `connect_component` itself to reverse the parameter order so either method being used "reads" the same, but I'm leaving that out here because I'll likely be experimenting with trying to make it possible to simply pass `"a"` instead of `["a"]` when declaring the "source" portion of the connection.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Existing and new unit tests.

## References

DADP-2

### DIFF
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -543,10 +543,8 @@ async fn add_baseline_metrics_pipeline_to_blueprint(
         // Components.
         .add_transform("metrics_enrich", metrics_enrich_config)?
         .add_encoder("dd_metrics_encode", dd_metrics_config)?
-        // Metrics.
-        .connect_component("dd_metrics_encode", ["metrics_enrich"])?
-        // Forwarding.
-        .connect_component("dd_out", ["dd_metrics_encode"])?;
+        // Metrics, then forwarding.
+        .connect_components_in_order(["metrics_enrich", "dd_metrics_encode", "dd_out"])?;
 
     Ok(())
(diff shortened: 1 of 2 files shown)


thieman (approved):
Yes this pleases my primitive monkey brain

PR #1797, opened 2026-06-02
feat(io): add vsock support for Agent IPC connections

jszwedko:
## Summary
Adds support for connecting to the Datadog Agent's IPC endpoint via vsock (AF_VSOCK), enabling communication from guest VMs (e.g., Nitro Enclaves) to host-side Agent processes.

This implementation mirrors the Agent's vsock listener logic, allowing configuration of a vsock Context ID (CID) to route IPC connections through AF_VSOCK instead of TCP. The port is taken from the configured endpoint URI, similar to how the Agent's `vsock_addr` configuration works.

## Change Type
- [x] New feature
- [ ] Bug fix
- [ ] Performance

## How did you test this PR?
- Added vsock feature flag to `lib/saluki-io` and `lib/datadog-agent-commons`
- Implemented `Transport::Vsock` variant in the HTTP connector to handle vsock connections alongside TCP and Unix socket transports
- Added vsock CID configuration to `RemoteAgentClientConfiguration` with proper platform and feature gating
- Updated `HttpsCapableConnectorBuilder` with `with_vsock_cid()` method to enable vsock routing
- All vsock code is conditionally compiled behind `#[cfg(all(feature = "vsock", target_os = "linux"))]` to ensure it only builds on supported platforms

## References
- Based on Agent implementation: `comp/api/api/apiimpl/listener/common.go`
- Slack discussion: vsock support for connecting to Agent IPC endpoint in guest VM environments

---

PR by Bits - [View session in Datadog](https://app.datadoghq.com/code/ab7d1e51-0633-463a-9f85-81871f8ec872)

Comment @datadog to request changes

### DIFF
--- a/lib/datadog-agent-commons/src/ipc/config.rs
+++ b/lib/datadog-agent-commons/src/ipc/config.rs
@@ -235,10 +235,25 @@ impl RemoteAgentClientConfiguration {
         self.grpc_max_message_size
     }
 
-    /// Returns the vsock CID to use for connecting to the Agent IPC endpoint, if configured.
+    /// Returns the vsock address to use for connecting to the IPC endpoint, if configured.
+    ///
+    /// Combines the CID from `vsock_addr` with the port resolved from `endpoint()`. Returns
+    /// an error if `vsock_addr` is set but the endpoint has no explicit port.
+    ///
+    /// # Errors
+    ///
+    /// If the configured endpoint has no explicit port.
     #[cfg(target_os = "linux")]
-    pub fn vsock_cid(&self) -> Option<u32> {
-        self.vsock_addr
+    pub fn vsock_addr(&self) -> Result<Option<tokio_vsock::VsockAddr>, GenericError> {
+        let Some(cid) = self.vsock_addr else {
+            return Ok(None);
+        };
+        let port = self
+            .endpoint()?
+            .port_u16()
+            .map(u32::from)
+            .ok_or_else(|| saluki_error::generic_error!("vsock requires an explicit port in the IPC endpoint"))?;
+        Ok(Some(tokio_vsock::VsockAddr::new(cid, port)))
     }
 }
 
@@ -403,6 +418,7 @@ mod tests {
     #[cfg(target_os = "linux")]
     #[tokio::test]
     async fn vsock_addr_valid_values() {
+        // (vsock_addr input, expected CID — port always comes from cmd_port=5001)
         let cases: &[(&str, Option<u32>)] = &[
             ("", None),
             ("host", Some(2)),
@@ -413,8 +429,12 @@ mod tests {
         for (input, expected_cid) in cases {
             let mut values = serde_json::Map::new();
             values.insert("vsock_addr".to_string(), (*input).into());
+            values.insert("cmd_port".to_string(), 5001u16.into());
             let config = config_from_values(values).await;
-            assert_eq!(config.vsock_cid(), *expected_cid, "input: {input:?}");
+            let result = config
+                .vsock_addr()
+                .expect("vsock_addr() should not error with cmd_port set");
+            assert_eq!(result.map(|a| a.cid()), *expected_cid, "input: {input:?}");
         }
     }
 
--- a/lib/datadog-agent-commons/src/ipc/config.rs
+++ b/lib/datadog-agent-commons/src/ipc/config.rs
@@ -206,7 +206,7 @@ impl RemoteAgentClientConfiguration {
 
         #[cfg(not(target_os = "linux"))]
         if this.vsock_addr.is_some() {
-            warn!("`vsock_addr` is configured but vsock is only supported on Linux; the setting will be ignored");
+            warn!("`vsock_addr` is configured but vsock is only supported on Linux. Setting will be ignored.");
         }
 
         Ok(this)
--- a/lib/datadog-agent-commons/src/ipc/config.rs
+++ b/lib/datadog-agent-commons/src/ipc/config.rs
@@ -160,7 +160,7 @@ pub struct RemoteAgentClientConfiguration {
     /// Common CID values: `2` (VMADDR_CID_HOST) to connect from a guest to the host.
     ///
     /// Defaults to unset (TCP connection).
-    #[cfg(all(feature = "vsock", target_os = "linux"))]
+    #[cfg(target_os = "linux")]
     vsock_addr: Option<u32>,
 }
 
@@ -200,7 +200,7 @@ impl RemoteAgentClientConfiguration {
     }
 
     /// Returns the vsock CID to use for connecting to the Agent IPC endpoint, if configured.
-    #[cfg(all(feature = "vsock", target_os = "linux"))]
+    #[cfg(target_os = "linux")]
     pub fn vsock_cid(&self) -> Option<u32> {
         self.vsock_addr
     }
--- a/lib/datadog-agent-commons/src/ipc/config.rs
+++ b/lib/datadog-agent-commons/src/ipc/config.rs
@@ -149,6 +149,19 @@ pub struct RemoteAgentClientConfiguration {
         default = "default_grpc_max_message_size"
     )]
     grpc_max_message_size: usize,
+
+    /// vsock Context ID (CID) for connecting to the Agent IPC endpoint via AF_VSOCK.
+    ///
+    /// When set, the IPC client connects over a vsock socket using this CID with the port taken
+    /// from the configured endpoint. This mirrors the Datadog Agent's `vsock_addr` configuration,
+    /// which enables communication with Agent processes running in a host or hypervisor from within
+    /// a guest VM (e.g., on Nitro Enclaves or Terrapin microVM environments).
+    ///
+    /// Common CID values: `2` (VMADDR_CID_HOST) to connect from a guest to the host.
+    ///
+    /// Defaults to unset (TCP connection).
+    #[cfg(all(feature = "vsock", target_os = "linux"))]
+    vsock_addr: Option<u32>,
 }
 
 impl RemoteAgentClientConfiguration {
@@ -184,6 +197,12 @@ impl RemoteAgentClientConfiguration {
     /// Returns the maximum message size for gRPC.
     pub fn grpc_max_message_size(&self) -> usize {
         self.grpc_max_message_size
+    }
+
+    /// Returns the vsock CID to use for connecting to the Agent IPC endpoint, if configured.
+    #[cfg(all(feature = "vsock", target_os = "linux"))]
+    pub fn vsock_cid(&self) -> Option<u32> {
+        self.vsock_addr
     }
 }
 
--- a/lib/datadog-agent-commons/src/ipc/config.rs
+++ b/lib/datadog-agent-commons/src/ipc/config.rs
@@ -150,18 +150,40 @@ pub struct RemoteAgentClientConfiguration {
     )]
     grpc_max_message_size: usize,
 
-    /// vsock Context ID (CID) for connecting to the Agent IPC endpoint via AF_VSOCK.
+    /// vsock address for connecting to the Agent IPC endpoint via AF_VSOCK.
     ///
-    /// When set, the IPC client connects over a vsock socket using this CID with the port taken
-    /// from the configured endpoint. This mirrors the Datadog Agent's `vsock_addr` configuration,
-    /// which enables communication with Agent processes running in a host or hypervisor from within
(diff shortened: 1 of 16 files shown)


lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -446,6 +523,8 @@ impl HttpsCapableConnectorBuilder {
             error_telemetry: self.error_telemetry.clone(),
             #[cfg(unix)]
             unix_socket_path: self.unix_socket_path.map(PathBuf::into_boxed_path).map(Arc::from),
+            #[cfg(target_os = "linux")]
+            vsock_cid: self.vsock_cid,
```
jszwedko: **Bug: DNS resolver initialized unconditionally even when vsock is configured**

The `build()` method above calls `HickoryResolver::from_system_conf()` unconditionally and propagates its error via `?`. When `vsock_cid` is set, `call()` short-circuits to the vsock path and never touches the DNS/TCP stack — but `build()` still fails if the system resolver config is unavailable.

In a Nitro Enclave or microVM (the exact environments this feature targets), `/etc/resolv.conf` may not exist, causing `build()` to return an error even though DNS is never used.

Suggest guarding the resolver construction:
```rust
// Only needed for the TCP path; skip in vsock-only environments.
let mut hickory_resolver = if self.vsock_cid.is_none() {
    let mut r = HickoryResolver::from_system_conf()
        .error_context("Failed to load system DNS configuration...")?;
    if let Some(et) = &self.error_telemetry {
        r = r.with_lookup_errors_counter(et.dns_errors());
    }
    r
} else {
    HickoryResolver::noop_or_skip()
};
```

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -246,13 +266,18 @@ impl Service<Uri> for InnerConnector {
...
+        // consider the service immediately ready.
         #[cfg(unix)]
         if self.unix_socket_path.is_some() {
             return Poll::Ready(Ok(()));
         }
 
+        #[cfg(target_os = "linux")]
+        if self.vsock_cid.is_some() {
+            return Poll::Ready(Ok(()));
+        }
```
jszwedko, line 282: **Latent bug: `unix_socket_path` and `vsock_cid` can be set simultaneously with no error**

Both `poll_ready` and `call()` check `unix_socket_path` first. If a caller chains both `.with_unix_socket_path()` and `.with_vsock_cid()`, the vsock setting is silently ignored — unix wins with no warning.

There is no mutual-exclusion guard in `build()` or the builder methods. Consider asserting exclusivity in `build()`:
```rust
#[cfg(all(unix, target_os = "linux"))]
if self.unix_socket_path.is_some() && self.vsock_cid.is_some() {
    return Err(generic_error!("unix_socket_path and vsock_cid are mutually exclusive"));
}
```

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -280,6 +305,41 @@ impl Service<Uri> for InnerConnector {
             });
         }
 
+        #[cfg(target_os = "linux")]
+        if let Some(cid) = self.vsock_cid {
+            let connect_timeout = self.connect_timeout;
+            let error_telemetry = self.error_telemetry.clone();
```
jszwedko, line 314: Nit: `error_telemetry` is cloned here before the `port_u16()` check on line 314. If the port is absent the clone is immediately dropped. Moving the port extraction above the clone avoids the unnecessary `Arc` bump on the error path:
```rust
let port = match dst.port_u16() {
    Some(port) => u32::from(port),
    None => return Box::pin(std::future::ready(Err(...))),
};
let connect_timeout = self.connect_timeout;
let error_telemetry = self.error_telemetry.clone();
```

lib/saluki-io/src/net/client/http/conn.rs
jszwedko: It looks like the Agent prioritizes vsock so we'll do that too:

https://github.com/DataDog/datadog-agent/blob/7685c986ac84afb9e478cb475cf1de2ecf53de66/comp/api/api/apiimpl/listener/common.go#L30-L58

jszwedko: We should test that vsock is preferred if both are set though to ensure we don't regress the precedence.

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -424,10 +487,36 @@ impl HttpsCapableConnectorBuilder {
...
+        // without system DNS configuration (e.g., Nitro Enclaves).
+        #[cfg(target_os = "linux")]
+        let mut hickory_resolver = if self.vsock_cid.is_some() {
+            HickoryResolver::noop()
+        } else {
+            HickoryResolver::from_system_conf()
+                .error_context("Failed to load system DNS configuration when creating DNS resolver for HTTP client.")?
+        };
+        #[cfg(not(target_os = "linux"))]
         let mut hickory_resolver = HickoryResolver::from_system_conf()
```
jszwedko: Nit: the `from_system_conf().error_context("...")` call and its error string are duplicated across the two `#[cfg]` branches. Extracting to a helper eliminates the copy:
```rust
fn build_dns_resolver(error_telemetry: &Option<HttpTransactionErrorTelemetry>) -> Result<HickoryResolver, GenericError> {
    let mut r = HickoryResolver::from_system_conf()
        .error_context("Failed to load system DNS configuration when creating DNS resolver for HTTP client.")?;
    if let Some(et) = error_telemetry {
        r = r.with_lookup_errors_counter(et.dns_errors());
    }
    Ok(r)
}
```
Then in `build()`:
```rust
#[cfg(target_os = "linux")]
let hickory_resolver = if self.vsock_cid.is_some() {
    HickoryResolver::noop()
} else {
    build_dns_resolver(&self.error_telemetry)?
};
#[cfg(not(target_os = "linux"))]
let hickory_resolver = build_dns_resolver(&self.error_telemetry)?;
```

jszwedko: Fixed in refactor(net): extract build_dns_resolver helper to remove cfg duplication (https://github.com/DataDog/saluki/commit/d4d13f16b892bb25015bb2a09bc0022cb4efba7b)

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -257,6 +283,41 @@ impl Service<Uri> for InnerConnector {
...
+                        io::ErrorKind::InvalidInput,
+                        "vsock: destination URI must have an explicit port",
+                    )) as BoxError)));
+                }
+            };
+            let connect_timeout = self.connect_timeout;
+            let error_telemetry = self.error_telemetry.clone();
+            return Box::pin(async move {
+                let addr = tokio_vsock::VsockAddr::new(cid, port);
+                let stream = tokio::time::timeout(connect_timeout, tokio_vsock::VsockStream::connect(addr))
```
tobz: Import these types so we can ditch the fully-qualified paths.

lib/datadog-agent-commons/src/ipc/config.rs
```diff
@@ -185,6 +234,12 @@ impl RemoteAgentClientConfiguration {
     pub fn grpc_max_message_size(&self) -> usize {
         self.grpc_max_message_size
     }
+
+    /// Returns the vsock CID to use for connecting to the Agent IPC endpoint, if configured.
+    #[cfg(target_os = "linux")]
+    pub fn vsock_cid(&self) -> Option<u32> {
```
tobz: We should just make this return `Option<VsockAddr>`, since we should have `cmd_port` available to us... which also makes me think we should make that non-optional because we're always going to be getting it from the Core Agent.

lib/datadog-agent-commons/src/ipc/config.rs
```diff
@@ -158,9 +200,16 @@ impl RemoteAgentClientConfiguration {
...
     pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
-        config
+        let this = config
             .as_typed::<Self>()
-            .error_context("Failed to parse Datadog Agent IPC client configuration.")
+            .error_context("Failed to parse Datadog Agent IPC client configuration.")?;
+
+        #[cfg(not(target_os = "linux"))]
+        if this.vsock_addr.is_some() {
+            warn!("`vsock_addr` is configured but vsock is only supported on Linux; the setting will be ignored");
```
tobz:
```suggestion
            warn!("`vsock_addr` is configured but vsock is only supported on Linux. Setting will be ignored.");

File: lib/saluki-io/src/net/client/http/conn.rs [hunk 8]
```diff
@@ -424,15 +498,38 @@ impl HttpsCapableConnectorBuilder {
...
+    ///
+    /// When set, the connector will connect via AF_VSOCK using this CID, with the port taken from
+    /// the destination URI. This allows connecting to a Datadog Agent running in a host or
+    /// hypervisor context from within a guest VM (for example, Nitro Enclaves or microVM environments).
+    ///
+    /// Mirrors the Agent's `vsock_addr` configuration key.
+    ///
+    /// Defaults to unset (TCP connections via DNS).
+    #[cfg(target_os = "linux")]
+    pub fn with_vsock_cid(mut self, cid: u32) -> Self {
```
tobz: Just mentioning here specifically that we should be taking `VsockAddr`. I don't think we want to generically allow nabbing the port from the URI that the request is being sent to, but that it should be explicit and upfront.

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -424,15 +498,38 @@ impl HttpsCapableConnectorBuilder {
         self
     }
 
+    /// Sets a vsock Context ID (CID) to route all connections through.
+    ///
+    /// When set, the connector will connect via AF_VSOCK using this CID, with the port taken from
+    /// the destination URI. This allows connecting to a Datadog Agent running in a host or
+    /// hypervisor context from within a guest VM (for example, Nitro Enclaves or microVM environments).
+    ///
+    /// Mirrors the Agent's `vsock_addr` configuration key.
```
tobz: Nit: Datadog Agent-ism.

tobz (changes requested):

lib/datadog-agent-commons/src/ipc/config.rs
jszwedko: Mmm, true, we should always be getting it from the Core Agent, but it will be `""` in the default case which I think we'd like to represent as `None`. 👍  on using the type though.

lib/saluki-io/src/net/client/http/conn.rs
jszwedko: Fixed in fix(docs): address review feedback on vsock doc comments (https://github.com/DataDog/saluki/commit/460196ebed68ef41fd93420d34647fea7822c7ce)

lib/saluki-io/src/net/client/http/conn.rs
jszwedko: Fixed in fix(docs): address review feedback on vsock doc comments (https://github.com/DataDog/saluki/commit/460196ebed68ef41fd93420d34647fea7822c7ce)

lib/datadog-agent-commons/src/ipc/config.rs
jszwedko: Applied via your suggestion commit, also covered in fix(docs): address review feedback on vsock doc comments (https://github.com/DataDog/saluki/commit/460196ebed68ef41fd93420d34647fea7822c7ce)

lib/saluki-io/src/net/client/http/conn.rs
jszwedko: 👍 since I keep making this mistake, I also added a note to AGENTS.md in this PR.

lib/datadog-agent-commons/src/ipc/config.rs
jszwedko: Fixed in refactor(net): take VsockAddr directly instead of extracting port from URI (https://github.com/DataDog/saluki/commit/9565eef198ccb6830349b273f51c2b86c7b4b4be)

lib/saluki-io/src/net/client/http/conn.rs
jszwedko: Fixed in refactor(net): take VsockAddr directly instead of extracting port from URI (https://github.com/DataDog/saluki/commit/9565eef198ccb6830349b273f51c2b86c7b4b4be)

tobz (approved):

tobz (approved):

PR #1801, opened 2026-06-02
feat(integration): add Windows runtime to integration tests

thieman:
## Summary

Adds Windows-container support to the integration test suite. The Panoramic harness gains a `windows` runtime that builds a Windows ADP image (layered onto the Datadog Agent LTSC2022 base) and runs the existing test cases against it via Docker on a Windows runner. Two minor naming cleanups ride along: the existing `docker` integration runtime is renamed to `linux`, and the test YAML `container.image` field is removed (the harness now selects the image from the active runtime).

The Windows GitLab CI job runs **27 of the 29 integration tests** (passing locally on the Datadog Windows runner). The two excluded tests are documented below.

### Harness

- Add a `windows` integration-test runtime to Panoramic; the existing `docker` runtime is renamed to `linux`.
- Teach Airlock about Linux vs Windows container defaults: separate network drivers (`bridge` vs `nat`), separate bind-mount sets, separate `pid_mode` handling.
- Replace the `target_is_windows_container: bool` on `AssertionContext` with `target_os: Option<airlock::driver::ContainerOs>` so future container OSes are an enum variant rather than a new boolean.
- HTTP, TCP port, and file-read assertions now select an OS-appropriate probe strategy: Linux containers are reached from the runner via mapped host ports; Windows containers are probed from inside via `docker exec` (curl.exe, .NET TcpClient/UdpClient, PowerShell `Get-Content`).
- New `DockerExitCodeCell = Arc<OnceLock<i64>>` mirrors `airlock::unix::ExitCodeCell` for the docker path so `adp_exits` no longer has to walk three optional layers.

### Test config simplification

- Remove `container.image` from every integration test YAML. The harness picks `saluki-images/datadog-agent:testing-devel` for the `linux` runtime and `saluki-images/agent-data-plane:testing-windows` for the `windows` runtime; tests no longer participate in image selection.
- Make the entire `container` block optional in the YAML schema.

### CI

- New `test-integration-windows-amd64` GitLab job that builds Panoramic and ADP for Windows, builds the Windows ADP test image, and runs Panoramic with `--runtime windows --no-tui -p 4`.
- Shared YAML anchors for the build-image inspect/pull script and shared `WINDOWS_PROTOC_VERSION`/`WINDOWS_PROTOC_SHA256` variables across the two Windows jobs.
- Job uses `needs: []` so it doesn't wait on unrelated Linux build/test stages.
- Sets `APP_FULL_NAME`/`APP_SHORT_NAME`/`APP_GIT_HASH` for the Windows build so `saluki-metadata` populates correctly (otherwise ADP renders its log subagent prefix as `UNKNOWN` instead of `DATAPLANE`).

## Excluded from the Windows runtime

- **`dogstatsd-forwarding`**: the test depends on a Linux Python fixture, UDS, and `/tmp` paths. Porting requires a Windows fixture container and named-pipe transport — out of scope.
- **`adp-rar-disabled`**: asserts that ADP stays alive across an exit-and-restart cycle. That contract requires a supervisor (s6 in the converged Linux Agent image). The Windows entrypoint runs ADP as a direct child without a supervisor, same as the `mac` runtime where this test is also excluded.

## Test plan

- `make fmt && make check-all`
- `cargo nextest run -p panoramic -p airlock`
- `cargo run -p panoramic -- list -d test/integration/cases --runtime linux --json` → 29 tests
- `cargo run -p panoramic -- list -d test/integration/cases --runtime windows --json` → 27 tests
- `make test-integration` (Linux runtime) on the `linux` integration runner
- Windows GitLab job `test-integration-windows-amd64`: 27 passed, 0 failed


webern (commented):
With this and the macOS tests, how are we segmenting them for local developer workflow? Like... I guess what I'm getting at is it would be nice if `make test-integration` and `make test-correctness` still work on one's local machine, where we expect that local dev machine to be either a Linux host or macOS host.

I haven't done a lot of development that crosses the Windows / *nix divide, so I'm not sure the best practices for keeping developer workflows clean.

thieman:
@webern I haven't reviewed this one yet but the pattern from the Mac integration PR is that `panoramic` takes a `--runtime` arg that controls running integration tests on different platforms. So if you're on a Mac, you can actually either run `mac` (native) or `docker` (Linux within Docker). If you run through the Makefile then you get your platform default runtime. That's probably not quite the actual flow we'd want for Mac local tests as we should probably layer in Tart VMs somewhere so that it doesn't mess with your local setup at all, but I left that as a TODO.

Unit tests are driven by `cargo` so the platform-specific invocations for them will probably just be in the Makefile. Currently not planning on porting `correctness` to other runtimes so it's Docker-Linux only still.

.gitlab/windows.yml
```diff
@@ -52,3 +52,59 @@ unit-tests-windows-amd64:
...
+      -e GITLAB_CI
+      -e CI_JOB_ID
+      -e CI_PIPELINE_ID
+      -e CI_COMMIT_SHA
+      -e CI_PROJECT_NAME
+      -e CI_COMMIT_REF_NAME
+      -e WINDOWS_CI_CARGO_HOME="c:\mnt\.ci-cache\cargo"
+      -e RUSTUP_HOME="c:\mnt\.ci-cache\rustup"
+      -e CARGO_TARGET_DIR="c:\mnt\target\windows-ci"
+      -e WINDOWS_ADP_BASE_IMAGE="registry.datadoghq.com/agent:7.78.0-ltsc2022"
```
thieman, line 100: Is it possible to have this tag as a constant somewhere and share it across wherever it's used? Same question for the protoc fields

.gitlab/windows.yml
```diff
@@ -52,3 +52,59 @@ unit-tests-windows-amd64:
...
+test-integration-windows-amd64:
+  extends: .windows-amd64-test-job
+  stage: e2e
+  needs: []
+  cache:
+    key: windows-amd64-integration-cargo
+    paths:
+      - .ci-cache/cargo/
+      - .ci-cache/protoc/
+      - .ci-cache/rustup/
```
thieman, line 73: Verify that the cache is working as you expect

bin/correctness/airlock/src/driver.rs
```diff
@@ -52,6 +52,15 @@ impl fmt::Display for ExitStatus {
     }
 }
 
+/// Container operating system for a driver target.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum ContainerOs {
```
thieman, line 57: Is this how this should be formatted according to Rust standards? I'd think `ContainerOS` would be better

bin/correctness/airlock/src/driver.rs
```diff
@@ -306,21 +318,69 @@ impl DriverConfig {
...
+    }
+
+    fn needs_shared_volume_permission_fixup(&self) -> bool {
+        self.container_os == ContainerOs::Linux
+    }
+
+    fn network_driver(&self) -> &'static str {
+        match self.container_os {
+            ContainerOs::Linux => "bridge",
+            ContainerOs::Windows => "nat",
```
thieman, line 352: Explain why these are different

bin/correctness/airlock/src/driver.rs
```diff
@@ -306,21 +318,69 @@ impl DriverConfig {
...
         self
     }
+
+    /// Sets the operating system defaults used when creating this container.
+    pub fn with_container_os(mut self, container_os: ContainerOs) -> Self {
+        self.container_os = container_os;
+        self
+    }
+
+    fn needs_shared_volume_permission_fixup(&self) -> bool {
```
thieman, line 341: Everything related to this needs explanatory docstrings

bin/correctness/airlock/src/driver.rs
```diff
@@ -306,21 +318,69 @@ impl DriverConfig {
...
 
 /// Detailed information about the spawned container.
 #[derive(Debug, Default)]
 pub struct DriverDetails {
     container_name: String,
+    container_ip: Option<String>,
     port_mappings: Option<HashMap<String, u16>>,
 }
 
+fn record_port_mapping(
```
thieman: `record`? What does this actually do, that's a weird verb

bin/correctness/panoramic/src/assertions/file_contains.rs
```diff
@@ -98,7 +98,11 @@ impl Assertion for FileContainsAssertion {
             let read_result = if ctx.is_host_process {
                 read_file_local(&self.path).await
             } else {
-                read_file_in_container(&ctx.container_name, &self.path).await
+                if ctx.target_is_windows_container {
+                    read_file_in_windows_container(&ctx.container_name, &self.path).await
+                } else {
+                    read_file_in_container(&ctx.container_name, &self.path).await
+                }
```
thieman: It seems like this branch could be taken care of inside the `read_file_in_container` function since they get the same args? Would that be simpler?

bin/correctness/panoramic/src/assertions/http_check.rs
```diff
@@ -83,20 +94,24 @@ impl Assertion for HttpCheckAssertion {
...
-            Ok(client) => client,
-            Err(e) => {
-                return AssertionResult {
-                    name: self.name().to_string(),
-                    passed: false,
-                    message: format!("Failed to build HTTP client: {}.", e),
-                    duration: started.elapsed(),
-                };
+        let client = if ctx.target_is_windows_container {
+            None
```
thieman: What is going on here, are http assertions not viable for Windows?

bin/correctness/panoramic/src/assertions/mod.rs
```diff
@@ -102,6 +102,14 @@ pub struct AssertionContext {
...
     /// Port mappings from internal port to host port.
     pub port_mappings: std::collections::HashMap<String, u16>,
+    /// Container IP address on its primary Docker network, if known.
+    pub container_ip: Option<String>,
+    /// Whether the target container runs Windows. Switches assertion helpers to Windows-flavored
+    /// implementations: HTTP checks and TCP port checks run via `docker exec` inside the
+    /// container instead of from the host, file reads use PowerShell `Get-Content` instead of
+    /// `cat`, and HTTP/TCP probes target `<container_ip>:<internal_port>` instead of the
+    /// mapped ephemeral host port.
+    pub target_is_windows_container: bool,
```
thieman: This should use the runtime enum instead, we may expand to other platforms

bin/correctness/panoramic/src/assertions/mod.rs
```diff
@@ -112,6 +120,8 @@ pub struct AssertionContext {
     /// path or while the process is still running; `Some(None)` if the process was killed by
     /// signal; `Some(Some(code))` if it exited normally.
     pub host_process_exit_code: Option<airlock::unix::ExitCodeCell>,
+    /// Exit code of a top-level Docker target process, populated once the container exits.
+    pub docker_container_exit_code: Option<std::sync::Arc<std::sync::RwLock<Option<i64>>>>,
```
thieman: Confirm this type signature is actually necessary, seems complex

bin/correctness/panoramic/src/assertions/port_listening.rs
```diff
@@ -106,15 +140,107 @@ impl Assertion for PortListeningAssertion {
...
     match tokio::net::UdpSocket::bind("0.0.0.0:0").await {
-        Ok(socket) => socket.connect(("127.0.0.1", port)).await.is_ok(),
+        Ok(socket) => socket.connect((host, port)).await.is_ok(),
         Err(_) => false,
     }
 }
+
+async fn check_tcp_port_in_container(container_name: &str, port: u16) -> bool {
+    // The Datadog Agent LTSC image does not ship the NetTCPIP module, so we
+    // probe with a .NET TcpClient against loopback inside the container.
```
thieman, line 185: Could we fix this in the image and then remove the custom code path here?

bin/correctness/panoramic/src/assertions/port_listening.rs
```diff
@@ -106,15 +140,107 @@ impl Assertion for PortListeningAssertion {
...
+            }
+        }
+    }
+
+    async fn run(&self, container_name: &str) -> bool {
+        match *self {
+            Self::HostTcp { port } => check_tcp_port("127.0.0.1", port).await,
+            Self::HostUdp { port } => check_udp_port("127.0.0.1", port).await,
+            Self::InContainerTcp { port } => check_tcp_port_in_container(container_name, port).await,
+            Self::InContainerUdp { port } => check_udp_port_in_container(container_name, port).await,
```
thieman, line 165: It seems like the two in-container versions are also Windows-specific, is that codified somewhere?

bin/correctness/panoramic/src/config.rs
```diff
@@ -155,28 +157,48 @@ fn default_integration_runtimes() -> Vec<String> {
 /// platforms will get their own identifiers (for example, `linux`).
 pub const MAC_RUNTIME: &str = "mac";
 
-/// Runtime identifier for integration tests that run inside a Docker container.
+/// Runtime identifier for integration tests that run inside a Linux Docker container.
 pub const DOCKER_RUNTIME: &str = "docker";
 
+/// Runtime identifier for integration tests that run inside a Windows Docker container.
+pub const WINDOWS_RUNTIME: &str = "windows";
```
thieman, line 164: Our naming is getting inconsistent here, let's rename the `docker` runtime to `linux` instead. So we'll have `linux`, `mac`, and `windows` but then `linux` and `windows` actually both use containers under the hood.

Be thorough in applying this rename, it'll probably touch quite a bit.

bin/correctness/panoramic/src/runner.rs
```diff
@@ -38,6 +38,65 @@ pub(crate) type EventSender = mpsc::UnboundedSender<TestEvent>;
 /// The amount of time a test has to clean up after cancellation or timing out.
 const GRACE_TIME: Duration = Duration::from_secs(30);
 
+/// Maps shared `DD_DATA_PLANE_*` test env keys to their Windows-image-native nested form.
+///
+/// The Linux converged image's s6 entrypoint normalizes flat `DD_DATA_PLANE_*` keys into the
+/// nested `DD_DATA_PLANE__X__Y` shape ADP reads natively. We don't have that entrypoint inside
+/// the Windows test image, so we apply the same translation here. Test YAML files continue to use
+/// the flat shape for portability across runtimes.
```
thieman: Provide more context in reply to this comment as to why this is necessary and why the Windows env vars are apparently different

bin/correctness/panoramic/src/runner.rs
```diff
@@ -38,6 +38,65 @@ pub(crate) type EventSender = mpsc::UnboundedSender<TestEvent>;
...
+    ),
+    (
+        "DD_DATA_PLANE_OTLP_PROXY_LOGS_ENABLED",
+        "DD_DATA_PLANE__OTLP__PROXY__LOGS__ENABLED",
+    ),
+    ("DD_DATA_PLANE_LOG_FILE", "DD_DATA_PLANE__LOG_FILE"),
+];
+
+/// Defaults applied to every Windows-runtime test, unless the test sets them explicitly.
+const WINDOWS_ENV_DEFAULTS: &[(&str, &str)] = &[
```
thieman: I'd prefer not to have these defaults, is there a problem with adding these to the YAML files when necessary? Is there similar defaulting going on for the other runtimes?

.gitlab/windows.yml
thieman: [Claude Opus 4.7] Done in bbfdc5afb8. WINBUILDIMAGE was already shared as a base-job variable; lifted PROTOC_VERSION/SHA into the same place (renamed WINDOWS_PROTOC_VERSION/SHA256 for clarity) and turned the `docker image inspect ... else docker pull` block into a YAML anchor.

.gitlab/windows.yml
thieman: [Claude Opus 4.7] Two observations:

1. The two jobs use *different* cache keys (`windows-amd64-cargo` vs `windows-amd64-integration-cargo`) on the same paths, so the integration job does not benefit from the unit-test job's cargo cache. They build many of the same crates, so unifying the keys would shave time off the integration job. I left it alone for now since merging caches across jobs has historically been a separate concern (concurrent writes from parallel pipelines, eviction policy, etc.) and felt out of scope for this PR. Happy to follow up.

2. Once a cache exists, the inspect-or-pull block reuses the WINBUILDIMAGE only across runs on the same runner; protoc and the cargo registry are restored from the GitLab cache. The Windows job log confirms the rustup/cargo home reuse on subsequent runs.

Let me know if you'd like me to consolidate the cache keys.

bin/correctness/airlock/src/driver.rs
thieman: [Claude Opus 4.7] Per [Rust API Guidelines (RFC 430 C-CASE)](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case), acronyms count as a single word in camel-case identifiers, so `Os` (not `OS`) is the conventional spelling. `std` itself uses this: `OsStr`, `OsString`, `IoError`, `TcpStream`. So `ContainerOs` matches the convention and `ContainerOS` would deviate from it. No change.

bin/correctness/airlock/src/driver.rs
thieman: [Claude Opus 4.7] Done in 5b98cd3dfa. Added a doc comment on `network_driver` explaining that Linux containers use `bridge` and Windows containers use `nat` (the only Windows driver that supports container-to-container traffic on a single host).

bin/correctness/airlock/src/driver.rs
thieman: [Claude Opus 4.7] Done in 5b98cd3dfa. Added explanatory docstrings to `with_container_os`, `needs_shared_volume_permission_fixup`, `network_driver`, and `container_binds_from`, plus an inline comment on the `pid_mode=host` divergence in `create_container_inner`.

bin/correctness/airlock/src/driver.rs
thieman: [Claude Opus 4.7] Renamed to `insert_port_mapping_if_parseable` in ce1339ea14. New name describes both what the helper does (insert into the map) and the precondition (the host port string parses as a `u16`); also added a docstring.

bin/correctness/panoramic/src/assertions/file_contains.rs
thieman: [Claude Opus 4.7] Good call — done in 962c06ba39. Folded the two readers into a single `read_file_in_container(name, path, is_windows)` that picks the docker-exec command (`cat` vs PowerShell `Get-Content`) inline. Drops ~70 lines of copy/pasted exec wiring.

bin/correctness/panoramic/src/assertions/http_check.rs
thieman: [Claude Opus 4.7] Refactored in 1df0755750. The previous code had an `Option<reqwest::Client>` that existed only on the Linux path plus a separate Windows branch in the request loop, which read awkwardly. Replaced with an `HttpProbe` enum constructed once before the loop:

```rust
enum HttpProbe {
    HostClient { client: reqwest::Client },           // linux/mac
    InContainerCurl { container_name: String },       // windows: docker exec curl.exe
}
```

Windows HTTP assertions are viable, just routed differently: the listener is bound to the container's internal port and not exposed on the runner host, so we shell out to `curl.exe` via `docker exec` inside the container instead of using `reqwest` from the runner.

bin/correctness/panoramic/src/assertions/mod.rs
thieman: [Claude Opus 4.7] Done in 805f40e546. Replaced `target_is_windows_container: bool` with `target_os: Option<airlock::driver::ContainerOs>` (`None` for host-process runtimes like `mac`). Call sites use a small `ctx.target_is_windows()` helper to keep the read sites concise. Adding another container OS in the future is now an enum variant + helper, not a new boolean.

bin/correctness/panoramic/src/assertions/mod.rs
thieman: [Claude Opus 4.7] Simplified in 94668a66dd. The previous `Option<Arc<RwLock<Option<i64>>>>` had three nesting levels (present-or-not, locked, populated-or-not). The exit code is written exactly once when the container exits, so I introduced a `DockerExitCodeCell = Arc<OnceLock<i64>>` type alias mirroring the existing `airlock::unix::ExitCodeCell` and used it in `AssertionContext`. The consumer in `adp_exits.rs` collapses from a six-level if-let nest to one `.get()` call.

bin/correctness/panoramic/src/assertions/port_listening.rs
thieman: [Claude Opus 4.7] We don't own the image — the Windows test image is built `FROM registry.datadoghq.com/agent:7.78.0-ltsc2022`, a third-party Datadog Agent image. Modifying it would mean either:

1. Forking it and maintaining our own copy (heavy), or
2. Adding a custom build step on top that installs the `NetTCPIP` Windows feature (slow — Windows-feature install adds minutes per build).

`Test-NetConnection` lives in the `NetTCPIP` module, which only ships with Windows PowerShell 5.1 by default; the Datadog Agent image's `pwsh` (PowerShell 7) doesn't have it. The .NET `TcpClient` approach is two lines and avoids touching the image, so I'd rather keep it. (One alternative: switch the docker-exec calls from `pwsh` to `powershell.exe`, which has `Test-NetConnection` natively. Same code complexity, different tradeoff. Happy to swap if you prefer.)

bin/correctness/panoramic/src/assertions/port_listening.rs
thieman: [Claude Opus 4.7] Yes — codified by the runtime enum from comment #9. The two `InContainer*` `Probe` variants are constructed only when `ctx.target_is_windows()` is true; the dispatch site enforces it. Updated in 805f40e546.

bin/correctness/panoramic/src/config.rs
thieman: [Claude Opus 4.7] Done in 077d075440. Renamed `DOCKER_RUNTIME` -> `LINUX_RUNTIME` (value `"docker"` -> `"linux"`), `DEFAULT_DOCKER_TARGET_IMAGE` -> `DEFAULT_LINUX_TARGET_IMAGE`, and updated all 30 test YAMLs from `runtimes: [docker, ...]` to `runtimes: [linux, ...]`. Also updated `default_host_runtime()`, the README, related docstrings, and unit tests. The unrelated `kubernetes_in_docker` correctness runtime keeps its name.

bin/correctness/panoramic/src/runner.rs
thieman: [Claude Opus 4.7] Expanded the docstring in ecccf6dbdd. The short version: ADP reads its own configuration as a nested object, so `data_plane.enabled` is `DD_DATA_PLANE__ENABLED` (single underscore between path levels, double underscore separating the `data_plane` prefix). The Core Agent's config loader uses single underscores throughout and exposes the same setting as `DD_DATA_PLANE_ENABLED`. On the Linux converged image, an s6 init script translates the flat shape into the nested shape before launching ADP; on the Windows ADP-only image we built in this repo, there's no s6, so the harness applies the same translation here. Test YAML continues to use the flat form so it stays portable across all three runtimes.

bin/correctness/panoramic/src/runner.rs
thieman: [Claude Opus 4.7] Dropped both defaults in d9ab6477c4 — they were redundant.

- `DD_DATA_PLANE__USE_NEW_CONFIG_STREAM_ENDPOINT=false` matches ADP's own default (`unwrap_or(false)` in `bin/agent-data-plane/src/config.rs`), so setting it explicitly was a no-op.
- `DD_DATA_PLANE__REMOTE_AGENT_ENABLED=false` was a workaround from when the Windows test image had no Core Agent and ADP's RAR registration would fail bootstrap. With the Datadog Agent base image we now use, RAR works on Windows (the `adp-rar-registration` test passes against it), so ADP's natural default of `true` is the right behavior. Tests that exercise RAR-disabled behavior already set the env explicitly.

For the second part of your question: the only similar runtime-side defaulting today is `port_isolation_env()` in `test_env.rs`, which shifts well-known DogStatsD/API ports to avoid collisions with the host Core Agent. That one is genuinely cross-cutting (every container test wants it) and is shared across all runtimes, not Windows-only.

bin/correctness/airlock/src/driver.rs
thieman: Okay just as long as you understand this sounds like a breakfast cereal

test/integration/cases/adp-cmd-port/config.yaml
```diff
@@ -26,8 +26,6 @@ env:
   DD_DATA_PLANE_USE_NEW_CONFIG_STREAM_ENDPOINT: "true"
   DD_CMD_PORT: "7777"
 
-container:
-  image: "saluki-images/datadog-agent:testing-devel"
```
thieman, line 30: Removed `image` from the test files since it is a) always the same on Linux, and b) needs to be a different always-the-same value on Windows. This is set through code based on the runtime now

test/integration/cases/adp-cmd-port/config.yaml
```diff
@@ -15,7 +15,7 @@ type: integration
 name: "adp-cmd-port"
 description: "Verifies ADP connects to the correct port when cmd_port is set"
 timeout: 120s
-runtimes: [docker, mac]
+runtimes: [linux, mac, windows]
```
thieman, line 18: Renamed the `docker` runtime to `linux` everywhere. Both `linux` and `windows` run within containers, if we want to expand those later we can create `linux_native` or something

test/integration/cases/adp-logging-windows-default-path/config.yaml
```diff
@@ -0,0 +1,25 @@
+type: integration
```
thieman, line 1: Created separate Windows versions of these since they need to pass Windows-specific log paths

test/integration/cases/dogstatsd-bind-custom-hostname/config.yaml
```diff
@@ -24,25 +24,27 @@ type: integration
...
   DD_DATA_PLANE_ENABLED: "true"
   DD_DATA_PLANE_STANDALONE_MODE: "true"
   DD_DATA_PLANE_DOGSTATSD_ENABLED: "true"
+  # PANORAMIC_DYNAMIC_* values are interpreted by the test runtime, not by the test image:
+  #  - On Linux, each value is executed as a shell command inside the container at startup.
+  #    CUSTOM_HOSTNAME's command writes "foo.local -> eth0 IP" into /etc/hosts and echoes
+  #    "foo.local" as the captured value used to template DD_BIND_HOST below.
+  #  - On Windows, the harness ignores these values and resolves the variables by name:
+  #    CONTAINER_IP looks up the container's primary IPv4 address; CUSTOM_HOSTNAME adds the
+  #    same hosts entry to the Windows hosts file and returns "foo.local".
```
thieman, line 43: Hack and a half, probably fine? Can exclude Windows from this one if you'd rather

bin/correctness/panoramic/src/assertions/adp_exits.rs
```diff
@@ -135,6 +135,26 @@ impl AdpExitsWithAssertion {
                     duration: started.elapsed(),
                 };
             }
+            if ctx.container_exit_token.is_cancelled() {
+                if let Some(code) = ctx
+                    .docker_container_exit_code
+                    .as_ref()
+                    .and_then(|cell| cell.get().copied())
```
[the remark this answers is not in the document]
thieman: [Claude Opus 4.7] Confirmed bug. Fixed in 44a0c88411. Gated the docker_container_exit_code fallback on `ctx.target_is_windows()` so the linux runtime continues to require the s6 supervisor log line. Updated the docstring to call out why: on linux the container exit code is the s6 exit code (s6 keeps the container alive across ADP restarts), while on windows ADP is the top-level process so the container exit code is ADP's.

bin/correctness/panoramic/src/assertions/http_check.rs
```diff
@@ -149,3 +174,86 @@ impl Assertion for HttpCheckAssertion {
...
+    let docker = docker::connect().map_err(|e| format!("Failed to connect to Docker: {}", e))?;
+    let endpoint = endpoint.replace("localhost", "127.0.0.1");
+    let exec = docker
+        .create_exec(
+            container_name,
+            CreateExecOptions::<String> {
+                cmd: Some(vec![
+                    "curl.exe".to_string(),
+                    "-k".to_string(),
+                    "-s".to_string(),
```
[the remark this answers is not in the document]
thieman: [Claude Opus 4.7] Confirmed bug. Fixed in ef992837a5. Plumbed `insecure_skip_verify` through the `HttpProbe::InContainerCurl` variant into `get_status_in_container`, which now only adds `-k` when the flag is set. Matches the host path's `danger_accept_invalid_certs` behavior.

tobz (approved):
Not so egregious that we can't improve it in the future... 😅

thieman:
> Not so egregious that we can't improve it in the future... 😅

"At least the functional requirements are clear now," he said generously

PR #1805, opened 2026-06-03
chore(core): improve `TopologyBlueprint::connect_component`

tobz:
## Summary

This PR updates `TopologyBlueprint::connect_component` to change the order of arguments to be "source/upstream components" followed by "destination/downstream" components (instead of the reverse), to accept multiple component IDs in either position, _and_ to rename it to `connect_components` to reflect its newfound capabilities.

Prior to this PR, `connect_component` took the destination component ID and source components IDs, in that order, which looks like this:

```rust
blueprint.connect_component("destination", ["source_a", "source_b"])?;
```

At least for people who natively speak languages that are written and read left-to-right (which is the outright majority of people working on Saluki), we also, similarly, describe unidirectional graphs from left-to-right. I don't know if there's a defined phenomenon that explains this -- like how _Royal Order of Adjectives_ effectively describes the unwritten-but-inherently-preferred order of adjectives in English -- but suffice to say it's a thing, and so we want to improve our parameter order. :)

This means that with this PR, you would instead write:

```rust
blueprint.connect_components(["source_a", "source_b"], "destination")?;
```

However, one additional change we've made here is that we can support a single _or_ multiple IDs for either the "from" or "to" arguments:

```rust
// One-to-one connection:
blueprint.connect_components("source", "destination")?;

// One-to-many connection:
blueprint.connect_components("source", ["destination_a", "destination_b"])?;

// Many-to-one connection:
blueprint.connect_components(["source_a", "source_b"], "destination")?;

// Many-to-many connection:
blueprint.connect_components(["source_a", "source_b"], ["destination_a", "destination_b"])?;
```

This means that, in conjunction with `connect_components_in_order` (#1796), we now have two methods for connecting components whose arguments match the left-to-right data flow of the graph they're describing, _and_ we can represent all the different possible connection shapes (point-to-point, fan-in, fan-out, fan-in-and-out) in a single `connect_components` call.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Existing and new unit tests.

## References

DADP-2

### DIFF
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -589,9 +589,12 @@ fn add_mrf_metrics_pipeline_to_blueprint(
         .add_transform("mrf_metrics_gateway", mrf_gateway_config)?
         .add_encoder("mrf_metrics_encode", mrf_metrics_config)?
         .add_forwarder("mrf_dd_out", mrf_forwarder_config)?
-        .connect_components("mrf_metrics_gateway", ["metrics_enrich"])?
-        .connect_components("mrf_metrics_encode", ["mrf_metrics_gateway"])?
-        .connect_components("mrf_dd_out", ["mrf_metrics_encode"])?;
+        .connect_components_in_order([
+            "metrics_enrich",
+            "mrf_metrics_gateway",
+            "mrf_metrics_encode",
+            "mrf_dd_out",
+        ])?;
 
     Ok(())
 }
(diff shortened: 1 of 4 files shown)


atanzu:
`memcpy(dest, src, len)` - that's why.

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -589,9 +589,9 @@ fn add_mrf_metrics_pipeline_to_blueprint(
         .add_transform("mrf_metrics_gateway", mrf_gateway_config)?
         .add_encoder("mrf_metrics_encode", mrf_metrics_config)?
         .add_forwarder("mrf_dd_out", mrf_forwarder_config)?
-        .connect_component("mrf_metrics_gateway", ["metrics_enrich"])?
-        .connect_component("mrf_metrics_encode", ["mrf_metrics_gateway"])?
-        .connect_component("mrf_dd_out", ["mrf_metrics_encode"])?;
+        .connect_components("mrf_metrics_gateway", ["metrics_enrich"])?
```
atanzu: Shall we reverse the order here as well?

tobz: Ah yeah, missed that one which got added yesterday. 👀 Will fix.

tobz: Fixed in e9b5c99397.

aqian01 (approved):

PR #1807, opened 2026-06-03
docs(agent-data-plane): classify enable_json_stream_shared_compressor…

### LINKED ISSUE #1749: Investigate serializer config key: `enable_json_stream_shared_compressor_buffers`
## Problem

The \`enable_json_stream_shared_compressor_buffers\` config key pre-allocates shared \`bytes.Buffer\` instances in the Go JSON stream serializer to reduce GC pressure. ADP uses a Rust encoder; this may be not applicable or may map to a buffer pre-allocation configuration.

This issue was split from #1686.

## What needs to happen

Determine whether ADP's encoder has a configurable equivalent buffer pre-allocation, or whether this is a Go-specific optimization with no ADP analog.

lucastemb:
…_buffers as N/A

## Summary

This PR updates documentation ADP documentation to reflect the following: 

> ADP doesn't have a configurable equivalent buffer pre-allocation, but it doesn't look like it needs one either. The garbage collection pressure from repeated buffer allocations isn't a problem here because of ownership model in Rust. ADP already pre-allocates buffers as hardcoded constants ([here](https://github.com/DataDog/saluki/blob/main/lib/saluki-components/src/common/datadog/io.rs#L48) and [here](https://github.com/DataDog/saluki/blob/main/lib/saluki-components/src/common/datadog/request_builder.rs#L12)) and owned by a [request-builder task](https://github.com/DataDog/saluki/blob/main/lib/saluki-components/src/common/datadog/request_builder.rs#L194).

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

N/A
## References

- Closes: https://github.com/DataDog/saluki/issues/1749

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1082,7 +1082,6 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
- "tokio-vsock",
  "tonic",
  "tracing",
  "windows-registry",
--- a/docs/agent-data-plane/configuration/dogstatsd.md
+++ b/docs/agent-data-plane/configuration/dogstatsd.md
@@ -55,6 +55,7 @@ architecture is fundamentally different or the feature is platform-specific.
 | `dogstatsd_queue_size`                         | Packet channel buffer size         | ADP uses async tasks                                         |
 | `dogstatsd_telemetry_enabled_listener_id`      | Per-listener telemetry tagging     | Not feasible to thread through                               |
 | `dogstatsd_workers_count`                      | Number of DSD processing workers   | ADP uses async tasks                                         |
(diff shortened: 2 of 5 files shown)


Cargo.lock
```diff
@@ -1082,6 +1082,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
+ "tokio-vsock",
```
jszwedko: Curious how this snuck in 🤔

jszwedko (approved):

Cargo.lock
lucastemb: Hm, weird. Looks like a product of my merge to fix a previous conflict. Fixed it now.

Issue #1363, opened 2026-04-13
Investigate mapping `forwarder_max_concurrent_requests` to ADP's concurrency model.


webern:
The core agent has two concurrency knobs for the forwarder:
- [`forwarder_num_workers`][1] (default 1) — spawns N goroutine [workers][3] per domain. Also sets [`MaxConnsPerHost`][4] on the HTTP transport.
- [`forwarder_max_concurrent_requests`][2] (default 10) — each worker holds a [semaphore][5] allowing up to this many in-flight HTTP requests.

The effective default is 1 worker × 10 concurrent = **10 in-flight requests per domain**.

ADP reads `forwarder_num_workers` (default 1) and uses it as Tower's [`.concurrency_limit()`][s1], meaning **1 in-flight request at a time**. `forwarder_max_concurrent_requests` is not currently read. The hyper connection pool has [`pool_max_idle_per_host(5)`][s2] hardcoded but this only governs idle connection reuse, not concurrency.

This is a meaningful behavioral difference under load — the core agent can push 10 requests in parallel while ADP processes them sequentially. A possible mapping (discussed internally): `forwarder_num_workers` → hyper pool size, `forwarder_max_concurrent_requests` → Tower concurrency limit. Or simply use the max of the two. Needs research to determine the right mapping, particularly considering HTTP/2 multiplexing which may change the picture.

[1]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/pkg/config/setup/common_settings.go#L1522
[2]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/pkg/config/setup/common_settings.go#L1524
[3]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/comp/forwarder/defaultforwarder/domain_forwarder.go#L226-L227
[4]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/comp/forwarder/defaultforwarder/domain_forwarder.go#L258
[5]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/comp/forwarder/defaultforwarder/worker.go#L69-L88
[s1]: https://github.com/DataDog/saluki/blob/dbb968c2db21/lib/saluki-components/src/common/datadog/io.rs#L268
[s2]: https://github.com/DataDog/saluki/blob/dbb968c2db21/lib/saluki-io/src/net/client/http/client.rs#L311

aqian01:
Closed by https://github.com/DataDog/saluki/pull/1811/s

PR #1811, opened 2026-06-03
feat(datadog): support `forwarder_max_concurrent_requests`

aqian01:
## What changed

Implements `forwarder_max_concurrent_requests` as the ADP forwarder endpoint concurrency setting for #1363.

- Reads `forwarder_max_concurrent_requests` into `ForwarderConfiguration::endpoint_concurrency()`.
- Defaults endpoint concurrency to `10`.
- Clamps `0` to `1` at accessor read time.
- Keeps the existing Tower `.concurrency_limit(...)` path and feeds it with the new setting.
- Removes `forwarder_num_workers` from ADP runtime config usage and marks it unsupported in docs/registry.
- Updates DogStatsD config docs and `known-configs.json` bookkeeping.

## Why

ADP previously used `forwarder_num_workers` as the request concurrency setting, which did not match the Agent-facing config key for max concurrent outbound requests. This change makes `forwarder_max_concurrent_requests` the setting that controls ADP request concurrency.

## Validation

Existing tests


tobz (changes requested):
Looking at the forwarder worker logic in DDA, it looks like the concurrency limit is set [per worker](https://github.com/DataDog/datadog-agent/blob/c8f020af3e6bb388033bfb0ae4905c8767e18bfa/comp/forwarder/defaultforwarder/impl/worker.go#L71), so I think we'd actually want to set our concurrency limit to `forwarder_num_workers * forwarder_max_concurrent_requests` to properly mimic the behavior.

aqian01:
@tobz I marked `forwarder_num_workers` as unsupported because ADP does not actually have forwarder workers. It only has endpoint I/O tasks and Tower concurrency.

Should we still use `forwarder_num_workers` to calculate effective concurrency as `forwarder_num_workers * forwarder_max_concurrent_requests`, even though it would not create real workers?

lib/saluki-components/src/config_registry/datadog/unsupported.rs
```diff
@@ -135,10 +135,10 @@ crate::declare_annotations! {
...
-    FORWARDER_MAX_CONCURRENT_REQUESTS = SalukiAnnotation {
-        schema: &schema::FORWARDER_MAX_CONCURRENT_REQUESTS,
-        // ADP concurrency model differs. #1363
+    /// `forwarder_num_workers` - forwarder worker count.
+    FORWARDER_NUM_WORKERS = SalukiAnnotation {
+        schema: &schema::FORWARDER_NUM_WORKERS,
+        // ADP does not use this setting. #1363
         support_level: SupportLevel::Incompatible(Severity::Medium),
         additional_yaml_paths: &[],
         env_var_override: None,
```
jszwedko: If we end up leaving this option marked as unsupported it should be moved to the `Low` level since we have no plans to implement it due to architectural differences.

jszwedko (commented):
I'm wondering if we should link the HTTP connection pool to `forwarder_num_workers` since in the Agent, in addition to number of workers, I believe it controls the connection count by virtue of there being one connection per worker.

aqian01:
@jszwedko Should we keep the name of the config `forwarder_num_workers` if we are only using it for controlling the size of the HTTP connection pool?

jszwedko:
> @jszwedko Should we keep the name of the config `forwarder_num_workers` if we are only using it for controlling the size of the HTTP connection pool?

@aqian01 Good question. I'd say yes, but we can document it as a `DIVERGENT` rather than `PARITY`.

lib/saluki-components/src/common/datadog/config.rs
```diff
@@ -245,7 +259,16 @@ impl ForwarderConfiguration {
 
     /// Returns the maximum number of concurrent requests for an individual endpoint.
     pub const fn endpoint_concurrency(&self) -> usize {
-        self.endpoint_concurrency
+        if self.endpoint_concurrency == 0 {
```
jszwedko: Would it make sense to use `NonZeroUsize` as the type to avoid needing to check for `0` here?

jszwedko (commented):
🤔 I think I'm with Toby here that the concurrency should be `forwarder_num_workers` * `forwarder_max_concurrent_requests` to match the Core Agent behavior since `forwarder_max_concurrent_requests` is per-worker so to maintain the same level of concurrency we should multiply the values.

lib/saluki-components/src/common/datadog/config.rs
aqian01: https://github.com/DataDog/datadog-agent/blob/main/comp/forwarder/defaultforwarder/impl/worker.go#L71-L75

I think if we use NonZeroUsize then we would diverge from the agent since the core agent just clamps the value to 1

jszwedko: Ah, I see. I wonder if that is due to lack of stronger types for configuration in the Agent, but 👍

jszwedko (approved):

tobz (approved):

PR #1815, opened 2026-06-04
enhancement(core): refactor topologies to run via supervision

tobz:
## Summary

This PR refactors how we build and spawn topologies to enforce doing so through supervision trees.

Currently, topology building looks roughly like this:

- build your blueprint (define the components and how they're connected) as a standalone thing
- "build" the topology (create the components, which is part config validation and part creating of futures/resources, etc)
- run the topology: spawn the necessary async tasks
- monitor the topology: wait for unexpected component failure, or wait for process shutdown

However, this all happens outside of the existing supervision tree, which means we lose out on things like:

- a supervisor managing the detection (and, eventually, restarting) of failed processes (components)
- access to the dataspace registry for dynamically asserting things (like the API handlers we expose from the DogStatsD Stats destination, etc)

This PR refactors all of the aforementioned code in order to allow us to do just that. We've made a number of changes, which are most easily explained in list form (sorry):

- `TopologyBlueprint` is the primary artifact: it now implements `Supervisable` and can be added directly as a worker to a supervisor
- `BuiltTopology` and `RunningTopology` are no longer public types that have to be dealt with
- `TopologyContext` gains a new method, `dataspace`, for getting a reference to the dataspace of the supervision tree so that components don't need to do the `DataspaceRegistry::try_current` dance: they're _always_ in a supervision tree, and _always_ have access to a dataspace
- the health registry got some updates to allow for waiting for readiness on a _subset_ of components (using a name-based predicate) so that the topology can wait for environment provider-related processes to become ready before starting the topology

This means that to build/spawn a topology, all that needs to happen is to add the blueprint as a worker, and then when the supervisor it's been added to is run, the building/running of the topology itself occurs. _Tada._

Like previous PRs, we're starting out here with the initial mechanical refactoring, which means we're getting things running under supervision, but we aren't fully exploiting all the benefits _yet_:

- no components are using the dataspace registry to publish their dynamic API routes
- we're spawning components in a `JoinSet` instead of as direct child workers on the topology supervisor
- topology components (and the topology supervisor itself) cannot be restarted: trying to restart/respawn them throws an error that causes the process to exit, same as it currently would

Each of these improvements have varying level of difficulty, which is why we're deferring them to follow-up PRs.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## Hw did you test this PR?

- [x] Added new unit tests (environment provider readiness, etc)
- [x] Existing unit, integration, and correctness tests.

## References

DADP-2

### DIFF
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -196,7 +196,7 @@ pub async fn handle_run_command(
     //
     // We run our internal supervisor (control plane, environment provider, etc) and our topology supervisor
     // side-by-side, which means everyone has access to the same dataspace. This is crucial for allowing processes
-    // in the topology supervisor to (eventually)
+    // in the topology supervisor to (eventually) interact with components in the control plane, and vise versa.
     blueprint
         .with_health_registry(health_registry.clone())
         .with_memory_limiter(memory_limiter)
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -33,11 +33,10 @@ use saluki_components::{
 };
 use saluki_config::{ConfigurationLoader, GenericConfiguration};
 use saluki_core::health::HealthRegistry;
-use saluki_core::runtime::{Supervisor, SupervisorError};
+use saluki_core::runtime::{RestartMode, RestartStrategy, Supervisor};
 use saluki_core::topology::TopologyBlueprint;
 use saluki_env::EnvironmentProvider as _;
 use saluki_error::{generic_error, ErrorContext as _, GenericError};
-use tokio::{select, sync::oneshot};
 use tracing::{debug, error, info, trace, warn};
 
 use crate::{
@@ -50,7 +49,7 @@ use crate::{
     },
     internal::{
         create_internal_supervisor, logging::LoggingConfigurationTranslator, remote_agent::RemoteAgentBootstrap,
-        DogStatsDControlSurface,
+        DogStatsDControlSurface, TopologyControlSurfaces,
     },
 };
 use crate::{config::DataPlaneConfiguration, internal::env::ADPEnvironmentProvider};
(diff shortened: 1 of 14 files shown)


bin/agent-data-plane/src/cli/run.rs
```diff
@@ -200,104 +192,51 @@ pub async fn handle_run_command(
...
-        },
-        early_result = &mut internal_supervisor_handle => {
-            return Err(generic_error!("Internal supervisor completed unexpectedly: {:?}", early_result))
-        }
-    }
+    // Assemble the full supervision tree and run it.
+    //
+    // We run our internal supervisor (control plane, environment provider, etc) and our topology supervisor
+    // side-by-side, which means everyone has access to the same dataspace. This is crucial for allowing processes
+    // in the topology supervisor to (eventually)
```
webern: Sentence fragment.

webern (approved):

bin/agent-data-plane/src/cli/run.rs
tobz: Good catch, will fix.

tobz: Fixed in 743f7d496c.

lib/saluki-core/src/health/mod.rs
```diff
@@ -374,6 +374,48 @@ impl HealthRegistry {
...
+    /// Waits until all currently registered components whose name matches `predicate` are ready.
+    ///
+    /// This is a scoped variant of [`all_ready`][Self::all_ready] that only considers components whose name satisfies
+    /// the given predicate, which is useful for waiting on a specific subsystem's components to become ready without
+    /// waiting on every other component in the registry.
+    ///
+    /// If no registered component matches the predicate, the method returns immediately. As with
+    /// [`all_ready`][Self::all_ready], components can be registered while this method is waiting, so callers should
+    /// ensure all components they care about have been registered before calling this method.
+    pub async fn all_ready_matching<F>(&self, predicate: F)
```
thieman, line 367: Would it make sense to either remove `all_ready` or swap out its impl for a call to this with an always-true predicate?

tobz: Ah yeah, that's something I clearly missed during review/cleanup. Will fix.

thieman (approved):

lib/saluki-core/src/health/mod.rs
tobz: Fixed in b32f188ac3.

PR #1819, opened 2026-06-04
feat(agent-data-plane): added workaround for tls handshake timeout config option 

### LINKED ISSUE #178: Allow configuring the TLS handshake timeout for HTTP clients.

lucastemb:
## Summary

Added workaround to allow for `tls_handshake_timeout` config option. `tokio-rustls` provides no built-in timeout, so handshakes could be potentially infinite. The idea here is to replace `hyper-rustls`'s `HttpsConnector` with direct ownership of the TLS layer so that we can wrap it in a timeout with the time specified through the config option. 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests

## References

- Closes: https://github.com/DataDog/saluki/issues/178

### DIFF
--- a/lib/saluki-components/src/common/datadog/config.rs
+++ b/lib/saluki-components/src/common/datadog/config.rs
@@ -25,6 +25,10 @@ const fn default_endpoint_buffer_size() -> usize {
     16
 }
 
+const fn default_tls_handshake_timeout_secs() -> u64 {
+    10
+}
+
 const fn default_forwarder_connection_reset_interval() -> u64 {
     0
 }
@@ -166,6 +170,13 @@ pub struct ForwarderConfiguration {
     #[serde(default = "default_request_timeout_secs", rename = "forwarder_timeout")]
     request_timeout_secs: u64,
 
+    /// TLS handshake timeout, in seconds.
+    ///
+    /// Defaults to 10 seconds. If the TLS handshake does not complete within this duration after the
+    /// TCP connection is established, the connection attempt fails with a timeout error.
+    #[serde(default = "default_tls_handshake_timeout_secs", rename = "tls_handshake_timeout")]
+    tls_handshake_timeout_secs: u64,
+
     /// Maximum number of pending requests for an individual endpoint.
     ///
     /// Defaults to 16.
@@ -251,6 +262,11 @@ impl ForwarderConfiguration {
     /// Returns the request timeout.
     pub const fn request_timeout(&self) -> Duration {
         Duration::from_secs(self.request_timeout_secs)
+    }
+
+    /// Returns the TLS handshake timeout.
+    pub const fn tls_handshake_timeout(&self) -> Duration {
+        Duration::from_secs(self.tls_handshake_timeout_secs)
     }
 
     /// Returns the maximum number of pending requests for an individual endpoint.
--- a/lib/saluki-components/src/common/datadog/config.rs
+++ b/lib/saluki-components/src/common/datadog/config.rs
@@ -1,7 +1,7 @@
 use std::time::Duration;
 
 use facet::Facet;
-use saluki_config::GenericConfiguration;
+use saluki_config::{DurationString, GenericConfiguration};
 use saluki_error::GenericError;
 use saluki_io::net::client::http::{HttpProtocol, TlsMinimumVersion};
 use serde::Deserialize;
@@ -25,8 +25,8 @@ const fn default_endpoint_buffer_size() -> usize {
     16
 }
 
-const fn default_tls_handshake_timeout_secs() -> u64 {
-    10
+const fn default_tls_handshake_timeout() -> DurationString {
+    DurationString::new(Duration::from_secs(10))
 }
 
 const fn default_forwarder_connection_reset_interval() -> u64 {
@@ -170,12 +170,13 @@ pub struct ForwarderConfiguration {
     #[serde(default = "default_request_timeout_secs", rename = "forwarder_timeout")]
     request_timeout_secs: u64,
 
-    /// TLS handshake timeout, in seconds.
+    /// TLS handshake timeout.
     ///
     /// Defaults to 10 seconds. If the TLS handshake does not complete within this duration after the
     /// TCP connection is established, the connection attempt fails with a timeout error.
-    #[serde(default = "default_tls_handshake_timeout_secs", rename = "tls_handshake_timeout")]
-    tls_handshake_timeout_secs: u64,
+    #[serde(default = "default_tls_handshake_timeout", rename = "tls_handshake_timeout")]
+    #[facet(opaque)]
+    tls_handshake_timeout: DurationString,
 
     /// Maximum number of pending requests for an individual endpoint.
     ///
@@ -266,7 +267,7 @@ impl ForwarderConfiguration {
 
     /// Returns the TLS handshake timeout.
     pub const fn tls_handshake_timeout(&self) -> Duration {
-        Duration::from_secs(self.tls_handshake_timeout_secs)
+        self.tls_handshake_timeout.as_duration()
     }
 
     /// Returns the maximum number of pending requests for an individual endpoint.
@@ -488,6 +489,21 @@ mod tests {
 
         let proxies = config.proxy().as_ref().unwrap().build().unwrap();
         assert_eq!(proxies[0].uri().to_string(), PROXY_B_URI);
+    }
+
+    #[tokio::test]
+    async fn tls_handshake_timeout_accepts_duration_string() {
+        let config =
+            forwarder_config_from(config_with(serde_json::json!({ "tls_handshake_timeout": "10s" })), None).await;
+
+        assert_eq!(config.tls_handshake_timeout(), Duration::from_secs(10));
+    }
+
+    #[tokio::test]
+    async fn tls_handshake_timeout_defaults_to_10_seconds() {
+        let config = forwarder_config_from(base_config(), None).await;
+
+        assert_eq!(config.tls_handshake_timeout(), Duration::from_secs(10));
     }
 
     #[tokio::test]
--- a/lib/saluki-components/src/common/datadog/io.rs
+++ b/lib/saluki-components/src/common/datadog/io.rs
@@ -161,6 +161,7 @@ where
         let endpoints = config.build_routable_endpoints(live_config.clone())?;
         let mut client_builder = HttpClient::builder()
             .with_request_timeout(config.request_timeout())
+            .with_tls_handshake_timeout(config.tls_handshake_timeout())
             .with_min_tls_version(config.min_tls_version())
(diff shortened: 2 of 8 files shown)


lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -161,6 +161,7 @@ where
         let endpoints = config.build_routable_endpoints(live_config.clone())?;
         let mut client_builder = HttpClient::builder()
             .with_request_timeout(config.request_timeout())
+            .with_tls_handshake_timeout(config.tls_handshake_timeout())
```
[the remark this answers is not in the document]
lucastemb: Worth a follow up, since it would involve making a PR against the `hyper-http-proxy` fork.

lib/saluki-components/src/config_registry/classifier.rs
```diff
@@ -112,12 +112,9 @@ mod tests {
     }
 
     #[test]
-    // To whoever implements this config in the future: sorry! We just wanted to make sure this is
-    // working correctly by giving it a currently unsupported key. You can delete the unsupported
-    // tests or choose a different key.
```
jszwedko, line 117: I'd leave this comment since `AGGREGATOR_BUFFER_SIZE` may end up being supported in the future too 😅

lucastemb: Good catch. That's my bad. I think Claude may have taken that snippet out without me noticing when refactoring.

lib/saluki-components/src/common/datadog/config.rs
```diff
@@ -166,6 +170,13 @@ pub struct ForwarderConfiguration {
     #[serde(default = "default_request_timeout_secs", rename = "forwarder_timeout")]
     request_timeout_secs: u64,
 
+    /// TLS handshake timeout, in seconds.
+    ///
+    /// Defaults to 10 seconds. If the TLS handshake does not complete within this duration after the
+    /// TCP connection is established, the connection attempt fails with a timeout error.
+    #[serde(default = "default_tls_handshake_timeout_secs", rename = "tls_handshake_timeout")]
```
jszwedko: Will this parse durations correctly? The config coming from the Core Agent will be a String "Go duration" like `10s`.

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -537,49 +575,22 @@ impl HttpsCapableConnectorBuilder {
...
+            tls_config: Arc::new(tls_config),
+            tls_handshake_timeout,
             bytes_sent: self.bytes_sent,
             error_telemetry: self.error_telemetry,
             conn_age_limit: self.conn_age_limit,
         })
     }
 }
 
-#[cfg(test)]
```
jszwedko, line 558: Did you mean to delete these tests?

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -611,28 +622,44 @@ pub(super) fn check_connection_state(captured_conn: CaptureConnection) {
...
+    fn test_tls_config() -> rustls::ClientConfig {
         rustls::ClientConfig::builder_with_provider(rustls::crypto::aws_lc_rs::default_provider().into())
             .with_safe_default_protocol_versions()
             .expect("AWS-LC default protocol versions should be valid")
             .with_root_certificates(rustls::RootCertStore::empty())
             .with_no_client_auth()
     }
 
-    #[test]
-    fn auto_protocol_advertises_h2_and_http1_alpn() {
```
jszwedko, line 625: Seems like we dropped some tests here too.

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -378,27 +382,51 @@ impl Service<Uri> for HttpsCapableConnector {
     }
 
     fn call(&mut self, dst: Uri) -> Self::Future {
```
jszwedko, line 384: I'd like to get some other eyes from @tobz or @webern given the nuance involved here.

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -378,27 +382,51 @@ impl Service<Uri> for HttpsCapableConnector {
...
+                let server_name =
+                    ServerName::try_from(hostname.to_string()).map_err(|e| BoxError::from(io::Error::other(e)))?;
+
+                let tls_connector = TlsConnector::from(tls_config);
+                let tls_stream = tokio::time::timeout(
+                    tls_handshake_timeout,
+                    tls_connector.connect(server_name, TokioIo::new(transport)),
+                )
+                .await
+                .map_err(|_| BoxError::from(io::Error::new(io::ErrorKind::TimedOut, "TLS handshake timed out")))?
```
jszwedko: **Bug: TLS handshake timeout does not increment `tls_error` telemetry**

When `tokio::time::timeout` fires, the error is mapped and returned via `?` on this line — before the second `.map_err` (line 412–417) that calls `increment_tls_error()`. The telemetry counter therefore only fires when `tls_connector.connect()` itself returns an error, not when the timeout expires.

To fix, increment the counter in the timeout arm as well:
```rust
.map_err(|_| {
    if let Some(error_telemetry) = &error_telemetry {
        error_telemetry.increment_tls_error();
    }
    BoxError::from(io::Error::new(io::ErrorKind::TimedOut, "TLS handshake timed out"))
})?
```

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -537,49 +575,22 @@ impl HttpsCapableConnectorBuilder {
...
-        // Create the HTTPS connector.
-        let https_connector_builder = HttpsConnectorBuilder::new().with_tls_config(tls_config).https_or_http();
-        let https_connector = match self.http_protocol {
-            HttpProtocol::Auto => https_connector_builder
-                .enable_all_versions()
-                .wrap_connector(inner_connector),
-            HttpProtocol::Http1 => https_connector_builder.enable_http1().wrap_connector(inner_connector),
+        tls_config.alpn_protocols = match self.http_protocol {
+            HttpProtocol::Auto => vec![b"h2".to_vec(), b"http/1.1".to_vec()],
+            HttpProtocol::Http1 => vec![],
```
jszwedko: **Bug: `HttpProtocol::Http1` no longer advertises `http/1.1` in TLS ALPN**

Previously `hyper_rustls::enable_http1()` set `alpn_protocols = ["http/1.1"]`. Now `vec![]` means the client sends no ALPN extension at all for HTTP/1.1 connections.

Most servers tolerate this, but servers that strictly require ALPN negotiation (or that use ALPN to distinguish protocol versions) will reject the connection. Consider:

```rust
HttpProtocol::Http1 => vec![b"http/1.1".to_vec()],
```

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -378,27 +382,51 @@ impl Service<Uri> for HttpsCapableConnector {
...
+
+            let stream = if is_https {
+                let host = dst.host().unwrap_or_default();
+                let mut hostname = host;
+                if let Some(trimmed) = hostname.strip_prefix('[').and_then(|h| h.strip_suffix(']')) {
+                    hostname = trimmed;
                 }
-            }
+                let server_name =
+                    ServerName::try_from(hostname.to_string()).map_err(|e| BoxError::from(io::Error::other(e)))?;
```
jszwedko: **Cleanup: unnecessary `String` allocation for server name**

`rustls::pki_types::ServerName` implements `TryFrom<&str>`, so `.to_string()` is not needed:

```rust
ServerName::try_from(hostname).map_err(|e| BoxError::from(io::Error::other(e)))?
```

This avoids a heap allocation on every TLS connection.

lib/saluki-io/src/net/client/http/conn.rs
lucastemb: I did because `configure_tls_alpn_for_http_protocol` was no longer a standalone function after refactoring, but I should've probably modified the tests instead of deleting altogether.

lib/saluki-components/src/common/datadog/config.rs
lucastemb: Good call out. I tried to base it off of some of the other config keys in `ForwarderConfiguration`, but I think the same logic does not apply since things like `forwarder_timeout` are based off just pure seconds. _As a result,_ _I also changed the name to `tls_handshake_timeout` to reflect this_

lib/saluki-components/src/common/datadog/config.rs
```diff
@@ -166,6 +170,14 @@ pub struct ForwarderConfiguration {
     #[serde(default = "default_request_timeout_secs", rename = "forwarder_timeout")]
     request_timeout_secs: u64,
 
+    /// TLS handshake timeout.
+    ///
+    /// Defaults to 10 seconds. If the TLS handshake does not complete within this duration after the
+    /// TCP connection is established, the connection attempt fails with a timeout error.
+    #[serde(default = "default_tls_handshake_timeout", rename = "tls_handshake_timeout")]
```
jszwedko, line 177:
```suggestion
    #[serde(default = "default_tls_handshake_timeout"]
```

I don't think we need the rename now that the field name matches.

lib/saluki-io/src/net/client/http/client.rs
```diff
@@ -171,6 +171,19 @@ impl HttpClientBuilder {
         self
     }
 
+    /// Sets the timeout for completing the TLS handshake after a connection is established.
+    ///
+    /// Defaults to 10 seconds.
+    ///
```
jszwedko, line 177:
```suggestion
```

There is no default for this function that I can see 🤔

lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -643,8 +736,6 @@ mod tests {
     async fn vsock_takes_priority_over_unix_when_both_set() {
         use std::sync::Arc;
 
-        use tower::Service as _;
-
```
jszwedko, line 647: Did you mean to remove this? I think it is actually needed 🤔

lib/saluki-io/Cargo.toml
```diff
@@ -79,5 +79,6 @@ tokio-vsock = { workspace = true }
 http-body-util = { workspace = true }
 proptest = { workspace = true }
 rand_distr = { workspace = true }
+saluki-metrics = { workspace = true, features = ["test"] }
```
jszwedko, line 82: Is this needed? I'm not seeing `saluki-metrics` used in `saluki-io`.

jszwedko (commented):
A few more comments, and I'd still like more review of the changes to `conn.rs`.

jszwedko:
Superseded by https://github.com/DataDog/saluki/pull/2303

Issue #1788, opened 2026-06-01
Centralize Datadog config inventory with build-time metadata support.


webern:
## Problem

Updating a single Datadog config key in ADP currently requires touching several files. There are two
sources of truth with no sync mechanism between them.

Three additional problems compound this:

- The config registry conflates production code with test support, requiring `allow(dead_code)`,
  `cfg(test)`, and `feature = "config-test-support"` guards throughout (see #1623).
- All config metadata (which keys are supported, which pipelines they affect, what they're used for)
  is encoded in hand-written Rust structs and is only visible at runtime.
- The ledger and documentation exist as separate artifacts with no generation step.

## Proposed Solution

Replace all existing per-key annotation sources with a single hand-maintained `schema_overlay.yaml`,
joined at build time with the vendored Datadog schema. The overlay will be the single source of
truth for Datadog config coverage, documentation, runtime classification, and smoke-test support.

`lib/datadog-agent-commons/build.rs` will read both files and:

- Validate that every key in the vendored schema appears exactly once in the overlay.
- Generate config registry annotation constants, replacing the hand-written `datadog/*.rs` files.
- Generate the documentation markdown.
- Generate a `DatadogConfiguration` struct (via typify or a custom generator, depending on the
  schema language).
- Generate a "witness trait" (`DatadogConfigConsumer`) trait with one method per supported key, plus
  a `drive` function — forcing a human to explicitly handle every supported key at compile time.

A hand-written `Translator` will implement the trait, converting `DatadogConfiguration` into the
internal topology and dataplane config structs. Adding a key to the overlay without updating the
translator will result in a compilation error. Removing a key leaves an orphan impl method — also a
compilation error.

The audit ledger will be deleted. Documentation will be generated. The registry annotation files
will be deleted at the end of the process when the smoke tests switch over to using the new code.

## Layout

```
lib/datadog-agent-commons/config/schema.yaml          # vendored, untouched
lib/datadog-agent-commons/config/schema_overlay.yaml  # hand-written, single source of truth
lib/datadog-agent-schema-model/                       # types for deserializing the overlay
```

## Anticipated PRs

If everything goes according to plan, then the following sequence makes sense to me. If it goes off
the rails, all bets are off!

- [x] (small): New `datadog-agent-schema-model` crate: Rust types for deserializing
  `schema_overlay.yaml`
- [x] (small): Infra scaffold: move schema to `datadog-agent-commons/config/`, add `build.rs` that
[...]

PR #1828, opened 2026-06-05
chore(config): consolidate config maintenance to `schema_overlay.yaml`

### LINKED ISSUE #1623: Separate prod from test/build concerns in the config registry.
The config registry started its life for the purpose of create the config smoke tests. It exported test support symbols in its lib.rs for this. Then in #1612 if grew to support a production runtime check in ADP.

So we have two types of symbols:
- pub for test
- pub for agent-data-plane

`cfg(test)` doesn't work across crate boundaries. It's only for compiling stuff in and out when the *current* crate is under test.

The conflation of these two things is evidenced in the config registry by the presence of
- `allow(dead_code)`
- `cfg(test)`
- `feature=config-test-support`

In various places.

Ideally we would cleanly separate these two use cases so that there can be a small interface for agent-data-plane and a different interface for testing scenarios.

Or perhaps a more extensive redesign. But for now this issue could be considered done if we didn't have a whole bunch of compile guards or dead_code allowances in the config registry.

webern:
## Human Summary

This PR consolidates the ledger files and the config registry into a single source of truth, *the
overlay* (`schema_overlay.yaml`). It generates the `config_registry`, moving it to
`datadog/agent/config-testsupport`. The smoke tests remain and pass, but edits to the config
registry now flow through the overlay file.

Code for the runtime classification of unsupported configuration for ADP is generated separately by
`datadog-agent/config`. This allows for a slimmer and purpose-built structure for the production
binary that does not have all of the test support metadata in it.

The documentation page that we have been maintaining by hand, `dogstatsd.md`, is now also generated.
I went to great lengths to preserve its character as much as possible, though it had drifted from
reality in some cases. Order of the main sections was preserved, but order of the prose edits was
not. Unfortunately, this is about as much as I could reduce the diff. Going forward, it will diff
mechanically with edits to the overlay.

I handled the consolidation and conflict resolution of our data with painstaking care. The config
registry was considered more authoratative than documentation or the ledger. I have been
incorporating recent changes to the data from recent PRs, but we should be vigilent against PRs that
are landing while this one is open.

A git check has been added to CI to catch in-tree code gen diffs that haven't been checked in.

### Reviewer's Guide

Worth looking at:
- `schema_overlay.yaml` to see its shape. `datadog-agent/config-overlay-model/lib.rs` restates the shape in Rust.
- `classifier_data.rs`: this is the actual data we check against at runtime to either warn or exit. It's being generated from the overlay.

Worth skipping:
- Claude skills, fallout in the `/confkey` skill, added a new skill for the model to look at when it's trying to understand configuration concerns and code gen, etc. referenced with one or two lines from AGENTS.md
- Smoke test invocations got another argument and fmt'ed due to stuff moving around.
- I put a couple of crates under `datadog-agent` which means `datadog-agent-commons` was moved to `datadog-agent/commons`
- `doc_gen.rs`, `registry_gen.rs` and `classifier_gen.rs` are ugly AI-written code generators, skip them. Their generating code in-tree so we see their outputs in the diff.

## AI Summary

Consolidate ADP's config metadata into a single `schema_overlay.yaml` and generate the config
registry + documentation from it at build time.

**Before:** Config metadata lived in multiple files (`ignored_keys.yaml`, `known-configs.json`,
`known-configs-not-applicable.json`, per-key annotation structs in `config_registry/datadog/*.rs`,
and `dogstatsd.md`). Updating a single key required touching up to five files with no
synchronization between them.

**After:** One hand-maintained YAML file (`schema_overlay.yaml`) is the single source of truth.
`build.rs` enforces completeness (every `core_schema.yaml` key must appear in the overlay) and
generates:
- The runtime classifier dataset (yaml_path, aliases, support_level, pipeline_affinity, default)
- The test-support annotation constants
- The configuration documentation (`dogstatsd.md`)

### New crates

- **`datadog-agent/config-overlay-model`** — Leaf crate with Rust types for deserializing and
  validating `schema_overlay.yaml`. Enforces structural invariants at load time.
- **`datadog-agent/config`** — Prod crate. Owns the overlay file, `core_schema.yaml`, and a
  `build.rs` that generates the slim classifier array.
- **`datadog-agent/config-testsupport`** — Test crate. Generates full annotation constants, the
  smoke test runner, and the configuration documentation from the same overlay.

### What was deleted

- `docs/agent-data-plane/configuration/dogstatsd/known-configs.json`
- `docs/agent-data-plane/configuration/dogstatsd/known-configs-not-applicable.json`
- Hand-written annotation structs in `config_registry/datadog/*.rs` (replaced by generated code)
- `saluki-components/src/config_registry/generated/` directory

### Key design decisions

- The overlay has four sections: `supported`, `unsupported`, `investigate`, `ignored`.
- `unsupported` entries carry mandatory `severity` (low/medium/high) for runtime warn/error behavior
  and mandatory `planned` (bool) for documentation rendering.
- `investigate` is a parking lot for keys not yet classified with optional severity (when present,
  flows through classifier as `Incompatible`).
- Documentation is generated from a `.tmpl` template with in-place output for clean PR diffs.
- `saluki-components` no longer has a `config_registry` module. Code that previously used
  `crate::config_registry::*` now imports from `datadog_agent_config_testsupport::config_registry`
  directly.

## Change Type

- [x] New feature

## How did you test this PR?

- A lot of painstaking work with the data by hand.
- Kept the smoke tests alive and they pass.
- Unit tests pass.
- I will fix any CI integration or correctness failures.
- Added a diff check after build/test in CI to make sure in-tree generated files don't have a diff.

I conducted an LLM audit of what keys emit a warning or error on main and on this PR which resulted in the following:

| Key                                                | Baseline `2ed5a5` | Changeset `acf87a` |
|----------------------------------------------------|-------------------|--------------------|
| `autoscaling.failover.enabled`                     | Warn              | Warn               |
| `autoscaling.failover.metrics`                     | Warn              | Warn               |
| `dogstatsd_disable_verbose_logs`                   | Warn              | Warn               |
| `dogstatsd_experimental_http.enabled`              | Exit              | Exit               |
| `dogstatsd_experimental_http.listen_address`       | Exit              | Exit               |
| `dogstatsd_metrics_stats_enable`                   | Warn              | Warn               |
| `dogstatsd_pipe_name`                              | Warn              | Warn               |
| `dogstatsd_stats_buffer`                           | Warn              | Warn               |
| `dogstatsd_stats_enable`                           | Warn              | Warn               |
| `dogstatsd_stats_port`                             | Warn              | Warn               |
| `dogstatsd_telemetry_enabled_listener_id`          | Warn              | Warn               |
| `dogstatsd_windows_pipe_security_descriptor`       | Warn              | Warn               |
| `forwarder_apikey_validation_interval`             | Warn              | Warn               |
| `forwarder_flush_to_disk_mem_ratio`                | Warn              | Warn               |
| `forwarder_low_prio_buffer_size`                   | Warn              | Warn               |
| `forwarder_max_concurrent_requests`                | Warn              | Warn               |
| `forwarder_retry_queue_capacity_time_interval_sec` | Warn              | Warn               |
| `heroku_dyno`                                      | Exit              | Exit               |
| `min_tls_version` (env: `DD_MIN_TLS_VERSION`)      | Warn              | Warn               |
| `sslkeylogfile`                                    | Warn              | Warn               |
| `tls_handshake_timeout`                            | Warn              | Warn               |

Exact match between main and this PR for observable behavior in `agent-data-plane`.

## References

- Closes #1623
- Progresses: #1788


lib/datadog-agent/config/build/classifier_gen.rs
```diff
@@ -0,0 +1,121 @@
...
+            Some(Severity::High) => "Severity::High",
+            None => continue,
+        };
+        let default_lit = schema_default_literal(yaml_path, schema_map);
+
+        writeln!(out, "    ClassifierEntry {{").unwrap();
+        writeln!(out, "        yaml_path: \"{}\",", yaml_path).unwrap();
+        writeln!(out, "        aliases: &[],").unwrap();
+        writeln!(out, "        support_level: SupportLevel::Incompatible({}),", severity).unwrap();
+        writeln!(out, "        pipeline_affinity: PipelineAffinity::CrossCutting,").unwrap();
```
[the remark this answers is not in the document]
webern: I'm going to say this is fine since I don't want to keep tweaking the data model forever. Basically I added `investigate` as a section when I realized that sometimes the reality is we don't quite know whether something needs to be supported or not. Then I realized that sometimes we are warning on things that we need to "investigate" so I added the ability to add a severity, but made it optional.

I think if we know enough to say what the severity is, and we really need to assign specific pipelines, we should probably move it up to the `unsupported` sections and leave `investigate` for the keys we know less about.

Not a bad AI catch, though.

Edit: changed the data model since this comment but both the original comment and my response are still relevant with the new data model.

.gitlab/test.yml
```diff
@@ -3,6 +3,7 @@ unit-tests-linux-amd64:
   stage: test
   script:
     - make test
+    - git diff --exit-code && test -z "$(git status --porcelain)" || (echo "Working tree is dirty after build+test. Generated files may be stale — rebuild locally and commit."; git status --short; exit 1)
```
tobz: This strikes me as something that should be a separate job with a name indicative of what we're actually checking for, which would then at least hint at how to "rebuild" the necessary files locally.

Cargo.toml
```diff
@@ -6,7 +6,10 @@ members = [
   "bin/correctness/millstone",
   "bin/correctness/panoramic",
   "bin/correctness/stele",
-  "lib/datadog-agent-commons",
+  "lib/datadog-agent/commons",
+  "lib/datadog-agent/config",
+  "lib/datadog-agent/config-overlay-model",
+  "lib/datadog-agent/config-testsupport",
```
tobz: The nittiest of nits: could we just having `testing` instead of `testsupport`? `testsupport` breaks the pattern of having hyphen-separated words in crate names, and is just a little simpler, I think. 😅

lib/datadog-agent/config-testing/build.rs
```diff
@@ -0,0 +1,56 @@
...
+
+fn main() {
+    let manifest_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    let doc_dir = manifest_dir
+        .join("..")
+        .join("..")
+        .join("..")
+        .join("docs")
+        .join("agent-data-plane")
+        .join("configuration");
```
tobz, line 29: [This](https://stackoverflow.com/a/74942075) might be a better way to calculate this while being resistant to the structure of where _this_ crate/build script live:

```rust
fn workspace_dir() -> PathBuf {
    let output = std::process::Command::new(env!("CARGO"))
        .arg("locate-project")
        .arg("--workspace")
        .arg("--message-format=plain")
        .output()
        .unwrap()
        .stdout;
    let cargo_path = Path::new(std::str::from_utf8(&output).unwrap().trim());
    cargo_path.parent().unwrap().to_path_buf()
}
```

tobz (approved):
Overall, looks good to me but left some take-it-or-leave-it-but-please-take-it nits. 😅

PR #1841, opened 2026-06-08
feat(datadog): validate API keys at runtime to track forwarder health

### LINKED ISSUE #1357: Support refreshing API keys like `forwarder_apikey_validation_interval` at runtime.
The core agent supports runtime API key refresh via [`forwarder_apikey_validation_interval`][1] (default 60 minutes). A [`forwarderHealth`][2] component periodically calls `v1/validate` for each API key/domain pair, and the forwarder watches for API key changes at runtime via config update callbacks.

This may be broader than a single config key — it touches runtime config refresh, validation endpoint calls, and propagating key changes to all components that hold a cached copy.

ADP has partial infrastructure on the config side (listeners for API key changes), but the forwarder does not act on updates. The API key is read at startup and cached in [`ResolvedEndpoint`][3]. There is no validation endpoint call, no periodic refresh, and no hot-reload path from config change to forwarder reconnection.

Document as a known caveat. Operators who rotate API keys will need to restart ADP until this is implemented.

[1]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/pkg/config/setup/common_settings.go#L1521
[2]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/comp/forwarder/defaultforwarder/forwarder_health.go
[3]: https://github.com/DataDog/saluki/blob/dbb968c2db21/lib/saluki-components/src/destinations/datadog/io.rs#L409

lucastemb:
## Summary

ADP now supports runtime API key refresh through a configurable `u64` key (which defaults to `60 minutes`).

Things to note: 

- Additional endpoints added during runtime will not be accounted for since it would involve dynamically altering the topology by creating a new forwarder and related components to support the new endpoint. This _also_ mimics Core Agent behavior.
- Non-zero integer values fall back to the default (`60`)
- Fake API key is not treated as invalid
- Forwarders' health now dictated by whether at least **one** API-key is valid. 
- `200` is the valid status code 
- `403` is invalid
- ADP will treat unexpected status codes as errors and not invalid keys. 
- Integration tests will now use the fake API key (32-char string of all 0s)

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests

## References

- Closes https://github.com/DataDog/saluki/issues/1357


lib/saluki-components/src/common/datadog/config.rs
```diff
@@ -388,6 +414,11 @@ impl ForwarderConfiguration {
     pub const fn allow_arbitrary_tags(&self) -> bool {
         self.allow_arbitrary_tags
     }
+
+    /// Returns the API key validation interval.
+    pub const fn api_key_validation_interval(&self) -> Duration {
+        Duration::from_secs(self.api_key_validation_interval_mins as u64 * 60)
```
tobz: Just use `Duration::from_mins`?

lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -230,6 +230,23 @@ where
...
+        Option<GenericConfiguration>,
+        Duration,
+    ) {
+        (
+            self.endpoints.clone(),
+            self.client.clone(),
+            self.live_config.clone(),
+            self.config.api_key_validation_interval(),
+        )
+    }
```
tobz, line 247: Can we actually just create a type under the `validation` module that holds all of these values, and can be used to spawn the underlying task that runs the validation loop?

Like `ApiKeyValidator`, and you call `ApiKeyValidator::new` with these pieces, hand back the caller the `ApiKeyValidator`, and then they call `validator.spawn()` and it gives them back the join handle and the receiver.

Basically, it just feels like we have a lot of the finer points of running this validator task spread out across various modules when we can constrain it a lot more.

lib/saluki-components/src/forwarders/datadog/mod.rs
```diff
@@ -139,15 +141,39 @@ impl Forwarder for Datadog {
...
+                        None => std::future::pending().await,
+                    }
+                } => match maybe_readiness {
+                    Some(ValidationReadiness::Ready) => health.mark_ready(),
+                    Some(ValidationReadiness::NotReady) => health.mark_not_ready(),
+                    None => {
+                        validation_readiness_rx = None;
+                        debug!("Datadog API key validation task stopped.");
+                    },
+                },
```
tobz: Can we actually add a handle type under `validation` that wraps the receiver and exposes a single async method that does this for us?

```rust
pub async fn wait_for_change(&mut self) -> ValidationReadiness {
    match self.readiness_rx {
        None => pending().await,
        Some(rx) => { ... },
    }
}
```

That sort of thing.

tobz (changes requested):
Overall, this is seeming reasonable to me, but I want to focus on a few things:

- update the PR description to actually reflect what _this_ PR is doing: we already support API key refresh, we're just adding _validation_ behavior now which properly drives readiness of the forwarder component
- dig deeper on whether or not we really need to accept negative values for `forwarder_apikey_validation_interval`, and if we can't simply use `u64` and let the normal deserialization behavior emit an error when we get a negative value (and link to the Agent codebase to justify why we need to keep `i64` otherwise)
- clean up/encapsulate the logic a little more

lucastemb:
> Overall, this is seeming reasonable to me, but I want to focus on a few things:
> 
> * update the PR description to actually reflect what _this_ PR is doing: we already support API key refresh, we're just adding _validation_ behavior now which properly drives readiness of the forwarder component
> * dig deeper on whether or not we really need to accept negative values for `forwarder_apikey_validation_interval`, and if we can't simply use `u64` and let the normal deserialization behavior emit an error when we get a negative value (and link to the Agent codebase to justify why we need to keep `i64` otherwise)
> * clean up/encapsulate the logic a little more

@tobz Done. Also, we don't really _need_ to accept negative values, but the Agent falls back to the default for non-positive integers, so we would need to keep `i64` to recreate this behavior. ([Source](https://github.com/DataDog/datadog-agent/blob/main/comp/forwarder/defaultforwarder/impl/default_forwarder.go#L190-L197))

docs/agent-data-plane/configuration/dogstatsd.md
```diff
@@ -121,6 +121,25 @@ rather than by a transaction count. ADP still preserves the same broad priority
 transactions are processed through the high-priority queue first, retries use the low-priority retry
 queue, and high-priority overflow is moved to the low-priority retry queue.
 
+### Forwarder API key validation (`forwarder_apikey_validation_interval`)
```
webern: This documentation will be moved to schema_overlay.yaml where `support: partial` will get it added to this section of the document.

lib/saluki-components/src/config_registry/datadog/forwarder.rs
```diff
@@ -176,6 +176,18 @@ crate::declare_annotations! {
         pipeline_affinity: PipelineAffinity::CrossCutting,
     };
 
+    /// `forwarder_apikey_validation_interval`—API key validation interval in minutes.
+    FORWARDER_APIKEY_VALIDATION_INTERVAL = SalukiAnnotation {
```
webern: This change and the unsupported.rs change will flow from `schema_overlay.yaml`.

webern (commented):
You can try these instructions, I think it may work:

```
Use the .claude/skills/config-management/SKILL.md skill to understand
the new config system. We need to rebase over the latest main in which
several of our edited files either no longer exist or became generated
files. Now all config registry and documentation changes flow from
schema_overlay.yaml

During the rebase, abandon our changes to the now-generated or deleted
files and move our documentation/support-level edits into
schema_overlay.yaml
```
Edit... uhh well, you would have to rebase to have that skill in the first place, so maybe this instead:

```
Look at the latest commit on upstream main and find this skill. Read
it .claude/skills/config-management/SKILL.md skill to understand the new
config system.

We need to rebase over the latest main in which several of our edited
files either no longer exist or became generated files. Now all config
registry and documentation changes flow from schema_overlay.yaml

During the rebase, abandon our changes to the now-generated or deleted
files and move our documentation/support-level edits into
schema_overlay.yaml
```

lib/saluki-components/src/config_registry/datadog/forwarder.rs
lucastemb: Niceee, the new `make build-schema-overlay`command is pretty neat 🙌

lib/saluki-components/src/common/datadog/config.rs
```diff
@@ -246,6 +254,15 @@ pub struct ForwarderConfiguration {
     /// outbound intake request. The data plane does not perform local tag validation based on this setting.
     #[serde(default)]
     allow_arbitrary_tags: bool,
+
+    /// API key validation interval, in minutes.
+    ///
+    /// Defaults to 60 minutes. Invalid `0` values fall back to 60 minutes.
```
tobz:
```suggestion
    /// API key validation interval, in minutes.
    ///
    /// All values that are less than or equal to zero will be ignored, and the default
    /// value will be used.
    ///
    /// Defaults to 60 minutes.
```

lib/saluki-components/src/common/datadog/validation.rs
```diff
@@ -0,0 +1,626 @@
...
+    for target in targets {
+        match validate_target(client, target).await {
+            KeyValidationResult::Valid => return ValidationReadiness::Ready,
+            KeyValidationResult::Invalid => {}
+            KeyValidationResult::Error => saw_error = true,
+        }
+    }
+
+    if saw_error {
+        ValidationReadiness::Ready
```
tobz, line 229: Do we actually want to indicate being ready if we only saw errors? (This code doesn't actually differentiate between "some errors" and "only errors", but I'm curious about the intended behavior.)

lib/saluki-components/src/common/datadog/validation.rs
```diff
@@ -0,0 +1,626 @@
...
+        ValidationReadiness::NotReady
+    }
+}
+
+fn collect_validation_targets(endpoints: &mut [RoutableEndpoint]) -> Vec<ValidationTarget> {
+    let mut seen = HashSet::new();
+    let mut targets = Vec::new();
+
+    for routable in endpoints {
+        let endpoint = routable.endpoint_mut();
```
tobz, line 241: Why are we mutably borrowing the endpoint? 🤔

lib/saluki-components/src/common/datadog/validation.rs
```diff
@@ -0,0 +1,626 @@
...
+use tracing::{debug, warn};
+use url::Url;
+
+use super::endpoints::RoutableEndpoint;
+
+const VALIDATE_PATH: &str = "/api/v1/validate";
+const FAKE_API_KEY: &str = "00000000000000000000000000000000";
+
+static DATADOG_API_DOMAIN_RE: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"([a-z]{2,}\d{1,2}\.)?(datadoghq\.[a-z]+|ddog-gov\.com)\.?$").unwrap());
```
tobz, line 27: Can we add a TODO comment here about moving this to `datadog-agent-commons` in the future?

This is the sort of thing that we use in multiple places and should really be shared in a cleaner way.

lib/saluki-config/src/lib.rs
```diff
@@ -297,6 +297,12 @@ impl ConfigurationLoader {
                 .push(ProviderSource::Static(ArcProvider(Arc::new(Serialized::defaults(
                     default_dict.clone(),
                 )))));
+            if let Some(string_values) = string_preserved_environment_values(&prefix) {
+                self.provider_sources
+                    .push(ProviderSource::Static(ArcProvider(Arc::new(Serialized::defaults(
+                        string_values,
+                    )))));
+            }
```
tobz: So I understand, conceptually, why we care about doing this -- Figment will parse the fake API key as an integer otherwise -- but this is too specific of a change to make in `saluki-config`.

What we should do instead (IMO) is: change over any integration tests that run via standalone mode to... not enable standalone mode. With that, ADP will always get the config from the Core Agent, which will send over the API key as a string.

tobz (changes requested):

lib/saluki-components/src/common/datadog/validation.rs
lucastemb: Yes. This mimics Core Agent behavior. The rationale is validation transport errors, timeouts, and unexpected statuses are inconclusive. It doesn't necessarily on its own prove the key is bad. We only mark forwarder as `NotReady` if all validation targets were checked and they were **all** invalid

lib/saluki-components/src/common/datadog/validation.rs
lucastemb: `ResolvedEndpoint::api_key()` refreshes the cached API key from live config before returning it, and since that can also update`self.api_key`, that endpoint needs to be borrowed mutably.

tobz: Ahhhh yes, right. 👍🏻

tobz (approved):

PR #1843, opened 2026-06-08
chore(ci): publish agent-data-plane darwin tarballs

thieman:
## Human Summary

Adds build and publish jobs for darwin tarballs, storing them in the same S3 bucket as existing Linux artifacts. On dev pipelines, these are manually triggered and use `release` profile. On release pipelines, the build is run automatically with `optimized-release` and publish is run manually.

## What

Adds darwin (`amd64` + `arm64`) variants of the existing `push-release-tarball-linux-*` flow,
so we publish a darwin tarball next to the linux ones at `s3://binaries-ddbuild-io-prod/saluki/`.

## Why

Today ADP only ships on Linux. We want to extend the Datadog Agent's macOS build
([datadog-agent#omnibus/config/software/datadog-agent-data-plane.rb](https://github.com/DataDog/datadog-agent/blob/main/omnibus/config/software/datadog-agent-data-plane.rb))
to include ADP, which means the agent's omnibus needs a downloadable darwin
tarball with the same layout as the linux one. This PR provides exactly that.

## How

Two-stage flow that parallels the linux one at a high level (linux: `build-adp-image-*`
always produces a docker image artifact, the tagged release job re-packs it into a tarball
and uploads to s3; darwin: `build-release-tarball-darwin-*` always produces the tarball
directly as a GitLab artifact, the tagged release job uploads it to s3).

### `build-release-tarball-darwin-{amd64,arm64}` (build stage, `.gitlab/build.yml`)

Runs on the macOS runner. Builds `agent-data-plane` from source via `make build-adp-host`
(host-arch cargo build), assembles a tarball with the same layout as the linux artifact, and
exposes the tarball as a GitLab job artifact (1-week retention).

- Auto-runs on tagged release pipelines so the artifact is ready when a maintainer clicks
  the manual publish job in the release stage.
- Manual + `allow_failure: true` on dev pipelines, so anyone can validate a PR's darwin
  tarball end-to-end with one click — without burning macOS runner capacity by default and
  without ever touching s3.

Test gating (`unit-tests-macos-*`, `miri`, `test-integration-macos-*`, `check-deny`,
`check-licenses`) is attached here, guarding both the dev-pipeline manual click and the
tagged-release auto-run.

### `push-release-tarball-darwin-{amd64,arm64}` (release stage, `.gitlab/release.yml`)

Consumes the build job's tarball artifact via `needs: [{ artifacts: true }]`, runs in the
same `${DOCKER_BUILD_IMAGE}` Linux container as `.push-release-tarball-definition`, and does
nothing more than `aws s3 cp` + `sha256sum`. Manual on tag only.

This means the s3 upload uses the same IAM-role credential flow that's already known to work
for the existing linux push jobs (no need to verify whether IAM creds are reachable from
inside the macOS Tart VM — the upload doesn't run there).

### Why mac-only requires this split

The linux release tarball is a thin re-pack of an already-published docker image (built by
`build-adp-image-*` on every pipeline). For darwin there's no equivalent docker image to
re-pack from, so we have to build the binary somewhere — the macOS runner is the only place
that can. By making that build a separate, artifact-producing stage, we get free dev-pipeline
validation without S3 side effects.

### Tarball layout

Matches the linux artifact bit-for-bit:

```
opt/datadog-agent/embedded/bin/agent-data-plane
opt/datadog/agent-data-plane/NOTICE
opt/datadog/agent-data-plane/LICENSE
opt/datadog/agent-data-plane/LICENSE-3rdparty.csv
opt/datadog/agent-data-plane/LICENSES/THIRD-PARTY-*
```

Per-license `THIRD-PARTY-*` files are reproduced inline (replicating the `license-builder`
Docker stage from `docker/Dockerfile.agent-data-plane`), pinned to the same SPDX
`license-list-data` version (`3.28.0`).

### Profile parity with linux

`make build-adp-host` is parameterized to honor `BUILD_PROFILE` so the darwin tarball gets
`optimized-release` on tagged release pipelines and `release` on dev pipelines, matching the
workflow rules in `.gitlab-ci.yml` that drive the linux Docker build's profile selection.
The default remains `release` to preserve historical behavior for the integration-test jobs
that already use this target.

### Out of scope

- **FIPS-on-darwin**: no FIPS variant of the darwin build exists upstream and the consumer
  (datadog-agent omnibus) does not request one.

## Testing

Locally verified:

- `make build-adp-host` still works at default profile (`release`) and honors
  `BUILD_PROFILE=optimized-release` env override (`make -n` shows the right cargo invocation
  in both cases).
- The license pipeline (the `tail | awk | awk | grep | sed | sort | uniq | xargs cp` line)
  produces 15 distinct SPDX license names against the current `LICENSE-3rdparty.csv`, all of
  which exist as `text/<name>.txt` in `license-list-data` v3.28.0.
- Both `.gitlab/release.yml` and `.gitlab/build.yml` parse cleanly with all `!reference` tags
  intact.
- Pre-commit hooks (fmt, doc) passed on both commits.

End-to-end CI validation will happen when the PR pipeline runs and a reviewer clicks the
manual `build-release-tarball-darwin-{amd64,arm64}` jobs to download and inspect the
artifacts. No S3 writes occur until tag time + a maintainer's manual click.

### DIFF
--- a/.gitlab/release.yml
+++ b/.gitlab/release.yml
@@ -132,14 +132,19 @@ push-release-tarball-linux-arm64-fips:
     - aws s3 cp ${TARBALL_NAME} ${TARGET_BUCKET_PATH}
     - sha256sum ${TARBALL_NAME}
 
+# Test gating mirrors `.push-release-tarball-definition` (linux): unit + miri + check-deny +
+# check-licenses + integration tests, all attached to the publish job rather than the build job
+# so the build can run in parallel with the test stage. Plus the build-stage tarball artifact
+# itself, pulled in via `artifacts: true`.
 push-release-tarball-darwin-amd64:
   extends: [.build-common-variables, .push-release-tarball-darwin-definition]
-  # Integration tests live in the e2e stage and can't gate the build-stage `build-release-
-  # tarball-darwin-amd64`, so they gate the publish step here instead. This mirrors how the linux
-  # push job gates on the e2e-stage `test-correctness-dsd-*` jobs.
   needs:
     - job: build-release-tarball-darwin-amd64
       artifacts: true
+    - unit-tests-macos-amd64
+    - unit-tests-miri-macos-amd64
+    - check-deny
+    - check-licenses
     - test-integration-macos-amd64
   variables:
     TARGET_ARCH: "amd64"
@@ -149,6 +154,10 @@ push-release-tarball-darwin-arm64:
   needs:
     - job: build-release-tarball-darwin-arm64
       artifacts: true
+    - unit-tests-macos-arm64
+    - unit-tests-miri-macos-arm64
+    - check-deny
+    - check-licenses
     - test-integration-macos-arm64
   variables:
     TARGET_ARCH: "arm64"
--- a/.gitlab/release.yml
+++ b/.gitlab/release.yml
@@ -106,3 +106,79 @@ push-release-tarball-linux-arm64-fips:
     TARGET_ARCH: "arm64"
     SOURCE_IMAGE: ${ADP_FULL_IMAGE_TAG_FIPS}
     SOURCE_IMAGE_VERSION: ${ADP_IMAGE_VERSION_FIPS}
+
+# Mirrors `.push-release-tarball-definition` but builds `agent-data-plane` natively on a macOS
+# runner instead of extracting it from a Linux container image. The output tarball layout
+# (opt/datadog-agent/embedded/bin/agent-data-plane plus opt/datadog/agent-data-plane/{LICENSES,*})
+# matches the Linux artifact bit-for-bit so downstream consumers (the Datadog Agent omnibus build,
+# see omnibus/config/software/datadog-agent-data-plane.rb in DataDog/datadog-agent) can use the
+# same software definition for both platforms.
+#
+# Differences from the Linux template:
+#   - No `image:` — macOS jobs run natively on the runner host (Tart VM on arm64,
+#     bare-metal on amd64), not inside a container.
+#   - `awscli` is not preinstalled on macOS runners, so we install it via Homebrew on demand.
+#   - The binary is built from source via `make build-adp-host` (host-arch cargo build) instead
+#     of `crane export`-ing it from a prebuilt Linux image. BUILD_PROFILE flows through to cargo
+#     so tagged releases get `optimized-release` and dev pipelines get `release`, matching the
+#     Linux Docker build's profile selection driven by the workflow rules in .gitlab-ci.yml.
+#   - The per-license THIRD-PARTY-* files are produced inline (replicating the `license-builder`
+#     stage from docker/Dockerfile.agent-data-plane) since we have no Docker stage to copy from.
+#   - FIPS is not in scope for darwin in v1 (no FIPS-on-darwin variant exists upstream).
+.push-release-tarball-darwin-definition:
+  stage: release
+  rules:
+    - if: !reference [.on_official_release, rules, if]
+      when: manual
+  variables:
+    TARGET_BUCKET_PATH: "s3://binaries-ddbuild-io-prod/saluki/"
+    TARGET_OS: "darwin"
+    # Aliased so the script body parallels `.push-release-tarball-definition`. The Linux job
+    # picks this up from each concrete job's variables (sourced from the image tag); on darwin
+    # there is no image, so the canonical pipeline version variable is used directly.
+    SOURCE_IMAGE_VERSION: ${ADP_IMAGE_VERSION}
+    # Pinned to match docker/Dockerfile.agent-data-plane so the tarball ships the same set of
+    # THIRD-PARTY-* license texts as the Linux artifact. Bump in lockstep with the Dockerfile.
+    SPDX_LICENSES_VERSION: "3.28.0"
+  script:
+    - command -v aws >/dev/null 2>&1 || brew install awscli
+    - mkdir -p /tmp/tarball/opt/datadog-agent/embedded/bin /tmp/tarball/opt/datadog/agent-data-plane/LICENSES
+    # Build natively for the host (uname-derived target). BUILD_PROFILE comes from the workflow
+    # rules in .gitlab-ci.yml: `release` for dev pipelines, `optimized-release` for tagged
+    # releases. The output path matches the profile name (cargo target/${BUILD_PROFILE}/...).
+    - make build-adp-host
+    - cp target/${BUILD_PROFILE}/agent-data-plane /tmp/tarball/opt/datadog-agent/embedded/bin/
+    # Replicate the `license-builder` stage from docker/Dockerfile.agent-data-plane on the host
+    # so the tarball ships the same per-license THIRD-PARTY-* files as the Linux artifact.
+    - curl -sSfL -o /tmp/spdx.tar.gz https://github.com/spdx/license-list-data/archive/refs/tags/v${SPDX_LICENSES_VERSION}.tar.gz
+    - tar -C /tmp -xzf /tmp/spdx.tar.gz
+    # Pipeline below mirrors the `license-builder` stage in docker/Dockerfile.agent-data-plane verbatim
+    # (single-occurrence sed paren strip and all) so the resulting THIRD-PARTY-* set is identical.
+    - tail -n +2 LICENSE-3rdparty.csv | awk -F ',' '{print $3}' | awk -F' ' '{for(i=1;i<=NF;i++) print $i}' | grep -v -E "(OR|AND|WITH|Custom|LLVM-exception)" | sed s/[\(\)]// | sort | uniq | xargs -I {} cp /tmp/license-list-data-${SPDX_LICENSES_VERSION}/text/{}.txt /tmp/tarball/opt/datadog/agent-data-plane/LICENSES/THIRD-PARTY-{}
+    - cp NOTICE LICENSE LICENSE-3rdparty.csv /tmp/tarball/opt/datadog/agent-data-plane/
+    - tar -C /tmp/tarball -czf /tmp/agent-data-plane-${SOURCE_IMAGE_VERSION}-${TARGET_OS}-${TARGET_ARCH}.tar.gz .
+    - aws s3 cp /tmp/agent-data-plane-${SOURCE_IMAGE_VERSION}-${TARGET_OS}-${TARGET_ARCH}.tar.gz ${TARGET_BUCKET_PATH}
+    # `sha256sum` is GNU coreutils (Linux); macOS uses BSD `shasum -a 256` to print the same digest format.
+    - shasum -a 256 /tmp/agent-data-plane-${SOURCE_IMAGE_VERSION}-${TARGET_OS}-${TARGET_ARCH}.tar.gz
+
+push-release-tarball-darwin-amd64:
+  extends: [.macos-amd64-test-job, .push-release-tarball-darwin-definition]
+  needs:
+    - unit-tests-macos-amd64
+    - unit-tests-miri-macos-amd64
+    - test-integration-macos-amd64
+    - check-deny
+    - check-licenses
+  variables:
+    TARGET_ARCH: "amd64"
+
+push-release-tarball-darwin-arm64:
+  extends: [.macos-arm64-test-job, .push-release-tarball-darwin-definition]
+  needs:
+    - unit-tests-macos-arm64
+    - unit-tests-miri-macos-arm64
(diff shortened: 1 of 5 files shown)


.gitlab/release.yml
```diff
@@ -106,3 +106,58 @@ push-release-tarball-linux-arm64-fips:
...
+      allow_failure: true
+      variables:
+        TARGET_BUCKET_PATH: "s3://binaries-ddbuild-io-prod/saluki/internal/"
+  image: ${DOCKER_BUILD_IMAGE}
+  variables:
+    TARBALL_NAME: "agent-data-plane-${ADP_IMAGE_VERSION}-darwin-${TARGET_ARCH}.tar.gz"
+  script:
+    - apt-get update && DEBIAN_FRONTEND=noninteractive apt install -yy --no-install-recommends awscli >/dev/null
+    - aws s3 cp ${TARBALL_NAME} ${TARGET_BUCKET_PATH}
+    - sha256sum ${TARBALL_NAME}
```
thieman, line 137: What's the point of doing this after the cp?

.gitlab/release.yml
```diff
@@ -106,3 +106,58 @@ push-release-tarball-linux-arm64-fips:
...
+        TARGET_BUCKET_PATH: "s3://binaries-ddbuild-io-prod/saluki/"
+    - when: manual
+      allow_failure: true
+      variables:
+        TARGET_BUCKET_PATH: "s3://binaries-ddbuild-io-prod/saluki/internal/"
+  image: ${DOCKER_BUILD_IMAGE}
+  variables:
+    TARBALL_NAME: "agent-data-plane-${ADP_IMAGE_VERSION}-darwin-${TARGET_ARCH}.tar.gz"
+  script:
+    - apt-get update && DEBIAN_FRONTEND=noninteractive apt install -yy --no-install-recommends awscli >/dev/null
```
thieman, line 135: Does the Linux job also need to manually install this?

ci/tooling/package-adp-tarball.sh
```diff
@@ -0,0 +1,70 @@
...
+echo "[*] Harvesting THIRD-PARTY-* license texts"
+tail -n +2 LICENSE-3rdparty.csv \
+  | awk -F ',' '{print $3}' \
+  | awk -F' ' '{for(i=1;i<=NF;i++) print $i}' \
+  | grep -v -E "(OR|AND|WITH|Custom|LLVM-exception)" \
+  | sed s/[\(\)]// \
+  | sort \
+  | uniq \
+  | xargs -I {} cp "${spdx_text_dir}/{}.txt" \
+      "${stage}/tarball/opt/datadog/agent-data-plane/LICENSES/THIRD-PARTY-{}"
```
thieman, line 63: Write a short comment above this code explaining what this is actually doing

Makefile
```diff
@@ -604,16 +604,35 @@ list-integration-tests: ## Lists available ADP integration tests
...
 		APP_IDENTIFIER="$(ADP_APP_IDENTIFIER)" \
 		APP_GIT_HASH="$(ADP_APP_GIT_HASH)" \
 		APP_VERSION="$(ADP_APP_VERSION)" \
 		APP_BUILD_DATE="$(ADP_APP_BUILD_DATE)" \
-		cargo build --release --bin agent-data-plane
+		cargo build --profile $(BUILD_PROFILE) --bin agent-data-plane
+
+# SPDX license-list-data tag pinned to match docker/Dockerfile.agent-data-plane so the
+# host-built tarball's THIRD-PARTY-* texts match the linux Docker artifact.
+ADP_SPDX_LICENSES_VERSION := 3.28.0
```
thieman: Hoist this to the top of the file with other constants and ideally source it from an already-existing one

Makefile
```diff
@@ -604,16 +604,17 @@ list-integration-tests: ## Lists available ADP integration tests
...
+build-adp-host: ## Builds the agent-data-plane binary for the current host (Cargo profile from $$BUILD_PROFILE, default: release)
+	@echo "[*] Building agent-data-plane ($(BUILD_PROFILE), host target)..."
 	@APP_FULL_NAME="$(ADP_APP_FULL_NAME)" \
 		APP_SHORT_NAME="$(ADP_APP_SHORT_NAME)" \
 		APP_IDENTIFIER="$(ADP_APP_IDENTIFIER)" \
 		APP_GIT_HASH="$(ADP_APP_GIT_HASH)" \
 		APP_VERSION="$(ADP_APP_VERSION)" \
 		APP_BUILD_DATE="$(ADP_APP_BUILD_DATE)" \
-		cargo build --release --bin agent-data-plane
+		cargo build --profile $(BUILD_PROFILE) --bin agent-data-plane
```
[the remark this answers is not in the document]
thieman: [Claude Sonnet 4.5] Good catch — this is a real bug. `test-integration-macos-run` hardcodes `target/release/{panoramic,agent-data-plane}` as its binary paths, but with `build-adp-host` now honoring `BUILD_PROFILE`, a tagged release pipeline (`BUILD_PROFILE=optimized-release` in workflow rules) would build into `target/optimized-release/` and the test runner wouldn't find anything.

Fixed in 59a1e3a4da by pinning `test-integration-macos-ci` to `BUILD_PROFILE=release` via `override` so the integration tests always build into `target/release/` regardless of the pipeline-level setting. The integration tests don't care about release vs optimized-release functionally, only about the path.

.gitlab/release.yml
```diff
@@ -106,3 +106,79 @@ push-release-tarball-linux-arm64-fips:
...
+    # Aliased so the script body parallels `.push-release-tarball-definition`. The Linux job
+    # picks this up from each concrete job's variables (sourced from the image tag); on darwin
+    # there is no image, so the canonical pipeline version variable is used directly.
+    SOURCE_IMAGE_VERSION: ${ADP_IMAGE_VERSION}
+    # Pinned to match docker/Dockerfile.agent-data-plane so the tarball ships the same set of
+    # THIRD-PARTY-* license texts as the Linux artifact. Bump in lockstep with the Dockerfile.
+    SPDX_LICENSES_VERSION: "3.28.0"
+  script:
+    - command -v aws >/dev/null 2>&1 || brew install awscli
+    - mkdir -p /tmp/tarball/opt/datadog-agent/embedded/bin /tmp/tarball/opt/datadog/agent-data-plane/LICENSES
```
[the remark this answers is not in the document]
thieman: [Claude Sonnet 4.5] This was reviewing commit 5e0ef73863, which had the build and push as a single inline-script job that ran on the macOS runner with `mkdir -p /tmp/tarball/...` and a tar of the whole directory. The concern was correct for that shape.

The current shape (since 03e5222658 / 5f1b9e04d7) sidesteps both issues:

- Tarball assembly moved to `ci/tooling/package-adp-tarball.sh` which uses `mktemp -d` for staging plus a `trap rm -rf` cleanup, so each invocation gets a fresh per-run directory and never reuses leftover state.
- The push job moved to `.gitlab/release.yml` and runs in `${DOCKER_BUILD_IMAGE}` (a fresh Linux container), not on the bare-metal macOS runner — there's no shared filesystem state to pick up stale files from in the first place.

.gitlab/release.yml
thieman: [Claude Sonnet 4.5] Fair point — moved sha256sum *before* aws s3 cp in 59a1e3a4da. The digest now lands in the job log unconditionally, so when an upload fails we still know what file we tried to send. (Linux release push has it after — that's mirrored from the original linux template; we can do better here.)

.gitlab/release.yml
thieman: [Claude Sonnet 4.5] Yes — the linux push job (`.push-release-tarball-definition` in this same file, line ~44) has the identical `apt-get update && DEBIAN_FRONTEND=noninteractive apt install -yy --no-install-recommends awscli` line. `${DOCKER_BUILD_IMAGE}` (`registry.ddbuild.io/docker:27.3.1`) doesn't ship with awscli preinstalled, so both push jobs install it on demand. Could move into the build image itself, but that's outside scope here.

ci/tooling/package-adp-tarball.sh
thieman: [Claude Sonnet 4.5] Added in 59a1e3a4da — a what-comment now sits above the pipeline explaining each pipe stage: walk the third CSV column (SPDX expressions), split on whitespace, drop join keywords (OR/AND/WITH) and tokens we don't have texts for (Custom, LLVM-exception), strip grouping parens, dedupe, then `xargs cp` each remaining identifier's text into the tarball as `THIRD-PARTY-<spdx-id>`. The previous comment only documented *why* the pipeline mirrors the Dockerfile, not *what* each stage was for.

Makefile
thieman: [Claude Sonnet 4.5] Hoisted in 59a1e3a4da — `ADP_SPDX_LICENSES_VERSION` now lives at the top of the Makefile alongside `ADP_APP_VERSION` and friends.

I checked whether we could source it from an already-existing constant and the answer is no: the only other place this version appears is `docker/Dockerfile.agent-data-plane` as a hardcoded literal (`v3.28.0`), and that Dockerfile gets invoked directly via `docker buildx build` (not via the Makefile), so there's no clean way to share a single source between them without plumbing a `--build-arg` through CI for one constant. Kept the dual source-of-truth and put a comment on the Makefile constant calling out the cross-file pin.

Makefile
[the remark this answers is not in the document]
thieman: [Claude Sonnet 4.5] Walking back the previous reply — pinning to `release` was the wrong direction. Tagged release pipelines would have shipped an `optimized-release` binary that no integration tests ever touched, which defeats the point of test gating.

Linux runs its integration tests against whatever profile the pipeline is building: `build-adp-image` honors `BUILD_PROFILE` (so on tag it builds with `optimized-release`), and the e2e correctness pipeline consumes that exact image via `build-bundled-agent-adp-image`. Tests exercise what's about to ship.

Mirrored that for darwin in 1310181b50:

- Dropped the `override BUILD_PROFILE := release` on `test-integration-macos-ci` so the pipeline-level `BUILD_PROFILE` flows through.
- Made `test-integration-macos-run` honor `BUILD_PROFILE` for the ADP binary path (`target/\$(BUILD_PROFILE)/agent-data-plane`). Dev pipelines resolve to `target/release/` as before; tagged releases track `target/optimized-release/`.
- `panoramic` stays pinned at `target/release/panoramic` — it's the test harness, not the SUT, and `build-panoramic` always builds with `--profile release`. Same model as linux's `build-panoramic-binary`.

Makefile
```diff
@@ -604,21 +609,42 @@ list-integration-tests: ## Lists available ADP integration tests
...
+# ADP path tracks $$BUILD_PROFILE so a tagged release pipeline (BUILD_PROFILE=optimized-release
+# in .gitlab-ci.yml workflow rules) tests the same binary it's about to ship — mirroring how the
+# linux flow builds and tests build-adp-image with whatever BUILD_PROFILE the pipeline sets.
+# panoramic stays at target/release/ unconditionally because it's the test harness, not the SUT;
+# build-panoramic always builds with --profile release, same as linux's build-panoramic-binary.
+test-integration-macos-run: BUILD_PROFILE ?= release
+test-integration-macos-run: ## Runs the macOS host-process integration tests using already-built binaries (assumes target/$$BUILD_PROFILE/agent-data-plane and target/release/panoramic exist). Defaults to all `mac`-runtime-eligible tests; narrow with CASE=<name>.
 	@echo "[*] Running macOS host-process integration tests..."
-	@ADP_BINARY_PATH="$(CURDIR)/target/release/agent-data-plane" \
+	@ADP_BINARY_PATH="$(CURDIR)/target/$(BUILD_PROFILE)/agent-data-plane" \
```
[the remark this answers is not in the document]
thieman: [Claude Sonnet 4.5] Real catch — I introduced this hole when I made `test-integration-macos-run` track `BUILD_PROFILE` (1310181b50). The `pkill -f /target/release/agent-data-plane` pattern is effectively a literal substring search; `/target/release/agent-data-plane` is not a substring of `/target/optimized-release/agent-data-plane` (the `release` in the latter is preceded by `optimized-`, not `/`), so a stranded ADP on a tagged release pipeline would survive the sweep on the bare-metal amd64 runner.

Fixed in e1379f27e4 by adding a second `pkill -9 -f /target/optimized-release/agent-data-plane` line. Both possible output paths now get swept regardless of which profile the prior pipeline was using; updated the surrounding comment to reflect that the path varies by `BUILD_PROFILE`. arm64 is unaffected (Tart VM, fresh state per job).

tobz (approved):

Issue #1362, opened 2026-04-13
Investigate `forwarder_high_prio_buffer_size` default divergence and `forwarder_low_prio_buffer_size` gap.


webern:
**Core agent:**
- [`forwarder_high_prio_buffer_size`][1] (default 100) — sizes a [Go channel][3] for new incoming transactions
- [`forwarder_low_prio_buffer_size`][2] (default 100) — sizes a [Go channel][4] for retry transactions
- Workers select from high prio before low prio, giving fresh transactions priority
- When high prio is [full][5], transactions overflow to the retry queue

**ADP:**
- Reads `forwarder_high_prio_buffer_size` but renames it to `endpoint_buffer_size`, [default 16][s1] (not 100)
- Uses it as [`VecDeque::with_capacity(max_enqueued)`][s2] for [`PendingTransactions.high_priority`][s3]
- Low priority is a [`RetryQueue`][s4] — **not sized by `forwarder_low_prio_buffer_size`** at all (it's sized by `queue_max_size_bytes`)
- Same priority semantics: [`pop()`][s5] drains high prio first, [overflow from high prio][s6] goes to low prio (retry queue)

**Key divergences:**
1. Default differs: 16 vs 100
2. `forwarder_low_prio_buffer_size` is completely ignored — ADP's retry queue is byte-bounded, not count-bounded
3. The architectures are different enough that a 1:1 mapping may not make sense

Probably just needs documentation. The priority semantics match; the sizing model is intentionally different.

[1]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/pkg/config/setup/common_settings.go#L1541
[2]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/pkg/config/setup/common_settings.go#L1542
[3]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/comp/forwarder/defaultforwarder/domain_forwarder.go#L205
[4]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/comp/forwarder/defaultforwarder/domain_forwarder.go#L206
[5]: https://github.com/DataDog/datadog-agent/blob/0f960f116880/comp/forwarder/defaultforwarder/domain_forwarder.go#L365-L369
[s1]: https://github.com/DataDog/saluki/blob/dbb968c2db21/lib/saluki-components/src/common/datadog/config.rs#L18-L19
[...]

PR #1844, opened 2026-06-08
fix(datadog): match forwarder high-priority buffer default

aqian01:
## What changed

- Changes ADP's `forwarder_high_prio_buffer_size` default from `16` to `100`, matching the Core Agent default.
- Updates the DogStatsD configuration docs and `known-configs.json` so `forwarder_high_prio_buffer_size` is classified as parity for #1362.
- Leaves the existing low-priority/retry queue behavior unchanged: ADP still uses the byte-bounded retry queue and does not implement `forwarder_low_prio_buffer_size`.

## Why

Follow-up from the discussion on #1827: there was no specific reason for ADP to diverge from the Core Agent default, so this keeps the default aligned while leaving any future queue behavior changes for separate evaluation.

Related to #1362.

## Validation

- `PATH=$HOME/.cargo/bin:$HOME/.rustup/toolchains/nightly-aarch64-apple-darwin/bin:$PATH make fmt`
- `PATH=$HOME/.cargo/bin:$HOME/.rustup/toolchains/nightly-aarch64-apple-darwin/bin:$PATH cargo check --workspace`
- `PATH=$HOME/.cargo/bin:$HOME/.rustup/toolchains/nightly-aarch64-apple-darwin/bin:$PATH cargo check --workspace --tests`
- `PATH=$HOME/.cargo/bin:$HOME/.rustup/toolchains/nightly-aarch64-apple-darwin/bin:$PATH cargo nextest run -p saluki-components common::datadog::io::tests`
  - 15 passed; nextest reported one leaky test and exited successfully.


docs/agent-data-plane/configuration/dogstatsd.md
```diff
@@ -108,19 +107,6 @@ doesn't support TLS 1.0 or TLS 1.1.
...
-ADP supports `forwarder_high_prio_buffer_size` for the shared Datadog forwarder, but uses a
-different default from the core agent. The core agent defaults both its high-priority and
-low-priority forwarder worker channels to `100`. ADP defaults its high-priority per-endpoint
-pending queue to `16`.
-
-ADP does not implement `forwarder_low_prio_buffer_size`. Low-priority pending transactions use the
-forwarder's retry queue, which is bounded by payload bytes and optional disk persistence settings
-rather than by a transaction count. ADP still preserves the same broad priority behavior: new
-transactions are processed through the high-priority queue first, retries use the low-priority retry
-queue, and high-priority overflow is moved to the low-priority retry queue.
```
jszwedko: I think this part is still true, no? We only implement `forwarder_high_prio_buffer_size`. I think this section needs to be updated to mention how ADP sizes both queues.

jszwedko (approved):
One comment about the docs. Otherwise 👍

docs/agent-data-plane/configuration/dogstatsd.md
tobz: From what I can see, the changed text does accurately reflect the behavior?

- we have a high-priority queue, sized by `forwarder_high_prio_buffer_size`
- we have a low-priority "retry" queue, but sized by `forwarder_retry_queue_payloads_max_size` (size, not count-based)

tobz (approved):
LGTM.

As mentioned in Slack, we should actually read this configuration setting, but it's better to at least have the defaults aligned until that happens.

PR #1845, opened 2026-06-08
feat(agent-data-plane): support sslkeylogfile

### LINKED ISSUE #1372: Support `sslkeylogfile` to dump TLS key material for network debugging.
The core agent exposes [`sslkeylogfile`][1] (default empty/disabled). When set to a file path, it [opens the file][2] and sets it as the [`KeyLogWriter`][3] on the Go TLS config. This writes TLS session keys in NSS Key Log Format, allowing outbound HTTPS traffic to be decrypted in tools like Wireshark.

ADP has no equivalent. Rustls supports this via the `key_log` field on `ClientConfig`, but ADP's [TLS builder][s1] does not expose it.

[1]: https://github.com/DataDog/datadog-agent/blob/141f12062790/pkg/config/setup/common_settings.go#L1088
[2]: https://github.com/DataDog/datadog-agent/blob/141f12062790/pkg/util/http/transport.go#L70-L76
[3]: https://github.com/DataDog/datadog-agent/blob/141f12062790/pkg/util/http/transport.go#L80
[s1]: https://github.com/DataDog/saluki/blob/e9373ef98013/lib/saluki-tls/src/lib.rs#L118-L148

aqian01:
Implements `sslkeylogfile` support in ADP. When set (via config or `DD_SSLKEYLOGFILE`), the Datadog intake forwarder writes TLS session secrets in NSS key log format so packet captures can be decrypted in tools like Wireshark. Protect the generated file as sensitive data.

The file is opened once per path with `0o600` permissions on Unix, and concurrent handshakes share a single write lock so NSS lines never interleave. A bad path warns once and continues without key logging. FIPS builds warn and skip — a leftover setting does not block startup.

Fixes #1372.

### Tests

Unit tests

### Manual QA

Ran ADP in standalone mode so the live Datadog Agent config stream would not override local DogStatsD socket and `sslkeylogfile` settings:

```fish
env \
  DD_API_KEY=<redacted> \
  DD_SSLKEYLOGFILE=/tmp/adp-sslkeylogfile.keys \
  DD_DATA_PLANE_ENABLED=true \
  DD_DATA_PLANE_STANDALONE_MODE=true \
  DD_DOGSTATSD_PORT=0 \
  DD_DOGSTATSD_SOCKET=/tmp/adp-dogstatsd-dgram.sock \
  DD_DOGSTATSD_STREAM_SOCKET=/tmp/adp-dogstatsd-stream.sock \
  RUST_LOG=info \
  target/devel/agent-data-plane run
```

Observed a single "TLS key logging enabled" warning after the file was opened, both DogStatsD sockets started:

```text
TLS key logging enabled; TLS session secrets will be written to disk.
listen_addr:"unix:///tmp/adp-dogstatsd-stream.sock" | DogStatsD listener started.
listen_addr:"unixgram:///tmp/adp-dogstatsd-dgram.sock" | DogStatsD listener started.
```

Sent a DogStatsD metric to force a forwarder flush:

```fish
python3 -c 'import socket; path="/tmp/adp-dogstatsd-dgram.sock"; sock=socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM); sock.sendto(b"adp.sslkeylogfile.manual:1|c|#source:manual\n", path)'
```

Verified the key log file was created with owner-only permissions and contained NSS key log entries:

```text
$ ls -l /tmp/adp-sslkeylogfile.keys
-rw-------@ 1 andrew.qian  wheel  528 Jun 11 10:09 /tmp/adp-sslkeylogfile.keys

$ cat /tmp/adp-sslkeylogfile.keys
CLIENT_RANDOM 82a4030012955596c7629a38b0b26022c26892e00ec89e472e3fd5de1676b059 20cc2069789ed8cc792c7819001eb49b6e4e18137e032cf157599602a4721b5b77a60d6a64079f68ce9b0f45f99f3fbc
CLIENT_RANDOM 4268951d8db643139a44e311b57baa041c07d165d8d6f122617b8b3044574267 3260ef89f1f8bfdadeb73717465a5c22e3488708d4c98498ecd36dce7f83597bb478d732fe4ebe046b9c259eeae82aa6
CLIENT_RANDOM 36afbc1f31d82487dd034f9eb8bc0d3ce0eb2cd8f8e21150b110566b7a053c6f 462a9998c7347853d5d993758daa6f05667bdf3de1060ac8ad69c8379ba2d7abb21673092a5a1568131da5f1899d09e4
```


lib/saluki-components/src/common/datadog/config.rs
```diff
@@ -228,6 +211,14 @@ pub struct ForwarderConfiguration {
     #[serde(default)]
     skip_ssl_validation: bool,
 
+    /// TLS key log file path for Datadog intake forwarding.
+    ///
+    /// Defaults to empty, which disables TLS key logging. If set to a path, HTTPS clients built for the shared Datadog
+    /// forwarder write TLS session secrets in NSS key log format. This compromises TLS security and should only be
+    /// enabled for debugging packet captures.
```
tobz:
```suggestion
    /// File path to write TLS key material to for all HTTPS connections to the
    /// Datadog backend.
    ///
    /// When non-empty, enables the logging of TLS key material to the given file path,
    /// in the [NSS Key Log][nss_key_log] format, which can be used for debugging TLS
    /// issues, as well as decrypting captured TLS traffic in tools such as Wireshark.
    ///
    /// Defaults to empty.
    ///
    /// [nss_key_log]: https://nss-crypto.org/reference/security/nss/legacy/key_log_format/index.html
```

lib/saluki-tls/src/lib.rs
```diff
@@ -100,13 +174,23 @@ pub struct ClientTLSConfigBuilder {
...
             min_tls_version: TlsMinimumVersion::default(),
             root_cert_store: None,
             danger_accept_invalid_certs: false,
         }
     }
 
+    /// Enables TLS key logging to the given file in NSS key log format.
+    ///
+    /// Defaults to disabled. This writes TLS session secrets that external tools can use to decrypt captured TLS
+    /// traffic, so it should only be enabled for debugging.
```
tobz:
```suggestion
    /// Enables logging of TLS key material to the given file path.
    ///
    /// TLS key material will be logged to the the given file path in the [NSS Key Log][nss_key_log]
    /// format, which can be used for debugging TLS issues, as well as decrypting captured
    /// TLS traffic in tools such as Wireshark.
    ///
    /// The file's permissions will be set to owner read/write only.
    ///
    /// [nss_key_log]: https://nss-crypto.org/reference/security/nss/legacy/key_log_format/index.html
```

tobz (commented):
Two things related to the doc comment wording.

tobz (approved):
LGTM, just looks like there's some docs conflicts to resolve.

PR #1848, opened 2026-06-09
fix(env): properly clean up cgroups poller thread when collector is dropped

tobz:
## Summary

This PR fixes a bug where the cgroups polling task for the cgroups metadata collector was not being properly shutdown when the collector was dropped.

Prior to this PR, we attempted to ensure that the blocking task we spawned in the cgroups metadata collector was then cleaned up (aborted, specifically) when the collector was dropped. However, I made an oversight when doing this work which is that blocking tasks _can't_ be aborted. Doh!

This PR fixes the issue by passing in a new shutdown handle type which is used to signal to the blocking task to stop, and is also done so via drop: the collector should always manage to signal the blocking task to stop no matter what causes the collector to be stopped/dropped.

We've added this new shutdown type (two types, actually) to a new `sync` module in `saluki-common`. I plan to move some related/duplicate shutdown types from `saluki-core` to this new module in a follow-up PR.. but for now, I leaned towards keeping this PR narrowly scoped.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

New and existing unit tests.

## References

DADP-2


thieman:
There's one more instance of `spawn_blocking` here we should probably audit https://github.com/DataDog/saluki/blob/d25b9a122c5fadedd2c7def1b5aec5681d4d7962/lib/saluki-io/src/net/util/retry/queue/persisted.rs#L333

thieman (approved):

PR #1849, opened 2026-06-10
chore(config): update core schema version and overlay

webern:
## Summary

Update the core schema with the latest version. The schema has been split into multiple subsystem files since the first time we vendored it. This might give us an opportunity to wholesale omit subsystems in the future, but for now I just vendored the whole thing. If the shapes are stable on the next update, I may introduce a blocklist for subsystem files that we know are not relevant to ADP.

I isolated the overlay and key handling to a single commit for easier reviewing. The most important thing to look at in this PR is the changes to `schema_overlay.yaml`. If you want to change the metadata for a key, that would be the place to leave a comment.

## Change Type

## How did you test this PR?

`make build-schema-overlay && make fmt && make check-all && make test-all`

## References

Indirectly related to #1788 since this is the first refresh of the schema overlay system.


lib/datadog-agent/config-overlay-model/src/lib.rs
```diff
@@ -365,27 +365,40 @@ impl SchemaOverlay {
...
             if let Some(name) = k.as_str() {
                 let full_key = if prefix.is_empty() {
                     name.to_string()
                 } else {
                     format!("{}.{}", prefix, name)
                 };
+                // Resolve $ref: <filename> by loading the referenced file and recursing.
+                if let Some(ref_path) = v.get("$ref").and_then(|rv| rv.as_str()) {
+                    let ref_file = schema_dir.join(ref_path);
+                    if let Ok(ref_contents) = std::fs::read_to_string(&ref_file) {
```
webern: sonnet 4.6: The `$ref` resolution here silently swallows both I/O errors and parse failures. If a subsystem file goes missing, the code falls through the `continue` and eventually inserts the parent key (e.g. `admission_controller`) as a leaf in the schema key set. Overlay validation then fails with a confusing "key not in schema" error for every `admission_controller.*` entry rather than a clear "failed to load `$ref`" message.

Contrast with `schema_gen.rs` line 79, which panics explicitly. Both paths are build-time only; this one should do the same — either `panic!` or propagate an `Err` up through `schema_keys`.

lib/datadog-agent/config-overlay-model/src/schema_gen.rs
```diff
@@ -65,18 +72,31 @@ fn collect_entries(mapping: &serde_yaml::Mapping, path_parts: &[&str], out: &mut
         let mut parts = path_parts.to_vec();
         parts.push(key_str);
 
+        // Resolve $ref: <filename> by loading the referenced file and recursing into its properties.
+        if let Some(ref_path) = value.get("$ref").and_then(|v| v.as_str()) {
```
webern: sonnet 4.6: The `$ref` resolution logic is now implemented independently in both `collect_entries` here and `collect_schema_keys` in `lib.rs`. They already diverge on error handling (panic vs. silent ignore). Worth extracting a shared helper before this pattern solidifies further — even a small `fn load_ref_properties(dir: &Path, ref_path: &str) -> ...` would keep them in sync.

tobz (approved):

PR #1850, opened 2026-06-10
feat(agent-data-plane): added observability for retry_queue

### LINKED ISSUE #1365: Investigate whether `forwarder_retry_queue_capacity_time_interval_sec` telemetry is needed.
This key (default 900 = 15 minutes) controls a purely observability feature in the core agent: estimating how many seconds of data the retry queue can buffer if the network goes down.

The core agent's [`QueueDurationCapacity`][1] works as follows:
1. For every transaction, [`OnTransaction()`][2] records its payload size into a per-domain sliding window accumulator spanning the configured interval (default 15 minutes).
2. Periodically, [`ComputeCapacity()`][3] calculates per-domain statistics:
   - **Bytes per second**: total bytes in the sliding window / window duration
   - **Available space**: in-memory queue max + proportional share of disk space (proportioned by each domain's [relative traffic rate][4])
   - **Capacity in seconds**: available space / bytes per second
3. The results are [emitted as three telemetry gauges][5] per domain:
   - `datadog.agent.retry_queue_duration.capacity_secs`
   - `datadog.agent.retry_queue_duration.bytes_per_sec`
   - `datadog.agent.retry_queue_duration.capacity_bytes`

This does not affect retry queue behavior — it is strictly observability. It answers the question "if the network drops right now, how long can we buffer data before we start dropping?"

ADP's retry queue reports [insertion/removal counters and dropped entry counts][s1] but has no equivalent rate or capacity estimation.

Investigate whether this is something that is likely to be requested of ADP. Implement if so, or document that ADP does not have it if not.

[1]: https://github.com/DataDog/datadog-agent/blob/84f2acad550c/comp/forwarder/defaultforwarder/internal/retry/queue_duration_capacity.go#L26-L35
[2]: https://github.com/DataDog/datadog-agent/blob/84f2acad550c/comp/forwarder/defaultforwarder/internal/retry/queue_duration_capacity.go#L60
[3]: https://github.com/DataDog/datadog-agent/blob/84f2acad550c/comp/forwarder/defaultforwarder/internal/retry/queue_duration_capacity.go#L89
[4]: https://github.com/DataDog/datadog-agent/blob/84f2acad550c/comp/forwarder/defaultforwarder/internal/retry/queue_duration_capacity.go#L119
[5]: https://github.com/DataDog/datadog-agent/blob/84f2acad550c/comp/forwarder/defaultforwarder/default_forwarder.go#L601-L609
[s1]: https://github.com/DataDog/saluki/blob/e9373ef98013/lib/saluki-components/src/common/datadog/telemetry.rs#L78-L98

lucastemb:
## Summary

Adds observability to ADP estimating how many seconds of data the retry queue can buffer if the network goes down. Emits three Datadog Agent-compatible telemetry gauges per domain:
 - `datadog.agent.retry_queue_duration.capacity_secs`
 - `datadog.agent.retry_queue_duration.bytes_per_sec`
 - `datadog.agent.retry_queue_duration.capacity_bytes` 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests

## References

- Closes: https://github.com/DataDog/saluki/issues/1365

### DIFF
--- a/lib/saluki-components/src/common/datadog/io.rs
+++ b/lib/saluki-components/src/common/datadog/io.rs
@@ -37,6 +37,7 @@ use super::{
     config::ForwarderConfiguration,
     endpoints::{EndpointRoute, ResolvedEndpoint, RoutableEndpoint},
     middleware::{for_resolved_endpoint, with_allow_arbitrary_tags, with_version_info},
+    retry_capacity::{TrafficRateWindow, RETRY_QUEUE_CAPACITY_BUCKET_DURATION_SECS},
     telemetry::{ComponentTelemetry, SharedTransactionQueueTelemetry, TransactionQueueTelemetry},
     transaction::{Metadata, Transaction, TransactionBody},
     METRIC_INTAKE_PATHS,
@@ -46,9 +47,6 @@ use super::{
 ///
 /// Used to influence the size of chunks in `ChunkedBytesBuffer`.
 pub const RB_BUFFER_CHUNK_SIZE: usize = 32 * 1024; // 32 KB
-
-const RETRY_QUEUE_CAPACITY_HISTORY_DURATION_SECS: u64 = 15 * 60;
-const RETRY_QUEUE_CAPACITY_BUCKET_DURATION_SECS: u64 = 10;
 
 /// A handle to the transaction forwarder.
 pub struct Handle<B>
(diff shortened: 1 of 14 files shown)


lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -692,7 +631,7 @@ impl<T: Retryable> PendingTransactions<T> {
     ///
     /// If the high-priority queue is full, the transaction will be pushed into the low-priority queue.
     pub async fn push_high_priority(&mut self, transaction: T) -> Result<PushResult, GenericError> {
-        self.record_incoming_transaction_size(transaction.size_bytes());
+        self.record_incoming_transaction_size(transaction.size_bytes()).await;
```
tobz, line 634: One thing that strikes me is: should we actually be calling this unless we have to push to the low priority side, aka the retry queue?

We'll end up calling this even when we aren't pushing into the retry queue, which seems suboptimal since, if I'm understanding it correctly, none of the values would actually be changing for retry queue-related telemetry.

lucastemb: From what I understand, yes. We need to be constantly sampling transactions so that we can ultimately answer `"How long do we roughly have, given current network traffic, until we start dropping data"`

tobz: Hmm, yeah, that makes sense when put like that. 👍🏻

tobz (approved):

PR #1851, opened 2026-06-10
feat(agent-data-plane): support forwarder flush-to-disk ratio

### LINKED ISSUE #1364: Decide whether `forwarder_flush_to_disk_mem_ratio` needs an ADP equivalent.
The core agent exposes [`forwarder_flush_to_disk_mem_ratio`][1] (default 0.5). This controls the memory-to-disk spill strategy for the retry queue. When the in-memory retry queue exceeds its size limit and a new failed transaction needs to be added, this ratio determines what fraction of the queue (by payload size) gets [serialized to disk][2] to make room.

For example, with `maxMemSizeInBytes=100` and `flushToStorageRatio=0.6`: when adding a transaction that would push the queue past 100 bytes, the oldest transactions totaling at least 60 bytes are [written to a `.retry` file][3] on disk and removed from memory. The [README][4] has a worked example with diagrams.

It's a tuning knob for the tradeoff between disk I/O frequency and memory headroom — a lower ratio means more frequent but smaller disk writes, a higher ratio means fewer but larger writes.

ADP's [`RetryQueue`][s1] has disk persistence but manages it differently — bounded by [`storage_max_size_bytes`][s2] and [`storage_max_disk_ratio`][s3] with no configurable spill ratio. The spill behavior is handled internally without a tunable equivalent.

Worth checking whether this is used in practice and whether configurability is needed here. If not, documentation will suffice.

[1]: https://github.com/DataDog/datadog-agent/blob/4e3b17883b32/pkg/config/setup/common_settings.go#L1535
[2]: https://github.com/DataDog/datadog-agent/blob/4e3b17883b32/comp/forwarder/defaultforwarder/internal/retry/transaction_retry_queue.go#L95-L108
[3]: https://github.com/DataDog/datadog-agent/blob/4e3b17883b32/comp/forwarder/defaultforwarder/default_forwarder.go#L352-L376
[4]: https://github.com/DataDog/datadog-agent/blob/4e3b17883b32/comp/forwarder/defaultforwarder/internal/retry/README.md
[s1]: https://github.com/DataDog/saluki/blob/e9373ef98013/lib/saluki-components/src/common/datadog/io.rs#L275
[s2]: https://github.com/DataDog/saluki/blob/e9373ef98013/lib/saluki-components/src/common/datadog/retry.rs#L105-L108
[s3]: https://github.com/DataDog/saluki/blob/e9373ef98013/lib/saluki-components/src/common/datadog/retry.rs#L124-L127

aqian01:
## Summary

Adds ADP support for the Datadog Agent `forwarder_flush_to_disk_mem_ratio` config key.

- Deserializes the key in retry configuration with the Agent default of `0.5`
- Passes the configured ratio into the retry queue
- Flushes at least `max_in_memory_bytes * ratio` worth of old entries to disk when a disk-enabled retry queue overflows
- Preserves no-disk behavior by dropping only enough old entries to make room
- Marks the key as supported in the config registry and DogStatsD compatibility docs

Fixes #1364

## Validation

unit tests


lib/saluki-io/src/net/util/retry/queue/mod.rs
```diff
@@ -189,15 +206,55 @@ where
...
+        let bytes_to_remove = if required_bytes == 0 {
+            0
+        } else if using_disk {
+            required_bytes.max(flush_to_disk_bytes(
+                self.max_in_memory_bytes,
+                self.flush_to_disk_mem_ratio,
+            ))
+        } else {
+            required_bytes
+        };
```
tobz: This could just be:

```suggestion
        let bytes_to_remove = if using_disk {
            required_bytes.max(flush_to_disk_bytes(
                self.max_in_memory_bytes,
                self.flush_to_disk_mem_ratio,
            ))
        } else {
            required_bytes
        };
```

lib/saluki-io/src/net/util/retry/queue/mod.rs
```diff
@@ -189,15 +206,55 @@ where
...
+        {
             let oldest_entry = self.pending.pop_front().expect("queue is not empty");
             let oldest_entry_size = oldest_entry.size_bytes();
 
-            if let Some(persisted_pending) = &mut self.persisted_pending {
-                let persist_result = persisted_pending.push(oldest_entry).await?;
-                push_result.merge(persist_result);
-
-                debug!(entry.len = oldest_entry_size, "Moved in-memory entry to disk.");
+            if using_disk {
```
tobz, line 253: Are you _sure_ this is right?

It seems like we would still want this is fire if disk persistence is enabled, even if we're only moving a single entry to disk. (As in, we should simply gate this on `self.pending_persistence.is_some()`.)

tobz (changes requested):

lib/saluki-io/src/net/util/retry/queue/mod.rs
```diff
@@ -214,15 +231,53 @@ where
...
+                self.flush_to_disk_mem_ratio,
+            ))
+        } else {
+            required_bytes
+        };
+        let mut bytes_removed = 0;
+
+        while !self.pending.is_empty()
+            && (bytes_removed < bytes_to_remove
+                || self.total_in_memory_bytes.saturating_add(current_entry_size) > self.max_in_memory_bytes)
```
tobz: I don't understand why we still have this separate condition here.

If we're calculating the total number of bytes to forcefully drain out, then we should only be paying attention to that.

tobz (changes requested):
One more small thing. 😅

tobz (approved):

PR #1855, opened 2026-06-11
enhancement(checks ipc): add new check data types and harden the protocol

nathan-b:
## Summary
Implement improvements to the IPC protocol between ADP and ACR.

Adds new types (sketches for histograms, the new openmetrics histogram bucket, extend log type). Also tries to make the protocol more robust and build failure detection into it.

## How did you test this PR?
Ran ACR against both ADP and the core agent to confirm data types are correctly sent


lib/protos/datadog/proto/checks/v1/acr_ipc.proto
```diff
@@ -0,0 +1,99 @@
...
+  string check_id = 3;
+  string error = 4;
+}
+
+// Trivial request/response types for RPCs that don't need a body.
+message StreamConfigRequest {}
+message SendCheckResultResponse {}
+
+// gRPC service for ACR <-> ADP IPC.
+service AcrIpc {
```
tobz: Could we just call this `CheckRunner`, and drop the IPC part?

More generic, fits the fact that _realistically_ anything can implement it, avoids the stuttering of "IPC" in the name when that's already implied by how it's used, etc.

lib/protos/datadog/proto/checks/v1/acr_ipc.proto
```diff
@@ -0,0 +1,99 @@
...
+  rpc Handshake(Hello) returns (HelloResp);
+
+  // Check data submission with acknowledgement.
+  rpc SendCheckData(CheckDataMsg) returns (CheckDataAck);
+
+  // Check result reporting.
+  rpc SendCheckResult(CheckResultMsg) returns (SendCheckResultResponse);
+
+  // Server pushes config updates to the client as a stream.
+  rpc StreamConfig(StreamConfigRequest) returns (stream ConfigData);
```
tobz: I realize that this is unwritten, but: can we follow the existing naming scheme for gRPC request/response message types in the Agent ecosystem? `FooRequest` for requests and `FooResponse` for responses, such that they're always paired by prefix for the same RPC operation.

As well, don't truncate words like request, response, or message.

lib/protos/datadog/proto/checks/v1/checks.proto
```diff
@@ -3,25 +3,26 @@ syntax = "proto3";
...
     oneof data {
         datadog.checks.v1.metric.Metric metric = 1;
         datadog.checks.v1.log.Log log = 2;
         datadog.checks.v1.service_check.ServiceCheck service_check = 3;
         datadog.checks.v1.event.Event event = 4;
+        // Slot 5 was previously `histogram` (an unused stub). Repurposed
+        // to carry pre-aggregated DDSketches; the wire bytes from any
+        // legacy producer of the old Histogram message would be
+        // mis-parsed here, so don't rely on cross-version compatibility
+        // for slot 5 specifically.
```
tobz: Nit: comments that describe a change are a code smell of not reviewing LLM output.

We document to explain the current state of things, Git history is for everything that came before, unless there's a _strong_ reason to couple them together.

lib/protos/datadog/proto/checks/v1/histogram_bucket.proto
```diff
@@ -0,0 +1,49 @@
...
+    // When `monotonic=true`, indicates the receiver should emit the
+    // first observed value rather than wait for a second to compute a
+    // delta. Producers set this when they know the bucket started from
+    // zero recently (e.g. a freshly-scraped target).
+    bool flush_first_value = 8;
+
+    // Distinguishes the two sender upcalls. False (the default)
+    // routes to `Sender.OpenmetricsBucket`; true routes to
+    // `Sender.HistogramBucket`. Python `submit_histogram_bucket`
+    // always lands as false.
```
tobz: Nit: opaque comment that requires implicit cross-repository knowledge.

It's also... confusing at face value. The naming implies multiple buckets, but this type is, definitionally, a value representing a single bucket.

lib/saluki-components/src/sources/checks_ipc/mod.rs
```diff
@@ -26,24 +29,41 @@ use saluki_core::{
...
+use tracing::{debug, info, warn};
 
 const fn default_grpc_endpoint() -> ListenAddress {
     ListenAddress::any_tcp(5105)
 }
 
+/// AcrIpc protocol version this server speaks. Bumped on
+/// wire-incompatible changes; today the protocol is wire-additive so
+/// we stay on 1.
+const PROTOCOL_VERSION: u32 = 1;
```
tobz: Nit: we should probably define this as an enum in the checks proto stuff so that we have a consistent baseline of what protocol versions are valid or not.

lib/saluki-components/src/sources/checks_ipc/mod.rs
```diff
@@ -26,24 +29,41 @@ use saluki_core::{
...
+
+/// Identifier surfaced in the Handshake response so a client can tell
+/// which server flavor it connected to.
+const SERVER_ID: &str = "saluki-checks-ipc";
```
tobz: We should probably be deriving the server ID through the information exposed in `saluki_metadata` (similarly for client ID on the ACR side) so that we can get things like the actual version of ADP for free.

tobz (changes requested):

lib/protos/datadog/proto/checks/v1/checks.proto
nathan-b: Yup, there are some commits that are mostly AI generated. Most of the code has been reviewed piecemeal, but I opened the PR as a draft so that I could do a higher-level review (I like the ability to mark files as reviewed because it helps me remember where I left off). Sorry for sending you the slop!

tobz: It _is_ in draft so I definitely don't see it as you asking us/me to review slop. 😅 

If nothing else, this should mostly just serve as a reminder that LLMs love to do this sort of thing (code comments as a description of _what_ they changed, rather than the thing they changed something _to_) and that we need to be vigilant in cleaning that up.

lib/protos/datadog/proto/checks/v1/histogram_bucket.proto
nathan-b: This comment (and TBH this field) is a hot mess. It took me several passes and I'm still not sure I got it right, but my next commit should make this better.

PR #1858, opened 2026-06-11
fix(panoramic): avoid core/adp OTLP bind collision on unix

thieman:
## Summary
- stop forcing Core Agent OTLP receiver endpoints in the Unix integration runner
- keep runner-owned env forcing focused on auth/run-path and DogStatsD ownership only
- update the Unix-runner unit test to assert OTLP endpoints are not forcibly overridden
- align `aggregate_context_limit` for both baseline/comparison targets in the origin-detection correctness matrix base config

## Root cause
`build_core_agent_forced_env` unconditionally set `DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_{GRPC,HTTP}_ENDPOINT` to `127.0.0.1:14317/14318` whenever `DD_DATA_PLANE_OTLP_ENABLED=true`. In converged Unix tests, ADP also attempted to bind OTLP gRPC on `127.0.0.1:14317`, causing an immediate bind conflict and process exit.

Separately, the `dsd-origin-detection-matrix/unified-high-cardinality` correctness variant was intermittently diverging with far more unique metric contexts on comparison than baseline. The matrix config set `DD_AGGREGATE_CONTEXT_LIMIT=500000` only on comparison. Moving `aggregate_context_limit: 500000` into the shared base config removes that asymmetry so both targets run with the same context-eviction behavior.

## Validation
- `cargo test -p panoramic core_agent_env_does_not_force_otlp_receiver_endpoints`
- `cargo test -p panoramic unix_runner::tests`
- `cargo test -p panoramic`
- `cargo run --bin panoramic -- run -d test/correctness/cases -t dsd-origin-detection-matrix/unified-high-cardinality --no-tui` *(fails in this local environment with datadog-intake dump EOF during collect_data; CI remains source of truth)*

## Notes
- I attempted a local run of `adp-config-check-no-warn` on `--runtime mac`, but this environment does not have the Core Agent binary at `/tmp/saluki-dda/datadog-agent/bin/agent/agent`, so full integration reproduction/verification is CI-only here.

### DIFF
--- a/test/correctness/cases/dsd-origin-detection-matrix/config.yaml
+++ b/test/correctness/cases/dsd-origin-detection-matrix/config.yaml
@@ -17,7 +17,6 @@ comparison:
     - DD_DATA_PLANE_DOGSTATSD_ENABLED=true
     - DD_DATA_PLANE_REMOTE_AGENT_ENABLED=true
     - DD_DATA_PLANE_STANDALONE_MODE=false
-    - DD_AGGREGATE_CONTEXT_LIMIT=500000
 
 # Each variant expands into an independent correctness test. All variants share the same base
 # agent config (datadog-base.yaml) and millstone corpus; only the origin detection settings differ.
--- a/bin/correctness/panoramic/src/unix_runner.rs
+++ b/bin/correctness/panoramic/src/unix_runner.rs
@@ -395,11 +395,6 @@ fn build_core_agent_forced_env(
         forced.push(("DD_USE_DOGSTATSD", "false".to_string()));
     }
 
-    // Do not force Core Agent OTLP receiver endpoints here.
-    //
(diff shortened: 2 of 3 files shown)


test/correctness/cases/dsd-origin-detection-matrix/config.yaml
```diff
@@ -17,7 +17,6 @@ comparison:
     - DD_DATA_PLANE_DOGSTATSD_ENABLED=true
     - DD_DATA_PLANE_REMOTE_AGENT_ENABLED=true
     - DD_DATA_PLANE_STANDALONE_MODE=false
-    - DD_AGGREGATE_CONTEXT_LIMIT=500000
```
thieman, line 20: Attempting to address a flake on this test that happened in a previous commit on this PR

tobz (approved):

PR #1856, opened 2026-06-11
chore(common): unify multiple shutdown synchronization types

tobz:
## Summary

This PR unifies the various shutdown synchronization types across multiple crates into a single pair of types in `saluki-common`.

Prior to this PR, we had five distinct shutdown types/pairs of types:

1. `DynamicShutdownCoordinator`/`DynamicShutdownHandle`: allowed for registering shutdown handles dynamically
2. `ComponentShutdownCoordinator`/`ComponentShutdownHandle`: basically `DynamicShutdownCoordinator` with a different name (my brain wasn't firing on all cylinders when I wrote this, clearly)
3. `ShutdownHandle`/`ProcessShutdown`: a paired shutdown trigger (like a oneshot sender/receiver pair) specifically for use in `saluki_core::runtime`
4. `ShutdownTrigger`/`ShutdownHandle`: a paired shutdown trigger added to `saluki-common` for general use where needed
5. `ShutdownHandle`, exposed for use with `HttpServer` to trigger shutdown

Conceptually, these are all very similar:

- 1 and 2 are basically identical: support registering multiple shutdown handles, then triggering them all at once... potentially waiting for them to complete their shutdown
- 3 and 4 are also basically identical, _and_ they're a subset of 1 and 2: a trigger and a handle, except only one handle is ever registered
- 5 is just one half of the pair in 3 and 4

So I decided to unify things. In this PR, I've introduced a new pair of types for driving the sender/receiver sides of the equation: `ShutdownCoordinator` and `ShutdownHandle`. All instances of the previous five shutdown types have been replaced with these new types, fully unifying how we handle shutdown triggering/propagation. In no particular order:

- `ShutdownCoordinator` supports multiple handles being registered
- `ShutdownCoordinator` does a non-blocking shutdown trigger on drop, ensuring shutdown isn't missed when we hit errors/panics and don't reach the happy path shutdown call
- `ShutdownCoordinator` supports both a blocking and non-blocking shutdown call: the blocking call (insofar as it waits, not that it's literally synchronous code) will wait until all shutdown handles have been dropped before returning
- `ShutdownHandle` supports both synchronous and asynchronous code by exposing a method for synchronous code to check if shutdown has been triggered, while also being usable as a `Future` that resolves when shutdown is triggered
- `ShutdownHandle` also exhibits fused/latching behavior in its `Future` implementation, easing potential bugs arising from misuse (re-polling resolved futures is _generally_ bad)
[...]

PR #1874, opened 2026-06-15
enhancement(core): add support for dynamically added child processes in `Supervisor`

tobz:
## Summary

This PR adds support for dynamically adding child processes to a `Supervisor`, along with other changes to better align our supervisor behavior with the behavior of the granddaddy of all supervisors, `supervisor.erl` in Erlang/OTP.

Currently, `Supervisor` is designed to be configured _before_ running, and then, once running, only ever manage the workers it was configured with. This is sufficient for almost all of our usage, but it makes it impossible to support a small but reasonable use case: supervision of dynamically spawned tasks, like network connection handlers (DSD stream handlers, for example).

This PR introduces support for dynamically spawned workers, along with some other behavioral uplifts, to support a higher-fidelity implementation of supervisors:

- `SupervisorHandle`, a new type that can be used to dynamically spawn workers at runtime, when a supervisor is already running
- `ShutdownMode`, to support concurrently stopping processes during shutdown, which is useful in a world where a supervisor may be supervising _many_ dynamic child processes
- support for "significant" children and automatic shutdown, which allows triggering a supervisor to shutdown completely if special child processes exit (ab)normally at runtime, providing a loose form of _structured concurrency_

What this will enable us to do is actually spawn each topology component as a dedicated supervisor, with the traditional component task as a significant child in that supervisor, and then give the component a handle to its supervisor so it can spawn whatever other child processes it requires. In doing so, we can ensure all of those descendant child processes are shutdown when the component task itself exits (ab)normally.. but that we can track their association (at a high level) to the component that spawned them, rather than really having no good idea where some arbitrary async task originated from.

Most of the code in this PR is tests: we've altered some of the restart/shutdown behavior not only to support dynamic child processes, but also to properly match the behavior exhibited by Erlang/OTP's `supervisor.erl`.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

A lot of new and existing unit tests.

## References

DADP-2


tobz:
* **#2029** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2029?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1942** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1942?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1874** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1874?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1874?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1873** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1873?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1860** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1860?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1856** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1856?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/saluki-core/src/runtime/worker_state.rs
```diff
@@ -0,0 +1,283 @@
...
+                    tokio::time::sleep(timeout)
+                }
+                ShutdownStrategy::Brutal => {
+                    debug!(worker_id, "Forcefully aborting process.");
+                    abort_handle.abort();
+
+                    // We have to return a future that never resolves, since we're already aborting it. This is a little
+                    // hacky but it's also difficult to do an optional future, so this is what we're going with for now.
+                    tokio::time::sleep(Duration::MAX)
+                }
```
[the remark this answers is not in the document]
tobz: By itself, this wouldn't work because it doesn't match the return type on the other side of the branch... and so it would require us to box both sides, creating an allocation just for a slightly more technically correct sleep.

Not worth it. 🤷🏻

lib/saluki-core/src/runtime/supervisor.rs
```diff
@@ -531,106 +782,210 @@ impl Supervisor {
...
+            dynamic: true,
+        };
+        match worker_state.add_worker(id, &entry.spec) {
+            Ok(()) => {
+                if config.significant {
+                    *significant_remaining += 1;
+                }
+                self.active.fetch_add(1, Ordering::Relaxed);
+                children.insert(id, entry);
+                let _ = ack.send(Ok(()));
```
aqian01, line 828: I think this might cause an issue since we are sending ack when the child is initiated instead of when the child starts running. In other areas, we specify that we wait until the child starts running.

e.g.
`supervisor.rs` line 546
```
/// Spawns a new dynamic worker (temporary, non-significant) and waits until the supervisor has started it.
    ///
    /// Returns once the supervisor has registered and begun running the child. Use [`spawn_with`](Self::spawn_with) to
    /// configure the child's restart policy or significance.
```

Added a quick unit test to confirm this
```
async fn dynamic_spawn_acknowledges_before_async_initialization_failure() {
        let sup = Supervisor::new("dyn-sup").unwrap();
        let handle = sup.handle();
        let (_tx, run) = run_supervisor_with_trigger(sup).await;
        wait_running(&handle).await;

        let child_id = handle.spawn(MockWorker::init_failure("bad-dyn")).await;
        assert!(
            child_id.is_ok(),
            "spawn currently acknowledges registration before async initialization completes"
        );

        let result = timeout(Duration::from_secs(2), run).await.unwrap().unwrap();
        match result {
            Err(SupervisorError::FailedToInitialize { child_name, .. }) => {
                assert_eq!(child_name, "bad-dyn");
            }
            other => panic!("expected FailedToInitialize, got: {:?}", other),
        }
    }
```

tobz: This is intentional: the goal is to signal whether or not we were even able to get to the point of spawning the child process, not that it specifically was able to reach its running state.

This is similar to, say, `tokio::spawn`, where if you want to know that the future you spawned actually started executing, you need to have some specific mechanism used by the future to signal that back.

tobz: Woops, cut off my comment too early: I'll update the wording to highlight that this doesn't mean the process is running yet when the call returns.

tobz: Updated the docs in 33bac878.

webern (approved):
My visual reading of it isn't worth much, I more-or-less understand what's going on. Gave it an opus max pass with no bug findings as well.

aqian01 (approved):

Issue #1681, opened 2026-05-18
Investigate aggregator config knobs: buffer size, stop timeout, tags store, and parallel flush settings.


jszwedko:
## Problem

The core agent exposes several config keys for tuning its Go channel-based aggregator that have no confirmed equivalent in ADP:

- `aggregator_buffer_size` — depth of the channel buffers between listeners and aggregator workers
- `aggregator_stop_timeout` — timeout (seconds) for the aggregator to flush on shutdown
- `aggregator_use_tags_store` — toggle for a shared ref-counted tag deduplication store (memory vs CPU tradeoff)
- `aggregator_flush_metrics_and_serialize_in_parallel_buffer_size` — buffer size for the parallel flush+serialize goroutine pipeline
- `aggregator_flush_metrics_and_serialize_in_parallel_chan_size` — channel size for the same pipeline

ADP uses an async Tokio topology rather than Go goroutine pools. These knobs may have no direct mapping, or they may correspond to existing ADP config values under different names.

## What needs to happen

For each key, determine whether ADP's aggregate transform or topology wiring has an equivalent tunable. Document the mapping (or absence) for each. If ADP hardcodes values that the core agent exposes as config, note what the hardcoded values are so operators can calibrate expectations.

jszwedko:
Closed by https://github.com/DataDog/saluki/pull/1876

PR #1876, opened 2026-06-16
feat(config): support configurable ADP stop timeout

aqian01:
## What changed

Document the ADP behavior for the core Agent aggregator tuning keys tracked by #1681:

- `aggregator_buffer_size`
- `aggregator_use_tags_store`
- `aggregator_flush_metrics_and_serialize_in_parallel_buffer_size`
- `aggregator_flush_metrics_and_serialize_in_parallel_chan_size`

These keys still have no direct ADP equivalent and remain documented as no-ops.

Also support shutdown timeout compatibility:

- `aggregator_stop_timeout` and `forwarder_stop_timeout` now configure ADP's topology-wide graceful shutdown timeout.
- The default is `aggregator_stop_timeout + forwarder_stop_timeout`, which is `2 + 2 = 4` seconds by default.
- `data_plane.stop_timeout` is a new ADP-specific override for that combined timeout.

## Why

Most aggregator knobs tune Go channel-based internals in the core Agent. ADP uses Saluki topology interconnects, `EventsBuffer` batches, and context/tag-set sharing instead.

The stop-timeout keys are different: they map cleanly to ADP's existing topology shutdown timeout, which was previously hardcoded to 30 seconds.

## Behavior change

ADP's topology graceful shutdown timeout now defaults to 4 seconds instead of the previous hardcoded 30 seconds, matching the core Agent's default shutdown budget. Operators can set `data_plane.stop_timeout` to override it directly.

## Validation

- `make fmt`
- `make check-docs`
- `cargo check -p saluki-core -p agent-data-plane`
- `cargo test -p agent-data-plane --bin agent-data-plane config::tests`
- `cargo test -p datadog-agent-config-testing`
- `git diff --check`

### DIFF
--- a/bin/agent-data-plane/src/config.rs
+++ b/bin/agent-data-plane/src/config.rs
@@ -1,5 +1,7 @@
+use std::time::Duration;
+
 use saluki_config::GenericConfiguration;
-use saluki_error::GenericError;
+use saluki_error::{generic_error, GenericError};
 use saluki_io::net::ListenAddress;
 
 /// General data plane configuration.
@@ -9,6 +11,7 @@ pub struct DataPlaneConfiguration {
     standalone_mode: bool,
     use_new_config_stream_endpoint: bool,
     remote_agent_enabled: bool,
+    stop_timeout: Duration,
     api_listen_address: ListenAddress,
     secure_api_listen_address: ListenAddress,
     checks: DataPlaneChecksConfiguration,
@@ -37,6 +40,7 @@ impl DataPlaneConfiguration {
                 .try_get_typed("data_plane.use_new_config_stream_endpoint")?
                 .unwrap_or(true),
             remote_agent_enabled: config.try_get_typed("data_plane.remote_agent_enabled")?.unwrap_or(true),
+            stop_timeout: topology_stop_timeout_from_configuration(config)?,
             api_listen_address: config
                 .try_get_typed("data_plane.api_listen_address")?
                 .unwrap_or_else(|| ListenAddress::any_tcp(5100)),
@@ -67,6 +71,11 @@ impl DataPlaneConfiguration {
     /// Returns `true` if the data plane should register as a remote agent.
     pub const fn remote_agent_enabled(&self) -> bool {
         self.remote_agent_enabled
+    }
+
+    /// Returns the topology shutdown timeout.
+    pub const fn stop_timeout(&self) -> Duration {
+        self.stop_timeout
     }
 
     /// Returns a reference to the API listen address
@@ -153,6 +162,19 @@ impl DataPlaneConfiguration {
         // - OTLP is enabled and not in proxy mode or proxy mode is enabled and proxy traces are disabled
         self.otlp().enabled() && (!self.otlp().proxy().enabled() || !self.otlp().proxy().proxy_traces())
     }
+}
+
+fn topology_stop_timeout_from_configuration(config: &GenericConfiguration) -> Result<Duration, GenericError> {
+    if let Some(stop_timeout_secs) = config.try_get_typed::<u64>("data_plane.stop_timeout")? {
+        return Ok(Duration::from_secs(stop_timeout_secs));
+    }
+
+    let aggregator_stop_timeout_secs = config.try_get_typed::<u64>("aggregator_stop_timeout")?.unwrap_or(2);
+    let forwarder_stop_timeout_secs = config.try_get_typed::<u64>("forwarder_stop_timeout")?.unwrap_or(2);
+
+    Duration::from_secs(aggregator_stop_timeout_secs)
+        .checked_add(Duration::from_secs(forwarder_stop_timeout_secs))
+        .ok_or_else(|| generic_error!("Topology stop timeout overflowed."))
 }
 
 /// Checks-specific data plane configuration.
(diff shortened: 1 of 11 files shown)


docs/agent-data-plane/configuration/dogstatsd.md
```diff
@@ -89,6 +89,84 @@ architecture is fundamentally different or the feature is platform-specific.
...
+is enabled, the same timeout also bounds draining in-flight DogStatsD time sampler batches before
+that final flush.
+
+ADP has no equivalent aggregator-specific timeout. Shutdown is coordinated by the Saluki topology:
+sources stop first, downstream inputs close, and the aggregate transform performs its final flush
+when its input stream ends. Whether open aggregation windows are included is controlled by
+`aggregate_flush_open_windows`, also aliased as `dogstatsd_flush_incomplete_buckets`.
+
+The ADP shutdown grace period is currently topology-wide and hardcoded to 30 seconds. Setting
+`aggregator_stop_timeout` has no effect in ADP.
```
jszwedko: I think we should support, at least, configuring the global shutdown grace period (as `stop_timeout` to match existing config option naming). Maybe it could default to `forwarder_stop_timeout` + `aggregator_stop_timeout`? The comment on `aggregator_stop_timeout` seems to indicate users can generally expect the Agent to shutdown in the summation of both of those timeouts:

https://github.com/DataDog/datadog-agent/blob/0ff79b8f0b1e88ec7ee7983df8e473c831a1f70f/pkg/config/config_template.yaml#L283-L294

How difficult / awkward would it be to support the `aggregator_stop_timeout` and `forwarder_stop_timeout`?

jszwedko (commented):
One comment about supporting configuration of the shutdown timeouts but otherwise 👍

docs/agent-data-plane/configuration/dogstatsd.md
jszwedko: Confirming I do see a decent amount of use of `forwarder_timeout` by Agents. I see less use of `aggregator_stop_timeout`.

aqian01: I think this would be pretty easy to support, adding it now.

lib/datadog-agent/config-testing/src/config_registry/data_plane.rs
```diff
@@ -4,6 +4,14 @@ use super::schema;
 #[allow(unused_imports)]
 use super::*;
 
+static DATA_PLANE_STOP_TIMEOUT_SCHEMA: SchemaEntry = SchemaEntry {
+    schema: Schema::Saluki,
+    yaml_path: "data_plane.stop_timeout",
+    env_vars: &[],
+    value_type: ValueType::Integer,
+    default: Some("4"),
```
tobz: This doesn't actually a default though, right? 🤔

tobz (changes requested):

lib/datadog-agent/config-testing/src/config_registry/data_plane.rs
aqian01: Yeah true it just defaults to `aggregator_stop_timeout` +  `forwarder_stop_timeout` which if unset is 4

bin/agent-data-plane/src/config.rs
```diff
@@ -155,6 +164,19 @@ impl DataPlaneConfiguration {
     }
 }
 
+fn topology_stop_timeout_from_configuration(config: &GenericConfiguration) -> Result<Duration, GenericError> {
+    if let Some(stop_timeout_secs) = config.try_get_typed::<u64>("data_plane.stop_timeout")? {
+        return Ok(Duration::from_secs(stop_timeout_secs));
+    }
+
+    let aggregator_stop_timeout_secs = config.try_get_typed::<u64>("aggregator_stop_timeout")?.unwrap_or(2);
+    let forwarder_stop_timeout_secs = config.try_get_typed::<u64>("forwarder_stop_timeout")?.unwrap_or(2);
```
jszwedko, line 173: `forwarder_stop_timeout` default is actually 20. In practice the Agent should always send it though.

bin/agent-data-plane/src/config.rs
```diff
@@ -155,6 +164,19 @@ impl DataPlaneConfiguration {
     }
 }
 
+fn topology_stop_timeout_from_configuration(config: &GenericConfiguration) -> Result<Duration, GenericError> {
+    if let Some(stop_timeout_secs) = config.try_get_typed::<u64>("data_plane.stop_timeout")? {
```
jszwedko, line 168: Just a note that we'll need to add this new option to datadog-agent too, after this is merged. See https://github.com/DataDog/datadog-agent/pull/51522 for an example.

bin/agent-data-plane/src/config.rs
aqian01: I think you saw the [forwarder_timeout](https://github.com/DataDog/datadog-agent/blob/52dca247f01e3c54d1d8a441f7e6ed0663b4d6bd/pkg/config/config_template.yaml#L289-L293) as 20 and not the [forwarder_stop_timeout](https://github.com/DataDog/datadog-agent/blob/52dca247f01e3c54d1d8a441f7e6ed0663b4d6bd/pkg/config/config_template.yaml#L309-L320) as 20 😅

bin/agent-data-plane/src/config.rs
aqian01: Sounds good

bin/agent-data-plane/src/config.rs
jszwedko: Doh! You are right 😅

bin/agent-data-plane/src/config.rs
jszwedko: Note, I think the Agent will always send `data_plane.stop_timeout` via the configstream when we make that change so we may need to put the default to the addition of `aggregator_stop_timeout` and `forwarder_stop_timeout` over there too.

jszwedko (approved):

jszwedko:
(Approving the config surface; I'll defer to @tobz for the implementation)

tobz (approved):

PR #1877, opened 2026-06-16
chore(integration): test dynamic config propagation

webern:
## Human Summary

This test sets up ADP, then changes the `log_level` value by calling the core Agent endpoint, then asserts that the changed value has propagated to ADP's config map by calling ADP's `/config` endpoint. A new type was added to test config for this, an `Action` trait. So an integration test can consist of `Actions` and `Assertions` in whatever order.

Think of this test as a starting point from which we may continue to build more powerful configuration (and other API interaction) tests.

## AI Summary

Adds Panoramic support for sequential action steps and a targeted ADP `/config` polling assertion, then uses them in a new integration test that verifies a Core Agent runtime config mutation propagates into ADP's `/config` output.

The new test runs ADP through the Core Agent, verifies the initial `log_level`, changes it through the Core Agent config API, and polls ADP until the updated value appears.

## Change Type
- [x] Non-functional

## How did you test this PR?

```bash
cargo check -p panoramic
target/debug/panoramic run -d test/integration/cases --tests adp-dynamic-config-propagation --runtime linux --no-tui
make fmt
```

Human here: also I changes the assertion to `desmug` to see that it would fail and changed it back to `debug` to see it pass. So the assertion is good.

## References

Supports #1788


bin/correctness/panoramic/src/actions/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+                ctx.core_agent_auth_token_path.as_ref(),
+                self.timeout,
+                &ctx.cancel_token,
+                &ctx.container_exit_token,
+            )
+            .await
+        } else if ctx.target_is_windows() {
+            set_config_in_windows_container(&ctx.container_name, &self.endpoint(), &self.key, &value).await
+        } else {
+            set_config_in_linux_container(&ctx.container_name, &self.endpoint(), &self.key, &value).await
```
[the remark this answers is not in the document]
webern: pushed

bin/correctness/panoramic/src/config.rs
```diff
@@ -354,6 +392,30 @@ pub enum HttpStatusMatcher {
...
 }
 
+impl ActionConfig {
+    /// Replaces `{{PANORAMIC_DYNAMIC_*}}` placeholders in string fields with resolved values.
+    pub fn resolve_dynamic_vars(&mut self, vars: &HashMap<String, String>) {
+        match self {
+            ActionConfig::CoreAgentConfigSet { key, endpoint, .. } => {
+                crate::dynamic_vars::resolve_placeholders(key, vars);
+                crate::dynamic_vars::resolve_placeholders(endpoint, vars);
+            }
```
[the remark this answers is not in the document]
webern: pushed

bin/correctness/panoramic/src/actions/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+        } => Ok(Box::new(CoreAgentConfigSetAction::new(
+            key.clone(),
+            value.clone(),
+            endpoint.clone(),
+            timeout.0,
+        ))),
+    }
+}
+
+struct CoreAgentConfigSetAction {
```
tobz: Since I can inevitably see us wanting to add more "actions", can we just move this action code out into its own file from the get-go?

test/integration/cases/adp-config-dynamic-basic/config.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+container:
+  exposed_ports:
+    - "55101/tcp"
+
+assertions:
+  - type: adp_config_key_equals
+    key: log_level
+    value: info
+    timeout: 60s
```
tobz: I will say that it's a bit muddy to go from `type: foo` meaning "assertion `foo`" to then having `action: bar` meaning "action `bar`".

Would be good to switch to `assertion: foo` in a follow-up PR, I think.

tobz (changes requested):

test/integration/cases/adp-config-dynamic-basic/config.yaml
webern: But I think you want that ordering: assert this, then action that, then assert this, in procedural order. Am I missing something?

Edit: oh, I see... Maybe rename `assertions` (the array) to something like, hmmm, procedures? I see the ugliness here now, but I think we want to make it procedural, I think the model is just bad.

webern: Pushed a better model for this: procedure with assertions and actions.

tobz: Yeah, it was more just about having the config structure look a little more consistent, if possible, for how we define both assertions and actions in the list of steps.

I think the changes you made (call them "procedures", and make it consistent in how we select either an assertion or action) are right on the money. 🎯

tobz (approved):

PR #1885, opened 2026-06-18
chore(config): experiment - mapless config system implementation 3a

webern:
## Summary

Go away.

## Change Type

## How did you test this PR?
I am using CI for this purpose.

## References

#1788


webern:
This implementation was inferior. It has a few findings worth keeping but not worth spending any more resources on it.

PR #1886, opened 2026-06-18
chore(config): experiment - mapless config system implementation 3b

webern:
## Summary

This is a prototype of a system that eliminates untyped, stringly access to a generic configuration map and produces a strongly typed barrier between `agent-data-plane` and Datadog Agent config. It encapsulates knowledge of Datadog Agent configuration into a single system, isolated from `agent-data-plane` and provides the extension point where other configuration dialects may be inserted (such as OTeL, for example).

This PR is not mergeable, but I need to keep it here and keep it healthy and rebased. The "real" PRs will essentially burn down this diff in more manageable chunks.

## Change Type

## How did you test this PR?
This PR is basically a test, in itself, of the design and the ability to pass CI with a prototype of the design.

## References

#1788


webern:
Closing. This is no longer needed to prove the concept...

PR #1889, opened 2026-06-18
enhancement(ci): build ADP binaries through `cargo-auditable` for embedded SBOM metadata

tobz:
## Summary

This PR builds ADP binaries in CI using `cargo-auditable` to add embedded SBOM metadata.

Currently, when it comes to supply chain provenance of ADP, we only generate license data for our third-party dependencies. However, we lack metadata about the _specific_ version of dependencies that were used to build ADP, which leaves a critical gap in the ability to perform vulnerability scanning on ADP.

This PR introduces the use of [`cargo-auditable`](https://github.com/rust-secure-code/cargo-auditable), which is a Cargo plugin for building binaries and then embedding their SBOM metadata in a linker section of the binary so that it travels along _directly in the binary itself: a small (single digit kilobytes) embedded blob of JSON that accurately lists which dependencies were used and their versions. This is natively supported by tools like Trivy, which can discover the metadata and use it when performing vulnerability scans.

We've updated our CI build jobs/scripts/tooling to use `cargo-auditable` such that any ADP binary we generate -- whether for an internal build or intended for a published artifact -- has the embedded SBOM metadata. Local developer builds stay as-is because there's no real need for auditing local builds. 🤷🏻 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

- [ ] Build ADP in CI for all three supported platforms and manually audit their resulting binaries for the embedded SBOM metadata.

## References

DADP-2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,10 +65,7 @@ export CARGO_TOOL_VERSION_cargo-sort ?= 1.0.9
 export CARGO_TOOL_VERSION_dummyhttp ?= 1.1.0
 export CARGO_TOOL_VERSION_cargo-machete ?= 0.9.1
 export CARGO_TOOL_VERSION_rustfilt ?= 0.2.1
-# cargo-auditable embeds an SBOM (the dependency tree) into the built binary. Used only by CI ADP
-# builds (host + Linux Docker) so released artifacts are auditable; local builds stay plain.
-# Keep >= 0.6.2: older versions panic when a rustc wrapper (buildcache) probes them with `rustc -vV`.
-export CARGO_TOOL_VERSION_cargo-auditable ?= 0.7.5
+export CARGO_TOOL_VERSION_cargo-auditable ?= 0.7.4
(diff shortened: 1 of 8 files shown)


jszwedko (approved):

Makefile
```diff
@@ -55,6 +65,7 @@ export CARGO_TOOL_VERSION_cargo-sort ?= 1.0.9
 export CARGO_TOOL_VERSION_dummyhttp ?= 1.1.0
 export CARGO_TOOL_VERSION_cargo-machete ?= 0.9.1
 export CARGO_TOOL_VERSION_rustfilt ?= 0.2.1
+export CARGO_TOOL_VERSION_cargo-auditable ?= 0.7.4
```
jszwedko, line 68: Not related to this PR specifically, but it might be worth configuring renovate to open PRs to update these build tools as new versions are available.

PR #1894, opened 2026-06-19
chore(config): rename saluki-config to saluki-config-tools

webern:
## Summary

`saluki-config` is a utilities crate for configuration mechanics (providers, loaders,
dynamic watchers). A follow-on PR series introduces typed configuration structs, at which
point the name `saluki-config` would conflate with those new crates. This rename preempts
that ambiguity.

## Change Type
- [x] Non-functional (chore, refactoring, docs)

## How did you test this PR?

This is a mechanical rename with no logic changes. `make fmt && make check-all` passed.

## References

PR #1895, opened 2026-06-19
chore(config): scaffold the configuration translation system

webern:
## Human Summary

This one adds the crates and a single example of "native" configuration (`dogstatsd::Config`). Nothing is introduced into production code yet in this PR and some things need a `dead_code` guard. The "witness" trait is introduced here and it is shown how DogStatsD configuration flows from from DatadogConfiguration to the native struct through the witness trait implementation.

## AI Summary

Introduces the scaffolding for the typed configuration translation system that will
replace ADP's direct access to the raw configuration map, and wires the first
component config through it end-to-end. Purely additive: nothing in the binary
consumes this yet, so there is no behavior change.

New crates:

- `saluki-component-config` - leaf crate for component-native config structs (resolved,
  source-agnostic data, no source key names or serde) and the `DynamicConfig<T>`
  fixed-or-live runtime handle. `DogStatsDConfig` is the first such struct.
- `agent-data-plane-config` - the ADP-native model: `SalukiConfiguration { control,
  components }`, plus `SalukiOnlyConfiguration` and its `seed()` for Saluki-schema-only
  values.
- `agent-data-plane-config-system` - the translation facade. `Translator` implements the
  generated witness trait; `translate()` seeds a base from the Saluki-only config, then
  drives the Datadog source over it to produce a `SalukiConfiguration`.

Also adds the generated witness trait `DatadogConfigConsumer` and its fallible `drive`
to `datadog-agent-config`. Every supported Datadog key gets a generated `consume_<key>`
method, so the translator cannot silently drop a key: adding a key is a compile-time
obligation to choose a destination.

Scope limits, by design:

- Only the DogStatsD source keys have a native destination so far; the remaining witness
  methods are temporary no-ops until their destination structs land in later changes.
- The remaining component-native structs, the full native model, and the config-system
  lifecycle/loading follow incrementally.
- No architecture tests in this change.

## Change Type

## How did you test this PR?

- `cargo test` across the three new crates (`DynamicConfig` handle behavior, `seed()`
  assignment, and seed-then-drive translation including the clobber-not-merge semantics
  and error surfacing).
- `make fmt`, `make check-all` (clippy, Vale, unused-deps, cargo-deny), and
  `make build-schema-overlay` all clean.

## References

- Progresses: #1788
- Stacked on #1894


webern:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/DataDog/saluki/1895?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1919** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1919?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1912** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1912?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1898** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1898?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1895** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1895?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1895?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1894** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1894?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1863** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1863?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/saluki-component-config/src/dogstatsd.rs
```diff
@@ -0,0 +1,289 @@
+//! Component-native configuration for the DogStatsD source.
```
tobz, line 1: Which copy of the configuration is intended to survive after cutover: this one (as part of this crate) or the version in `saluki-components`?

I have reservations about separating the configuration of a component from the actual component code.

lib/saluki-component-config/src/dynamic.rs
```diff
@@ -0,0 +1,185 @@
...
+            DynamicConfig::Fixed(_) => pending().await,
+            DynamicConfig::Live { rx, .. } => {
+                // A closed channel resolves immediately with an error; treat it as "no further
+                // changes" by parking so callers do not spin on a dead channel.
+                if rx.changed().await.is_err() {
+                    pending().await
+                }
+            }
+        }
+    }
```
tobz, line 137: This bifurcation seems ripe for TOCTOU: I think it's generally worth ensuring watchers get a chance to observe any individual change, rather than just the latest change.

tobz (commented):
I left two questions/comments, but my overarching feeling is that I'm worried that the unidirectional data flow -- datadog config -> translator -> opaque saluki to hold it all -- is forcing us into a state where we will end up just having a slightly more generalized version of what we already have?

For example, the way it's written now, the witness trait implementation needs a place to shove the value as it witnesses/consumes it, so we almost definitionally end up with a somewhat mirrored version of the source Datadog configuration. If there's a configuration setting that's needed for multiple components, our consume method for that setting would have to go and specifically shove it into multiple different places _or_ keep it in a suitably generic struct that, as mentioned above, doesn't end up looking much different than the source.

This feels like it could end up becoming fairly brittle, where, for example, three components all needing to access the same five values means each of those five consume methods has to update three distinct versions of that field (in each component config), extrapolate that out across all the potential areas where this might be true.

Admittedly, I think it's hard to get a good grip on the design without also seeing more of it, but narrowly focused on _this_ PR, it's hard to wrap my head around how this is our stepping stone to breaking free from rigid configuration.

webern:
Closing, though I will try to remember to reference post design-change since this has the feedback that precipitated the design change. But I think the PR-pathway will look a little different and it won't be terribly convenient to reuse this PR.

Issue #1368, opened 2026-04-13
Investigate migrating `agent_ipc_endpoint` to the standard `agent_ipc.*` settings.


webern:
Note: the human doesn't understand this one and suspects that the machine might not understand what it's saying either. Here's what it says...

ADP uses a custom [`agent_ipc_endpoint`][s1] setting (default `https://127.0.0.1:5001`) to control where it connects to the core agent over IPC — to register as a remote agent and receive config. This key was originally added because the core agent's [`agent_ipc.host`/`agent_ipc.port`][a1] settings were not enabled by default when ADP first needed IPC connectivity. Rather than requiring changes to the core agent's config, ADP introduced its own key so it could set the endpoint independently.

ADP's default of port 5001 actually connects to the core agent's [`cmd_port`][a2] (default 5001), not the `agent_ipc.*` listener. The `agent_ipc.*` namespace is a separate, newer IPC mechanism that [defaults to disabled][a3] (`agent_ipc.port` defaults to `0`, and the listener returns early when port is 0). So the current setup works — ADP connects to `cmd_port` which is always enabled.

The core agent now exposes the full [`agent_ipc.*`][a1] namespace (`agent_ipc.host`, `agent_ipc.port`, `agent_ipc.socket_path`, `agent_ipc.use_socket`, `agent_ipc.config_refresh_interval`). Investigate whether ADP should migrate to this newer mechanism or continue using `cmd_port`. If migrating, the Agent Runtimes team would need to enable `agent_ipc.*` by default first.

[a1]: https://github.com/DataDog/datadog-agent/blob/141f12062790/pkg/config/setup/common_settings.go#L1121-L1125
[a2]: https://github.com/DataDog/datadog-agent/blob/141f12062790/pkg/config/setup/common_settings.go#L1120
[a3]: https://github.com/DataDog/datadog-agent/blob/141f12062790/comp/api/api/apiimpl/listener/common.go#L71-L74
[s1]: https://github.com/DataDog/saluki/blob/e9373ef98013/lib/saluki-env/src/helpers/remote_agent/client.rs#L55-L64

jszwedko:
Closed by https://github.com/DataDog/saluki/pull/1900

PR #1900, opened 2026-06-22
fix(agent-data-plane): remove legacy agent IPC endpoint

aqian01:
## What changed

Removes the ADP-only `agent_ipc_endpoint` override and makes `cmd_port` the single supported Core Agent CMD API port used for ADP gRPC IPC.

This keeps the same default endpoint behavior by defaulting `cmd_port` to `5001`, and updates the generated ADP config docs to describe `cmd_port` more precisely.

## Why

Issue #1368 asks whether ADP should migrate from `agent_ipc_endpoint` to the standard `agent_ipc.*` settings. Review feedback pointed out that `agent_ipc_endpoint` appears to have been an internal development-era misalignment rather than a setting we should keep.

ADP should use `cmd_port` today, not `agent_ipc.*`, because the Core Agent CMD API server hosts the gRPC services ADP uses.

Code proof:

- Saluki now builds the remote-agent endpoint from `cmd_port`: https://github.com/DataDog/saluki/blob/andrewq/document-adp-cmd-port-ipc/lib/datadog-agent/commons/src/ipc/config.rs#L205-L209
- Core Agent defaults `cmd_port` to `5001`, while `agent_ipc.port` and `agent_ipc.config_refresh_interval` default to `0`: https://github.com/DataDog/datadog-agent/blob/2e29f97b26737c6217babd4ea629e0f8bdb1aed2/pkg/config/setup/common_settings.go#L1180-L1188
- The CMD API server mounts the gRPC server, which registers `AgentSecure` and `RemoteAgent`: https://github.com/DataDog/datadog-agent/blob/2e29f97b26737c6217babd4ea629e0f8bdb1aed2/comp/api/api/apiimpl/server_cmd.go#L35-L64, https://github.com/DataDog/datadog-agent/blob/2e29f97b26737c6217babd4ea629e0f8bdb1aed2/comp/api/grpcserver/impl-agent/grpc.go#L126-L147
- The separate IPC API server only mounts `/config/v1/` HTTP routes: https://github.com/DataDog/datadog-agent/blob/2e29f97b26737c6217babd4ea629e0f8bdb1aed2/comp/api/api/apiimpl/server_ipc.go#L28-L44

## Validation

- `cargo build -p datadog-agent-config-testing -p datadog-agent-config`
- `make fmt`
- `cargo test -p datadog-agent-commons ipc::config`
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1084,7 +1084,6 @@ dependencies = [
  "backon",
  "datadog-protos",
  "futures",
- "http-serde-ext",
  "pin-project-lite",
  "rustls",
  "rustls-pki-types",
--- a/docs/agent-data-plane/configuration/dogstatsd.md
+++ b/docs/agent-data-plane/configuration/dogstatsd.md
@@ -566,7 +566,7 @@ compressed wire payload bytes.
 | `cluster_agent.enabled`                                        | Enable Cluster Agent communication                 |
 | `cluster_agent.kubernetes_service_name`                        | Cluster Agent Kubernetes service name              |
 | `cluster_agent.url`                                            | Cluster Agent HTTPS endpoint                       |
-| `cmd_port`                                                     | Datadog Agent IPC/CMD API port                     |
+| `cmd_port`                                                     | Core Agent CMD API port for ADP gRPC IPC           |
 | `cri_connection_timeout`                                       | CRI container runtime connection timeout (s)       |
 | `cri_query_timeout`                                            | CRI container runtime query timeout (s)            |
 | `data_plane.api_listen_address`                                | Unprivileged API listen address                    |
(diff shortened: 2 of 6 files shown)


webern (commented):
Maybe we should just get rid of ADP-only `agent_ipc_endpoint`. I think it was just a misalignment mistake that it was named `agent_ipc_endpoint` on ADP and `agent_ipc.port` on the Agent.

aqian01:
@webern 
Right now we use the `agent_ipc_endpoint` as a fallback when `cmd_port` isn't set.

```
if let Some(cmd_port) = self.cmd_port {
    ...
} else {
    Ok(self.ipc_endpoint.clone())
}
```
Do we need to check if any customers are using `agent_ipc_endpoint` before removing it?

jszwedko (commented):
Agreed with @webern . Let's drop `agent_ipc_endpoint`. I don't see any usages of it; it may have just been used during development of the IPC feature internally. Otherwise 👍

jszwedko (approved):

PR #1908, opened 2026-06-23
chore(ci): base converged Agent image on -full

jszwedko:
> Stacked on #1906 (`jszwedko/renovate-agent-version-pins`). Review/merge that first; this PR will retarget to `main` automatically once it lands.

## Summary

Switch to the `-full` image which contains both JMX and DDOT for correctness tests and the converged image we publish. This removes the need for the OTLP correctness tests to pin to published Agent `-full` images. It also enables deployment of the converged ADP image on internal clusters that also run DDOT.

## Test plan

- [x] Verified `-full` includes JMX: both `-jmx` and `-full` carry the same `jmxfetch.jar` + Java 11 JRE; `-full` additionally has `otel-agent`.
- [x] Rebuilt the converged `testing-release` image on `-full` and confirmed `otel-agent` (DDOT) is present.
- [x] Ran locally on the `-full` converged image — all pass: `otlp-traces-ottl-filtering`, `otlp-traces-ottl-transform` (baseline now runs DDOT from the converged image), plus `otlp-traces`, `otlp-metrics`, and `dsd-plain` to confirm no regression from the base-image change.


renovate.json5
```diff
@@ -50,25 +50,6 @@
   //     weekly non-major group)
   "osvVulnerabilityAlerts": true,
 
-  // Paths Renovate skips entirely, before any manager (built-in or custom) extracts from them.
```
webern, line 53: Is the renovate diff a mistake in this PR?

webern (commented):
Makes sense to me, but wasn't sure about the renovate diff.

tobz (approved):

renovate.json5
jszwedko: Ah, this is actually expected. Given we no longer have a separate image tag specified in the correctness tests, we no longer need to override the default ignore paths (which include `**/test/**` by default).

PR #1914, opened 2026-06-23
fix(config): publish dynamic updates after applying config

webern:
## Human Summary

Turns out, this might matter for the config system later, and it seems like an isolated and sensible change.

## AI Summary

Dynamic config update events are currently broadcast before the backing `GenericConfiguration` figment map is updated. This means subscribers that react to an event by reading the full config can observe the previous value.

This PR updates the dynamic config updater to:

- compute the changed fields;
- write the rebuilt figment map;
- then broadcast the corresponding change events.

It also adds a regression test that blocks the figment write and verifies no update event is published until the write can complete.

## Change Type

- [x] Bug fix

## How did you test this PR?

Exercised the new test before the fix: `red -> green`

```text
cargo test -p saluki-config update_events_are_sent_after_the_figment_map_is_updated -- --nocapture
cargo test -p saluki-config
```

## References

Kinda supports #1788

### DIFF
--- a/lib/saluki-config/src/lib.rs
+++ b/lib/saluki-config/src/lib.rs
@@ -631,18 +631,22 @@ async fn run_dynamic_config_updater(
         let new_config: figment::value::Value = new_figment.clone().extract().unwrap();
 
         if current_config != new_config {
-            for change in dynamic::diff_config(&current_config, &new_config) {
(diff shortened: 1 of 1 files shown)


tobz (approved):
Yeah, good catch. LGTM. 👍🏻

PR #1920, opened 2026-06-24
chore(datadog-protos): move V3 protocol into a separate crate

StephenWakely:
## Summary

This extracts the V3 Writer implemented in #1175  into a standalone crate so that it can be reused by other Rust based projects.

Ideas on what this crate should be named and if it is in the correct position in the source tree would be welcomed.

I don't think the `generate-api-docs` CI failure is caused by this PR.

## Questions

1. What should the scope be in the PR title?

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Using this crate in a different Rust based project, I setup a pipeline of host metrics -> Datadog. This project can send transactions as both V2 and V3, validating the payloads contain the same metrics.


lib/saluki-components/src/encoders/datadog/metrics/v3/constants.rs
```diff
@@ -1,33 +1,7 @@
...
-pub const SKETCH_BIN_CNTS_FIELD_NUMBER: u32 = 22;
-pub const SOURCE_TYPE_NAME_FIELD_NUMBER: u32 = 23;
-pub const ORIGIN_INFO_FIELD_NUMBER: u32 = 24;
-pub const DICT_UNIT_STR_FIELD_NUMBER: u32 = 25;
-pub const UNIT_REFS_FIELD_NUMBER: u32 = 26;
-
+// Display names for the V3 columns, indexed by their Protocol Buffers field number (from
+// `lib/protos/datadog/proto/agent-payload/intake_v3.proto`, also mirrored in
+// `datadog_metrics_v3::writer` as the `*_FIELD_NUMBER` constants). Index 0 is unused since
+// field numbers start at 1.
```
tobz: Could we actually move this to the new crate as well, with a unit test that asserts that `SOME_CONSTANT_FOR_FIELD_XYZ` is equal to `"NAME_FOR_XYZ"`?

Basically expose something that we can index into with the field number to get the column name, and ensure we test it to keep the mappings in sync.

lib/datadog-agent/datadog-metrics-v3/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "datadog-metrics-v3"
+version = "0.1.0"
+edition = "2021"
+description = "V3 columnar protobuf codec for Datadog metrics"
+license = "Apache-2.0"
```
tobz: Can we actually map these to the workspace settings? Here's how we normally do it:

```
edition = { workspace = true }
license = { workspace = true }
```

Our repo-wide license is already Apache-2.0, so this notation just ensures it stays in sync automatically with the workspace settings... ditto for edition and the repository metadata.

lib/datadog-agent/datadog-metrics-v3/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "datadog-metrics-v3"
```
tobz: Can we actually move this crate to `lib/datadog-agent/metrics-v3`?

We try to avoid stuttering in file paths on disk, but the name of the crate can stay as-is since that only shows up in imports.

lib/datadog-agent/datadog-metrics-v3/src/interner.rs
```diff
@@ -2,7 +2,8 @@
 
 use std::{borrow::Borrow, hash::Hash};
 
-use saluki_common::collections::FastHashMap;
+pub type FastBuildHasher = foldhash::quality::RandomState;
+pub type FastHashMap<K, V> = std::collections::HashMap<K, V, FastBuildHasher>;
```
tobz: Why are these public?

lib/datadog-agent/datadog-metrics-v3/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+mod constants;
+pub mod interner;
```
tobz: Why is `interner` public?

tobz (changes requested):

aqian01:
Since we're extracting this for reuse, should we add `#![deny(missing_docs)]` plus a crate-level doc comment and document the public API?

lib/datadog-agent/datadog-metrics-v3/src/interner.rs
StephenWakely: Copy and paste error..

Cargo.lock
```diff
@@ -1058,9 +1058,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
```
StephenWakely, line 1061: Need this to keep `cargo deny check` happy.

tobz (approved):

PR #1922, opened 2026-06-24
feat(agent-data-plane): document dogstatsd.* as not planned and telemetry.dogstatsd_origin as planned but unsupported

### LINKED ISSUE #1679: Investigate DogStatsD internal telemetry config: `telemetry.dogstatsd.*` and `telemetry.dogstatsd_origin`.
## Problem

The core agent exposes four config keys for fine-tuning DogStatsD internal telemetry:

- `telemetry.dogstatsd.aggregator_channel_latency_buckets` — histogram buckets for DSD-to-aggregator channel lag
- `telemetry.dogstatsd.listeners_channel_latency_buckets` — histogram buckets for listener packet-channel latency
- `telemetry.dogstatsd.listeners_latency_buckets` — histogram buckets for listener processing latency
- `telemetry.dogstatsd_origin` — enables per-UDS-origin breakdown of `dogstatsd.processed` counters

The first three tune Prometheus histogram bucket boundaries for Go-channel-based latency metrics. ADP uses a Tokio async topology with no equivalent channel stage.

`telemetry.dogstatsd_origin` is a real feature gap: it adds per-origin labels to processed-metric counters, which is relevant for debugging origin detection in production.

## What needs to happen

For `telemetry.dogstatsd_origin`: determine whether ADP's OpenMetrics endpoint already exposes per-origin counters, and if not, whether it should.

For the histogram bucket keys: determine whether ADP has equivalent latency histograms that expose configurable buckets, or whether these are not applicable to ADP's architecture.

lucastemb:
## Summary

`telemetry.dogstatsd.*` config keys are not applicable to ADP, and thus, won't be implemented. These config keys define static bucket bounds for histograms intended to capture latency; however, ADP employs dynamic, logarithmically calculated bounds. 

`telemetry.dogstatsd_origin` was implemented by threading the resolved origin as the `dogstatsd.processed` metric. If there is no origin, it is represented as `""`, like the Core Agent, but never omitted.   

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit testing. 

## References

- Closes: https://github.com/DataDog/saluki/issues/1679


lib/saluki-components/src/sources/dogstatsd/metrics.rs
```diff
@@ -1,9 +1,83 @@
...
+            ),
+        };
+
+        inner.counters.insert(origin.to_string(), counters.clone());
+        inner.order.push_back(origin.to_string());
+
+        while inner.order.len() > MAX_ORIGIN_COUNTERS {
+            if let Some(evicted_origin) = inner.order.pop_front() {
+                inner.counters.remove(&evicted_origin);
+            }
```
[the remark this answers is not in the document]
lucastemb: To fix this, this will require a larger change. We'll have to add support for deletion of the global per-origin counter (mimicking Core Agent behavior). The alternative is having a designated "overflow" origin, but then we lose information about the origin (which from my understanding is the _entire_ point of this config key).

lib/saluki-components/src/sources/dogstatsd/metrics.rs
```diff
@@ -1,9 +1,111 @@
...
+                    ("origin", origin.to_string()),
+                ],
+            ),
+        };
+
+        inner.counters.insert(origin.to_string(), counters.clone());
+        inner.order.push_back(origin.to_string());
+
+        while inner.order.len() > MAX_ORIGIN_COUNTERS {
+            if let Some(evicted_origin) = inner.order.pop_front() {
```
webern: Below is what sonnet considers a bug. I have to admit I don't know this domain at all, so I hope it's not just wrong or something, but it's saying that the deletion method you added misses the key and the cache grows unbounded. This probably wouldn't be an issue in most environments but if there are 10's of thousands of container IDs churning, perhaps it could be:

---

`origin_processed_metric_key` builds a key with only 3 tags (`message_type`, `listener_type`, `origin`), but the counters were registered through `MetricsBuilder`, which prepends `component_id` and `component_type` as default tags. The keys don't match, so `delete_counter` always misses and the registry grows unbounded.

Fix: store the correct key at registration time rather than reconstructing it at eviction. Add a `make_key` method to `MetricsBuilder` that mirrors `register_counter_with_tags` but returns the `Key`:

```rust
// in saluki-metrics MetricsBuilder
pub fn make_key<I, T>(&self, metric_name: &'static str, additional_tags: I) -> Key
where I: IntoIterator<Item = T>, T: MetricTag,
{
    let mut tags = (*self.default_tags).clone();
    tags.extend(additional_tags.into_iter().map(MetricTag::into_label));
    Key::from_parts(metric_name, tags)
}
```

Then extend `OriginMetricCounters` to carry `ok_key: Key` and `error_key: Key`, populate them alongside the counters in `get_or_create`, and replace these two `delete_counter` calls with `delete_counter(&evicted.ok_key)` / `delete_counter(&evicted.error_key)`. The two free functions (`origin_processed_metric_key`, `origin_processed_error_key`) can then be removed.

lib/saluki-components/src/sources/dogstatsd/metrics.rs
```diff
@@ -109,15 +224,27 @@ pub(super) fn build_metrics(listen_addr: &ListenAddress, component_context: &Com
...
+                ("origin", ""),
+            ],
         ),
         service_checks_received: builder.register_counter_with_tags(
             "component_events_received_total",
-            [("message_type", "service_checks"), ("listener_type", listener_type)],
+            [
+                ("message_type", "service_checks"),
+                ("listener_type", listener_type),
+                ("origin", ""),
```
webern: Sonnet: Verified parity with Core Agent: `dogstatsd.processed` for events and service checks is declared with three label dimensions (`message_type`, `state`, `origin`) in server.go:283-284, and both paths always call `tlmProcessed.Inc("events"/"service_checks", "ok"/"error", "")` with a hardcoded empty string for origin. So Core Agent also always emits `origin:""` on these series — `origin:""` here is correct and consistent. 👍

webern (commented):
AI aided, seems like maybe there's a bug with evicting from the cache. Otherwise looks good.

lib/saluki-components/src/sources/dogstatsd/metrics.rs
lucastemb: Good catch. Seems like Sonnet was right. I just put up the fix in my newest commit.

tobz (changes requested):
I hate vetoing work, but, could we drop the `dogstatsd_origin` changes and just mark this as unsupported for now?

I have reservations about adding a metric with unbounded cardinality metric after our last issue with the task name memory leak stuff... at least before the metrics registry rework stuff is in place to deal with this in a consistent way.

jszwedko:
> I hate vetoing work, but, could we drop the `dogstatsd_origin` changes and just mark this as unsupported for now?
> 
> I have reservations about adding a metric with unbounded cardinality metric after our last issue with the task name memory leak stuff... at least before the metrics registry rework stuff is in place to deal with this in a consistent way.

I'm alright with this, but can we create a follow-up issue to add it after the metrics registry rework? We seem to use this in internal deployments.

tobz (approved):

jszwedko:
Follow-up issue created here: https://github.com/DataDog/saluki/issues/1935

PR #1925, opened 2026-06-25
fix(core): wake elastic pool waiters after shrink

aqian01:
## Summary
Follow up to this [comment](https://github.com/DataDog/saluki/pull/1901/s#discussion_r3469080436)

- add an `active_decreased` notification path to `ElasticObjectPool`
- make slow-path acquires retry the fast/on-demand path when shrinking lowers `active`
- add a regression test for the shrinker acquiring an idle permit before a waiter enters the slow path

## Root cause

Slow-path waiters only waited on semaphore permits, which represent returned idle items. When the shrinker deleted an idle item, it lowered `active` and created room for on-demand allocation, but no semaphore permit was produced to wake waiters.

## Validation

Unit tests

### DIFF
--- a/lib/saluki-core/src/pooling/elastic.rs
+++ b/lib/saluki-core/src/pooling/elastic.rs
@@ -16,7 +16,7 @@ use std::{
 use pin_project::pin_project;
 use resource_accounting::ResourceGroupToken;
 use tokio::{
-    sync::{OwnedSemaphorePermit, Semaphore},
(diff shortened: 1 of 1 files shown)


webern (approved):
I spent some time in an agent making sure I understand what this object is, what the bug is, etc. I'm convinced this is clean. Curious, how did you find the bug? Did we see consequences of it in the wild?

aqian01:
@webern Codex caught the bug [here](https://github.com/DataDog/saluki/pull/1901/s#discussion_r3468185672), and we decided to fix it in a follow up pr for isolation

PR #1936, opened 2026-06-26
fix(stringtheory): avoid double reclamation bug due to faulty ref counting

tobz:
## Summary

This PR fixes an issue with both interner implementations in `stringtheory` around concurrent drop/reclaim logic.

Currently, there's a bug with how the logic for dropping an interned string -- which potentially triggers reclamation if it's the last instance of that string -- interacts with a concurrent interning operation that is trying to reuse that same entry, leading to a double reclamation that ultimately causes a panic due to underflow of the entry count.

This PR fixes that bug (but in slightly different ways) but ensuring that the "active reference" count for an entry is altered at the right time such that a concurrent interning operating cannot mistakenly identify an entry as valid for reuse.

We handle this in slightly different ways for the fixed-size vs generic map interner as we have to compensate for the difference in how entries are stored. However, the underlying bug (and the new tests for validating that we fixed the bug) are the same.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

New and existing unit tests.

## References

DADP-2
VULN-1709

### DIFF
--- a/lib/stringtheory/src/interning/fixed_size.rs
+++ b/lib/stringtheory/src/interning/fixed_size.rs
@@ -219,6 +219,16 @@ impl EntryHeader {
     fn decrement_active_refs(&self) -> bool {
         self.refs.fetch_sub(1, AcqRel) == 1
     }
+
+    /// Attempts to increment the active reference count, but only if it is currently non-zero.
+    ///
+    /// Returns `true` if the reference count was incremented. A return value of `false` means the entry has already
+    /// been released -- its reference count reached zero -- and is pending reclamation, so it must not be reused.
+    fn try_increment_active_refs(&self) -> bool {
+        self.refs
+            .fetch_update(AcqRel, Acquire, |refs| (refs != 0).then_some(refs + 1))
+            .is_ok()
+    }
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


lib/stringtheory/src/interning/fixed_size.rs
```diff
@@ -219,6 +219,16 @@ impl EntryHeader {
     fn decrement_active_refs(&self) -> bool {
         self.refs.fetch_sub(1, AcqRel) == 1
     }
+
+    /// Attempts to increment the active reference count, but only if it is currently non-zero.
+    ///
+    /// Returns `true` if the reference count was incremented. A return value of `false` means the entry has already
+    /// been released -- its reference count reached zero -- and is pending reclamation, so it must not be reused.
+    fn try_increment_active_refs(&self) -> bool {
```
thieman, line 227: So what's the behavior of the new writer if it hits this case? Does the string end up not interned at all? And that's not a huge deal because the next instance of the string _will_ be interned? Or am I not seeing it

thieman (approved):

lib/stringtheory/src/interning/fixed_size.rs
tobz: It just keeps searching for free space in the spare capacity of the interner, so we might end up with some less efficient interning when these concurrent operations collide... but we don't leak capacity or anything.

PR #1946, opened 2026-06-29
feat(agent-data-plane): agent flare collects diagnostic artifacts

### LINKED ISSUE #1703: Expand Agent flare to collect Agent Data Plane telemetry, config, and runtime state
# Expand Agent flare to collect Agent Data Plane telemetry, config, and runtime state

## Background

Agent Data Plane (ADP) runs as a separate process alongside the Core Agent. Today, when a user runs `agent flare`, ADP's logs, status, and live telemetry land in the flare. We'd like to expand the data that is collected to make debugging easier.

The plumbing for ADP→Agent flare exchange already exists:

- Agent side: `comp/core/remoteagentregistry/` implements gRPC clients for three services — `StatusProvider`, `TelemetryProvider`, `FlareProvider` — and calls each on every registered remote agent.
- ADP side: `bin/agent-data-plane/src/internal/remote_agent.rs` registers all three services. `get_status_details` and `get_telemetry` are implemented; `get_flare_files` is a stub returning an empty `HashMap`.

The flare-builder hook on the Agent side (`comp/core/remoteagentregistry/impl/services.go:fillFlare`) already absorbs whatever `get_flare_files` returns and writes it into the archive. Because ADP's stub returns nothing, no flare-time ADP artifacts beyond logs and live telemetry are captured.

## Currently collected

- `logs/agent-data-plane.log` (active), `logs/agent-data-plane.log.1` (rotated) — picked up by the generic log-dir sweep in `pkg/flare/common/common.go:90`.
- `telemetry.log` — already contains ADP's Prometheus telemetry via the registry's `GetTelemetry` fan-out and the `registryCollector` (`comp/core/remoteagentregistry/impl/services.go:122`).
- `status.log` — already contains ADP's status section (version, git commit, architecture, start time, DSD metrics) populated by `RemoteAgentImpl::get_status_details` in `bin/agent-data-plane/src/internal/remote_agent.rs:444`.

## Proposed additions

All artifacts below ship as entries in the `HashMap<String, Vec<u8>>` returned by ADP's `get_flare_files`. The Core Agent does not scrape ADP endpoints directly; everything flows through the existing Remote Agent Registry gRPC interface. Files land under a `data-plane/` subdirectory (path encoded in the map key returned by ADP).

### Configuration

- `data-plane/runtime_config_dump.yaml` — resolved ADP config (equivalent of `/config` privileged-API endpoint output). Mirrors the existing core-agent `runtime_config_dump.yaml`.

### Health and runtime state

- `data-plane/health.yaml` — capture of `/health`, `/ready`, `/live` results.
- `data-plane/memory_status.json` — capture of `/memory/status`. Useful for OOM diagnosis; no current equivalent in the core flare.

### Workload / tagger

- `data-plane/workload-tags-dump.json` — tagger state.
- `data-plane/workload-external-data-dump.json` — external-data resolver state.

These complement the core agent's existing `tagger-list.json` and `workload-list.log` and are critical for debugging origin-detection / tag-cardinality discrepancies between ADP and core.

### Process info

- `data-plane/runtime_debug_info.log` — pid, uptime, RSS, args, fd count, thread count. Mirrors `runtime_debug_info.log`.
- `data-plane/process_tree.txt` — Linux-only: pstree fragment rooted at ADP pid. Useful when ADP forks worker subprocesses.

### Async runtime / stack equivalent

- `data-plane/runtime-dump.txt` — Rust tokio task dump or stack snapshot. Mirrors `go-routine-dump.log`. (Open question — see below.)

## Implementation notes

Two-side change:

**ADP side (saluki repo):**
1. Implement `RemoteAgentImpl::get_flare_files` in `bin/agent-data-plane/src/internal/remote_agent.rs:492`. Build the `HashMap<String, Vec<u8>>` by gathering the artifacts listed above (config, health, memory, workload dumps, process info). Use map keys of the form `data-plane/<artifact>` so they land in a single subdirectory inside the flare archive.
2. No scrubbing on the ADP side. Return raw bytes; the Core Agent scrubs.

**Agent side (datadog-agent repo):**
1. No code change strictly required — `comp/core/remoteagentregistry/impl/services.go:fillFlare` already absorbs whatever ADP returns and writes it into the flare archive. Scrubbing is applied by the existing `flarebuilder.FlareBuilder.AddFile` path (`pkg/util/scrubber`).
2. Gate the entire ADP fan-out on `data_plane.enabled = true`. If false, skip silently.
3. When ADP is enabled but unreachable (crashed, gRPC connection fails, timeout), write `data-plane/UNREACHABLE.txt` containing the error message returned by the gRPC client. Do not block or fail core-agent flare collection.
4. Document in `docs/dev/agent-data-plane/` what each artifact contains.

## Privacy and scrubbing

Scrubbing is the Core Agent's responsibility. ADP returns raw bytes via `get_flare_files`; the existing Go-side `flarebuilder.FlareBuilder.AddFile` path runs every artifact through `pkg/util/scrubber` before writing to the archive. This matches how every other flare-builder caller works today.

Expected exposure profile (all already mitigated by the existing scrubber):
- `data-plane/runtime_config_dump.yaml` — API keys, app keys, proxy passwords, secret URLs. Scrubbed identically to the core agent's `runtime_config_dump.yaml`.
- Workload tag dumps — container IDs and pod names. Same exposure as `tagger-list.json` today.

## Open questions

1. Does ADP expose a pprof / task-dump endpoint equivalent to Go's `/debug/pprof/goroutine`? If not, is it worth adding one specifically for flare diagnostics?
2. Size budget: the core flare has rough soft limits. ADP artifacts can be sizable on high-cardinality hosts. Cap per-artifact at some bound (e.g. 5MB) with truncation marker?

## Decisions

- **Scrubbing:** Core Agent only. ADP returns raw bytes.
- **Collection mechanism:** Remote Agent Registry `FlareProvider` gRPC. No direct endpoint scraping from Core Agent.
- **`data_plane.enabled = false`:** skip ADP flare fan-out entirely.
- **ADP unreachable:** write `data-plane/UNREACHABLE.txt` containing the gRPC error message; continue with the rest of the flare.

## Acceptance criteria

- Running `agent flare` on a host with `data_plane.enabled: true` produces a `data-plane/` subdirectory containing at minimum: config dump, health, memory status, workload tag dump, and process info.
- Secret scrubbing applied by the Core Agent. No raw API keys in any artifact.
- Flare succeeds with `data-plane/UNREACHABLE.txt` containing the gRPC error when ADP is unreachable; does not block or hang core-agent flare collection.
- Flare on `data_plane.enabled: false` hosts contains no `data-plane/` subdirectory.
- Documentation in `docs/dev/` describes each artifact and what to look for during triage.

## References

- `comp/core/remoteagentregistry/impl/services.go:fillFlare` — Agent-side flare fan-out
- `comp/core/remoteagentregistry/impl/client.go:31-35` — service name constants
- `bin/agent-data-plane/src/internal/remote_agent.rs:444` (saluki) — already-implemented `get_status_details`
- `bin/agent-data-plane/src/internal/remote_agent.rs:469` (saluki) — already-implemented `get_telemetry`
- `bin/agent-data-plane/src/internal/remote_agent.rs:492` (saluki) — `get_flare_files` stub to fill in
- `pkg/flare/common/common.go:90` — existing log-dir sweep that picks up ADP logs today

lucastemb:
## Summary

`agent flare` is a diagnostic tool that captures logs, status, and live telemetry. Previously, this command was returning an empty map; however, this pull request seeks to populate that map appropriately with the following data: current config, memory status, health, tagger state, runtime debug info, process tree, and Rust tokio task dump/ stack snapshot. 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests and local testing:

Screenshot 2026-06-29 at 3 12 43 PM
Screenshot 2026-06-29 at 3 13 11 PM

## References

- Closes: https://github.com/DataDog/saluki/issues/1703

### DIFF
--- a/bin/agent-data-plane/src/internal/env/workload/api.rs
+++ b/bin/agent-data-plane/src/internal/env/workload/api.rs
@@ -211,8 +211,7 @@ impl Supervisable for RemoteAgentWorkloadAPIWorker {
         let workload_route = DynamicRoute::http(EndpointType::Privileged, &self.handler);
 
         let state = self.handler.state.clone();
-        let tags_handle =
-            DiagnosticHandle::new("workload-tags-dump.json", move || state.tags_dump_json().into_bytes());
+        let tags_handle = DiagnosticHandle::new("workload-tags-dump.json", move || state.tags_dump_json().into_bytes());
 
         let state = self.handler.state.clone();
         let eds_handle = DiagnosticHandle::new("workload-external-data-dump.json", move || {
--- a/bin/agent-data-plane/src/internal/env/workload/api.rs
+++ b/bin/agent-data-plane/src/internal/env/workload/api.rs
@@ -61,8 +61,8 @@ impl RemoteAgentWorkloadState {
     /// Returns a JSON snapshot of the current tag store state.
     ///
     /// Lists every active entity with its tags at each cardinality level. This is the same data served by
-    /// `/workload/remote_agent/tags/dump`, collected directly from the in-process store for use in flare
-    /// artifacts.
+    /// `/workload/remote_agent/tags/dump`, collected directly from the in-process store for use as a
+    /// diagnostic artifact.
     fn tags_dump_json(&self) -> String {
         self.get_tags_dump_response()
     }
@@ -70,8 +70,8 @@ impl RemoteAgentWorkloadState {
     /// Returns a JSON snapshot of the current external data store state.
     ///
     /// Lists every pod/container identity mapping tracked by the external data store. This is the same data served
-    /// by `/workload/remote_agent/external_data/dump`, collected directly from the in-process store for use in
-    /// flare artifacts.
+    /// by `/workload/remote_agent/external_data/dump`, collected directly from the in-process store for use as a
+    /// diagnostic artifact.
     fn eds_dump_json(&self) -> String {
         self.get_eds_dump_response()
     }
--- a/bin/agent-data-plane/src/internal/env/workload/api.rs
+++ b/bin/agent-data-plane/src/internal/env/workload/api.rs
@@ -13,7 +13,10 @@ use saluki_context::{
     origin::OriginTagCardinality,
     tags::{SharedTagSet, TagSet},
 };
-use saluki_core::runtime::{state::DataspaceRegistry, InitializationError, Supervisable, SupervisorFuture};
+use saluki_core::{
+    diagnostic::DiagnosticHandle,
+    runtime::{state::DataspaceRegistry, InitializationError, Supervisable, SupervisorFuture},
+};
 use saluki_env::workload::{
     entity::HighestPrecedenceEntityIdRef,
     stores::{ExternalDataStoreResolver, TagStoreQuerier},
@@ -55,6 +58,24 @@ pub struct RemoteAgentWorkloadState {
 }
 
 impl RemoteAgentWorkloadState {
+    /// Returns a JSON snapshot of the current tag store state.
+    ///
+    /// Lists every active entity with its tags at each cardinality level. This is the same data served by
+    /// `/workload/remote_agent/tags/dump`, collected directly from the in-process store for use in flare
+    /// artifacts.
+    fn tags_dump_json(&self) -> String {
+        self.get_tags_dump_response()
+    }
+
+    /// Returns a JSON snapshot of the current external data store state.
+    ///
+    /// Lists every pod/container identity mapping tracked by the external data store. This is the same data served
+    /// by `/workload/remote_agent/external_data/dump`, collected directly from the in-process store for use in
+    /// flare artifacts.
+    fn eds_dump_json(&self) -> String {
+        self.get_eds_dump_response()
+    }
+
     fn get_tags_dump_response(&self) -> String {
         let mut active_entities = FastHashSet::default();
         let mut entity_aliases = FastHashMap::default();
@@ -189,10 +210,22 @@ impl Supervisable for RemoteAgentWorkloadAPIWorker {
     async fn initialize(&self, process_shutdown: ShutdownHandle) -> Result<SupervisorFuture, InitializationError> {
         let workload_route = DynamicRoute::http(EndpointType::Privileged, &self.handler);
 
+        let state = self.handler.state.clone();
+        let tags_handle =
+            DiagnosticHandle::new("workload-tags-dump.json", move || state.tags_dump_json().into_bytes());
+
+        let state = self.handler.state.clone();
+        let eds_handle = DiagnosticHandle::new("workload-external-data-dump.json", move || {
+            state.eds_dump_json().into_bytes()
+        });
+
         Ok(Box::pin(async move {
-            DataspaceRegistry::try_current()
-                .ok_or_else(|| generic_error!("Dataspace not available."))?
-                .assert(workload_route, "workload-api");
+            let dataspace =
+                DataspaceRegistry::try_current().ok_or_else(|| generic_error!("Dataspace not available."))?;
+
+            dataspace.assert(workload_route, "workload-api");
+            dataspace.assert(tags_handle, "diag-workload-tags");
+            dataspace.assert(eds_handle, "diag-workload-eds");
 
             process_shutdown.await;
             Ok(())
--- a/bin/agent-data-plane/src/internal/remote_agent.rs
+++ b/bin/agent-data-plane/src/internal/remote_agent.rs
@@ -579,6 +579,7 @@ impl FlareProvider for RemoteAgentImpl {
                 // Grab and collect all asserted diagnostic handles 
                 if let Some(dataspace) = self.dataspace.get() {
                     let handles = dataspace.current_values::<DiagnosticHandle>(IdentifierFilter::all());
+                    let total_handles = handles.len();
                     let deadline = tokio::time::Instant::now() + DIAGNOSTIC_COLLECT_TIMEOUT;
 
                     for handle in handles {
@@ -601,6 +602,10 @@ impl FlareProvider for RemoteAgentImpl {
                         };
 
                         files.insert(artifact_name, cap_artifact(bytes));
+                    }
+
+                    if files.len() < total_handles {
+                        warn!("Diagnostic artifact collection timed out; collected {}/{total_handles} artifacts.", files.len());
                     }
                 }
 
--- a/bin/agent-data-plane/src/internal/remote_agent.rs
+++ b/bin/agent-data-plane/src/internal/remote_agent.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
-use std::{collections::hash_map::Entry, time::Duration};
+use std::sync::OnceLock;
+use std::{collections::hash_map::Entry, sync::Arc, time::Duration};
 
 use async_trait::async_trait;
 use chrono::{DateTime, Utc};
@@ -13,11 +14,18 @@ use datadog_protos::agent::{
     ConfigSnapshot,
 };
 use futures::StreamExt;
+use process_memory::Querier as MemoryQuerier;
 use prost_types::value::Kind;
+use saluki_common::sync::shutdown::ShutdownHandle;
 use saluki_common::task::spawn_traced_named;
 use saluki_config::{dynamic::ConfigUpdate, upsert, GenericConfiguration};
-use saluki_core::observability::metrics::{
-    get_shared_metrics_state, AggregatedMetricsProcessor, Reflector, TelemetryProcessor,
+use saluki_core::{
+    diagnostic::DiagnosticHandle,
+    observability::metrics::{get_shared_metrics_state, AggregatedMetricsProcessor, Reflector, TelemetryProcessor},
+    runtime::{
+        state::{DataspaceRegistry, IdentifierFilter},
+        InitializationError, Supervisable, SupervisorFuture,
+    },
 };
 use saluki_error::{generic_error, GenericError};
 use saluki_io::net::GrpcTargetAddress;
@@ -58,6 +66,7 @@ pub struct RemoteAgentBootstrap {
     client: RemoteAgentClient,
     session_id: SessionIdHandle,
     internal_metrics: Reflector<AggregatedMetricsProcessor>,
+    dataspace: Arc<OnceLock<DataspaceRegistry>>,
 }
 
 impl RemoteAgentBootstrap {
@@ -108,6 +117,7 @@ impl RemoteAgentBootstrap {
             client,
             session_id,
             internal_metrics: get_shared_metrics_state().await,
+            dataspace: Arc::new(OnceLock::new()),
         })
     }
 
@@ -117,6 +127,21 @@ impl RemoteAgentBootstrap {
             internal_metrics: self.internal_metrics.clone(),
             processor: Mutex::new(TelemetryProcessor::new().with_remapper_rules(get_datadog_agent_remappings())),
             session_id: self.session_id.clone(),
+            dataspace: Arc::clone(&self.dataspace),
+        }
+    }
+
+    /// Creates a worker that captures the dataspace from the supervisor context and makes it
+    /// available to the flare service.
+    ///
+    /// This worker must be added to the control plane supervisor. When it initializes, it runs
+    /// inside the supervisor process where the dataspace task-local is set, and fills the shared
+    /// [`OnceLock`] so that `get_flare_files` can subscribe to flare handles at request time.
+    /// Without this, the gRPC handler tasks spawned by tonic would not have access to the
+    /// dataspace since tonic's `tokio::spawn` does not propagate task-locals.
+    pub fn create_dataspace_anchor(&self) -> DataspaceAnchorWorker {
+        DataspaceAnchorWorker {
+            dataspace: Arc::clone(&self.dataspace),
         }
     }
 
@@ -356,6 +381,7 @@ pub struct RemoteAgentImpl {
     internal_metrics: Reflector<AggregatedMetricsProcessor>,
     processor: Mutex<TelemetryProcessor>,
     session_id: SessionIdHandle,
+    dataspace: Arc<OnceLock<DataspaceRegistry>>,
 }
 
 impl RemoteAgentImpl {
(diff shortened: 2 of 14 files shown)


bin/agent-data-plane/src/internal/remote_agent.rs
```diff
@@ -471,17 +528,197 @@ impl TelemetryProvider for RemoteAgentImpl {
...
+                //
+                // Serialises the fully-resolved ADP configuration to JSON so that the exact settings
+                // in effect at flare-time are captured. Scrubbing (API keys, proxy passwords, etc.) is
+                // the Core Agent's responsibility; we return raw bytes here.
+                if let Some(config) = &self.config {
+                    let bytes = match config.as_typed::<serde_json::Value>() {
+                        Ok(val) => serde_json::to_vec_pretty(&val).unwrap_or_default(),
+                        Err(e) => format!("error serialising configuration: {e}").into_bytes(),
+                    };
+                    files.insert("data-plane/runtime_config_dump.yaml".to_string(), cap_artifact(bytes));
```
tobz: We shouldn't need to prefix any of the file names with `data-plane` since the flare provider for RAR does that for us based on the registered remote agent name.

bin/agent-data-plane/src/internal/remote_agent.rs
```diff
@@ -471,17 +528,197 @@ impl TelemetryProvider for RemoteAgentImpl {
...
+                        "data-plane/runtime-dump.txt".to_string(),
+                        cap_artifact(task_dump.into_bytes()),
+                    );
+                }
+
+                // Artifact H: process tree.
+                //
+                // Linux-only: a pstree fragment rooted at ADP's PID. Useful if ADP ever
+                // forks worker subprocesses. Omitted from the archive entirely if pstree
+                // is not installed in the container.
```
tobz: Let's skip adding speculative things like this.

lucastemb: Sounds good.

Just to be clear, is the idea that this would be implementing observability for a capability that ADP doesn't have yet, so it's one of those things that we'll deal with "once we cross that bridge" - if at all?

lib/saluki-app/src/config.rs
```diff
@@ -87,10 +90,20 @@ impl Supervisable for ConfigWorker {
     async fn initialize(&self, process_shutdown: ShutdownHandle) -> Result<SupervisorFuture, InitializationError> {
         let config_route = DynamicRoute::http(EndpointType::Privileged, &self.handler);
 
+        let config = self.handler.state.config.clone();
+        let flare_handle = FlareDiagnosticHandle::new("runtime_config_dump.yaml", move || {
+            config
+                .as_typed::<serde_json::Value>()
+                .map(|v| serde_json::to_vec_pretty(&v).unwrap_or_default())
```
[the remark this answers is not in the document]
lucastemb: Secrets are redacted by the Core Agent. This is supported by the following statement in the related issue  

> Scrubbing is the Core Agent's responsibility. ADP returns raw bytes via get_flare_files; the existing Go-side flarebuilder.FlareBuilder.AddFile path runs every artifact through pkg/util/scrubber before writing to the archive. This matches how every other flare-builder caller works today.

and was verified through manual testing.

lib/saluki-core/src/runtime/state/dataspace.rs
```diff
@@ -356,6 +356,34 @@ impl DataspaceRegistry {
     ///
     /// Returns a [`Subscription`] that can be used to asynchronously receive updates. Any assertions that match the
     /// filter at the time of subscribing will be immediately yielded.
```
tobz: Seems like this was leftover during the modification here to add `current_values`?

lib/saluki-core/src/runtime/state/dataspace.rs
```diff
@@ -356,6 +356,34 @@ impl DataspaceRegistry {
...
+    /// Returns all currently asserted values of type `T` whose identifier matches `filter`.
+    ///
+    /// This is a synchronous point-in-time read that locks the registry, collects matching values,
+    /// and returns immediately without creating any subscription or channel. Use this when you need
+    /// a snapshot of what is currently asserted rather than ongoing notifications of future changes.
+    ///
+    /// This is the correct primitive for pull-based collection patterns such as flare artifact
+    /// gathering, where a consumer wants to read all currently registered handles exactly once
+    /// without leaving any state behind in the registry.
```
tobz: Nit: irrelevant clanker comment.

lib/saluki-core/src/runtime/state/dataspace.rs
```diff
@@ -411,6 +439,20 @@ impl<T> Subscription<T>
...
+    /// the filter are immediately replayed into an internal pending queue. This method drains that
+    /// queue synchronously, returning only the values that were present at subscribe time. It does
+    /// not wait for future assertions.
+    ///
+    /// This is the correct way to collect all currently registered handles in `get_flare_files`,
+    /// where we want a point-in-time snapshot of everything asserted so far without blocking on
+    /// future updates.
+    pub fn drain_pending(&mut self) -> impl Iterator<Item = AssertionUpdate<T>> + '_ {
+        self.pending.drain(..)
+    }
```
tobz: Seems like leftover code? Can't find where we actually use it.

lib/saluki-core/src/flare.rs
```diff
@@ -0,0 +1,59 @@
+//! Flare diagnostic handle for participating in flare collection via the dataspace.
```
tobz: Flares are a Datadog Agent-ism, _and_ this is a module-level doc comment, so we should just use super generic language here. For example:

```suggestion
//! Diagnostics collection.
```

lib/saluki-core/src/flare.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use std::sync::Arc;
+
+/// A handle that a component asserts into the dataspace to participate in flare collection.
+///
+/// When `get_flare_files` is called, the flare service subscribes to all asserted
+/// `FlareDiagnosticHandle`s via the dataspace (using `IdentifierFilter::prefix("flare.")`) and
+/// calls [`collect`][Self::collect] on each one. Components assert a handle when they start up
+/// and the flare service automatically discovers them, no manual wiring required.
```
tobz: Nit: irrelevant clanker comment.

lib/saluki-core/src/flare.rs
```diff
@@ -0,0 +1,59 @@
+//! Flare diagnostic handle for participating in flare collection via the dataspace.
+
+use std::sync::Arc;
+
+/// A handle that a component asserts into the dataspace to participate in flare collection.
```
tobz:
```suggestion
/// A handle for exposing diagnostics from a component.

File: lib/saluki-core/src/flare.rs [hunk 10]
```diff
@@ -0,0 +1,59 @@
...
+///             serde_json::to_vec(&my_state).unwrap_or_default()
+///         }),
+///         "flare.my-component",
+///     );
+/// ```
+///
+/// The `artifact_name` becomes the filename in the flare archive. The server process that collects
+/// the flare adds any subdirectory prefix, so callers should not include it.
+///
+/// [`Supervisable::initialize`]: crate::runtime::Supervisable::initialize
```
tobz: 1. Assertions can be made at literally at time during the lifecycle of a process so we shouldn't be prescriptive here about when to assert. That part reads to me like the clanker was _assuming_ that it needs to happen that way because we have a lot of code that _does_ do it that way... but that's not an official convention.

2. A lot of Datadog Agent-isms here with the talk of flare. We should only go as far as saying that the artifact name is the component-provided name for the given diagnostic instance, and is generally suitable (but not guaranteed) to be used as a filename.

lib/saluki-core/src/flare.rs
```diff
@@ -0,0 +1,59 @@
...
+#[derive(Clone)]
+pub struct FlareDiagnosticHandle {
```
tobz: `FlareDiagnosticHandle` -> `DiagnosticHandle`

lib/saluki-core/src/flare.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// and reasonably fast, it runs directly inside `get_flare_files` and any delay will delay
+    /// the entire flare response.
+    pub fn new(artifact_name: impl Into<String>, collect_fn: impl Fn() -> Vec<u8> + Send + Sync + 'static) -> Self {
+        Self {
+            artifact_name: artifact_name.into(),
+            collect_fn: Arc::new(collect_fn),
+        }
+    }
+
+    /// Returns the artifact filename.
```
tobz: We should just say "artifact name" instead of "artifact filename."

lib/saluki-core/src/lib.rs
```diff
@@ -12,6 +12,7 @@ pub mod reexport {
 pub mod components;
 pub mod constants;
 pub mod data_model;
+pub mod flare;
```
tobz: Module should be called `diagnostic`.

lib/saluki-core/src/flare.rs
```diff
@@ -0,0 +1,59 @@
...
+    artifact_name: String,
+    collect_fn: Arc<dyn Fn() -> Vec<u8> + Send + Sync>,
+}
+
+impl FlareDiagnosticHandle {
+    /// Creates a new handle with the given artifact name and collection closure.
+    ///
+    /// `collect_fn` is called at flare time to produce the artifact bytes. It must be synchronous
+    /// and reasonably fast, it runs directly inside `get_flare_files` and any delay will delay
+    /// the entire flare response.
```
tobz: This comment should be updated to remove the Datadog Agent-isms (mentioning flares, etc) but the bit about the collect function running synchronously, and how it should be reasonably fast, etc... that _should_ stay.

bin/agent-data-plane/src/internal/remote_agent.rs
```diff
@@ -471,22 +497,216 @@ impl TelemetryProvider for RemoteAgentImpl {
...
+                // and drain all currently-registered handles synchronously — `drain_pending()`
+                // returns the values that were already asserted at subscribe time without blocking
+                // on future updates.
+                //
+                // Components that assert handles:
+                //   - ConfigWorker         → "runtime_config_dump.yaml"
+                //   - HealthRegistryWorker → "health.json"
+                //   - ResourceTelemetryWorker → "memory_status.json"
+                //   - RemoteAgentWorkloadAPIWorker → "workload-tags-dump.json",
+                //                                    "workload-external-data-dump.json"
```
tobz: We should drop this whole doc comment and just write it as more like... "Grab all asserted diagnostic handles and collect each one.

We don't want to have a hard-coded list of what handles we _think_ should exist since that's fragile and likely to change over time, and we also don't need to explain how "current values" works... its doc comment already does that for us.

bin/agent-data-plane/src/internal/remote_agent.rs
```diff
@@ -471,22 +497,216 @@ impl TelemetryProvider for RemoteAgentImpl {
...
+                if let Some(dataspace) = self.dataspace.get() {
+                    let handles =
+                        dataspace.current_values::<FlareDiagnosticHandle>(IdentifierFilter::prefix("flare."));
```
tobz: I would actually drop the prefix filter here and grab everything... the idea is that we already sort of know that we're filtering for diagnostic handles by virtue of getting only the `FlareDiagnosticHandle` assertions.

If there's every a time where we start asserting diagnostic values that _shouldn't_ be part of a flare, then we might want to filter... but right now, all we have wired up is stuff we intend to go into flares.

bin/agent-data-plane/src/internal/remote_agent.rs
```diff
@@ -471,22 +497,216 @@ impl TelemetryProvider for RemoteAgentImpl {
...
+
+                    for handle in handles {
+                        let artifact_name = handle.artifact_name().to_string();
+
+                        // Run the collect closure on a blocking thread so that a slow or
+                        // deadlocked closure cannot block the tokio executor. The timeout
+                        // ensures that a hung closure does not stall the remaining artifacts.
+                        let bytes = match tokio::time::timeout(
+                            FLARE_COLLECT_TIMEOUT,
```
tobz: This is blocking _per_ diagnostic handle, when we really want it to be for the overall call. We also need it to be _under_ the RPC operation timeout on the Core Agent side, so the timeout here should probably be 4 seconds instead: Core Agent's timeout is 5 seconds, and then being conservative to avoid just missing the cutoff lands us at 4 seconds.

bin/agent-data-plane/src/internal/env/workload/api.rs
```diff
@@ -189,10 +210,22 @@ impl Supervisable for RemoteAgentWorkloadAPIWorker {
...
+
         Ok(Box::pin(async move {
-            DataspaceRegistry::try_current()
-                .ok_or_else(|| generic_error!("Dataspace not available."))?
-                .assert(workload_route, "workload-api");
+            let dataspace =
+                DataspaceRegistry::try_current().ok_or_else(|| generic_error!("Dataspace not available."))?;
+
+            dataspace.assert(workload_route, "workload-api");
+            dataspace.assert(tags_handle, "flare.workload-tags");
```
tobz: For this, and all other assertions for diagnostics, this should be a more generic prefix, like `diag-` or something.

tobz (changes requested):

bin/agent-data-plane/src/internal/remote_agent.rs
lucastemb: Sounds good. The call now times out after 4 seconds. The only thing I can foresee now, given my level of understanding, is that if one of the earlier handles takes too long, it could affect the others by leaving them with too little time to complete the request (so those just won't produce an artifact). However, I don't think this should happen in practice.

tobz: Right, so I thought about this too, and we could also just spawn all of the handle collections in parallel. This ends up being a little uglier/more difficult for synchronous calls compared to if we, for example, made the signature of the collection function give back a future: `Fn() -> F where F: Future<Output = Vec<u8>>`, etc.

The flipside is that, while I don't think any of this stuff is actually likely to add up to anywhere close to 4 seconds... because almost all of it is actually synchronous code -- like there's no I/O to be asynchronous with -- running them concurrently doesn't really give any actual benefit.

So I think you're right -- there's a _potential_ risk -- but I'm not sure how likely it is that we're going to hit it in practice. We could hedge this bet by tracking how many handles we're able to collect within the timeout, and if we hit the timeout but there's still handles we didn't collect... we emit a warning log to that effect so that we can see that things are likely to be missing and are unexpectedly slow.

lucastemb: Added a small debug log line that tells you how many artifacts were produced if collection were to time out.

tobz (approved):

PR #1947, opened 2026-06-29
enhancement(core): rework internal metrics registry to support evicting expired/idle metrics

tobz:
## Summary

This PR introduces a new internal metrics registry to support evicting expired/idle metrics that would otherwise lead to unbounded memory growth over time.

Currently, we utilize the registry implementation from `metrics-util` which, while fast and optimized for low registration overhead, has one main limitation: no mechanism for automatically removing metrics which are "idle." This is approximated with helpers like `Recency` and `GenerationalAtomicStorage` (also from `metrics-util`) but can be somewhat convoluted and still requires user-managed integration to do correctly. Spurred by #1902, we wanted a way to automatically clean up metrics which have come and gone -- in particular: sparse, high-cardinality metrics -- so that we could limit the fallout of such misconfigured metrics in the future.

This PR introduces a new implementation of the internal metrics registry that supports detection (and eviction) of "idle" metrics as a first-class feature. At a high level, it looks like this:

- we created our own `MetricsRegistry`, which is just a synchronized set of hashmaps for each metric type (essentially identical to `metrics_util::registry::Registry`)
- likewise, we have our own metric handle types which now track the "idleness" of the metric (and, relatedly but not germane to the stated benefits: the level of the metric as well)
- all metric registration operations update the "idle" state of a metric: how recently was that metric retouched (this is mainly to cover cases where a metric is short-lived)
- during flush, we track the "idleness" of a metric: how many active references exist for this metric, how long ago was it last registered, did we see any updates to the metric for this flush
- every flush calculates both a set of metric updates (changes to active metrics, or new metrics) and a set of evictions (which metrics are now deemed truly idle and should be removed)
- the downstream code (`AggregatedMetricsProcessor`, etc) deals with this new flush format and appropriately upserts metric changes _or_ removes evicted ones 

I've tried to hew towards simplicity here: while the concurrent code is legitimately detail-oriented to ensure we don't lose updates, we own all of the code now, and we aren't using inherently complex techniques or code... just basic stuff like mutexes and atomics.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Existing and new unit tests.

## References

DADP-2

### DIFF
--- a/lib/saluki-core/src/observability/metrics/aggregated.rs
+++ b/lib/saluki-core/src/observability/metrics/aggregated.rs
@@ -13,8 +13,8 @@ use tokio::sync::OnceCell;
 
 use super::histogram::AggregatedHistogram;
 use super::reflector::{Processor, Reflector};
-use super::MetricsStream;
-use crate::data_model::event::{metric::MetricValues, Event};
+use super::{MetricsSnapshot, MetricsStream};
+use crate::data_model::event::metric::MetricValues;
 
 /// Aggregated metric value.
 #[derive(Clone, Debug)]
@@ -226,7 +226,7 @@ impl AggregatedMetricsState {
 pub struct AggregatedMetricsProcessor;
 
 impl Processor for AggregatedMetricsProcessor {
-    type Input = Event;
+    type Input = MetricsSnapshot;
     type State = AggregatedMetricsState;
 
     fn build_initial_state(&self) -> Self::State {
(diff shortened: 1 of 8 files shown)


lib/saluki-core/src/observability/metrics/aggregated.rs
```diff
@@ -238,16 +238,27 @@ impl Processor for AggregatedMetricsProcessor {
...
+                    );
+                }
             }
         }
+
+        // Remove any "evicted" metrics: they were idle long enough without anyone holding a handle to them,
+        // so they're considered dead and should be be removed to avoid unbounded growth.
+        for context in input.evictions {
+            metrics.remove(&context);
+        }
```
thieman, line 261: Any concerns with doing this on every process? Should we instead do it at most once every N seconds or something?

tobz: We only call `process` on a flush, and we flush once a second... so it's basically already that.

thieman (approved):

lib/saluki-core/src/observability/metrics/mod.rs
```diff
@@ -229,91 +431,174 @@ impl Stream for MetricsStream {
...
+                std::sync::atomic::fence(Ordering::Acquire);
             }
+            let delta = handle.consume();
 
-            let histograms = state.registry.get_histogram_handles();
-            for (_, histogram) in histograms {
-                histogram.clear();
+            if orphaned && delta == 0 && idle >= idle_threshold {
+                evicted_keys.push(key.clone());
+                return false;
```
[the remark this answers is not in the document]
tobz: This is true, but lagging and rare and I'd rather approach fixing this by actually changing internal metrics emission to not use a lossy transport like broadcast channels, since we should expect more resilience/durability/correctness/etc from our own internal observability.

tobz:
* **#1947** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1947?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1947?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1955, opened 2026-06-30
fix(dogstatsd): handle unsupported UDS credentials on macOS

thieman:
## Summary

- stop counting non-Linux `UnsupportedPlatform` UDS credential results as per-packet DogStatsD origin-detection telemetry errors
- add a startup warning when UDS origin detection is enabled on a platform that cannot expose sender PIDs through Unix socket credentials
- document the macOS UDS origin-detection limitation and the Linux-only on-demand PID resolver/containerd alias nuance

## Test plan

- `cargo test -p saluki-io telemetry_error`
- `cargo test -p saluki-components origin_detection_telemetry_errors`
- `cargo test -p saluki-components origin_detection_on_non_linux`
- `cargo check -p datadog-agent-config-testing`
- `make fmt`
- `cargo check --workspace`
- `make check-all` *(failed in feature-matrix FIPS checks because local machine is missing `cmake`, required by `aws-lc-fips-sys`; earlier `make check-all` stages passed: fmt, clippy, docs style, deny/license checks, license file check, cargo-machete, and API docs build)*

### DIFF
--- a/docs/agent-data-plane/configuration/dogstatsd.md
+++ b/docs/agent-data-plane/configuration/dogstatsd.md
@@ -491,6 +491,22 @@ while syslog logging is enabled, ADP uses the platform default local syslog sock
 `unixgram:///dev/log` on Linux and `unixgram:///var/run/syslog` on macOS. Set `syslog_rfc: true`
 when the receiving syslog daemon expects the Agent's RFC-style header.
 
+### UDS origin detection on macOS
+
+ADP supports PID-based UDS origin detection on Linux. On macOS and other non-Linux platforms,
+Unix socket credentials do not expose the sender PID: `LOCAL_PEERCRED` and `getpeereid()` provide
+UID and GID only. ADP still accepts DogStatsD traffic over UDS on those platforms, but it cannot
+derive a process ID from the socket and therefore cannot use that path for origin enrichment.
+
+When `dogstatsd_origin_detection` is enabled with a UDS listener on a non-Linux platform, ADP logs a
+startup warning and treats the missing PID as a static platform limitation rather than a per-packet
+origin detection error.
+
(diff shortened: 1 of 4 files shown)


thieman:
[GPT-5] Agent parity note for the telemetry behavior in this PR:

I checked the Datadog Agent DogStatsD UDS path. On non-Linux platforms, Agent does not emit per-packet UDS origin-detection telemetry errors for the unsupported-platform case.

The Agent flow is:

1. UDS listener setup reads `dogstatsd_origin_detection`.
2. If enabled, setup calls `enableUDSPassCred(conn)`.
3. On non-Linux, `enableUDSPassCred` returns `ErrLinuxOnly` (`"only implemented on Linux hosts"`).
4. `setupUnixConn` logs `dogstatsd-uds: error enabling origin detection: ...` and flips `originDetection = false` for that listener.
5. Because origin detection is disabled at setup time, the packet read loop does not allocate OOB credential buffers and does not call `processUDSOrigin(...)` per packet. As a result, the Agent does not increment either `dogstatsd-uds.OriginDetectionErrors` or `dogstatsd.uds_origin_detection_error` for every packet on non-Linux.

On Linux, Agent still increments UDS origin-detection telemetry errors when origin detection was enabled successfully but per-packet origin processing fails (bad ancillary data, invalid credentials, zero PID, lookup errors, etc.). A normal “no container match” is converted to `NoOrigin, nil` and is not counted as an error.

So this PR aligns ADP with the Agent’s effective behavior: unsupported platform is treated as a static setup/platform capability limitation, not as a packet-by-packet origin-detection failure. ADP still keeps `UnsupportedPlatform` visible as a credential error for display/debugging, but it no longer increments DogStatsD origin-detection error telemetry for that static platform condition.

tobz (approved):

PR #1956, opened 2026-06-30
feat(dogstatsd): add Windows named pipe intake

thieman:
## Summary
- Add a reusable Windows named-pipe listen address/transport in `saluki-io`.
- Wire DogStatsD config to enable `dogstatsd_pipe_name` and `dogstatsd_windows_pipe_security_descriptor`.
- Use core-Agent-compatible raw newline framing for Windows named-pipe DogStatsD payloads.
- Drop oversized named-pipe frames that fill the I/O buffer before a newline arrives, matching the core Agent's oversized-message behavior.
- Match core Agent pipe creation behavior by applying `dogstatsd_buffer_size` as the Windows pipe input buffer and using an output buffer size of 0.
- Accept either bare pipe names or full `\\.\pipe\...` values when constructing named-pipe listen addresses.
- Use `npipe://` for URL-style display/parsing of Windows named-pipe addresses, matching existing Agent conventions for named-pipe socket paths.
- Update DogStatsD config inventory/docs from planned unsupported to supported and cover named-pipe newline/stream behavior.
- Add a Windows integration case that executes a PowerShell named-pipe client via `target_exec`, writes to the named pipe, and asserts ADP decodes the frame.

## Test plan
- `cargo test -p panoramic`
- `cargo run --release --bin panoramic -- list -d test/integration/cases --runtime windows --json | rg 'dogstatsd-named-pipe-windows'`
- `cargo check --workspace`
- `cargo test -p saluki-io`
- `cargo test -p saluki-components dogstatsd`
- `cargo xwin check -p saluki-io --target x86_64-pc-windows-msvc`
- `cargo xwin check -p saluki-components --target x86_64-pc-windows-msvc`
- `make fmt`

CI investigation note: earlier Windows integration runs proved the action connected/wrote to the pipe, then exposed two bugs: ADP initially treated named-pipe bytes as length-delimited stream frames, and a mounted helper script used a container path Docker rejected on Windows. The test now runs an inline PowerShell client through `target_exec`, and ADP uses raw newline framing for named pipes.


lib/datadog-agent/config/src/generated.rs
```diff
@@ -1422,6 +1432,9 @@ pub mod defaults {
     pub(super) fn datadog_configuration_dogstatsd_tag_cardinality() -> String {
         "low".to_string()
     }
+    pub(super) fn datadog_configuration_dogstatsd_windows_pipe_security_descriptor() -> String {
+        "D:AI(A;;GA;;;WD)".to_string()
```
thieman, line 1454: Explain this or provide references to datadog-agent behavior doing the same

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -593,6 +622,7 @@ pub struct DogStatsDConfiguration {
 struct EolRequired {
```
thieman, line 639: Explain what this is for and why we need it for named pipe but not for some of the other transports

lib/saluki-io/src/net/addr.rs
```diff
@@ -172,6 +210,14 @@ impl<'a> TryFrom<&'a str> for ListenAddress {
 
                 Ok(Self::Unix(path_buf))
             }
+            "npipe" => {
```
thieman, line 232: should this be `named_pipe`? if `npipe` is right, provide references back to datadog-agent proving it

lib/datadog-agent/config/src/generated.rs
thieman: [OpenAI API model] This default is generated from the vendored Agent schema/config inventory and matches the core Agent default for `dogstatsd_windows_pipe_security_descriptor`: `D:AI(A;;GA;;;WD)`. Datadog Agent binds the same default here: https://github.com/DataDog/datadog-agent/blob/f928a470d2d657b94ed67db74b8644b54345906e/pkg/config/setup/common_settings.go#L1641-L1658 and reads it when constructing the Windows named-pipe listener here: https://github.com/DataDog/datadog-agent/blob/f928a470d2d657b94ed67db74b8644b54345906e/comp/dogstatsd/listeners/named_pipe_windows.go#L39-L49. I also added a source comment on the ADP default helper explaining that it is the core Agent default.

lib/saluki-components/src/sources/dogstatsd/mod.rs
thieman: [OpenAI API model] Added a code comment explaining this. The core Agent exposes `named_pipe` as its own `dogstatsd_eol_required` target (separate from `udp` and `uds`), so ADP needs a separate flag to let Windows named-pipe clients require newline-terminated pipe frames without changing UDS behavior. TCP remains false because the Agent setting does not expose a `tcp` token. Reference for the core Agent tokens: https://github.com/DataDog/datadog-agent/blob/f928a470d2d657b94ed67db74b8644b54345906e/pkg/config/setup/common_settings.go#L1655-L1658 and handling: https://github.com/DataDog/datadog-agent/blob/f928a470d2d657b94ed67db74b8644b54345906e/comp/dogstatsd/server/impl/server.go#L250-L264.

lib/saluki-io/src/net/addr.rs
thieman: [OpenAI API model] Changed this from `npipe://` to `named_pipe://`. I did not find DogStatsD-specific evidence that `npipe://` is the right display/config scheme here; it is clearer to match the listener type/config token. I also added normalization so callers may pass either a bare pipe name or a full `\\.\pipe\...` path, while display remains `named_pipe://...`.

thieman: [OpenAI API model] I re-checked this instead of relying on the previous inference. `npipe://` is the convention used in `datadog-agent` for URL-style Windows named-pipe socket paths: `pkg/config/env/environment_containers.go` defines `winNamedPipePrefix = "npipe://"`, and kubelet helpers document Windows socket paths as requiring `npipe://` (`pkg/util/kubernetes/kubelet/podresources.go`, `deviceplugin.go`). I reverted the display/parse scheme back to `npipe://`, kept `listener_type()` as `named_pipe` for DogStatsD telemetry/config labels, and added a test for parsing `npipe:////./pipe/datadog-dogstatsd`.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -593,6 +622,10 @@ pub struct DogStatsDConfiguration {
...
     udp: bool,
     uds: bool,
+    // The core Agent exposes `named_pipe` as a separate `dogstatsd_eol_required` target.
+    // Named pipes use raw newline framing like UDP datagrams, but they have their own switch
+    // so Windows clients can require newline-terminated pipe frames without affecting UDS.
```
thieman: Remove this comment

webern (commented):
Config changes look good. You did the right thing with the overlay and the generated diff flows from there and looks like what we expect.

thieman:
I'll tell Claude he did a great job

lib/saluki-components/src/sources/dogstatsd/framer.rs
```diff
@@ -36,6 +36,7 @@ pub fn get_framer(listen_address: &ListenAddress, eol_required: bool) -> DsdFram
         ListenAddress::Udp(_) => DsdFramer::NonStream(newline_framer),
         ListenAddress::Unixgram(_) => DsdFramer::NonStream(newline_framer),
         ListenAddress::Unix(_) => DsdFramer::Stream(NestedFramer::new(newline_framer, LengthDelimitedFramer)),
+        ListenAddress::NamedPipe { .. } => DsdFramer::NonStream(newline_framer),
```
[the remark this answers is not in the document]
thieman: [OpenAI API model] Verified this against the ADP read loop and agree. Named pipes use raw newline framing but are connection-oriented, so a no-newline payload that fills the fixed I/O buffer needs an explicit drop path. I added a guard before frame extraction that detects a full named-pipe buffer with no newline, increments framing errors, clears the buffer, and continues reading. Added regression coverage for full/no-newline drop, partial-with-capacity retention, and full-with-newline retention.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1362,6 +1401,13 @@ async fn drive_stream(
                         bytes_read
                     );
 
+                    if should_drop_oversized_named_pipe_frame(&listen_addr, io_buffer) {
+                        metrics.framing_errors().increment(1);
+                        debug!(%listen_addr, %peer_addr, "DogStatsD named pipe frame exceeded the configured buffer size. Dropping frame.");
+                        io_buffer.clear();
+                        continue 'read;
+                    }
```
tobz, line 1435: Hmm, can you link to where we do this in the Agent?

This feels a lot like UDP, but for UDP we'll try and parse the payload even if there's no EOL delimiter _unless_ `dogstatsd_eol_delimiter` is enabled... so this would be a divergence from that, I believe?

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1464,6 +1513,12 @@ async fn drive_stream(
     debug!(%listen_addr, "Stream handler stopped.");
 }
 
+fn should_drop_oversized_named_pipe_frame(listen_addr: &ListenAddress, buffer: &BytesBuffer) -> bool {
+    matches!(listen_addr, ListenAddress::NamedPipe { .. })
+        && buffer.remaining_mut() == 0
+        && !buffer.chunk().contains(&b'\n')
```
tobz: We should use `memchr::memchr` here.

lib/saluki-components/src/sources/dogstatsd/mod.rs
thieman: [OpenAI API model] I checked the core Agent named-pipe path. It differs from UDP here: the Windows named-pipe listener only calls `PacketAssembler.AddMessage(buffer[:messageSize])` when `bytes.LastIndexByte(buffer[:endIndex], '\n') + 1` is > 0, and if the pending partial message reaches the read buffer size it resets `startWriteIndex = 0`, dropping that oversized partial before continuing. Reference: https://github.com/DataDog/datadog-agent/blob/f928a470d2d657b94ed67db74b8644b54345906e/comp/dogstatsd/listeners/named_pipe_windows.go#L196-L214. So the explicit full-buffer/no-newline drop is intended to match Agent named-pipe behavior, not UDP behavior. The current guard only triggers when the named-pipe buffer is full and has no newline; partial frames with spare capacity keep waiting, and full buffers containing a newline still decode normally (covered by tests).

lib/saluki-components/src/sources/dogstatsd/mod.rs
thieman: [OpenAI API model] Updated to use `memchr::memchr` in the named-pipe oversized-frame guard and added the direct `memchr` workspace dependency to `saluki-components`.

tobz (approved):

PR #1957, opened 2026-07-01
chore(config): typed config foundation

webern:
## Human Summary

This PR serves as the foundation for strongly-typed configuration for ADP. It does not affect the binary, but sets up the base for future PRs that will migrate the binary from map access with `GenericConfiguration` to typed access with `SalukiConfiguration`.

This PR sets up:
- the translation target: `SalukiConfiguration`
- a compile time auditable translation system: `DatadogConfigWitness`
- implements the translation of `DatadogConfiguration` to `SalukiConfiguration`
- Discovers and promotes a bunch of keys that were missing from our inventory (which, I think, sort of demonstrates the need for this project to succeed).

The shape of `SalukiConfiguration` is likely to change as we progress. 

**Challenges**

One difficult aspect of this project is that the discovery and inventorying of all of the configs we are using messy. The full audit and shape of `SalukiConfiguration` can only be fully realized once it is being pushed down to components and usage sites, so consider this a sketch of its final form.

The other big challenge is churn. It will be costly for me to keep it up-to-date as the codebase changes underneath it. I have written skills to help with this, but it is not easy. Once the system is complete and high-churn areas like DogStatsD are swung over to typed config, I don't think it will be hard for team members to use, but it is not intuitive in a transitory state. As such, I recommend we refrain from merging until enough of the PR stack is approved to swing us over to using this.

That being said, it's important to get early feedback before the stack gets too deep.

 ## AI Summary

 This PR adds the foundation for translating Datadog Agent configuration into a typed ADP
 configuration model.

 The change is additive: the existing runtime config path is not cut over in this PR.

 What changed:

 - Adds `agent-data-plane-config`, a source-agnostic data model centered on
   `SalukiConfiguration { control, shared, domains }`. The model derives `PartialEq` to support
   later per-subtree change detection.
 - Adds generated witness support in `datadog-agent-config`:
   - `DatadogConfigWitness`
   - generated `drive(...)`, which reports every failed key, not just the first
   - `TranslateError` (one failed key) and `TranslateErrors` (the aggregate `drive` returns)
 - Adds `agent-data-plane-config-system`, with the initial Datadog translator
   (`DatadogTranslator::translate`, an inherent method that returns the fully-populated model plus
   any translation errors).
 - Separates the two config writers:
   - Datadog-schema keys are consumed through the generated witness driver.
   - Saluki-only fields are copied through `SalukiOnly::seed`.
 - Promotes 24 additional Datadog config keys into the supported inventory so they can be represented
   in the typed model.
 - Adds the `NO_SMOKE` config-testing sentinel for supported keys that do not yet fit the existing
   smoke-test model.
 - Moves generated Datadog config output under `datadog_agent_config::generated` so generated config
   modules are grouped together.

 Out of scope for this PR:

 - wiring the new config system into `bin/agent-data-plane`
 - migrating components away from `GenericConfiguration`
 - dynamic config routing
 - removing the old raw-map access path

 ## Change Type

 ## How did you test this PR?

 - `make check-all`
 - Verified each commit builds independently.
 - Regenerated config output and verified the generated tree is stable.

 ## References

 - Progresses: #1788


lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -0,0 +1,1077 @@
...
+
+    fn consume_dogstatsd_log_file_max_size(&mut self, value: String) {
+        match value.parse::<ByteSize>() {
+            Ok(size) => self.config.domains.dogstatsd.debug_log.log_file_max_size = size.as_u64(),
+            Err(reason) => self.record_error(
+                TranslatorSnafu {
+                    key: "dogstatsd_log_file_max_size",
+                    reason,
+                }
+                .build(),
```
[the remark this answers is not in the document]
webern: Comeon copilot, let the compiler decide what compiles 🙄

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -0,0 +1,1077 @@
...
+            additional,
+        } = self;
+
+        let url = if !dd_url.is_empty() {
+            dd_url
+        } else if !site.is_empty() {
+            site
+        } else {
+            "datadoghq.com".to_string()
+        };
```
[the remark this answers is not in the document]
webern: For the moment I flattened this to get the algorithm out of this PR. Everytime I look at config the problem gets bigger. Currently I suspect that site is inert in ADP (see below). However, reading the message below, it sounds like maybe we do have access to a field that needs to be added to `DatadogConfiguration` or something. It sites `source`, so I need to look into that. The scope keeps growing...

*AI written below:*

This looks likely correct, and possibly broader than this PR. ADP cannot currently distinguish whether a value like `dd_url` arrived as a user setting or as the core agent's schema default: the config stream carries a `source` per key, but it is dropped on ingest, so the `site` fallback may already be unreachable in ADP today, independent of this change. Tracking in #1965. For this PR the resolution logic is removed and `site`/`dd_url` are modeled as separate optional fields with a TODO referencing that issue.

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -0,0 +1,1077 @@
...
+            api_key,
+            dd_url,
+            site,
+            additional,
+        } = self;
+
+        let url = if !dd_url.is_empty() {
+            dd_url
+        } else if !site.is_empty() {
+            site
```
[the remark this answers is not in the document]
webern: This looks likely correct, and possibly broader than this PR. ADP cannot currently distinguish whether a value like `dd_url` arrived as a user setting or as the core agent's schema default: the config stream carries a `source` per key, but it is dropped on ingest, so the `site` fallback may already be unreachable in ADP today, independent of this change. Tracking in #1965. For this PR the resolution logic is removed and `site`/`dd_url` are modeled as separate optional fields with a TODO referencing that issue.

lib/datadog-agent/config/src/lib.rs
```diff
@@ -1,10 +1,10 @@
...
-///
-/// Generated at build time from `core_schema.yaml` plus `schema_overlay.yaml`. Contains only keys
-/// inventoried as `support: full` or `support: partial`. Mostly unused until the configuration
-/// translator consumes it.
+/// Build-time generated code, produced from `core_schema.yaml` plus `schema_overlay.yaml`.
 pub mod generated;
 
-pub use generated::DatadogConfiguration;
+/// The translation error type recorded by the translator and surfaced by the witness driver.
+pub mod translate_error;
```
tobz: Do these modules actually need to be public themselves since we're already exporting the relevant types?

lib/datadog-agent/config/src/generated/datadog_configuration.rs
```diff
@@ -44,8 +44,8 @@ pub struct DatadogConfiguration {
     #[serde(default, skip_serializing_if = ":: std :: collections :: HashMap::is_empty")]
     pub additional_endpoints: HashMap<String, Vec<String>>,
 
-    #[serde(default, skip_serializing_if = "::std::option::Option::is_none")]
-    pub agent_ipc: Option<DatadogConfigurationAgentIpc>,
+    #[serde(default)]
+    pub agent_ipc: DatadogConfigurationAgentIpc,
```
tobz: For what it's worth: these autogenerated names have both a stutter (`datadog_agent::config::Datadog...`, but we know they're from DDA from the crate name itself) _and_ they have swapped precedence: `[Parent]Configuration[Child]` instead of `[Parent][Child]Configuration`.

Would be good to clean these up slightly, even if they're mostly hidden.

tobz (approved):

webern:
* **#1986** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1986?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1977** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1977?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1957** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1957?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1957?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/datadog-agent/config/src/lib.rs
webern: Good call, fixed. Also did away with snafu since I had to use `visibility(pub)` for this exported error and I hate doing that (even though this isn't a "real" library, still...)

webern:
I've spent many hours trying to preserve the separation of commits in this PR and it's just... a massive headache. So I'm going to squash it down to one commit so that future rebases and changes are less painful. Goodbye my beautiful commit history...

lib/datadog-agent/config/src/generated/datadog_configuration.rs
webern: I fixed by just going with e.g. `AgentIpc`. It's hidden and qualifying as config etc is just extra noise. Good call.

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -0,0 +1,1177 @@
...
+    fn consume_data_plane_secure_api_listen_address(&mut self, value: String) {
+        self.config.control.secure_api_listen_address = ListenAddress(value);
+    }
+
+    fn consume_data_plane_use_new_config_stream_endpoint(&mut self, value: bool) {
+        self.config.control.use_new_config_stream_endpoint = value;
+    }
+
+    fn consume_dd_url(&mut self, value: String) {
+        self.config.shared.endpoints.dd_url = non_empty(value).filter(|url| url != DEFAULT_PRIMARY_ENDPOINT);
```
webern: Pointing out that #1965 / #2028 are handled here.

lib/datadog-agent/config/schema/schema_overlay.yaml
```diff
@@ -708,6 +813,13 @@ inventory:
     documentation: "Experimental HTTP path for pre-aggregated DogStatsD metrics; planned when the DogStatsD-over-HTTP work resumes."
     issue: "#1682"
 
+  dogstatsd_expiry_seconds:
+    support: full
+    pipelines: [dogstatsd]
```
thieman, line 818: Are these pipeline declarations and the differentiation between "shared" and "domains" config structs still worth the lift? IIRC this was originally to help the smoke tests determine if something was used, but seems like that job is being moved to the witness?

thieman:
Is there a plan to compare the old vs new config structs to make sure type conversions etc are all handled equivalently?

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -0,0 +1,1177 @@
...
+        (self.config, errors)
+    }
+
+    /// Records a translation error encountered while consuming.
+    fn record_error(&mut self, error: TranslateError) {
+        self.errors.push(error);
+    }
+}
+
+/// Returns `Some(s)` when `s` is non-empty, mapping the empty string to "unset".
```
thieman: This seems like it's for the function immediately below

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -0,0 +1,1177 @@
...
+            .forwarder
+            .retry_queue_capacity_time_interval_sec = value.max(0) as u64;
+    }
+
+    fn consume_forwarder_retry_queue_max_size(&mut self, value: i64) {
+        self.config.shared.endpoints.forwarder.retry_queue_max_size = Some(value.max(0) as u64);
+    }
+
+    fn consume_forwarder_retry_queue_payloads_max_size(&mut self, value: i64) {
+        self.config.shared.endpoints.forwarder.retry_queue_payloads_max_size = Some(value.max(0) as u64);
```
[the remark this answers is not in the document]
tobz: Maybe we just move the logic here, similar-ish to what you did above for `dd_url`/`site`.

tobz (approved):
Took another pass on this, and I'm still generally OK with this.

Given the size, I think we should bias towards getting it merged since we can incrementally tidy things up if we decide that we really want to make stylistic / ergonomics-related changes before starting the next "phase" of the migration.

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
webern: Yes, thanks. Fixed in this push https://github.com/DataDog/saluki/compare/4d87f55dd1fa2b9d252799d7d97c5dde420dc0fb..a125c4028b0afd82f234bf44bd59064547fac1b2

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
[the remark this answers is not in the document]
webern: Added the logic and made it a little more readable in this push https://github.com/DataDog/saluki/compare/4d87f55dd1fa2b9d252799d7d97c5dde420dc0fb..a125c4028b0afd82f234bf44bd59064547fac1b2

Also note that it is inert at the moment, with this merge, and will get scrutinized again during any component cutovers.

I hope to get to #1965 properly soon, but I think I need to get rid of the config system's use of `GenericConfiguration` to make that really correct. Anyway, it's an issue we need to address as soon as we can.

lib/datadog-agent/config/schema/schema_overlay.yaml
webern: Aaaactually, these are used primarily, and only, by the mechanism we use to warn on incompatible configs. And I believe they do nothing for `support: full` keys 🙄 . The thing that *will* become unused, as components cutover, is the `used_by` field that names structs that are directly deserializing from the `GenericConfiguration` map. So as we cut those over to typed config, the overlay will need a sentinel like `MIGRATED_TO_CONFIG_SYSTEM` (or whatever), during the transitional state. Then `used_by` and the smoke tests will be deleted, and alternate forms of config system testing will be introduced.

lib/datadog-agent/config/src/duration_de.rs
```diff
@@ -0,0 +1,94 @@
...
+        fn visit_i64<E: de::Error>(self, nanos: i64) -> Result<Duration, E> {
+            Ok(Duration::from_nanos(nanos.max(0) as u64))
+        }
+
+        fn visit_f64<E: de::Error>(self, nanos: f64) -> Result<Duration, E> {
+            Ok(Duration::from_nanos(if nanos < 0.0 { 0 } else { nanos as u64 }))
+        }
+
+        fn visit_str<E: de::Error>(self, text: &str) -> Result<Duration, E> {
+            parse_duration(text).map_err(de::Error::custom)
```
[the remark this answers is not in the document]
webern: Pushed a fix in https://github.com/DataDog/saluki/compare/a125c4028b0afd82f234bf44bd59064547fac1b2..52d4d614ee5b54f21d893fae481c76c48ac18863

I also pushed some more stuff to the shared go-duration crate and made that a solo precursor commit.

webern:
Note: i tried to separate out a small commit for the go-time-duration change that is more generally applicable, but... the merge queue rejected rebase-merge (apparently only *after* testing it or something), so I just squash merged it after that.

PR #1958, opened 2026-07-01
fix(config): recognize duration keys at their default in the config classifier

jszwedko:
## Summary

The nightly integration suite (which runs against the latest Datadog Agent master rather than the pinned release) started failing because ADP emitted a spurious `Unsupported configuration key` warning for `tls_handshake_timeout`. The key was at its default the whole time — the classifier just couldn't tell, because it compares the Agent-supplied value against the schema default with exact JSON equality, and duration keys don't survive that comparison: the schema expresses their default as a Go duration string (`10s`) while the Agent transmits durations over the config stream as integer nanoseconds (`10000000000`). The result is a false "override" for any duration key the Agent populates at its default. This restores the intended behavior — a key sitting at its default is never treated as an incompatible override — by teaching the classifier that duration defaults and duration values are the same quantity in two encodings.

```mermaid
flowchart LR
    A["Agent config stream<br/>tls_handshake_timeout = 10000000000 (ns)"] --> C{is_default?}
    S["Schema default<br/>10s (Go duration string)"] --> C
    C -- "before: String != Number<br/>=> false => WARN" --> W["Unsupported configuration key"]
    C -- "after: normalize both to ns<br/>10e9 == 10e9 => true" --> OK["Recognized as default (no warning)"]
```

The fix threads the schema's `format: duration` marker through codegen into each `ClassifierEntry` (`is_duration`), and `is_default_value` normalizes both sides to nanoseconds before comparing. It applies to every duration-typed key in the classifier, not just `tls_handshake_timeout`. No schema/overlay edits were required.

## Test plan

- [x] New unit tests in `classifier.rs`: a duration key sent as nanoseconds is recognized as its default; a non-default nanosecond value is not; and `parse_go_duration_nanos` covers the common Go duration forms (`10s`, `500ms`, `1h30m0s`, `1.5h`, `250µs`, `-5s`, bare `0`, and malformed input).
- [x] Reproduced the nightly failure locally against Agent `master-py3-full`, then confirmed `adp-config-check-no-warn` and `adp-config-stream` pass with this change.

> [!NOTE]
> This addresses one of two independent regressions surfaced by the nightly. The remaining failures (`*-api-endpoints`, `adp-config-dynamic-basic`, `dogstatsd-forwarding`) are a separate issue — under the new Agent, ADP's API binds default ports instead of honoring `DD_DATA_PLANE__API_LISTEN_ADDRESS` — and will be handled separately.

### DIFF
--- a/lib/datadog-agent/config/src/classifier/classifier.rs
+++ b/lib/datadog-agent/config/src/classifier/classifier.rs
@@ -45,16 +45,32 @@ impl ConfigClassifier {
         let entry = self.lookup.get(key)?;
         Some(Classification {
             support_level: entry.support_level,
-            is_default: is_default_value(entry.default, value),
+            is_default: is_default_value(entry.default, entry.is_duration, value),
             pipeline_affinity: entry.pipeline_affinity,
         })
     }
 }
 
-fn is_default_value(default: Option<&str>, value: &Value) -> bool {
+fn is_default_value(default: Option<&str>, is_duration: bool, value: &Value) -> bool {
     match default {
         Some(default_str) => match serde_json::from_str::<Value>(default_str) {
-            Ok(default_value) => *value == default_value,
+            Ok(default_value) => {
+                if *value == default_value {
+                    return true;
+                }
+                // Duration keys need special handling: the schema default is a Go duration string
+                // (for example, `"10s"`), but the Datadog Agent transmits durations as integer
+                // nanoseconds. Normalize both sides to nanoseconds before comparing so a value at
+                // its default doesn't get flagged as an override.
+                if is_duration {
+                    if let (Some(default_ns), Some(value_ns)) =
+                        (duration_value_as_nanos(&default_value), duration_value_as_nanos(value))
+                    {
+                        return default_ns == value_ns;
+                    }
+                }
+                false
+            }
             Err(_) => false,
         },
         None => match value {
@@ -62,6 +78,113 @@ fn is_default_value(default: Option<&str>, value: &Value) -> bool {
             Value::String(s) => s.is_empty(),
             _ => false,
         },
+    }
+}
+
+/// Normalizes a duration-typed config value to nanoseconds.
+///
+/// Accepts a Go duration string (for example, `"10s"`, `"1m30s"`) or a JSON number already
+/// expressed in nanoseconds (the form the Datadog Agent sends over the config stream). Returns
+/// `None` for any other shape.
+fn duration_value_as_nanos(value: &Value) -> Option<i128> {
+    match value {
+        Value::String(s) => parse_go_duration_nanos(s),
+        Value::Number(n) => n.as_i64().map(i128::from),
+        _ => None,
+    }
+}
+
+/// Parses a Go `time.Duration` string (for example, `"10s"`, `"500ms"`, `"1h30m0s"`) into
+/// nanoseconds.
+///
+/// Supports the unit suffixes Go's `time.ParseDuration` accepts (`ns`, `us`/`µs`/`μs`, `ms`, `s`,
+/// `m`, `h`), an optional leading sign, and fractional components. Returns `None` if the string is
+/// not a well-formed duration.
+fn parse_go_duration_nanos(input: &str) -> Option<i128> {
+    let mut s = input.trim();
+    if s.is_empty() {
+        return None;
+    }
+
+    let negative = match s.as_bytes()[0] {
+        b'-' => {
+            s = &s[1..];
+            true
+        }
+        b'+' => {
+            s = &s[1..];
+            false
+        }
+        _ => false,
+    };
+
+    // Go accepts a bare "0" (no unit) as a special case.
+    if s == "0" {
+        return Some(0);
+    }
+
+    let mut total_nanos: f64 = 0.0;
+    let mut chars = s.char_indices().peekable();
+    let mut consumed_any = false;
+
+    while let Some(&(start, _)) = chars.peek() {
+        // Parse the numeric portion (digits with an optional single decimal point).
+        let mut end = start;
+        let mut seen_digit = false;
+        let mut seen_dot = false;
+        while let Some(&(idx, c)) = chars.peek() {
+            if c.is_ascii_digit() {
+                seen_digit = true;
+                end = idx + c.len_utf8();
+                chars.next();
+            } else if c == '.' && !seen_dot {
+                seen_dot = true;
+                end = idx + c.len_utf8();
+                chars.next();
+            } else {
+                break;
+            }
+        }
+        if !seen_digit {
+            return None;
+        }
+        let number: f64 = s[start..end].parse().ok()?;
+
+        // Parse the unit portion (letters, including the micro sign variants).
+        let unit_start = end;
+        let mut unit_end = end;
+        while let Some(&(idx, c)) = chars.peek() {
+            if c.is_ascii_alphabetic() || c == 'µ' || c == 'μ' {
+                unit_end = idx + c.len_utf8();
+                chars.next();
+            } else {
+                break;
+            }
+        }
+        let unit = &s[unit_start..unit_end];
+        let unit_nanos = match unit {
+            "ns" => 1.0,
+            "us" | "µs" | "μs" => 1_000.0,
+            "ms" => 1_000_000.0,
+            "s" => 1_000_000_000.0,
+            "m" => 60.0 * 1_000_000_000.0,
+            "h" => 3_600.0 * 1_000_000_000.0,
+            _ => return None,
+        };
+
+        total_nanos += number * unit_nanos;
+        consumed_any = true;
+    }
+
+    if !consumed_any {
+        return None;
+    }
+
+    let total_nanos = total_nanos.round();
+    if negative {
+        Some(-(total_nanos as i128))
+    } else {
+        Some(total_nanos as i128)
     }
 }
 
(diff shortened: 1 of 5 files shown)


lib/datadog-agent/config/src/classifier/classifier.rs
```diff
@@ -65,6 +81,113 @@ fn is_default_value(default: Option<&str>, value: &Value) -> bool {
...
+    }
+}
+
+/// Parses a Go `time.Duration` string (for example, `"10s"`, `"500ms"`, `"1h30m0s"`) into
+/// nanoseconds.
+///
+/// Supports the unit suffixes Go's `time.ParseDuration` accepts (`ns`, `us`/`µs`/`μs`, `ms`, `s`,
+/// `m`, `h`), an optional leading sign, and fractional components. Returns `None` if the string is
+/// not a well-formed duration.
+fn parse_go_duration_nanos(input: &str) -> Option<i128> {
```
jszwedko, line 103: This can be refactored out to share the implementation from saluki-config.

lib/datadog-agent/config/src/classifier/mod.rs
```diff
@@ -154,6 +154,10 @@ pub struct ClassifierEntry {
     pub pipeline_affinity: PipelineAffinity,
     /// JSON-encoded default value from the Agent schema, if present.
     pub default: Option<&'static str>,
+    /// Whether this key holds a `format: duration` value. The Agent sends durations as integer
+    /// nanoseconds while `default` is a Go duration string, so the default check normalizes both
+    /// sides before comparing.
+    pub is_duration: bool,
```
webern, line 160: I spent some time on designing what I think is a proper fix and that design is below. It solves a few problems and models the actual data type correctly. You can merge your fix as is, but if you do I might suggest posting this as a follow up issue (which I can take a stab at). It might have enough fidelity to nearly one-shot it as is from the write-up, though there's a decision buried in there about how we stop rewriting the "Go duration" algorithm over and over again 😅 

The major benefits of the design below
- We catch this at compile time. If the format is duration and default is not parsable as a duration, our build breaks.
- We properly model the specified datatype at the source and all downstream codegen can do the right thing™ with it.

### 1. Model `format: duration` as an effective field type

The vendored Agent schema is JSON Schema in YAML form. I checked the resolved schema, and the only `format` value we currently see is `duration`. Those entries are all `type: number, format: duration`. (Well, actually on main it is now corrected to `type: string, format: duration`)

Given that, I think `is_duration` is modeling this at the wrong level. The source schema has `type + format`, but our generated `FieldType` is already the effective type used by codegen/smoke tests/classification. So I would rather fold `format: duration` into `FieldType::Duration` than carry `FieldType::Float` plus an `is_duration` side channel.

`lib/datadog-agent/config-overlay-model/src/schema_gen.rs`
```rust
pub enum FieldType {
    String,
    Bool,
    Integer,
    Float,
    StringList,
    Duration,
    Unknown,
}
```

`lib/datadog-agent/config-overlay-model/src/schema_gen.rs`
```rust
fn parse_value_type(value: &Value) -> FieldType {
    match (
        value.get("type").and_then(|v| v.as_str()),
        value.get("format").and_then(|v| v.as_str()),
    ) {
        (Some("string"), Some("duration")) => FieldType::Duration,
        // Should we break the build on nonsense like this?
        (_, Some("duration")) => panic!("illogical duration field {show_helpful_stuff}"),
        (Some("string"), _) => FieldType::String,
        (Some("boolean"), _) => FieldType::Bool,
        (Some("integer"), _) => FieldType::Integer,
        (Some("number"), _) => FieldType::Float,
        (Some("array"), _) => {
            let item_type = value.get("items").and_then(|v| v.get("type")).and_then(|v| v.as_str());
            if item_type == Some("string") {
                FieldType::StringList
            } else {
                FieldType::Unknown
            }
        }
        _ => FieldType::Unknown,
    }
}
```

If we later want raw JSON Schema fidelity, I would model that explicitly as something like `json_type + format`. But `FieldType + is_duration` is neither the raw schema model nor a clean effective-type model; it is a one-off side channel that future schema gotchas are likely to copy.

### 2. Teach smoke tests about duration too

I would also make the config-testing registry understand this as a real value type. That avoids needing bespoke `test_json` values for every future duration key.

`lib/datadog-agent/config-testing/src/config_registry/mod.rs`
```rust
pub enum ValueType {
    Bool,
    String,
    Integer,
    Float,
    StringList,
    Duration,
}
```

`lib/datadog-agent/config-testing/src/smoke_test.rs`
```rust
fn test_json_value(value_type: ValueType) -> serde_json::Value {
    match value_type {
        ValueType::String => json!(TEST_STRING_VALUE),
        ValueType::Bool => json!(TEST_BOOL_VALUE),
        ValueType::StringList => json!(TEST_STRING_LIST_VALUE),
        ValueType::Integer => json!(42i64),
        ValueType::Float => json!(1.5f64),
        ValueType::Duration => json!("42s"),
    }
}
```

`lib/datadog-agent/config-testing/src/smoke_test.rs`
```rust
fn json_value_to_env_string(value: &serde_json::Value, value_type: ValueType) -> String {
    match value_type {
        ValueType::Bool => value
            .as_bool()
            .map(|b| b.to_string())
            .unwrap_or_else(|| "true".to_string()),
        ValueType::Integer => value
            .as_i64()
            .map(|n| n.to_string())
            .unwrap_or_else(|| "42".to_string()),
        ValueType::Float => value
            .as_f64()
            .map(|f| f.to_string())
            .unwrap_or_else(|| "1.5".to_string()),
        ValueType::String => value.as_str().unwrap_or(TEST_STRING_VALUE).to_string(),
        ValueType::StringList => value
            .as_array()
            .map(|arr| arr.iter().filter_map(|v| v.as_str()).collect::<Vec<_>>().join(" "))
            .unwrap_or_else(|| TEST_STRING_LIST_VALUE.join(" ")),
        ValueType::Duration => value.as_str().unwrap_or("42s").to_string(),
    }
}
```

Individual overlay entries should only need `test_json: '"20s"'` or similar when the generic duration test value is bad for that specific key.

### 3. Store parsed duration defaults, not parse schema defaults during classification

The other thing I would change is the default representation. Duration defaults are static generated metadata. I do not think the classifier should parse the schema default on every classification. Codegen should parse and canonicalize the duration default once, and fail the build if the schema default is invalid.

For example:

`lib/datadog-agent/config/src/classifier/mod.rs`
```rust
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub enum DefaultValue {
    Missing,
    Json(&'static str),
    DurationNanos(u64),
}
```

Then generated classifier data can carry a typed default:

`lib/datadog-agent/config/src/classifier/classifier_data.rs`
```rust
ClassifierEntry {
    yaml_path: "tls_handshake_timeout",
    aliases: &[],
    pipeline_affinity: PipelineAffinity::CrossCutting,
    default: DefaultValue::DurationNanos(10_000_000_000),
}
```

Non-duration defaults can stay JSON-backed:

`lib/datadog-agent/config/src/classifier/classifier_data.rs`
```rust
ClassifierEntry {
    yaml_path: "min_tls_version",
    aliases: &[],
    support_level: SupportLevel::Partial,
    pipeline_affinity: PipelineAffinity::CrossCutting,
    default: DefaultValue::Json("\"tlsv1.2\""),
}
```

Then the runtime default check switches on the already-typed default:

`lib/datadog-agent/config/src/classifier/classifier.rs`
```rust
fn is_default_value(default: DefaultValue, value: &Value) -> bool {
    match default {
        DefaultValue::Json(default_str) => serde_json::from_str::<Value>(default_str)
            .map(|default_value| *value == default_value)
            .unwrap_or(false),

        DefaultValue::DurationNanos(default_ns) => duration_value_as_nanos(value)
            .map(|value_ns| value_ns == default_ns)
            .unwrap_or(false),

        DefaultValue::Missing => match value {
            Value::Null => true,
            Value::String(s) => s.is_empty(),
            _ => false,
        },
    }
}
```

Runtime still needs to normalize the incoming value because the Agent config stream may send durations as integer nanoseconds:

`lib/datadog-agent/config/src/classifier/classifier.rs`
```rust
fn duration_value_as_nanos(value: &Value) -> Option<u64> {
    match value {
        Value::Number(n) => n.as_u64(),
        Value::String(s) => parse_duration(s).ok().map(|d| d.as_nanos() as u64),
        _ => None,
    }
}
```

### 4. Avoid duplicating `DurationString`

This PR's duration parser duplicates the existing `saluki-config::DurationString` / `parse_duration` behavior. I agree `datadog-agent-config` should not depend on `saluki-config`; that would pull a leaf-ish Datadog config crate toward a much heavier, higher-level crate.

There are also existing crates that may solve the leaf-crate problem without making our crates depend on each other:

- https://crates.io/crates/go-parse-duration
- https://crates.io/crates/duration-str

If one of these is acceptable after careful vetting, I would prefer taking that dependency over continuing to rewrite this parser in multiple places. The vetting should include at least: license, maintenance/activity, dependency footprint, exact Go `time.ParseDuration` compatibility, overflow behavior, negative duration behavior, and whether the API returns the unit we want without lossy conversion.

This is especially attractive because duration parsing has a pile of small edge cases (`µs` vs `μs`, fractional units, signs, overflow, bare `0`, malformed input). The existence of this PR is a good signal that we should stop locally reimplementing that algorithm.

One possible shape if we pick a crate:

`lib/datadog-agent/config/Cargo.toml`
```toml
[dependencies]
go-parse-duration = "0.1"
```

`lib/saluki-config/Cargo.toml`
```toml
[dependencies]
go-parse-duration = "0.1"
```

Then `saluki-config::DurationString` can remain our domain type, but delegate parsing to the vetted implementation:

`lib/saluki-config/src/duration_string.rs`
```rust
fn parse_string(s: &str) -> Result<Duration, ParseDurationError> {
    let nanos = go_parse_duration::parse_duration(s).map_err(|err| ...)?;
    Duration::from_nanos(nanos.try_into().map_err(|_| ParseDurationError::Negative)?).into()
}
```

And classifier codegen can use the same parser for schema defaults without depending on `saluki-config`:

`lib/datadog-agent/config/build/classifier_gen.rs`
```rust
fn schema_default_expr(yaml_path: &str, schema_map: &IndexMap<String, FieldInfo>) -> String {
    let Some(info) = schema_map.get(yaml_path) else {
        return "DefaultValue::Missing".to_string();
    };

    match (&info.value_type, &info.default) {
        (FieldType::Duration, Some(default)) => {
            let default_value: serde_json::Value =
                serde_json::from_str(default).expect("schema duration default must be JSON");
            let default_str = default_value
                .as_str()
                .expect("schema duration default must be a string");

            let nanos = go_parse_duration::parse_duration(default_str)
                .expect("schema duration default must be a valid Go duration");

            format!("DefaultValue::DurationNanos({})", nanos)
        }
        (_, Some(default)) => {
            format!("DefaultValue::Json(\"{}\")", escape_str(default))
        }
        (_, None) => "DefaultValue::Missing".to_string(),
    }
}
```

If neither crate passes vetting, then I think the fallback should be to move our existing duration parsing primitive into a smaller lower-level shared crate/module and have both `saluki-config` and `datadog-agent-config` use it. The important part is that the parser has a single owner.

The main point: I like the bugfix direction, but I would rather put the duration-specific behavior into the generated type/default model than add `is_duration` and reparsing logic to the classifier path.

webern (approved):
Approved as a quick fix, with a recommended follow-up (see long comment).

Basically, I think this is closer to right than #1959, but the design I pasted in (even if it's not 100% right as written, not sure, but...) gets us even closer than this to actually dealing with this datatype.

lib/datadog-agent/config/src/classifier/mod.rs
jszwedko: 👍 I like the explicit modeling of duration types. Started a draft follow-up with this direction here: https://github.com/DataDog/saluki/pull/1966

PR #1959, opened 2026-07-01
fix(config): normalize duration defaults to nanoseconds in classifier codegen [alt approach]

jszwedko:
## Summary

Same fix as #1958, different implementation. The nightly integration suite started failing because ADP emitted a spurious `Unsupported configuration key` warning for `tls_handshake_timeout`. The root cause: the classifier stores schema defaults as JSON literals and compares them against Agent-supplied values with exact equality — but duration keys don't survive that comparison because the schema expresses their default as a Go duration string (`10s`) while the Agent transmits durations as integer nanoseconds (`10000000000`). This approach fixes it by normalizing duration defaults to nanoseconds **at codegen time**, so the generated classifier entry for a duration key holds an integer literal that directly matches what the Agent sends. The runtime comparison stays a plain equality check with no added logic.

Companion to #1958 (runtime duration-aware compare) for design comparison.

| | **This PR (codegen)** | **#1958 (runtime)** |
|---|---|---|
| Parser lives in | `schema_gen.rs` (build-time) | `classifier.rs` (runtime) |
| `ClassifierEntry` changes | None | Adds `is_duration: bool` |
| Generated default | `10000000000` (ns integer) | `"10s"` (Go string, unchanged) |
| Readability of generated data | Opaque integers | Human-readable strings |
| Files changed | 3 (schema_gen + classifier tests + generated) | 5 (schema_gen, classifier_gen, mod.rs, classifier.rs, generated) |

## Test plan

- [x] New unit tests in `schema_gen.rs`: Go duration parser covers all common forms (`10s`, `500ms`, `1h30m0s`, `1.5h`, `250µs`, bare `0`, negative, and malformed input).
- [x] New unit tests in `classifier.rs`: duration key sent as nanoseconds is recognized as its default; non-default nanosecond value is not.
- [x] Verified `adp-config-check-no-warn` and `adp-config-stream` pass locally against `agent-dev:master-py3-full`.

### DIFF
--- a/lib/datadog-agent/config-overlay-model/src/schema_gen.rs
+++ b/lib/datadog-agent/config-overlay-model/src/schema_gen.rs
@@ -32,6 +32,10 @@ pub struct FieldInfo {
     /// `no-env` tag.
     pub env_vars: Vec<String>,
     /// Default value serialised as a JSON literal, or `None` if the schema omits one.
+    ///
+    /// For `format: duration` fields the literal is normalized to an integer number of
+    /// nanoseconds (rather than the schema's Go duration string) so it matches the form the
+    /// Datadog Agent transmits over the config stream.
     pub default: Option<String>,
 }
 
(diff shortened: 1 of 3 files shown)


lib/datadog-agent/config-overlay-model/src/schema_gen.rs
```diff
@@ -119,6 +135,101 @@ fn parse_setting(path_parts: &[&str], value: &Value) -> (String, FieldInfo) {
     )
 }
 
+/// Parses a Go `time.Duration` string (for example, `"10s"`, `"500ms"`, `"1h30m0s"`) into
+/// nanoseconds.
+///
+/// Supports the unit suffixes Go's `time.ParseDuration` accepts (`ns`, `us`/`µs`/`μs`, `ms`, `s`,
+/// `m`, `h`), an optional leading sign, and fractional components. Returns `None` if the string is
+/// not a well-formed duration.
+fn parse_go_duration_nanos(input: &str) -> Option<i128> {
```
jszwedko, line 144: This can be refactored out to share the implementation from saluki-config.

tobz (approved):

webern (changes requested):
I think this is the inferior approach because it fixes the problem downstream of where the problem exists. While it's true that, currently, this may be the only downstream effect of the underspecified schema typing, we can't rely on that always being true and I think we should instead fix it at the source, more along the lines of #1958 (but I think that could be done differently and will comment over there next)

jszwedko:
Closing in-lieu of https://github.com/DataDog/saluki/pull/1958

PR #1962, opened 2026-07-01
fix(dogstatsd): preserve host as context dimension

thieman:
## Summary

Fixes ADP DogStatsD custom metric handling when metrics include a `host:<value>` tag.

Datadog Agent treats DogStatsD `host:<value>` as a first-class host dimension:

- `host:<value>` is extracted into the metric sample host and removed from normal tags.
- The host participates in aggregation context identity.
- The final payload emits host as the series host/resource, not as a normal tag.

This PR mirrors that model in ADP:

- adds a focused correctness case, `dsd-host-tag`, that reproduces the divergence
- adds an explicit host dimension to `saluki-context::Context`
- resolves DogStatsD metrics with the extracted host as context host, not as a visible tag
- flows the default hostname into the DogStatsD source so unset host and `host:<default>` share the same context host
- allocation self-review fix: unset host uses default context host without eagerly allocating metadata hostname
- added direct saluki-context coverage for host-preserving context rewrites and mutations
- preserves context host through name/tag/origin rewrites, including DogStatsD mapper remaps
- incorporates review feedback by avoiding per-packet `host:<value>` string reconstruction

## Datadog Agent references

Agent extraction path:

- `extractTagsMetadata` starts with the default hostname, extracts `host:<value>`, and removes the host tag from the metric tag slice: https://github.com/DataDog/datadog-agent/blob/89f7769ceb331b644985df7a0d2dd811eb086a27/comp/dogstatsd/server/impl/enrich.go#L45-L79
- `enrichMetricSample` stores the extracted host in `MetricSample.Host`: https://github.com/DataDog/datadog-agent/blob/89f7769ceb331b644985df7a0d2dd811eb086a27/comp/dogstatsd/server/impl/enrich.go#L144-L190

Agent aggregation/serialization path:

- `MetricSample.GetHost()` and `GetTags()` expose host separately from tags: https://github.com/DataDog/datadog-agent/blob/89f7769ceb331b644985df7a0d2dd811eb086a27/pkg/metrics/metric_sample.go#L122-L130
- context keys include `GetHost()`: https://github.com/DataDog/datadog-agent/blob/89f7769ceb331b644985df7a0d2dd811eb086a27/pkg/aggregator/context_resolver.go#L90-L92
- flushed series copy `context.Host` into `serie.Host`: https://github.com/DataDog/datadog-agent/blob/89f7769ceb331b644985df7a0d2dd811eb086a27/pkg/aggregator/time_sampler.go#L185-L203
- v2 payloads emit host as a `host` resource and tags separately: https://github.com/DataDog/datadog-agent/blob/89f7769ceb331b644985df7a0d2dd811eb086a27/pkg/serializer/internal/metrics/iterable_series.go#L288-L344

## Validation

- `target/release/panoramic run -d test/correctness/cases -t dsd-host-tag --no-tui`
  - before rebuilding the fixed ADP image, failed as expected with baseline-only `hosttag.metric` (empty host) and `hosttag.metric[host:custom]` contexts
  - after rebuilding correctness-tools, ADP, and converged Datadog Agent images, passed locally with the same metric name covering: unset host, `host:`, `host:correctness-testing`, and `host:custom`
- `cargo test --package saluki-context host_ -- --nocapture`
- `cargo test --package saluki-context`
- `cargo test --package saluki-components sources::dogstatsd::tests::metric_host_tag_disambiguates_contexts_without_remaining_tag`
- `cargo test --package saluki-components transforms::dogstatsd_mapper::tests::mapper_preserves_host_context_dimension`
- `cargo test --package saluki-context --package saluki-components`
- `make fmt`
- `cargo check --workspace`
- `git diff --check`
- Commit hook passed for the latest fix commit:
  - `make check-fmt`
  - `make check-clippy`
  - `make check-licenses`
  - `make check-deny` (existing warning for `antithesis_sdk` git source)
  - `make check-docs`
  - `make generate-api-docs`

## Local image rebuild note

The normal `make build-datadog-agent-image-release` path is currently blocked locally by the pinned `ca-certificates=20240203` package no longer being available in Ubuntu 26.04 package indexes. To validate locally, I built a temporary Ubuntu 26.04 runtime base with current CA certificates installed so the ADP app-stage CA install script no-ops, then rebuilt:

- `saluki-images/agent-data-plane:testing-release`
- `saluki-images/datadog-agent:testing-release`

This only affects local validation; no Dockerfile or Makefile changes are included for that workaround.

### DIFF
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -37,7 +37,7 @@ use saluki_config::{ConfigurationLoader, GenericConfiguration};
 use saluki_core::health::HealthRegistry;
 use saluki_core::runtime::{RestartMode, RestartStrategy, Supervisor};
 use saluki_core::topology::TopologyBlueprint;
-use saluki_env::EnvironmentProvider as _;
+use saluki_env::{EnvironmentProvider as _, HostProvider as _};
 use saluki_error::{generic_error, ErrorContext as _, GenericError};
 use tracing::{debug, error, info, trace, warn};
 
@@ -692,8 +692,14 @@ async fn add_dsd_pipeline_to_blueprint(
     //    │    (destination)    │    │                       (Datadog Platform)                        │
     //    └─────────────────────┘    └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘
 
+    let default_hostname = env_provider
+        .host()
+        .get_hostname()
+        .await
+        .error_context("Failed to get default hostname for DogStatsD source.")?;
     let dsd_config = DogStatsDConfiguration::from_configuration(config)
         .error_context("Failed to configure DogStatsD source.")?
+        .with_default_hostname(default_hostname)
         .with_workload_provider(env_provider.workload().clone())
         .with_capture_entity_resolver(env_provider.workload().clone());
     let dsd_prefix_filter_configuration = DogStatsDPrefixFilterConfiguration::from_configuration(config)?;
--- a/lib/saluki-components/src/sources/dogstatsd/mod.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/mod.rs
@@ -1776,6 +1776,7 @@ fn handle_metric_packet(
     let tags = get_filtered_tags_iterator(packet.tags, additional_tags);
 
     let hostname = well_known_tags.hostname.unwrap_or(default_hostname);
+    let metadata_hostname = well_known_tags.hostname.map(Arc::from);
 
     // Try to resolve the context for this metric.
     let maybe_context = context_resolver.resolve_with_host(packet.metric_name, hostname, tags, Some(origin));
@@ -1788,7 +1789,7 @@ fn handle_metric_packet(
                 .unwrap_or_else(MetricOrigin::dogstatsd);
             let metadata = MetricMetadata::default()
                 .with_origin(metric_origin)
-                .with_hostname(Some(Arc::from(hostname)))
+                .with_hostname(metadata_hostname)
                 .with_unit(packet.unit.map_or_else(MetaString::empty, MetaString::from_static));
 
             Some(Metric::from_parts(context, packet.values, metadata))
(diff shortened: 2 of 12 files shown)


lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1749,8 +1750,23 @@ fn handle_metric_packet(
 
     let tags = get_filtered_tags_iterator(packet.tags, additional_tags);
 
+    let host_context_tag = well_known_tags
```
tobz: Nah, we can't be allocating a string for every single packet just to reconstruct a host tag.

We should add an additional field to `WellKnownTags` called `hostname_raw`, that contains the full tag. We should only set that field (to the value of the `tag` itself) when we've matched against `HOST_TAG_KEY`.

Once we have that, then we can skip doing any sort of allocating here.

tobz (changes requested):

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -692,8 +692,14 @@ async fn add_dsd_pipeline_to_blueprint(
...
     //    └─────────────────────┘    └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘
 
+    let default_hostname = env_provider
+        .host()
+        .get_hostname()
+        .await
+        .error_context("Failed to get default hostname for DogStatsD source.")?;
     let dsd_config = DogStatsDConfiguration::from_configuration(config)
         .error_context("Failed to configure DogStatsD source.")?
+        .with_default_hostname(default_hostname)
```
tobz, line 702: I'm saying this out loud so we don't forget, but it's _not_ a blocking comment: we should do this when building the source, not ahead of time.

The reason I think we should do it then is that it lets us construct the _blueprint_ without not reaching out to actual external resources, which (IMO) makes it a little cleaner. If the source needs the hostname, it can just grab that from the environment provider directly during the building phase... since we're already giving it the environment provider anyways.

This _would_ necessitate a configuration to instruct it to do so, to replace the inherent enabled/disabled aspect that we get from calling `with_default_hostname`... but 🤷🏻.

tobz (approved):

PR #1964, opened 2026-07-01
fix(test): use single-underscore env for ADP listen addresses

jszwedko:
## Summary

With the addition of the `data_plane` keys to the Agent schema, ADP was reading the defaults rather than the test configuration due to the Agent not recognizing the `__` variants of environment variables like `DD_DATA_PLANE__API_LISTEN_ADDRESS`. This PR updates the test config to use the single-underscore variants.

For standalone mode, `saluki-config` already supports falling back to reading the single-underscore variants (like `DD_DATA_PLANE_API_LISTEN_ADDRESS`) if the double-underscore variant isn't set. A test is added to assert this.

**Converged mode** — the Core Agent reads `DD_DATA_PLANE_API_LISTEN_ADDRESS` (single-underscore, viper form) and propagates it as the authoritative `data_plane.api_listen_address` via the config stream. ADP resolves the nested key on the first try.

**Standalone mode** — no config stream; `from_environment("DD_")` splits on `__` and produces the flat figment key `data_plane_api_listen_address`. `get()` retries on `MissingField` with the flat form and resolves it. A regression test (added to saluki-config) documents and protects this fallback.

> [!NOTE]
> `DD_DATA_PLANE__API_LISTEN_ADDRESS` (double-underscore) remains fully functional — the nested key is checked before the flat fallback, so existing scripts using that form continue to work.

## Test plan

- [x] Added `test_underscore_fallback_on_get_multi_segment_key` to saluki-config, confirming `try_get_typed("data_plane.api_listen_address")` resolves from a single-underscore env var via the existing `get()` fallback.
- [x] All 5 previously-failing converged tests pass against `agent-dev:master-py3-full` with base-config ADP (no `config.rs` change): `privileged-api-endpoints`, `unprivileged-api-endpoints`, `telemetry-endpoint`, `adp-config-dynamic-basic`. (`dogstatsd-forwarding` forwarding all passes; only remaining failure is the separate regression #1 tls classifier bug.)
- [x] 3 standalone-mode tests (`basic-startup`, `adp-memory-mode-strict-within-limit`, `otlp-traces-enabled`) pass — ADP resolves the shifted listen address via the flat-key fallback.

### DIFF
--- a/bin/correctness/panoramic/src/test_env.rs
+++ b/bin/correctness/panoramic/src/test_env.rs
@@ -15,10 +15,12 @@ use std::collections::HashMap;
 /// 5001 -> 55001, etc.). The GUI is disabled outright since we don't exercise it.
 ///
 /// Note on env-var nesting: saluki-config (and figment) split env-var names on `__` to map to
-/// nested config keys. Single-underscore env vars like `DD_DATA_PLANE_API_LISTEN_ADDRESS` map
-/// to the flat key `data_plane_api_listen_address` and are silently ignored; we use double
-/// underscores at every dot boundary for the deep ADP / OTLP keys below. The top-level Agent
-/// env vars (`DD_CMD_PORT` etc.) are explicitly queried by the Agent so they don't need it.
+/// nested config keys. The ADP listen-address fields accept both the single-underscore Agent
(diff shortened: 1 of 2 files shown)


lucastemb (approved):

webern (approved):
Much less scary than #1958 and #1959!

PR #1966, opened 2026-07-01
chore(config): model duration defaults as a typed value

jszwedko:
## Summary

Follow-up to #1958 (now merged), taking the alternative design proposed in review (https://github.com/DataDog/saluki/pull/1958#discussion_r3508274012): model `format: duration` as a first-class type and store **parsed** duration defaults in the generated classifier data, instead of carrying an `is_duration` side channel and reparsing a schema string on every classification. The base type is modeled correctly, and an invalid duration default now breaks the build rather than misbehaving at runtime.

Key changes:

- **New `go-duration` leaf crate** owns Go `time.ParseDuration` parsing. `saluki-config`'s `DurationString` delegates to it, and the classifier build script uses it too — so the algorithm has a single owner instead of being reimplemented (the parser #1958 added to the classifier is removed). An upstream crate (for example `go-parse-duration`) can replace this later after vetting; keeping it in-tree for now avoids taking an unvetted external dependency.
- **`FieldType::Duration`** (plus a matching `ValueType::Duration` for the config-testing smoke tests) replaces the `is_duration` flag. `parse_value_type` folds `format: duration` in and panics on an illogical base type. Smoke tests use a generic `"42s"` value so future duration keys need no bespoke `test_json`.
- **Typed `DefaultValue`** (`Missing` / `Json(&'static str)` / `DurationNanos(u64)`) replaces `Option<&str>` + `is_duration` on `ClassifierEntry`. Codegen parses duration defaults to nanoseconds via `go-duration` and **fails the build** if a `format: duration` default isn't a valid Go duration. The runtime default check simply switches on the typed default; it never parses a schema default.

Behavior matches #1958 for the `tls_handshake_timeout` bug (a default-valued duration is recognized as default whether sent as nanoseconds or a Go duration string), but the model is cleaner and the failure mode moves to compile time.

## Test plan

- [x] `go-duration` unit tests (Go-style units, signs/fractions/zero, largest value, invalid/overflow, error messages).
- [x] `saluki-config` duration tests pass against the delegated parser.
- [x] Classifier unit tests updated for `DefaultValue` (per-variant `is_default_value` coverage; duration matched both as nanoseconds and as a Go duration string).
- [x] `cargo check --workspace --tests` and `cargo clippy` clean; config-testing smoke tests pass.


webern (approved):
LGTM on a quick pass and does what I think I need on my branch(es) 👍

lib/datadog-agent/config-overlay-model/src/schema_gen.rs
```diff
@@ -21,6 +21,11 @@ pub enum FieldType {
     Integer,
     Float,
     StringList,
+    /// A `format: duration` field. Modeled as its own effective type because the Datadog Agent
+    /// transmits durations as integer nanoseconds, while the schema default is a Go duration string
+    /// (for example, `10s`). Downstream codegen, smoke tests, and classification all need to treat
+    /// it differently from a plain number.
```
jszwedko:
```suggestion
```

PR #1975, opened 2026-07-02
fix(correctness): decode v3 tagset prefix references in the intake parser

jszwedko:
## Summary

Alternative to #1973 (which pinned the Core Agent back to v2). Instead of avoiding v3, this fixes the correctness harness so it correctly decodes the Agent's **v3 series payloads**, letting the v3 baseline and v2 (ADP) comparison be compared faithfully.

**Root cause.** The Agent enabled the v3 series API by default in 7.81.0 (datadog-agent commit `68ede29f685`, #52059). The intake's v3 tagset decoder (`bin/correctness/stele/src/metrics.rs::parse_tagsets`) assumed every entry in a `dictTagsets` tagset was a positive tag-string index. But the Agent **prefix-compresses composite tags**: when a metric's tags span two groups (e.g. host/low-cardinality tags *and* DogStatsD origin tags), the second group's tagset is encoded with a negative `-prefix_tagset_id` entry referencing the first group's tagset, then sorted and delta-encoded (see `internTags`/`internTags1` in `iterable_series_v3.go`). The decoder hit those negative entries and failed the entire payload with `Invalid negative tag index: -1`, dropping ~half the baseline metrics (recovered **630 of 3030**) and producing false mismatches. Origin-detection cases produce exactly these composite tags, which is why they were the ones failing.

**Fix.** Resolve a negative entry `-N` as a reference to the previously-interned tagset with 1-based id `N`, including its fully-resolved tags. The referenced tagset is always interned before the one referencing it, so a single forward pass resolves it.

## Test plan
- [x] New `stele` unit tests: positive-index tagsets, negative-prefix-reference resolution, and rejection of a forward reference.
- [x] Reproduced locally against `agent-dev:master-py3-full` with the Agent on v3: before, baseline processed **2/4** v3 payloads (630 metrics); after, **4/4** (3030), context sets match exactly (0 divergence), and `dsd-origin-detection` **passes**.
- [x] `cargo test -p stele -p datadog-intake` and `cargo clippy` clean.

> [!NOTE]
> A residual, unrelated flakiness remains under local arm64 emulation: a handful of histogram percentile summaries (`.avg`/`.median`/`.max`/`.95percentile`) differ due to flush-bucket timing skew (also present on the v2 path). It's an emulation artifact, not a v3-parsing issue, and does not occur on native CI.
>
> If this lands, #1973's v2 pin becomes unnecessary and can be dropped so the suite exercises v3 end-to-end.

### DIFF
--- a/bin/correctness/stele/src/metrics.rs
+++ b/bin/correctness/stele/src/metrics.rs
@@ -685,9 +685,18 @@ fn parse_dict_strings(data: &[u8]) -> Result<Vec<String>, GenericError> {
 
 /// Parse tagsets from the `dictTagsets` array using the tag dictionary.
 ///
-/// Each tagset in `dict_tagsets` is encoded as: length, then that many delta-encoded tag indices.
+/// Each tagset in `dict_tagsets` is encoded as: length, then that many sorted-then-delta-encoded
+/// entries. Each decoded entry is one of:
+///
+/// - A positive value: a 1-based index into the tag-string dictionary.
+/// - A negative value `-N`: a prefix reference to the previously-interned tagset with 1-based ID
+///   `N`, whose (already fully-resolved) tags are included in this one. The Agent emits these when
+///   it splits a metric's composite tags into two groups and prefix-compresses the second group
+///   against the first (see `internTags` in the Agent's `iterable_series_v3.go`). Because the
+///   referenced tagset is always interned before the tagset that references it, a single forward
+///   pass can resolve the reference against the tagsets parsed so far.
 fn parse_tagsets(dict_tagsets: &[i64], tags_dict: &[String]) -> Result<Vec<Vec<String>>, GenericError> {
-    let mut tagsets = Vec::new();
+    let mut tagsets: Vec<Vec<String>> = Vec::new();
     let mut offset = 0;
     while offset < dict_tagsets.len() {
         let count = i64_to_usize(dict_tagsets[offset], "tagset length")?;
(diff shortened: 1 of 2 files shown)


lucastemb (approved):

bin/correctness/stele/src/metrics.rs
```diff
@@ -685,9 +685,18 @@ fn parse_dict_strings(data: &[u8]) -> Result<Vec<String>, GenericError> {
...
+/// Each tagset in `dict_tagsets` is encoded as: length, then that many sorted-then-delta-encoded
+/// entries. Each decoded entry is one of:
+///
+/// - A positive value: a 1-based index into the tag-string dictionary.
+/// - A negative value `-N`: a prefix reference to the previously interned tagset with 1-based ID
+///   `N`, whose (already fully resolved) tags are included in this one. The Agent emits these when
+///   it splits a metric's composite tags into two groups and prefix-compresses the second group
+///   against the first (see `internTags` in the Agent's `iterable_series_v3.go`). Because the
+///   referenced tagset is always interned before the tagset that references it, a single forward
+///   pass can resolve the reference against the tagsets parsed so far.
```
jszwedko, line 697:
```suggestion
/// - A positive value: a 1-based index into the tag-string dictionary.
/// - A negative value `-N`: a backreference to the previously interned tagset with 1-based ID
///   `N`, whose (already fully resolved) tags are included in this one. The Agent emits these when
///   it splits a metric's composite tags into two groups and encodes the second group
///   against the first (see `internTags` in the Agent's `iterable_series_v3.go`). Because the
///   referenced tagset is always interned before the tagset that references it, a single forward
///   pass can resolve the reference against the tagsets parsed so far.
```

PR #1977, opened 2026-07-02
chore(config): add runtime configuration updates

webern:
## Human Summary

Adds a typed configuration update subscription mechanism, known as `Live<T>`.

The contract is to get a Live<T> from the type system like this:

```rust
let live = config_system.live(|c| &c.domains.dogstatsd.debug_log)
```

Then you put `live.changes()` in your tokio `select!`

```rust
  _ = live.changed() => {
      let metrics_stats_enabled = live.current().metrics_stats_enable;
      self.metrics_stats_enabled = metrics_stats_enabled;
      debug!(metrics_stats_enabled, "Updated DogStatsD metrics stats debug logging gate.");
```

## AI Summary

Adds the typed runtime configuration library and `Live<T>` view without wiring it into the Agent Data Plane binary.

### Configuration system

The new `agent-data-plane-config-system` crate translates two configuration sources into the ADP-native `SalukiConfiguration` model:

- Datadog configuration, translated through the generated configuration witness
- Saluki-only configuration, which seeds fields not represented in the Datadog schema

`ConfigurationSystem`:

- translates an initial `GenericConfiguration` into a complete typed configuration
- subscribes to updates when the source map is dynamic
- coalesces bursts of updates before re-translating
- fails startup when the initial source cannot be translated
- keeps runtime updates durable by retaining defaults for invalid translated values while applying other valid changes
- exposes the current configuration through a shared `ArcSwap`

### Live view of the configuration

`Live<T>` in `agent-data-plane-config` provides a typed view over a projection of `SalukiConfiguration`:

- `Live::fixed` creates a value that never changes
- `Live::dynamic` tracks a shared configuration snapshot
- `current()` returns the latest projected value
- `Deref` reads the view's latest snapshot
- `changed()` waits for the projected value to change
- `project()` narrows a view to a child node or field

The view uses `ArcSwap` for consistent snapshots and a Tokio `watch` channel for update notifications. A view only wakes its consumer when its projected value changes.

This PR is library-only. It does not modify `bin/agent-data-plane`, wire the configuration system into startup, add the `/config/internal` endpoint, or add integration-test coverage. Those changes are deferred to PR-3b, where environment handling and runtime wiring will be addressed together.

## Change Type

## How did you test this PR?

- `make check-clippy`
- Unit tests in `agent-data-plane-config-system` cover startup translation, translation errors, update convergence, projections, field-level live views, fixed views, and startup values.

No integration test is included in this PR because the Agent Data Plane wiring is deferred to PR-3b.

## References

- Depends on https://github.com/DataDog/saluki/pull/1957

### DIFF
--- a/lib/agent-data-plane-config-system/src/system.rs
+++ b/lib/agent-data-plane-config-system/src/system.rs
@@ -1,0 +1,538 @@
+//! [`ConfigurationSystem`]: the runtime configuration, translated from the raw source map and kept
+//! current as the map streams updates.
+
+use std::sync::Arc;
+
+use agent_data_plane_config::{Live, SalukiConfiguration};
+use arc_swap::ArcSwap;
+use datadog_agent_config::{DatadogConfiguration, TranslateErrors};
+use saluki_config::dynamic::ConfigChangeEvent;
+use saluki_config::{ConfigurationError, GenericConfiguration};
+use serde::Deserialize;
+use snafu::Snafu;
+use tokio::sync::{broadcast, watch};
+use tracing::{debug, warn};
+
+use crate::saluki_only::SalukiOnly;
+use crate::translators::DatadogTranslator;
+
+/// An error building the translated configuration from the raw source map.
+#[derive(Debug, Snafu)]
+pub enum Error {
+    /// The merged configuration value could not be read from the raw configuration map.
+    #[snafu(context(false), display("{source}"))]
+    Source {
+        /// The underlying configuration error.
+        source: ConfigurationError,
+    },
+
+    /// A source model could not be deserialized from the merged configuration value.
+    #[snafu(context(false), display("{source}"))]
+    Deserialize {
+        /// The underlying deserialization error.
+        source: serde_json::Error,
+    },
+
+    /// Translating the sources into the model failed on one or more keys.
+    #[snafu(display("{source}"))]
+    Translate {
+        /// Every translation error recorded.
+        source: TranslateErrors,
+    },
+}
+
+type Result<T> = std::result::Result<T, Error>;
+
+/// The runtime configuration, translated from the raw source map and kept current.
+///
+/// Holds the raw source map, the current translated [`SalukiConfiguration`], and (when the source
+/// map is dynamic) a background task that re-translates on every committed source update. The
+/// current configuration lives in an [`ArcSwap`] cell so readers load a whole, self-consistent
+/// version with no lock, while the background task replaces it in one atomic store.
+pub struct ConfigurationSystem {
+    raw_map: GenericConfiguration,
+    current: Arc<ArcSwap<SalukiConfiguration>>,
+    // Fired once after each accepted update so live views wake and re-project. Shared with the
+    // router task via `Arc` because `watch::Sender` is not `Clone` and both the system (to mint
+    // views) and the task (to notify) need it.
+    tick: Arc<watch::Sender<()>>,
+}
+
+impl ConfigurationSystem {
+    /// Translates `raw_map` into the initial configuration and, if the map streams updates, starts
+    /// the task that keeps it current.
+    ///
+    /// Translation always yields a complete configuration; individual values that cannot be
+    /// converted keep their defaults and are logged.
+    ///
+    /// `async` because starting the update task requires a Tokio runtime; this keeps that
+    /// requirement visible at the call site instead of panicking deep inside `tokio::spawn`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the raw map cannot be deserialized into the source models.
+    pub async fn load(raw_map: GenericConfiguration) -> Result<Self> {
+        // Subscribing before reading, in one call, makes it impossible to lose a commit that lands
+        // between the two: see `subscribe_then_deserialize`.
+        let (maybe_rx, sources) = subscribe_then_deserialize(&raw_map);
+        let Sources { datadog, saluki } = sources?;
+        let (config, errors) = translate(&datadog, &saluki);
+
+        // Startup and runtime take opposite stances on translation errors. Startup is the strict
+        // gate: this is the first, authoritative Agent snapshot, so any error fails the load and we
+        // never boot on bad config. Runtime updates (see `apply`) must not fail or panic and take
+        // the system down, and the bad value is already committed to the shared map, so they log
+        // and keep operating on the valid config. The real fix is to sanitize Agent input before it
+        // enters the shared map, a larger change deferred until this system proves itself.
+        // TODO: sanitize source input before it reaches the shared config map.
+        if let Some(errors) = errors {
+            return Err(Error::Translate { source: errors });
+        }
+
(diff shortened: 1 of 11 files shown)


webern:
* **#1986** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1986?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1977** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1977?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/1977?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1957** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1957?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/agent-data-plane-config-system/src/system.rs
```diff
@@ -0,0 +1,521 @@
...
+        // The initial receiver is dropped immediately; `send_replace` works with zero receivers, and
+        // each live view subscribes its own receiver from the sender.
+        let (tick, _) = watch::channel(());
+        let tick = Arc::new(tick);
+
+        // A dynamic source map broadcasts one event per committed key change. The task ignores the
+        // event contents and re-reads the whole committed map, so it only exists when updates can
+        // arrive; a static map has no sender and needs no task.
+        if let Some(rx) = raw_map.subscribe_for_updates() {
+            spawn_router(rx, raw_map.clone(), Arc::clone(&current), Arc::clone(&tick));
```
tobz: `spawn_router` is a single line... so we should just move it into this, and make this function `async` so that we know we're only executing it in the context of an asynchronous runtime... making it little a little more misuse-resistant.

lib/agent-data-plane-config-system/src/system.rs
```diff
@@ -0,0 +1,521 @@
...
+        self.raw_map.clone()
+    }
+}
+
+/// Spawns the task that keeps `current` up to date as `raw_map` commits updates.
+fn spawn_router(
+    rx: broadcast::Receiver<ConfigChangeEvent>, raw_map: GenericConfiguration,
+    current: Arc<ArcSwap<SalukiConfiguration>>, tick: Arc<watch::Sender<()>>,
+) {
+    tokio::spawn(router_loop(rx, raw_map, current, tick));
```
tobz: Just a note: we'll eventually want to make this slot into the normal supervision tree.

Not a change to make here/now.

lib/agent-data-plane-config-system/src/system.rs
```diff
@@ -0,0 +1,521 @@
...
+/// instantiate the giant generated `DatadogConfiguration` visitor through figment once per model
+/// (measured ~1.6 MiB of binary). The source models use no figment magic types (path fields are
+/// plain strings), so nothing is lost.
+///
+/// Transitional: this seam goes away when `GenericConfiguration` is eliminated.
+fn deserialize_sources(
+    raw_map: &GenericConfiguration,
+) -> Result<(DatadogConfiguration, SalukiOnly), ConfigurationSystemError> {
+    let merged = raw_map.as_typed::<serde_json::Value>()?;
+    let datadog = DatadogConfiguration::deserialize(&merged)?;
```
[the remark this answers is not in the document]
tobz: At face value, this makes sense to me. It shouldn't matter in normal operation when configuration comes from the Core Agent (with the right types, etc) but would matter in standalone mode when we often set things via environment variable... and `figment` does a lot of heavy lifting on trying to coerce the completely stringified values into the fuzzy detected type.

tobz (changes requested):

lib/agent-data-plane-config-system/src/system.rs
webern: Nice catch. I think this is what you mean: https://github.com/DataDog/saluki/compare/4e175c2211e7a1df92fe1e88d380aded06dd2e15..fea786dca512345f8e1fd1736422f90984b5a5ec?diff=split&w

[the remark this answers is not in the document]
webern: Fixed in: https://github.com/DataDog/saluki/compare/fea786dca512345f8e1fd1736422f90984b5a5ec..1bae649157c0ef1ef4db471e270915c2ece6b23f

lib/agent-data-plane-config-system/src/system.rs
webern: That makes sense. If I understand right that will mean implementing a trait, at which point the supervisor can kill it, restart it, etc. 👍

lib/agent-data-plane-config-system/src/system.rs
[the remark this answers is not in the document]
webern: The backround is that `raw_map.as_typed::<DatadogConfiguration>()?` increased the binary size by 5 MB 😅 . And when I dug in a bit it was all Figment. It was... impressive. So I tried passing it through serde and it reduced the bloat. In the next stacked PR I actually start dealing with the environment variables directly in the deserialization.

I actually have a hard time wrapping my head around Figment and what it's doing exactly. I feel like I'm low-key on a path to replacing Figment just because I never quite seem to get it 😅 .

Clanker said the following... seems fine?

Clanker says:
>  I dug into this. Figment’s environment provider parses values before they enter the configuration map: 58125 becomes a number and true becomes
 a boolean. Converting the map to serde_json::Value preserves those types, and I verified that both DD_DOGSTATSD_PORT=58125 and
 DD_ALLOW_ARBITRARY_TAGS=true deserialize correctly through this path. Also, GenericConfiguration::get uses Figment’s non-lossy extractor, so
 switching to it would not provide the coercion described in the comment. There may be edge cases involving Figment-specific types, but the
 reported environment-variable failure does not appear to apply here.

lib/agent-data-plane-config-system/src/system.rs
tobz: Exactly, wrapping this up as a worker... and then also tying in the configuration system _somehow_ so it can be "spawned" (for some definition of "spawn") and this lil' guy gets spawned at _that_ point.

lib/agent-data-plane-config-system/src/system.rs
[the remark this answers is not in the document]
tobz: Still a little suspicious of it being able to work without the equivalent `figment` logic present, in terms of standalone mode... but if tests currently pass, that's good enough for me. 👍🏻

tobz (approved):

webern:
I removed the integration with `agent-data-plane` because it needs more work as @tobz and the bot recognized.

This push removes the `agent-data-plane` touch points so I can still land the reviewed `Live<T>` work.

https://github.com/DataDog/saluki/compare/32eb7749c8369b102967c36c78af0ddd8947de69..7f68a1a5eb845da6dfd863aebf740de0b6394df5

PR #1978, opened 2026-07-02
fix(metrics): store host on context only

thieman:
## Summary

Follow-up to #1962. Removes metric hostname from `MetricMetadata` and makes `Context.host` the canonical metric host source.

This PR:

- removes `MetricMetadata::hostname`, `with_hostname`, and `set_hostname`
- makes `Context.host` optional so unset host and explicit empty host remain distinguishable
- updates host enrichment to set `Context.host` instead of metadata hostname
- updates metric encoders to emit host from `metric.context().host()`
- updates metric producers/tests that previously used metadata hostname to use context host
- adds direct host enrichment tests for host-enrichment metric no-op behavior
- moves default host assignment for Checks IPC and native OTLP metrics into their source paths, so host enrichment no longer rekeys metric contexts after source emission
- adds OTLP metrics host permutation correctness coverage using a static base64 protobuf payload
- reduces source-level host allocations by carrying OTLP dimension hosts as `MetaString` and cloning default check host `MetaString` directly
- avoids cloning owned context host values before resolver cache lookup when hashing context keys
- moves resource-derived OTLP host strings into `MetaString` instead of copying from borrowed strings
- exposes `Context::host_meta()` so remap paths can preserve cheap `MetaString` host clones

## Why

#1962 adds host to metric context identity for DogStatsD correctness. Keeping hostname in both `Context` and `MetricMetadata` duplicates host storage and preserves the old per-metric metadata allocation path. This change makes context host the single source of truth for metric host identity and emission.

## Validation

- `cargo test --package saluki-context host_ -- --nocapture`
- `cargo test --package saluki-context`
- `cargo test --package saluki-components sources::dogstatsd::tests::metric_host_tag_disambiguates_contexts_without_remaining_tag`
- `cargo test --package saluki-components transforms::dogstatsd_mapper::tests::mapper_preserves_host_context_dimension`
- `cargo test --package saluki-components transforms::host_enrichment::tests:: -- --nocapture`
- `cargo test --package saluki-components sources::checks_ipc::tests::metric_hostname_propagates`
- `cargo test --package saluki-components sources::checks_ipc::tests::metric_empty_hostname_uses_default_host`
- `cargo test --package saluki-components sources::otlp::metrics::translator::tests::translate_metrics_ -- --nocapture`
- `cargo test --package saluki-context --package saluki-components`
- `make fmt`
- `cargo test --package millstone`
- `make build-correctness-tools-image`
- rebuilt ADP and converged Datadog Agent images using the local CA-base workaround from #1962
- `target/release/panoramic run -d test/correctness/cases -t dsd-host-tag --no-tui` — passed locally
- `target/release/panoramic run -d test/correctness/cases -t otlp-metrics-host --no-tui` — passed locally
- commit hook passed all configured checks

### DIFF
--- a/lib/saluki-components/src/sources/checks_ipc/mod.rs
+++ b/lib/saluki-components/src/sources/checks_ipc/mod.rs
@@ -1,4 +1,4 @@
-use std::sync::{Arc, LazyLock};
+use std::sync::LazyLock;
 use std::time::Duration;
 
 use async_trait::async_trait;
@@ -183,8 +183,11 @@ fn check_data_to_event(check_data: Data) -> Option<Event> {
             let metric_type = MetricType::try_from(r#type).ok()?;
 
             let tags = tags.into_iter().map(Tag::from).collect::<TagSet>();
-            let context = Context::from_parts(name, tags.into_shared());
-            let mut metric = match metric_type {
+            let mut context = Context::from_parts(name, tags.into_shared());
+            if !hostname.is_empty() {
+                context = context.with_host(Some(hostname.into()));
+            }
+            let metric = match metric_type {
                 MetricType::Counter => Metric::counter(context, (timestamp, value)),
                 MetricType::Gauge => Metric::gauge(context, (timestamp, value)),
                 MetricType::Rate => {
@@ -200,9 +203,6 @@ fn check_data_to_event(check_data: Data) -> Option<Event> {
                     return None;
                 }
             };
-            if !hostname.is_empty() {
-                metric.metadata_mut().set_hostname(Arc::from(hostname));
-            }
             Some(Event::Metric(metric))
         }
         Data::Log(log) => {
@@ -593,7 +593,7 @@ mod tests {
         let Event::Metric(m) = event else {
             panic!("expected Metric event");
         };
-        assert_eq!(m.metadata().hostname(), Some("host-a"));
+        assert_eq!(m.context().host(), Some("host-a"));
     }
 
     #[test]
@@ -603,7 +603,7 @@ mod tests {
         let Event::Metric(m) = event else {
             panic!("expected Metric event");
         };
-        assert_eq!(m.metadata().hostname(), None);
+        assert_eq!(m.context().host(), None);
     }
 
     #[test]
--- a/lib/saluki-components/src/sources/checks_ipc/mod.rs
+++ b/lib/saluki-components/src/sources/checks_ipc/mod.rs
@@ -100,7 +100,12 @@ struct ChecksIPC {
 
 #[async_trait]
 impl Source for ChecksIPC {
-    async fn run(mut self: Box<Self>, mut context: SourceContext) -> Result<(), GenericError> {
+    async fn run(self: Box<Self>, mut context: SourceContext) -> Result<(), GenericError> {
+        let ChecksIPC {
+            grpc_endpoint,
+            default_hostname,
+        } = *self;
+
         let global_shutdown = context.take_shutdown_handle();
         pin!(global_shutdown);
 
@@ -110,10 +115,10 @@ impl Source for ChecksIPC {
 
         let grpc_server = Server::builder().add_service(ChecksServer::new(ChecksService {
             events_tx,
-            default_hostname: self.default_hostname,
+            default_hostname,
         }));
 
-        let grpc_socket_addr = match self.grpc_endpoint {
+        let grpc_socket_addr = match grpc_endpoint {
             ListenAddress::Tcp(addr) => addr,
             _ => return Err(generic_error!("OTLP gRPC endpoint must be a TCP address.")),
         };
--- a/lib/saluki-components/src/sources/checks_ipc/mod.rs
+++ b/lib/saluki-components/src/sources/checks_ipc/mod.rs
@@ -43,6 +43,9 @@ const fn default_grpc_endpoint() -> ListenAddress {
 /// Checks IPC source.
 #[derive(Debug, Deserialize)]
 pub struct ChecksIPCConfiguration {
+    #[serde(skip)]
+    default_hostname: MetaString,
+
     #[serde(rename = "checks_ipc_endpoint", default = "default_grpc_endpoint")]
     grpc_endpoint: ListenAddress,
 }
@@ -51,6 +54,12 @@ impl ChecksIPCConfiguration {
     /// Creates a new `ChecksIPCConfiguration` from the given configuration.
     pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
         Ok(config.as_typed()?)
+    }
+
+    /// Sets the default hostname used when check metrics do not carry an explicit hostname.
+    pub fn with_default_hostname(mut self, hostname: impl Into<MetaString>) -> Self {
+        self.default_hostname = hostname.into();
+        self
     }
 }
 
@@ -72,6 +81,7 @@ impl SourceBuilder for ChecksIPCConfiguration {
     async fn build(&self, _context: ComponentContext) -> Result<Box<dyn Source + Send>, GenericError> {
         Ok(Box::new(ChecksIPC {
             grpc_endpoint: self.grpc_endpoint.clone(),
+            default_hostname: self.default_hostname.clone(),
         }))
     }
 }
@@ -85,6 +95,7 @@ impl MemoryBounds for ChecksIPCConfiguration {
 
 struct ChecksIPC {
     grpc_endpoint: ListenAddress,
+    default_hostname: MetaString,
 }
 
 #[async_trait]
@@ -97,7 +108,10 @@ impl Source for ChecksIPC {
 
         let (events_tx, mut events_rx) = mpsc::channel(16);
 
-        let grpc_server = Server::builder().add_service(ChecksServer::new(ChecksService { events_tx }));
+        let grpc_server = Server::builder().add_service(ChecksServer::new(ChecksService {
+            events_tx,
+            default_hostname: self.default_hostname,
+        }));
 
         let grpc_socket_addr = match self.grpc_endpoint {
             ListenAddress::Tcp(addr) => addr,
@@ -141,6 +155,7 @@ impl Source for ChecksIPC {
 
 struct ChecksService {
     events_tx: mpsc::Sender<Event>,
+    default_hostname: MetaString,
 }
 
 #[async_trait]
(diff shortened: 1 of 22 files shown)


test/correctness/cases/otlp-metrics-host/millstone.yaml
```diff
@@ -0,0 +1,8 @@
+seed: [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131]
+target: "grpc://$GROUP:4317/opentelemetry.proto.collector.metrics.v1.MetricsService/Export"
+aggregation_bucket_width_secs: 10
+volume: 1000
+corpus:
+  size: 1
+  payload:
+    static_base64: "Ci0KABIpEicKD290bHBob3N0Lm1ldHJpYyoUChIZAMqaOwAAAAAhAAAAAAAA8D8KPgoRCg8KCWhvc3QubmFtZRICCgASKRInCg9vdGxwaG9zdC5tZXRyaWMqFAoSGQDKmjsAAAAAIQAAAAAAAABAClEKJAoiCglob3N0Lm5hbWUSFQoTY29ycmVjdG5lc3MtdGVzdGluZxIpEicKD290bHBob3N0Lm1ldHJpYyoUChIZAMqaOwAAAAAhAAAAAAAACEAKRAoXChUKCWhvc3QubmFtZRIICgZjdXN0b20SKRInCg9vdGxwaG9zdC5tZXRyaWMqFAoSGQDKmjsAAAAAIQAAAAAAABBA"
```
thieman, line 10: Add a comment here explaining what you put in this payload

lib/saluki-context/src/context.rs
```diff
@@ -178,9 +178,29 @@ impl Context {
         &self.inner.name
     }
 
-    /// Returns the host of this context.
-    pub fn host(&self) -> &MetaString {
-        &self.inner.host
+    /// Returns the host of this context, if one has been set.
+    pub fn host(&self) -> Option<&str> {
```
thieman, line 182: Why the change off of MetaString?

lib/saluki-components/src/transforms/host_enrichment/mod.rs
```diff
@@ -99,13 +91,51 @@ impl HostEnrichment {
 impl SynchronousTransform for HostEnrichment {
     fn transform_buffer(&mut self, event_buffer: &mut EventsBuffer) {
         for event in event_buffer {
-            if let Some(metric) = event.try_as_metric_mut() {
-                self.enrich_metric(metric);
```
thieman, line 103: This transform no longer actually enriches metrics at all, if I'm understanding

test/correctness/cases/otlp-metrics-host/millstone.yaml
thieman: [GPT 5.5] Added a comment above the `static_base64` payload explaining that it is an `ExportMetricsServiceRequest` containing four `otlphost.metric` gauge points for unset host, `host.name=""`, `host.name="correctness-testing"`, and `host.name="custom"`.

lib/saluki-context/src/context.rs
thieman: [GPT 5.5] The later migration changed this to `Option<MetaString>` internally and `Option<&str>` externally so we can distinguish unset host (`None`) from explicit empty host (`Some("")`). That distinction matters because unset should default to the Agent hostname, while explicit empty must remain empty. Returning `Option<&str>` keeps callers from depending on the concrete `MetaString` storage while still exposing the semantic host state.

lib/saluki-components/src/transforms/host_enrichment/mod.rs
thieman: [GPT 5.5] Correct. I changed the design so metric hosts are assigned at the metric sources/translators (DogStatsD, Checks IPC, native OTLP) and host enrichment no longer rekeys metrics after source emission. I also updated the transform docs and added a test asserting metric context host is left unchanged. Host enrichment still enriches events and service checks.

lib/saluki-components/src/sources/checks_ipc/mod.rs
```diff
@@ -183,8 +198,14 @@ fn check_data_to_event(check_data: Data) -> Option<Event> {
             let metric_type = MetricType::try_from(r#type).ok()?;
 
             let tags = tags.into_iter().map(Tag::from).collect::<TagSet>();
-            let context = Context::from_parts(name, tags.into_shared());
-            let mut metric = match metric_type {
+            let mut context = Context::from_parts(name, tags.into_shared());
+            let hostname = if hostname.is_empty() {
+                default_hostname.as_ref()
+            } else {
```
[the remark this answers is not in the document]
thieman: [GPT 5.5] Fixed. Checks IPC now clones the existing default-host `MetaString` for empty/unset metric host and consumes the proto hostname `String` into `MetaString` for explicit host, avoiding the `&str -> MetaString` rebuild on the default path.

lib/saluki-components/src/sources/otlp/metrics/translator.rs
```diff
@@ -419,9 +424,9 @@ impl OtlpMetricsTranslator {
             identifier,
         }) = &source
```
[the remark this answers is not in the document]
thieman: [GPT 5.5] Fixed with a slightly different shape: OTLP metric dimensions now carry `Option<MetaString>` instead of `Option<String>`. The resource/default host is converted once per resource metrics batch and then cheaply cloned through derived dimensions, avoiding repeated `String` allocation/cloning while still owning the value safely across translation helpers.

lib/saluki-components/src/transforms/host_enrichment/mod.rs
```diff
@@ -99,13 +91,51 @@ impl HostEnrichment {
...
-            } else if let Some(eventd) = event.try_as_eventd_mut() {
+            if let Some(eventd) = event.try_as_eventd_mut() {
                 self.enrich_eventd(eventd);
             } else if let Some(service_check) = event.try_as_service_check_mut() {
                 self.enrich_service_check(service_check);
```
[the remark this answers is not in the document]
thieman: [GPT 5.5] Fixed. Event and service-check host enrichment now clones the stored `MetaString` directly instead of converting from `&str` on each enrichment.

lib/saluki-context/src/resolver.rs
```diff
@@ -488,19 +517,20 @@ impl ContextResolver {
...
     where
         N: AsRef<str> + CheapMetaString,
-        H: AsRef<str> + CheapMetaString,
+        H: AsRef<str> + CheapMetaString + Clone,
         I: IntoIterator<Item = T> + Clone,
         T: AsRef<str> + CheapMetaString,
     {
-        let (context_key, tagset_key) = self.create_context_key_with_host(&name, &host, tags.clone(), &origin_tags);
+        let (context_key, tagset_key) =
+            self.create_context_key_with_host(&name, host.clone(), tags.clone(), &origin_tags);
```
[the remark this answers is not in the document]
thieman: [GPT 5.5] Fixed. `create_context_key_with_host` now takes `Option<&H>` and `resolve_inner_with_host` passes `host.as_ref()` for key generation, so hashing no longer clones owned host values before cache lookup. I also tightened `static_base64` validation to decode during config validation for clearer fixture errors.

lib/saluki-context/src/context.rs
```diff
@@ -178,9 +178,29 @@ impl Context {
...
+        self.inner.host.as_deref()
+    }
```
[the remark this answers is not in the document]
thieman: [GPT 5.5] Fixed. I added `Context::host_meta()` so remap paths can preserve the underlying `MetaString`, and the DogStatsD mapper now passes that into the optional-host resolver path instead of reducing host to `&str`. This keeps cheap-clone behavior for already-interned context hosts.

lib/saluki-components/src/sources/checks_ipc/mod.rs
```diff
@@ -97,7 +108,10 @@ impl Source for ChecksIPC {
 
         let (events_tx, mut events_rx) = mpsc::channel(16);
 
-        let grpc_server = Server::builder().add_service(ChecksServer::new(ChecksService { events_tx }));
+        let grpc_server = Server::builder().add_service(ChecksServer::new(ChecksService {
+            events_tx,
+            default_hostname: self.default_hostname,
+        }));
```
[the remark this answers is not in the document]
thieman: [GPT 5.5] Cleaned this up. The code compiled before because it only used separate fields after the move, but I agree the partial move was confusing. `ChecksIPC::run` now destructures `*self` into local `grpc_endpoint` and `default_hostname` before constructing the gRPC service.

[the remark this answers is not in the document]
thieman: [GPT 5.5] Fixed in the latest branch state. `ChecksIPC::run` now destructures `*self` into local `grpc_endpoint` and `default_hostname` values before building the service/server, so there is no partial move through `self`.

lib/saluki-context/src/context.rs
```diff
@@ -178,9 +178,34 @@ impl Context {
...
+
+    /// Returns the underlying host metadata string of this context, if one has been set.
+    pub fn host_meta(&self) -> Option<&MetaString> {
+        self.inner.host.as_ref()
+    }
```
tobz: Can we actually remove this for now?

I understand the optimization being made, but it just feels a little too implementation-detail-y to expose and I'd rather hem towards the simpler approach (using `context.host()` directly, even if it's slightly suboptimal in some cases) for now.

We should put a TODO comment in DogStatsD Mapper, though, about revisiting this in the future.

tobz (changes requested):

tobz (approved):
Approving so it's ready to be merged once fixed and rebased on the PR to fix the nightly Rust version issue.

lib/saluki-context/src/context.rs
thieman: [GPT 5.5] Removed `Context::host_meta()` and switched DogStatsD mapper back to `context.host()`. Added a mapper TODO for revisiting the underlying-host-representation optimization only if host-bearing remaps show measurable allocation overhead. Validated with mapper host preservation test, `cargo check --workspace --tests`, `make fmt`, and the pre-commit checks.

webern (approved):
I think the description could better explain the observable behavior. Presumably they could become out-of-sync if the hostname changed and in some scenarios the stale hostname was being attached?

thieman:
@webern a bit more context in the prior PR on the `host` problem actually being solved. This one is more of a refactoring. Added correctness tests also show the problem scenario https://github.com/DataDog/saluki/pull/1962

PR #1979, opened 2026-07-04
chore(config): build the first components from the typed configuration model

webern:
## Human Summary 

This PR serves as a foundational model for cutting over components to typed config. It serves two purposes:
- Show human reviewers what it looks like to cut over two configurations to the typed model in a controlled diff.
- Serves as a reference for coding agents performing the rest of the cutovers.

An issue with dotted path environment variable support also had to be dealt with in this PR rather than laboriously shuttling it downward in the stack.

Key aliases that seemed to have arisen by mistake, such that we have a "saluki only" key for something that is Datadog schema-provided are treated as bugs as we go and the "saluki-only" aliases will be removed. I believe there are only a couple cases of this, but one of them happens to be in this PR.

## AI Summary

Establishes the reference pattern for constructing pipeline components from the typed
`SalukiConfiguration` model instead of the raw `GenericConfiguration` map, and cuts two DogStatsD
components over to it.

- **Aggregate transform (static):** `AggregateConfiguration` builds from the typed DogStatsD
  aggregation slice and shared histogram-encoding settings. Source serde behavior is removed from
  the component. Histogram parsing moves to a fallible conversion, and construction rejects a zero
  aggregation window.
- **Debug-log destination (dynamic):** `DogStatsDDebugLogConfiguration` builds from the typed
  debug-log slice and holds a `Live<DebugLog>`. The destination reacts to runtime changes through
  the typed live view instead of watching a raw string key.

Topology assembly now supplies each component its typed configuration subtree and, where needed, a
live view from `ConfigurationSystem`.

The aggregate cutover also fixes two duplicate configuration paths:

- `dogstatsd_flush_incomplete_buckets` now drives the aggregation field the transform reads; the
  duplicate Saluki-only path is removed.
- `dogstatsd_expiry_seconds` is the sole source for counter expiry; the duplicate Saluki-only key is
  removed.

Translation seeds Saluki-only values first and applies the Datadog source afterward, making the
Datadog source authoritative for every field it owns. Obsolete component-level configuration smoke
tests are removed; equivalent coverage lives in translation and component construction tests.

## Change Type

- [x] Non-functional (chore, refactoring, docs)

## How did you test this PR?

- `cargo test` for `saluki-components`, `agent-data-plane-config`, and
  `agent-data-plane-config-system`
- `cargo check -p agent-data-plane`
- Clippy on the touched crates
- Regenerated the schema overlay, configuration registry, and documentation

Integration and correctness suites were not run locally.

## References
- Progresses #1788
- Follows #1986

PR #1989, opened 2026-07-05
chore(config): migrate OTLP components to typed config

webern:
## Human Summary

There's some weird stuff going on here with preserving "absence" in an attempt to preserve exact behavioral parity, but I think it's OK.

## AI Summary

Cuts the OTLP components over from the raw `GenericConfiguration` map to the typed
`SalukiConfiguration` model, following the PR #1979 pattern: `OtlpConfig` (shared receiver
config), the OTLP source, decoder, forwarder, and relay.

Each component now builds from typed model slices (`domains.otlp`, `domains.traces.otlp`)
instead of deserializing the raw config:

- **Source / relay / decoder / forwarder**: `from_configuration` takes borrowed model slices;
  the call site in `run.rs` reads them from the config system. Component structs carry no serde
  and no defaults.
- **Seeded OTLP knobs** (context sizing, HTTP receiver transport, trace interner size,
  top-level-by-span-kind) are converted to the required-with-default doctrine: non-optional in
  `SalukiOnly` with the default defined once as a shared `const`/`fn` in `agent-data-plane-config`
  and referenced by both the model `Default` and the `SalukiOnly` serde default. The two byte-size
  interner keys are typed `ByteSize` so they accept both a bare integer and a suffixed string.
- **`OtlpTracesTranslator`** now takes the two native booleans it actually reads instead of the
  whole raw `TracesConfig`. `TracesConfig` is kept only for the not-yet-migrated Datadog trace
  encoder.
- The per-component config smoke tests are retired (coverage now lives in the translator and
  construction tests).

### Behavior notes

Because the OTLP receiver keys are witnessed against the vendored Datadog schema, their defaults
now come from that schema rather than Saluki's former hardcoded values. When a key is left unset:

- gRPC/HTTP receiver endpoints default to `localhost:4317` / `localhost:4318` (previously
  `0.0.0.0:...`), matching the Datadog Agent.
- `otlp_config.logs.enabled` defaults to `false` (previously `true`), matching the Datadog Agent.
- `otlp_config.receiver.protocols.grpc.max_recv_msg_size_mib` of `0` now maps to grpc-go's built-in
  4 MiB limit, matching the Agent and the schema documentation (previously `0` produced a 0-byte
  limit).

Deployments that set these keys explicitly (including the OTLP correctness cases, which pin the
gRPC endpoint to `0.0.0.0:4317`) are unaffected.

## Change Type

## How did you test this PR?

- `make build-schema-overlay && make fmt` (no generated drift), `make check-all`, `make check-docs`.
- Unit tests for `agent-data-plane-config`, `agent-data-plane-config-system`, `saluki-components`,
  and `agent-data-plane`, including new `SalukiOnly` transport/default tests (both byte-size input
  forms) and per-component construction tests.
- Statically reviewed the OTLP correctness cases: all send over gRPC on the explicitly configured
  `0.0.0.0:4317`, so the receiver default changes above do not affect them.

## References

- Progresses #1788
- Merges into `m/pr5-cutover`

### DIFF
--- a/lib/saluki-components/src/sources/otlp/mod.rs
+++ b/lib/saluki-components/src/sources/otlp/mod.rs
@@ -3,9 +3,10 @@ use std::sync::Arc;
 use std::sync::LazyLock;
 use std::time::Duration;
 
+use agent_data_plane_config::domains::otlp::Domain as OtlpDomain;
+use agent_data_plane_config::domains::traces::OtlpTraces;
 use async_trait::async_trait;
 use axum::body::Bytes;
-use bytesize::ByteSize;
 use otlp_protos::opentelemetry::proto::collector::logs::v1::ExportLogsServiceRequest;
 use otlp_protos::opentelemetry::proto::collector::metrics::v1::ExportMetricsServiceRequest;
 use otlp_protos::opentelemetry::proto::collector::trace::v1::ExportTraceServiceRequest;
@@ -16,7 +17,6 @@ use prost::Message;
 use resource_accounting::{MemoryBounds, MemoryBoundsBuilder};
 use saluki_common::sync::shutdown::{ShutdownCoordinator, ShutdownHandle};
 use saluki_common::task::HandleExt as _;
-use saluki_config::GenericConfiguration;
 use saluki_context::ContextResolver;
 use saluki_core::topology::interconnect::BufferedDispatcher;
 use saluki_core::{
@@ -31,15 +31,13 @@ use saluki_env::WorkloadProvider;
 use saluki_error::ErrorContext as _;
 use saluki_error::{generic_error, GenericError};
 use saluki_io::net::ListenAddress;
-use serde::Deserialize;
 use tokio::pin;
 use tokio::select;
 use tokio::sync::mpsc;
 use tokio::time::{interval, MissedTickBehavior};
 use tracing::{debug, error};
 
-use crate::common::otlp::config::OtlpConfig;
-use crate::common::otlp::{build_metrics, Metrics, OtlpHandler, OtlpServerBuilder};
+use crate::common::otlp::{build_metrics, grpc_max_recv_msg_size_bytes, Metrics, OtlpHandler, OtlpServerBuilder};
 
 mod logs;
 mod metrics;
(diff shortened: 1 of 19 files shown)


lib/saluki-components/src/sources/otlp/mod.rs
```diff
@@ -50,87 +48,52 @@ use self::resolver::build_context_resolver;
...
-    /// Creates a new `OTLPConfiguration` from the given configuration.
-    pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
-        let mut cfg: Self = config.as_typed()?;
-        cfg.otlp_config.traces.apply_env_overrides(config)?;
-        Ok(cfg)
+    /// Creates a new `OtlpConfiguration` from the resolved OTLP configuration.
+    pub fn from_configuration(otlp: &OtlpDomain, traces_otlp: &OtlpTraces) -> Result<Self, GenericError> {
+        Ok(Self {
+            metrics_enabled: otlp.receiver.metrics_enabled,
+            logs_enabled: otlp.receiver.logs_enabled,
```
webern, line 82: P1: This changes the default for OTLP logs from enabled to disabled. Before this cutover, `common::otlp::config::LogsConfig` used `default_logs_enabled() -> true`; the resolved Datadog model's `OtlpConfigLogs.enabled` defaults to `false`, and this constructor now consumes that value. With no explicit `otlp_config.logs.enabled`, the native source will therefore stop accepting/dispatching OTLP logs. Please preserve the old effective default in the typed configuration/translation and add an empty-config regression assertion.

lib/saluki-components/src/sources/otlp/mod.rs
```diff
@@ -50,87 +48,52 @@ use self::resolver::build_context_resolver;
...
+            traces_enabled: traces_otlp.enabled,
+            grpc_endpoint: otlp.receiver.grpc.endpoint.clone(),
```
webern, line 84: P1: The default listener addresses regress from `0.0.0.0:4317` (gRPC) and `0.0.0.0:4318` (HTTP) in the old `common::otlp::config` structs to the generated Datadog schema defaults of `localhost:4317` and `localhost:4318`. Since this constructor now consumes the resolved model, an otherwise empty config binds only loopback and remote OTLP clients can no longer connect. Please preserve the previous effective defaults in the typed config/translation (for both endpoints) and add an empty-config regression assertion.

webern (commented):
Claude Code operating on behalf of webern. I found two default-value regressions in the typed cutover that should be fixed before merging. Both are reachable with an otherwise empty OTLP configuration and change the externally observable native OTLP pipeline behavior.

lib/saluki-components/src/sources/otlp/mod.rs
webern, line 82: P1: This changes the default for OTLP logs from enabled to disabled. Before this cutover, `common::otlp::config::LogsConfig` used `default_logs_enabled() -> true`; the resolved Datadog model's `OtlpConfigLogs.enabled` defaults to `false`, and this constructor now consumes that value. With no explicit `otlp_config.logs.enabled`, the native source will therefore stop accepting/dispatching OTLP logs. Please preserve the old effective default in the typed configuration/translation and add an empty-config regression assertion.

webern (commented):
Claude Code operating on behalf of webern. I found two default-value regressions in the typed cutover that should be fixed before merging. Both are reachable with an otherwise empty OTLP configuration and change the externally observable native OTLP pipeline behavior.

lib/saluki-components/src/sources/otlp/mod.rs
webern, line 84: P1: The default listener addresses regress from `0.0.0.0:4317` (gRPC) and `0.0.0.0:4318` (HTTP) in the old `common::otlp::config` structs to the generated Datadog schema defaults of `localhost:4317` and `localhost:4318`. Since this constructor now consumes the resolved model, an otherwise empty config binds only loopback and remote OTLP clients can no longer connect. Please preserve the previous effective defaults in the typed config/translation (for both endpoints) and add an empty-config regression assertion.

PR #1992, opened 2026-07-05
chore(config): migrate DogStatsD source and mapper to typed config

webern:
## AI Summary

Build the DogStatsD source and metric mapper from the typed `SalukiConfiguration` model
(`domains.dogstatsd`) instead of the raw `GenericConfiguration` map.

- **Source** (`DogStatsDConfiguration`): `from_configuration` now takes `&domains::dogstatsd::Domain`
  plus a caller-derived default capture directory. `run_path` is deliberately unmodeled, so the call
  site reads it from the raw config and passes `run_path/dsd_capture` in; the component no longer
  touches `GenericConfiguration`. All source serde (`Deserialize`, renames, `default_*` fns) is
  stripped; injected runtime state (workload provider, capture/replay control) stays. Origin
  enrichment builds from the model via `OriginEnrichmentConfiguration::from_model`.
- **Mapper** (`DogStatsDMapperConfiguration`): `from_configuration` takes `&domains::dogstatsd::Mapper`
  and builds the `MetricMapper` from typed `MapperProfile`s. The interner byte budget parses as
  `ByteSize` in `SalukiOnly` (so `64KiB` and a bare integer both load).
- **Defaults**: the Saluki-only DogStatsD listener/context/mapper defaults (`buffer_count`,
  `buffer_count_max`, `permissive_decoding`, `cached_contexts_limit`, `cached_tagsets_limit`,
  `allow_context_heap_allocs`, `minimum_sample_rate`, mapper `string_interner_size`) move into
  hand-written model `Default` impls, so an absent key yields the value the components expect.
  Witnessed keys keep their schema-driven defaults.
- Retire the two component config smoke tests (the structs no longer deserialize); construction and
  parsing coverage now lives in the translator/`SalukiOnly` tests and the components' own tests.

## Change Type

## How did you test this PR?

- `make build-schema-overlay && make fmt` (no file changes)
- `make check-all`
- `make test`
- `make check-docs`

### Last of Several AI PR Reviews

Claude Code, operating on behalf of @webern.

Verified at `eec29828`: all three alternate-input-form regressions are resolved, and I think this is ready.

- `dogstatsd_string_interner_size_bytes` — the `SalukiOnly` field is now `Option<ByteSize>` with `.as_u64()` in `seed()`, mirroring the mapper interner, so the suffix-string form (`"2MiB"`) loads again alongside the bare integer.
- `dogstatsd_eol_required` — `normalize_datadog_input_forms` splits the space-separated string in the input seam (after the env overlay, before `DatadogConfiguration::deserialize`), matching `deserialize_space_separated_or_seq`.
- `dogstatsd_mapper_profiles` — the same helper parses the JSON-string form into the array the translator expects, leaving malformed JSON in place so the normal error still surfaces.

Each form has both alternate- and native-shape coverage through `ConfigurationSystem::load` / `seed`, replacing the deleted component-level tests at the right layer. Only `saluki_only.rs` and `system.rs` changed — no generated files or model types touched. The defaults and field mappings I checked earlier still hold.

The deterministic static gates are green (`check-schema-overlay`, `check-fmt`, `check-deny`, `check-docs`, `check-unused-deps`, mergegate); unit tests are still running. No further blocking items from me.

## References

- Progresses #1788
- Merges into `m/pr5-cutover`

### DIFF
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -738,12 +738,27 @@ async fn add_dsd_pipeline_to_blueprint(
 
     let saluki = config_system.config();
     // `run_path` is deliberately not modeled, so read it from the raw config to derive the default
-    // DogStatsD capture directory (used only when `dogstatsd_capture_path` is unset).
-    let default_dsd_capture_path = config
-        .try_get_typed::<PathBuf>("run_path")
-        .ok()
-        .flatten()
-        .map(|run_path| run_path.join(DOGSTATSD_CAPTURE_DIR));
+    // DogStatsD capture directory only when `dogstatsd_capture_path` is unset.
+    let default_dsd_capture_path = if saluki.domains.dogstatsd.listeners.capture_path.as_os_str().is_empty() {
+        match config.try_get_typed::<PathBuf>("run_path") {
+            Ok(Some(run_path)) => Some(run_path.join(DOGSTATSD_CAPTURE_DIR)),
+            Ok(None) => {
+                debug!(
+                    "`dogstatsd_capture_path` and `run_path` were empty. Default DogStatsD capture path is unavailable."
+                );
+                None
+            }
+            Err(error) => {
+                debug!(
+                    error = %error,
+                    "Failed to read `run_path` from configuration. Default DogStatsD capture path is unavailable."
+                );
+                None
+            }
+        }
+    } else {
+        None
+    };
     let dsd_config = DogStatsDConfiguration::from_configuration(&saluki.domains.dogstatsd, default_dsd_capture_path)
         .error_context("Failed to configure DogStatsD source.")?
         .with_workload_provider(env_provider.workload().clone())
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -28,7 +28,7 @@ use saluki_components::{
     },
     forwarders::{ClusterAgentForwarderConfiguration, DatadogForwarderConfiguration, OtlpForwarderConfiguration},
     relays::otlp::OtlpRelayConfiguration,
-    sources::{ChecksIPCConfiguration, DogStatsDConfiguration, OtlpConfiguration},
+    sources::{ChecksIPCConfiguration, DogStatsDConfiguration, OtlpConfiguration, DOGSTATSD_CAPTURE_DIR},
     transforms::{
         AggregateConfiguration, ApmStatsTransformConfiguration, AutoscalingFailoverGatewayConfiguration,
         ChainedConfiguration, DogStatsDMapperConfiguration, HostEnrichmentConfiguration,
@@ -736,17 +736,24 @@ async fn add_dsd_pipeline_to_blueprint(
     //    │    (destination)    │    │                       (Datadog Platform)                        │
     //    └─────────────────────┘    └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘
 
-    let dsd_config = DogStatsDConfiguration::from_configuration(config)
+    let saluki = config_system.config();
+    // `run_path` is deliberately not modeled, so read it from the raw config to derive the default
+    // DogStatsD capture directory (used only when `dogstatsd_capture_path` is unset).
+    let default_dsd_capture_path = config
+        .try_get_typed::<PathBuf>("run_path")
+        .ok()
+        .flatten()
+        .map(|run_path| run_path.join(DOGSTATSD_CAPTURE_DIR));
+    let dsd_config = DogStatsDConfiguration::from_configuration(&saluki.domains.dogstatsd, default_dsd_capture_path)
         .error_context("Failed to configure DogStatsD source.")?
         .with_workload_provider(env_provider.workload().clone())
         .with_capture_entity_resolver(env_provider.workload().clone());
     let dsd_prefix_filter_configuration = DogStatsDPrefixFilterConfiguration::from_configuration(config)?;
-    let dsd_mapper_config = DogStatsDMapperConfiguration::from_configuration(config)?;
+    let dsd_mapper_config = DogStatsDMapperConfiguration::from_configuration(&saluki.domains.dogstatsd.mapper)?;
     let dsd_enrich_config =
         ChainedConfiguration::default().with_transform_builder("dogstatsd_mapper", dsd_mapper_config);
     let dsd_tag_filterlist_config = TagFilterlistConfiguration::from_configuration(config)
         .error_context("Failed to configure metric tag filterlist transform.")?;
-    let saluki = config_system.config();
     let dsd_agg_config = AggregateConfiguration::from_configuration(
         &saluki.domains.dogstatsd.aggregation,
         &saluki.shared.metrics_encoding.histogram,
--- a/lib/saluki-components/src/sources/dogstatsd/mod.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/mod.rs
@@ -16,6 +16,7 @@ use std::{
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
+use agent_data_plane_config::domains::dogstatsd::{Domain as DogStatsDConfig, EnablePayloads};
 use async_trait::async_trait;
 use bytes::{Buf, BufMut};
 use bytesize::ByteSize;
@@ -24,7 +25,6 @@ use saluki_common::{
     sync::shutdown::{ShutdownCoordinator, ShutdownHandle},
     task::spawn_traced_named,
 };
-use saluki_config::{deserialize_space_separated_or_seq, GenericConfiguration};
 use saluki_context::{
     origin::RawOrigin,
     tags::{RawTags, RawTagsFilter},
@@ -54,8 +54,6 @@ use saluki_io::{
         ConnectionAddress, ListenAddress, ProcessCredentials, ProcessIdentity, Stream,
     },
 };
-use serde::{Deserialize, Deserializer};
-use serde_with::{serde_as, NoneAsEmptyString};
 use snafu::{ResultExt as _, Snafu};
 use stringtheory::MetaString;
 use tokio::{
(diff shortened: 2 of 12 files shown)


lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -687,10 +481,63 @@ async fn resolve_bind_host(host: &str) -> Result<std::net::IpAddr, Error> {
...
+        let contexts = &config.contexts;
+
+        let capture_path = if listeners.capture_path.as_os_str().is_empty() {
+            default_capture_path.unwrap_or_default()
+        } else {
+            listeners.capture_path.clone()
+        };
+
+        let mut dogstatsd_config = Self {
+            buffer_size: listeners.buffer_size,
```
webern, line 502: **Blocking — preserve numeric input validation in the config layer.** Before this cutover, deserializing directly into `u16`/`usize`/`u64` rejected negative and out-of-range values. This path now consumes values that `DatadogTranslator` silently clamps (`to_port` or `value.max(0)`), so bad input becomes valid configuration with different behavior. Examples include a negative `dogstatsd_buffer_size` becoming `0`, a negative `dogstatsd_capture_depth` becoming the 1024 fallback, an oversized port becoming 65535, and a negative context expiry/interner entry count becoming zero. Please make those narrowing conversions record a `TranslateError` instead of clamping (or otherwise preserve the old rejection semantics), and cover the invalid boundaries through the configuration system.

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
```diff
@@ -348,18 +277,25 @@ impl MetricMapper {
...
 impl DogStatsDMapperConfiguration {
-    /// Creates a new `DogstatsDMapperConfiguration` from the given configuration.
-    pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
-        Ok(config.as_typed()?)
+    /// Creates a new `DogStatsDMapperConfiguration` from the resolved mapper configuration.
+    pub fn from_configuration(mapper: &MapperConfig) -> Result<Self, GenericError> {
+        Ok(Self {
+            context_string_interner_bytes: ByteSize::b(mapper.string_interner_size),
+            cache_size: mapper.cache_size,
+            profiles: mapper.profiles.clone(),
```
webern, line 285: **Blocking — the typed profile path accepts input that the old component rejected.** `parse_mapper_profile` declares `RawProfile.mappings` with `#[serde(default)]`, while the removed `MappingProfileConfig` required `mappings` to be present. Once this line switches the component to the model, `{"name":"p","prefix":"x"}` silently becomes an empty profile instead of failing startup. Please remove that default or validate the required field during translation, and add a config-system test so the old sanitization contract is retained.

webern (commented):
Claude Code, operating on behalf of @webern.

I found two configuration-compatibility regressions that should be fixed before this joins the cutover branch. The field mapping and defaults otherwise look consistent with the previous component-owned configuration.

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
```diff
@@ -31,156 +28,106 @@ const MATCH_TYPE_REGEX: &str = "regex";
...
-            cache,
-        })
+        built_profiles.push(profile);
     }
+    let profiles = built_profiles;
+
+    let context_string_interner_size = NonZeroUsize::new(context_string_interner_bytes.as_u64() as usize)
+        .ok_or_else(|| generic_error!("context_string_interner_size must be greater than 0"))
+        .unwrap();
+
```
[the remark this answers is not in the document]
webern: This is not introduced by the cutover: the left side of the diff contains the same `ok_or_else(...).unwrap()` sequence. It is a pre-existing issue moved by the refactor, so I would not change it in this PR.

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -723,17 +723,24 @@ async fn add_dsd_pipeline_to_blueprint(
...
 
-    let dsd_config = DogStatsDConfiguration::from_configuration(config)
+    let saluki = config_system.config();
+    // `run_path` is deliberately not modeled, so read it from the raw config to derive the default
+    // DogStatsD capture directory (used only when `dogstatsd_capture_path` is unset).
+    let default_dsd_capture_path = config
+        .try_get_typed::<PathBuf>("run_path")
+        .ok()
+        .flatten()
+        .map(|run_path| run_path.join(DOGSTATSD_CAPTURE_DIR));
```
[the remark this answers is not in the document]
webern: Good catch. This does remove the previous debug diagnostic. The caller should preserve that signal by matching the `try_get_typed` result rather than discarding the error.

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
```diff
@@ -31,156 +28,106 @@ const MATCH_TYPE_REGEX: &str = "regex";
...
+                    i
+                ));
+            }
+            if mapping.metric_match.is_empty() {
+                return Err(generic_error!(
+                    "profile: {}, mapping num {}: match is required",
+                    config_profile.name,
+                    i
+                ));
             }
```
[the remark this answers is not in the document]
webern: This diagnostic bug is also present on the left side of the diff: the old implementation enumerates profiles as `i` and uses that value as the mapping number. The refactor preserves existing behavior, so it is outside this cutover review.

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
```diff
@@ -348,18 +277,25 @@ impl MetricMapper {
...
-    /// Creates a new `DogstatsDMapperConfiguration` from the given configuration.
-    pub fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
-        Ok(config.as_typed()?)
+    /// Creates a new `DogStatsDMapperConfiguration` from the resolved mapper configuration.
+    pub fn from_configuration(mapper: &MapperConfig) -> Result<Self, GenericError> {
+        Ok(Self {
+            context_string_interner_bytes: ByteSize::b(mapper.string_interner_size),
+            cache_size: mapper.cache_size,
+            profiles: mapper.profiles.clone(),
+        })
```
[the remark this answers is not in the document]
webern: Confirmed. The old `PickFirst<(DisplayFromStr, _)>` was an explicit reachability contract, and the typed source currently accepts only the array shape. This needs to be hoisted into the typed-config input path and covered through `ConfigurationSystem::load`.

lib/saluki-components/src/transforms/dogstatsd_mapper/mod.rs
webern: Verified at `daaf9a29fa`: `RawProfile.mappings` is required again, with production-path tests for a missing field, an explicit empty array, and the other required profile/mapping fields. This finding is fixed.

lib/saluki-components/src/sources/dogstatsd/mod.rs
webern: Verified at `daaf9a29fa`: the eight DogStatsD consumers now use checked conversions and record translation errors instead of clamping. The added startup-path tests cover negative values, oversized ports, and valid boundaries. This finding is fixed.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -687,10 +481,63 @@ async fn resolve_bind_host(host: &str) -> Result<std::net::IpAddr, Error> {
...
+            tcp_port: listeners.tcp_port,
+            statsd_forward_host: listeners.forward_host.clone().map(MetaString::from),
+            statsd_forward_port: listeners.forward_port,
+            socket_path: listeners.socket.clone(),
+            socket_stream_path: listeners.stream_socket.clone(),
+            stream_log_too_big: listeners.stream_log_too_big,
+            pipe_name: listeners.pipe_name.clone(),
+            windows_pipe_security_descriptor: listeners.windows_pipe_security_descriptor.clone(),
+            disable_verbose_logs: config.debug_log.disable_verbose_logs,
+            eol_required: listeners.eol_required.clone(),
```
webern, line 516: **Blocking — preserve the space-separated `dogstatsd_eol_required` form.** The removed component deserializer used `deserialize_space_separated_or_seq`, and this PR deletes the explicit `eol_required_accepts_space_separated_string` test. The generated Datadog source is now a plain `Vec<String>`, so a value such as `"udp uds"` fails before reaching this assignment instead of producing two listener types. Please hoist the string-or-sequence coercion into the typed-config input path and replace the deleted coverage with `ConfigurationSystem::load` tests for both forms.

webern (commented):
Claude Code, operating on behalf of @webern.

The fix commit correctly addresses the two blocking findings from my first review and restores the `run_path` diagnostic. The checked conversions, mapper-profile validation, and production-path tests all look good.

One reachability class still needs attention before merge: the removed component serde accepted alternate scalar forms that the typed source does not. Copilot correctly identified the JSON-string form of `dogstatsd_mapper_profiles`; I found the same problem for the space-separated form of `dogstatsd_eol_required` below. Both should be hoisted into the typed-config input path and covered through `ConfigurationSystem::load`.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -687,10 +481,63 @@ async fn resolve_bind_host(host: &str) -> Result<std::net::IpAddr, Error> {
...
+            bind_host: listeners.bind_host.clone(),
+            non_local_traffic: listeners.non_local_traffic,
+            autoscale_udp_listeners: listeners.autoscale_udp_listeners,
+            allow_context_heap_allocations: contexts.allow_context_heap_allocs,
+            no_aggregation_pipeline_support: config.aggregation.no_aggregation_pipeline,
+            context_string_interner_entry_count: contexts.string_interner_size,
+            context_string_interner_size_bytes: contexts.string_interner_size_bytes.map(ByteSize::b),
```
webern, line 523: Same reachability regression as `eol_required` / `mapper_profiles`, one more key. This wraps `contexts.string_interner_size_bytes` (`Option<u64>`) back into `ByteSize`, but the value now arrives through `SalukiOnly.dogstatsd_string_interner_size_bytes: Option<u64>`, which only deserializes a bare integer.

On the base branch this field was `Option<ByteSize>` on the component, so `dogstatsd_string_interner_size_bytes: "2MiB"` (or `"512KiB"`) loaded fine; after the cutover a suffixed value fails at startup. It's the same suffix-string budget form you already preserved for the mapper interner in this PR, and the sibling budgets (`memory_limit`, `otlp_string_interner_size`) are modeled as `ByteSize` too.

Fix mirrors the mapper change: make `SalukiOnly.dogstatsd_string_interner_size_bytes` an `Option<ByteSize>` and `.as_u64()` it in `seed()`, plus a config-system test that loads a suffixed string.

webern (commented):
Claude Code, operating on behalf of @webern — a fresh pass over the current head (b47e85c).

The defaults and field mappings hold up. I re-checked the witnessed keys against the vendored schema (`dogstatsd_port` 8125, `dogstatsd_buffer_size` 8192, `dogstatsd_string_interner_size` 4096, `dogstatsd_context_expiry_seconds` 20, `dogstatsd_mapper_cache_size` 1000, the Windows pipe SDDL, `dogstatsd_no_aggregation_pipeline` / `dogstatsd_origin_optout_enabled` / `enable_payloads.*` true, `dogstatsd_tag_cardinality` low) and the hand-written model `Default` impls for the Saluki-only knobs (`buffer_count` 128, `buffer_count_max` 256, `permissive_decoding` true, cached contexts/tagsets 500k, `allow_context_heap_allocs` true, `minimum_sample_rate`, mapper interner 65536). The empty-string→None coercions the old source did with `NoneAsEmptyString` / `deserialize_empty_metastring_as_none` are preserved through `non_empty` in the translator, so the deleted `statsd_forward_empty_host_disabled` case is still covered.

The outstanding item is the same reachability class flagged earlier: the old component serde accepted alternate scalar shapes that the typed path rejects. Two are already on the thread and remain unaddressed at this head:

- `dogstatsd_eol_required` — generated as a plain `Vec<String>`, so `"udp uds"` no longer parses (the old `deserialize_space_separated_or_seq`).
- `dogstatsd_mapper_profiles` — generated as `Vec<serde_json::Value>`, so the JSON-string form no longer parses (the old `PickFirst<(DisplayFromStr, _)>`).

I found one more instance of that class that isn't on the list yet — see the inline comment on `dogstatsd_string_interner_size_bytes`.

webern (commented):
Claude Code, operating on behalf of @webern.

Verified at `eec29828`: all three alternate-input-form regressions are resolved, and I think this is ready.

- `dogstatsd_string_interner_size_bytes` — the `SalukiOnly` field is now `Option<ByteSize>` with `.as_u64()` in `seed()`, mirroring the mapper interner, so the suffix-string form (`"2MiB"`) loads again alongside the bare integer.
- `dogstatsd_eol_required` — `normalize_datadog_input_forms` splits the space-separated string in the input seam (after the env overlay, before `DatadogConfiguration::deserialize`), matching `deserialize_space_separated_or_seq`.
- `dogstatsd_mapper_profiles` — the same helper parses the JSON-string form into the array the translator expects, leaving malformed JSON in place so the normal error still surfaces.

Each form has both alternate- and native-shape coverage through `ConfigurationSystem::load` / `seed`, replacing the deleted component-level tests at the right layer. Only `saluki_only.rs` and `system.rs` changed — no generated files or model types touched. The defaults and field mappings I checked earlier still hold.

The deterministic static gates are green (`check-schema-overlay`, `check-fmt`, `check-deny`, `check-docs`, `check-unused-deps`, mergegate); unit tests are still running. No further blocking items from me.

Issue #1965, opened 2026-07-01
Investigate whether `dd_url`'s schema default makes `site` unreachable in ADP.


webern:
## Edit

PR #2279 has landed and solves the configuration-provenance part of this problem. It preserves whether each value from the Agent configuration stream was sourced from a default (`default` or `schema`) or supplied explicitly, carries that distinction through typed configuration, and exposes it to consumers. A default-sourced `dd_url` can therefore be treated as no endpoint override while an explicitly configured `dd_url`—even the schema-default URL—remains an override. The legacy `GenericConfiguration` compatibility view is still value-only.

The remaining work is to migrate the affected configuration consumers to typed configuration. Until these migrations are complete, the raw consumers still cannot use the provenance that #2279 provides.

- [x] #2310 
- [x] #2312
- [x] #2313
- [x] #2314
- [x] #2315
- [x] #2316

The topology constructors in `bin/agent-data-plane/src/cli/run.rs` must pass the shared typed endpoint configuration through the primary, MRF, and autoscaling-failover paths. The MRF and Cluster Agent endpoint overrides remain explicit programmatic overrides.

## Original Issue

While reviewing #1957, automated reviewers noted that `site` may never affect the resolved primary
endpoint because `dd_url` always carries a value.

**Problem**

The Remote Agent config stream seems to send every key with a `source`,
including default-sourced values, while ADP's ingest keeps only the value and drops `source`. If so,
`dd_url` is always populated with its default, and the existing fallback to `site` may never fire
for stream-fed config.

Saluki models `dd_url` as an optional override and falls back to `site` only when it is `None`
([endpoints.rs L482-L483](https://github.com/DataDog/saluki/blob/7c042f913f3ae3d5d1462ff751eb24355561d9ad/lib/saluki-components/src/common/datadog/endpoints.rs#L482-L483), [L529-L537](https://github.com/DataDog/saluki/blob/7c042f913f3ae3d5d1462ff751eb24355561d9ad/lib/saluki-components/src/common/datadog/endpoints.rs#L529-L537)):

```rust
#[serde(default, alias = "url")]
dd_url: Option<String>,
```

```rust
pub(crate) fn configured_primary_endpoint(&self) -> String {
    match self.dd_url.as_deref() {
        Some(url) => url.to_string(),
        None => {
            let base_domain = if self.site.is_empty() { DEFAULT_SITE } else { &self.site };
            format!("https://app.{base_domain}")
        }
    }
}
```

[...]

PR #1993, opened 2026-07-05
chore(config): migrate the Datadog forwarder to typed config

webern:
## Human Summary

Agents are thrashing hard on #1965 and they have a hard time understanding that we can't fix it right now. I tried to get everything else fixed.

## AI Summary

Cuts the Datadog forwarder subsystem over from the raw `GenericConfiguration` map to the typed
`SalukiConfiguration` model. This is a tightly-coupled cluster, so it moves together in one PR.

Static construction (no more source serde, no restated defaults):

- `ForwarderConfiguration`, `EndpointConfiguration`, `ProxyConfiguration`, and `RetryConfiguration`
  build field-by-field from `shared.endpoints` (and `shared.metrics_encoding` for the V3 settings).
- `DatadogForwarderConfiguration` builds from the shared model; the Cluster Agent forwarder and the
  `run.rs` call sites are rewired to pass typed slices.

Runtime refresh (now reads a `Live<SalukiConfiguration>` view instead of a live
`GenericConfiguration`):

- `ResolvedEndpoint` refreshes its API key from a typed source: the shared primary key, the
  Multi-Region Failover override key, or the dual-shipping endpoint keys by URL/index.
- `TransactionForwarder` and `ApiKeyValidator` thread the live view; the retry policy's 403 secrets
  gate reads it too. `ApiKeyValidator` re-validates when any API-key source changes.

Config model:

- `secret_backend_command`, `secret_refresh_on_api_key_failure_interval`, and `run_path` are
  witnessed keys (in the vendored schema), promoted from `excluded:` to `inventory:` and modeled as
  `shared.secrets` and `shared.run_path`, with `consume_*` bodies in the Datadog translator.

The `ForwarderConfiguration` and `ProxyConfiguration` config smoke tests are retired; equivalent
coverage now lives in the construction and translator tests.

## Change Type

## How did you test this PR?

- `make build-schema-overlay && make fmt` (no changes)
- `make check-all`
- `make check-docs`
- `cargo nextest run` for `saluki-components` (`common::datadog`, `forwarders`),
  `agent-data-plane-config`, `agent-data-plane-config-system`, and the `agent-data-plane`
  config/cli modules

## References

- Progresses #1788
- Merges into `m/pr5-cutover`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4492,7 +4492,6 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_with",
- "serde_yaml",
  "simdutf8",
  "smallvec",
  "snafu",
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -4,7 +4,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use agent_data_plane_config::shared::{Compression, Endpoints, MetricsEncoding};
+use agent_data_plane_config::shared::Compression;
 use agent_data_plane_config_system::{ConfigurationSystem, EnvOverlayMode};
 use argh::FromArgs;
 use datadog_agent_commons::platform::PlatformSettings;
@@ -404,8 +404,10 @@ async fn create_topology(
         || dp_config.service_checks_pipeline_required()
         || dp_config.traces_pipeline_required()
     {
-        let dd_forwarder_config = DatadogForwarderConfiguration::from_configuration(config)
-            .error_context("Failed to configure Datadog forwarder.")?;
+        let saluki = config_system.config();
+        let dd_forwarder_config =
+            DatadogForwarderConfiguration::from_model(&saluki.shared, Some(config_system.live(|c| c)))
+                .error_context("Failed to configure Datadog forwarder.")?;
         blueprint.add_forwarder("dd_out", dd_forwarder_config)?;
     }
 
@@ -493,22 +495,15 @@ async fn add_baseline_metrics_pipeline_to_blueprint(
         // Metrics, then forwarding.
         .connect_components_in_order(["metrics_enrich", "dd_metrics_encode", "dd_out"])?;
 
-    add_mrf_metrics_pipeline_to_blueprint(blueprint, config, config_system)?;
+    add_mrf_metrics_pipeline_to_blueprint(blueprint, config_system)?;
     let saluki = config_system.config();
-    add_autoscaling_failover_metrics_pipeline_to_blueprint(
-        blueprint,
-        config,
-        &saluki.shared.autoscaling_failover,
-        &saluki.shared.cluster_agent,
-        &saluki.shared.metrics_encoding,
-        &saluki.shared.endpoints,
-    )?;
+    add_autoscaling_failover_metrics_pipeline_to_blueprint(blueprint, &saluki.shared)?;
 
     Ok(())
 }
 
 fn add_mrf_metrics_pipeline_to_blueprint(
-    blueprint: &mut TopologyBlueprint, config: &GenericConfiguration, config_system: &ConfigurationSystem,
+    blueprint: &mut TopologyBlueprint, config_system: &ConfigurationSystem,
 ) -> Result<(), GenericError> {
     let saluki = config_system.config();
     let mrf_config = MrfConfiguration::from_configuration(&saluki.domains.multi_region_failover)
@@ -535,15 +530,10 @@ fn add_mrf_metrics_pipeline_to_blueprint(
         DatadogMetricsConfiguration::from_configuration(&saluki.shared.metrics_encoding, &saluki.shared.endpoints)
             .error_context("Failed to configure Multi-Region Failover Datadog Metrics encoder.")?;
 
-    let mrf_forwarder_config = DatadogForwarderConfiguration::from_configuration(config)
-        .map(|config| {
-            config.with_endpoint_override_and_api_key_refresh_config_path(
-                mrf_dd_url,
-                mrf_api_key,
-                "multi_region_failover.api_key",
-            )
-        })
-        .error_context("Failed to configure Multi-Region Failover Datadog forwarder.")?;
+    let mrf_forwarder_config =
+        DatadogForwarderConfiguration::from_model(&saluki.shared, Some(config_system.live(|c| c)))
+            .map(|config| config.with_multi_region_failover_endpoint_override(mrf_dd_url, mrf_api_key))
+            .error_context("Failed to configure Multi-Region Failover Datadog forwarder.")?;
 
     blueprint
         .add_transform("mrf_metrics_gateway", mrf_gateway_config)?
@@ -560,14 +550,11 @@ fn add_mrf_metrics_pipeline_to_blueprint(
 }
 
 fn add_autoscaling_failover_metrics_pipeline_to_blueprint(
-    blueprint: &mut TopologyBlueprint, config: &GenericConfiguration,
-    autoscaling_failover: &agent_data_plane_config::shared::AutoscalingFailover,
-    cluster_agent: &agent_data_plane_config::shared::ClusterAgent, metrics_encoding: &MetricsEncoding,
-    endpoints: &Endpoints,
+    blueprint: &mut TopologyBlueprint, shared: &agent_data_plane_config::shared::SharedConfiguration,
 ) -> Result<(), GenericError> {
-    let af_config = AutoscalingFailoverConfiguration::from_configuration(autoscaling_failover)
+    let af_config = AutoscalingFailoverConfiguration::from_configuration(&shared.autoscaling_failover)
         .error_context("Failed to configure autoscaling failover metrics pipeline.")?;
-    let ca_config = ClusterAgentConfiguration::from_configuration(cluster_agent)
+    let ca_config = ClusterAgentConfiguration::from_configuration(&shared.cluster_agent)
         .error_context("Failed to configure Cluster Agent metrics forwarding.")?;
 
     let Some((ca_url, ca_token)) = ca_config.endpoint_and_token() else {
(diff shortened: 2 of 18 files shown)


webern (commented):
Claude Code, reviewing on behalf of webern.

Overall this is a clean, faithful cutover: field-by-field construction, no restated defaults, tests rewritten to build from the typed model, and the runtime API-key refresh reworked into an explicit `ApiKeySource` that preserves the primary / MRF / additional-endpoint behaviors. I checked the promoted keys against the vendored schema (`run_path`, `secret_backend_command`, `secret_refresh_on_api_key_failure_interval` are all witnessed correctly), and the forwarder numeric defaults flowing through the generated `DatadogConfiguration` match the old serde defaults (concurrency 10, num_workers 1, timeout 20, buffer 100, backoff 2/2/64, recovery 2, flush ratio 0.5, disk ratio 0.8, outdated 10 days, capacity 900s).

There is one blocker and one narrower issue. Both were also flagged independently by Codex/Copilot and I agree with them.

## Blocker: `dd_url` schema default overrides `site` (data routing regression)

`consume_dd_url` sets `shared.endpoints.dd_url = non_empty(value)`, and the generated `DatadogConfiguration.dd_url` defaults to the schema value `https://app.datadoghq.com`. So when a user does **not** set `dd_url`, the typed model still carries `dd_url = Some("https://app.datadoghq.com")` rather than `None`.

`EndpointConfiguration::from_model` copies that through, and `calculate_resolved_endpoint` treats any `Some(dd_url)` as an explicit override that takes precedence over `site`:

```rust
let raw_endpoint = match override_url {
    Some(url) => url.to_string(),                 // dd_url wins, site ignored
    None => format!("https://app.{}", base_domain) // site path
};
```

Net effect: a deployment that sets only `site` (e.g. `datadoghq.eu`, `us5.datadoghq.com`, `ap1.datadoghq.com`) and leaves `dd_url` unset — the normal way non-US customers configure the intake — now has its primary endpoint silently routed to `app.datadoghq.com`. Before this PR the forwarder read the raw `dd_url` key, which stayed `None` when unset, so `site` was honored.

This is the root cause already documented in the `TODO(#1965)` comment on `Endpoints`: `consume_dd_url` can't distinguish "user set it to the default URL" from "schema default filled it in," so it can't safely populate `dd_url`. The forwarder was the last reader relying on the raw-config escape hatch, and this PR removes it.

Worth noting the test suite doesn't catch this because the forwarder tests build from `SharedConfiguration::default()` (where `dd_url` is `None`), not from a translated config where the schema default is applied. A test that sets only `site` and asserts the resolved primary endpoint would have surfaced it, and would be good to add alongside the fix.

## Narrower: `run_path` placeholder can seed a literal retry-queue path

`run_path`'s schema default is the literal template `${run_path}`, and `consume_run_path` stores it verbatim (saluki does no `${...}` expansion). `RetryConfiguration::from_model` then sees a non-empty run path and, when `forwarder_storage_path` is empty, derives `${run_path}/transactions_to_retry`. Previously an unset `run_path` read as `None` and left the storage path empty.

This is gated behind `forwarder_storage_max_size_in_bytes > 0`, so it's inert in the default (disk-persistence-off) configuration. But a deployment that enables on-disk retry without an explicit `forwarder_storage_path` and without `run_path` would initialize the queue under a literal `${run_path}` directory instead of leaving it unset. Treating `${run_path}` (and empty/whitespace) as unset in `consume_run_path` would restore the prior behavior.

## Secondary (perf, non-blocking)

The retry 403 gate clones the entire `SalukiConfiguration` on every evaluated 403 (`secrets_in_use` → `live.current()`). On a sustained-403 path that's a large clone per response; projecting to `Live<Secrets>` once when building the predicate would keep it to the small struct. Copilot also flagged the additional-endpoints `api_key()` refresh cloning the full map per request — comparable to the old per-call deserialize, so less of a change, but a borrowed read path on `Live` would help both. These are follow-ups, not merge blockers.

CI is green (only the ADP benchmark job pending).

webern:
Follow-up to my review: after reading #1965, I no longer consider the `dd_url`/`site` issue a blocker for this PR. It predates typed config, and fixing it correctly requires retaining configuration-source metadata at ingest, which is outside this cutover. We will fix the other two actionable findings here: treating the unresolved `${run_path}` placeholder as unset, and projecting the 403 retry gate to `Live<Secrets>` so it does not clone the full configuration per response.

webern (commented):
Claude Code, reviewing on behalf of webern — fresh clean-room pass on the current HEAD (`07733e3`).

First, the two items from the earlier round that the follow-up commits addressed, so they're off the table: `consume_run_path` now treats the `${run_path}` placeholder (and empty/whitespace) as unset, and the 403 retry gate projects to `Live<Secrets>` instead of cloning the whole `SalukiConfiguration` per response. Both look good.

The rest of the cutover holds up well: field-by-field construction, no restated defaults, the forwarder numeric defaults flowing through the generated `DatadogConfiguration` still match the old serde defaults, and the `ApiKeySource` rework faithfully preserves the primary / MRF / additional-endpoint refresh behavior. The promoted keys (`run_path`, `secret_backend_command`, `secret_refresh_on_api_key_failure_interval`) are witnessed correctly against the vendored schema.

There is still one blocker, plus a second instance of the same root cause that hasn't been called out yet.

## Blocker (still open): `dd_url` schema default shadows `site`

This is the same routing regression flagged last round, and it is unchanged at HEAD.

`consume_dd_url` sets `shared.endpoints.dd_url = non_empty(value)`, and the generated `DatadogConfiguration.dd_url` defaults to the schema value `https://app.datadoghq.com`. `drive()` calls the consumer unconditionally, so when a user does *not* set `dd_url` the typed model still carries `dd_url = Some("https://app.datadoghq.com")` rather than `None`.

`EndpointConfiguration::from_model` copies that through, and `calculate_resolved_endpoint` treats any `Some(dd_url)` as an explicit override that beats `site`:

```rust
let raw_endpoint = match override_url {
    Some(url) => url.to_string(),                    // dd_url wins, site ignored
    None => format!("https://app.{}", base_domain),  // site path
};
```

Net effect: a deployment that sets only `site` (`datadoghq.eu`, `us5.datadoghq.com`, `ap1.datadoghq.com`, ...) and leaves `dd_url` unset — the normal way non-US1 customers point at their intake — now has its primary endpoint silently routed to `app.datadoghq.com`. Before this PR the forwarder read the raw `dd_url` key, which stayed `None` when unset, so `site` was honored. This is the last reader that relied on that raw escape hatch, so this cutover is what activates the regression.

`TODO(#1965)` on `Endpoints` already documents the underlying cause: the translator can't tell "user set `dd_url` to the default URL" from "schema default filled it in." One workable option here is that the `dd_url` default (`https://app.datadoghq.com`) and the default `site` resolution (`datadoghq.com` -> `https://app.datadoghq.com`) produce the same URL, so treating a `dd_url` equal to the schema default as unset in `consume_dd_url` would restore `site` precedence in the common case without hurting the coinciding-default case. Either way this needs to be resolved (or explicitly deferred with the routing consequence understood) before it merges.

The forwarder tests don't catch it because they build from `SharedConfiguration::default()` (where `dd_url` is `None`), not from a translated config where the schema default is applied. A test that sets only `site` and asserts the resolved primary endpoint would surface it.

## Same root cause, not yet flagged: the deprecated `forwarder_retry_queue_max_size` fallback is now dead

`queue_max_size_bytes()` still prefers `retry_queue_payloads_max_size` and falls back to the deprecated `retry_queue_max_size`:

```rust
self.retry_queue_payloads_max_size
    .or(self.retry_queue_max_size)
    .unwrap_or(FORWARDER_RETRY_QUEUE_PAYLOADS_MAX_SIZE_BYTES)
```

That `.or()` chain depends on `retry_queue_payloads_max_size` being `None` when the user didn't set it. But `consume_forwarder_retry_queue_payloads_max_size` sets `Some(...)` unconditionally, and `drive()` always feeds it the generated default (`15728640`). So post-translation the model always carries `retry_queue_payloads_max_size = Some(15 MiB)`, the `.or()` never reaches the deprecated key, and a config that sets only `forwarder_retry_queue_max_size` now silently gets 15 MiB instead of the configured value. Previously that field deserialized as `None` when unset (no serde default), so the fallback worked.

Impact is much narrower than `dd_url` — it only bites a deprecated-key-only config — but it's the same defect: an `Option` that used to encode "was it set" is now always `Some(default)` after `drive()`. The rewritten `queue_max_size_bytes_fallback_behavior` test hides it by building `Forwarder::default()` (`payloads = None`), a shape that never occurs after translation. Whatever the fix for `dd_url`, this one wants the same treatment or an explicit note that the deprecated fallback is being dropped.

## Test loss worth relocating

The four `deser_additional_endpoints_*` tests were deleted, but `AdditionalEndpoints`'s serde parsing (JSON-string maps via `PickFirst<DisplayFromStr>`, and the single-string-or-seq value shape) is still live — it's used by the not-yet-migrated `DatadogMetricsConfiguration`, as the retained `#[serde_as]` and the `TODO` on the type note. Deleting those tests leaves that live parsing path with no coverage. Since the code isn't going away yet, relocating the deser tests (rather than dropping them) would keep the coverage until the metrics encoder migrates.

## Minor / non-blocking

- Copilot's note that `RetryConfiguration::from_model` uses `storage_path.parent().is_none()` (also true for `/`) — worth tidying, but it matches the pre-cutover behavior, so it's not a regression from this PR.
- Copilot's note that `consume_run_path` stores the untrimmed `value` after checking `trimmed` — cosmetic; a padded path was also possible before. Storing the trimmed value would be slightly cleaner.
- Copilot's per-request clone of the whole `additional_endpoints` map in `refreshed_api_key` — reasonable follow-up, comparable to the old per-call deserialize, not a merge blocker.

webern:
Claude Code, on behalf of webern — a follow-up now that the fixes have landed.

Two of the three points from my earlier review are resolved:

- **Retry-queue deprecated fallback** — fixed. `forwarder_retry_queue_payloads_max_size` and
  `forwarder_retry_queue_max_size` are now absence-aware (`saluki_overrides_default`), so a config
  that sets only the deprecated key resolves to that value again instead of the 15 MiB default, with
  coverage that runs through `translate()` (the path that exposed the regression).

- **`AdditionalEndpoints` deserialization tests** — withdrawn. Once this branch rebased onto #1990,
  `additional_endpoints` is witnessed as a plain `HashMap<String, Vec<String>>` and nothing
  deserializes through the `AdditionalEndpoints` serde impl anymore, so restoring those tests would
  only guard dead code. That commit was reverted and the now-unused `serde_yaml` dependency dropped.
  The behavior that actually needed protecting — the JSON-string and bare-string input shapes those
  serde impls used to accept (notably `DD_ADDITIONAL_ENDPOINTS` as a JSON string) — is preserved on
  the witnessed path in a separate PR into the base branch, #1994.

On the third point — **`dd_url` overriding `site`** — I was wrong to raise it as a blocker on this
PR. It's a known limitation of the current model (the translator can't yet distinguish an
operator-set `dd_url` from the schema default), tracked by #1965 and noted in the `Endpoints` TODO.
It's out of scope here and shouldn't hold up this cutover.

PR #1987, opened 2026-07-05
chore(config): cutover saluki-components to the new typed configuration system

webern:
## Summary

This PR aggregates multiple PRs, each of which was used to cut over, test and review one or more components, mostly from `saluki-components`, to the new typed configuration system. AI was used extensively to scrutinize each PR before merging it into this PR.

## Change Type 
- [x] Non-functional (chore, refactoring, docs)

## How did you test this PR?

Each cutover to typed config was tested and reviewed (aided by AI) in a PR before being merged into this one. Thus CI passes in the upstream PR before adding the component cutover to this aggregation and being tested again by CI in aggregate.

## References

- Progresses #1788
- Aggregates 
    - #1980
    - #1981
    - #1982

PR #1996, opened 2026-07-05
chore(config): migrate HostTagsConfiguration to typed config

webern:
## Human Summary

It thrashed a bit on preserving the exact behavior that `DurationString` was providing and on viper `time.Duration` parsing correctness. Overall the changes it made seem fine to me.

## AI Summary

Migrates `HostTagsConfiguration` from the raw `GenericConfiguration` map to the typed
`SalukiConfiguration` model, continuing the ADP config cutover.

`expected_tags_duration` is witnessed and already modeled at
`shared.tags.expected_tags_duration`, so the component now reads it directly from the typed
slice instead of parsing a `DurationString` off `GenericConfiguration`. The component's
hand-written zero default is dropped; the generated schema default (`0s`) is authoritative.

`RemoteAgentClientConfiguration` is a shared struct still consumed by several not-yet-migrated
remote-agent components, so it stays on `GenericConfiguration` for now. It is built at the call
site and passed into `HostTagsConfiguration`, which removes `GenericConfiguration` from the
component entirely.

## Change Type

## How did you test this PR?

`make check-all`, `make test`, and `make check-docs` all pass. Translation of
`expected_tags_duration` into the typed model is already covered by existing translator and
config-system tests.

## References

- Progresses #1788
- Merges into #1987

### DIFF
--- a/bin/agent-data-plane/src/components/host_tags/mod.rs
+++ b/bin/agent-data-plane/src/components/host_tags/mod.rs
@@ -3,10 +3,10 @@ use std::{
     time::{Duration, Instant},
 };
 
+use agent_data_plane_config::shared::GlobalTags;
 use async_trait::async_trait;
 use datadog_agent_commons::ipc::{client::RemoteAgentClient, config::RemoteAgentClientConfiguration};
 use resource_accounting::{MemoryBounds, MemoryBoundsBuilder};
-use saluki_config::{DurationString, GenericConfiguration};
 use saluki_context::tags::{SharedTagSet, Tag};
 use saluki_core::{components::transforms::*, topology::EventsBuffer};
 use saluki_core::{components::ComponentContext, data_model::event::metric::Metric};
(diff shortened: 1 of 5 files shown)


webern (commented):
Claude Code operating on behalf of webern. Clean-room review completed. I found one input-sanitization/behavioral regression that should be fixed before merging; config reachability and the zero default otherwise look preserved. The existing Copilot review contained no actionable inline finding, and the performance/binary-size bot reports are not relevant to this config cutover.

bin/agent-data-plane/src/components/host_tags/mod.rs
```diff
@@ -22,21 +22,13 @@ pub struct HostTagsConfiguration {
...
-            .unwrap_or(DEFAULT_EXPECTED_TAGS_DURATION);
-
-        Ok(Self {
+    /// Creates a new `HostTagsConfiguration` from the remote-agent client configuration and the shared tag settings.
+    pub fn from_configuration(client_config: RemoteAgentClientConfiguration, tags: &GlobalTags) -> Self {
+        Self {
             client_config,
-            expected_tags_duration,
-        })
+            expected_tags_duration: tags.expected_tags_duration,
```
webern, line 30: [P1] Preserve rejection of negative numeric durations. Before this migration, `GenericConfiguration::try_get_typed::<DurationString>` rejected negative numeric values and `HostTagsConfiguration::from_configuration` propagated that error. The typed `DatadogConfiguration` path uses `deserialize_go_duration`, whose numeric visitor clamps negative values to zero. Consequently, `expected_tags_duration: -1` now boots with host-tag enrichment silently disabled instead of rejecting the invalid configuration. Please preserve the old validation in the typed layer (including its `DurationString` overflow bounds) or otherwise make this cutover behavior-equivalent.

lib/datadog-agent/config/src/duration_de.rs
```diff
@@ -10,14 +10,23 @@
...
+const MAX_NANOS_U64: u64 = i64::MAX as u64;
+
 /// Deserialize a duration expressed as integer nanoseconds or a Go duration string.
 ///
 /// A numeric value is nanoseconds (matching the wire encoding and the classifier's
-/// `duration_value_as_nanos`); a string is parsed with the shared `go-duration` parser. Negative
-/// numeric values clamp to zero, since `Duration` cannot represent them.
+/// `duration_value_as_nanos`); a string is parsed with the shared `go-duration` parser.
+///
+/// Acceptance and rejection match `saluki_config::DurationString`, the type this replaced on the
```
webern: [P1] Preserve the legacy string coercion too. `DurationString` does more than call `parse_duration`: it trims whitespace and, if Go-duration parsing fails, accepts a bare integer string as nanoseconds. Thus `expected_tags_duration: "5"` (and env-var input `DD_EXPECTED_TAGS_DURATION=5`) previously produced 5 ns, while this typed path rejects it; padded input such as `" 5s"` also previously worked. That leaves the cutover behaviorally incompatible despite this statement. Please use equivalent string parsing here (preferably shared rather than duplicated) and add coverage for bare-integer and whitespace-padded strings, including overflow/negative fallback cases.

webern (commented):
Claude Code operating on behalf of webern. The negative-number and numeric-overflow fix is correct, but the follow-up review found one remaining `DurationString` compatibility gap in the same typed boundary.

PR #1998, opened 2026-07-05
chore(config): migrate DogStatsD metric filters to typed config

webern:
## Human Summary

There's a lot going on in this one. I'm not going to merge it into the aggregating PR, so we can treat this one on it's own (and last) if everything else merges.

## AI Summary

Migrates the three DogStatsD metric-filter transforms from the raw
`GenericConfiguration` map to the typed `SalukiConfiguration` model:

- `DogStatsDPrefixFilterConfiguration`
- `DogStatsDPostAggregateFilterConfiguration`
- `TagFilterlistConfiguration`

All three are dynamic (runtime-reactive) transforms. Each now takes borrowed
model slices for its static settings and a `Live<T>` view for the fields it
reacts to, replacing the stored `GenericConfiguration` and the per-key
`watch_for_updates` calls:

- The prefix and post-aggregate filters track `domains.dogstatsd.prefix_filter`
  (filterlist/blocklist and their prefix-match flags). The prefix filter reads
  the namespace and its exemption list once; the post-aggregate filter reads the
  histogram aggregate/percentile suffixes once from
  `shared.metrics_encoding.histogram`.
- The tag filterlist tracks `domains.dogstatsd.tag_filterlist` and reads the
  dedup-cache capacity from `domains.dogstatsd.aggregation`.

The shared `dogstatsd_filterlist` helper drops the now-unused config-key
constants and per-field setters and gains `EffectiveFilterlist::from_prefix_filter`.
Component-side serde, hand-written defaults (histogram aggregates/percentiles,
the metric-namespace blocklist, the tag-filter cache capacity), and the
prefix-filter smoke test are removed; the generated schema witness now supplies
those defaults. The prefix filter's per-field runtime update counter is
preserved by comparing the effective (active) filter values across updates.

Reactive tests drive a `Live` view through an `ArcSwap` + `watch` channel,
mirroring the reference cutover, in place of `ConfigurationLoader`.

## Change Type

## How did you test this PR?

- `make check-all` (fmt, clippy, docs, deny, licenses, unused-deps, api-docs, features)
- `make test` (all unit tests pass)
- `make build-schema-overlay && make fmt` leave the tree unchanged (no schema drift)
- Rewrote each component's construction and reactive tests against the typed model

## References

- Progresses #1788
- Merges into #1987


bin/agent-data-plane/src/components/tag_filterlist/mod.rs
```diff
@@ -652,29 +644,6 @@ mod tests {
...
-            "metric_name": "my.dist",
-            "tags": ["env"]
-        }))
-        .unwrap();
-
-        assert_eq!(entry.action, FilterAction::Exclude);
-    }
-
-    #[test]
-    fn invalid_action_defaults_to_exclude() {
```
webern, line 667: This test covered behavior, not merely component-local deserialization. Before this cutover, an unknown action was warned about and coerced to `Exclude`, so the initial configuration still loaded; `null` was also accepted as `Exclude`. The typed translator now records a `TranslateError` for an unknown action (which makes `ConfigurationSystem::load` reject startup) and rejects `null`. Its `RawEntry` also adds `#[serde(default)]` to `tags`, although this field was previously required; an entry such as `{ metric_name: "my.dist", action: "include" }` can now load and strip every ordinary tag from that metric. Please preserve the old accepted-input/coercion contract in the translator and move the regression tests there rather than deleting them.

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -765,19 +765,32 @@ async fn add_dsd_pipeline_to_blueprint(
...
     let dsd_enrich_config =
         ChainedConfiguration::default().with_transform_builder("dogstatsd_mapper", dsd_mapper_config);
-    let dsd_tag_filterlist_config = TagFilterlistConfiguration::from_configuration(config)
-        .error_context("Failed to configure metric tag filterlist transform.")?;
+    let dsd_tag_filterlist_config = TagFilterlistConfiguration::from_configuration(
+        saluki
+            .domains
+            .dogstatsd
+            .aggregation
+            .aggregator_tag_filter_cache_capacity,
```
webern, line 781: Claude: Please preserve the old rejection of negative cache capacities. The previous `try_get_typed::<usize>` failed on a negative value. The model translator currently uses `value.max(0) as usize`, so `-1` is silently accepted as zero, after which `build_context_cache` turns it into a one-entry cache. That is both an input-sanitization and behavioral regression. Translation should reject/record an error for negative values (and use a checked conversion), with a regression test.

webern (commented):
Claude Code operating on behalf of webern. I found two input-compatibility regressions that should be fixed before merge. The ordinary defaults and reachability for the prefix/block lists, histogram suffixes, namespace alias, and tag-filter cache default otherwise look preserved, and the `Live<T>` update paths look sound. CI is green; Copilot's two comments only concern verbose debug logging and do not cover these compatibility issues.

webern:
Deferred

PR #2000, opened 2026-07-05
chore(config): migrate logging translation to typed config

webern:
## Human Summary

There's crazy stuff happening in this one. I am not going to merge it into the aggregating PR that is part of the merge stack. This will me left as a follow up into main if we end up merging the stack.

## AI Summary

Cut `LoggingConfigurationTranslator` and `DynamicLogLevelWorker` over from the raw
`GenericConfiguration` map to the typed `SalukiConfiguration` model.

All eleven logging keys the translator reads (`log_level`, `log_format_json`,
`log_format_rfc3339`, `log_to_console`, `log_to_syslog`, `syslog_rfc`, `syslog_uri`,
`log_file_max_size`, `log_file_max_rolls`, `disable_file_logging`, `data_plane.log_file`) are
witnessed against the vendored Datadog schema and were already driven into `control.logging`.
The translator now builds `LoggingConfiguration` from that typed model instead of re-parsing raw
config, keeping only the ADP-specific business rules: first-party log-filter expansion, syslog
gating with a platform-default URI fallback, and the per-subagent log-file path with a
platform-default fallback (the Core Agent's own `log_file` is still never consulted).

`DynamicLogLevelWorker` no longer stores a `GenericConfiguration`; it reacts to a
`Live<String>` view of `control.logging.level` and re-applies the parsed filter on change.

Logging is initialized before the configuration system exists, so this adds a lenient one-shot
`translate_snapshot` to the config-system crate (translation errors keep field defaults, mirroring
the runtime update path) and uses it at the two bootstrap-phase sites: the initial logging setup in
`main` and the config-stream reload in `run`. The control-plane wiring now takes
`&ConfigurationSystem` and mints both the log-level view and the `/config/internal` handle from it.

## Change Type

## How did you test this PR?

- Rewrote the logging tests to build the typed model directly and cover filter expansion, syslog
  gating, log-file fallback/disable, and byte-size conversion; added a driven-`Live` test that the
  worker reacts to a live log-level change.
- Added translator tests covering the schema-driven logging defaults, driven values (including the
  `log_file_max_size` byte-size parse), and the invalid-byte-size error path.
- `make build-schema-overlay && make fmt` (no drift), `make check-all`, `make test`, `make check-docs`.

## References

- Progresses #1788
- Merges into #1987


webern (commented):
**Claude Code, operating on behalf of @webern.**

Clean-room review of the logging cutover. Overall this is a faithful migration: all eleven logging keys are witnessed with real `consume_*` bodies and land in `control.logging`, the syslog gating and default-URI fallback are preserved, the log-level parsing path is unchanged (`parse_adp_log_level` still routes plain levels through the first-party filter), and the test rewrites keep the behavioral coverage rather than just the deserialization. The `Live<String>` conversion for the dynamic worker is a clean equivalent of the old key watch.

Two things are worth resolving before this goes into the big-bang merge, plus a couple of low-severity notes.

## 1. Platform-specific default log-file path is lost on non-Linux

`bin/agent-data-plane/src/internal/logging.rs` still carries the `PlatformSettings::get_default_log_file_path()` fallback for an empty `logging.file`, but that branch is now unreachable in the default case.

`data_plane.log_file` carries a hardcoded schema default of `/var/log/datadog/agent-data-plane.log` (`core_schema.yaml`), so `drive()` populates `logging.file` with that Linux path even when the user configures nothing. `logging.file` is therefore never empty unless the user *explicitly* sets `data_plane.log_file: ""`, and the platform fallback only fires in that one corner.

- On Linux this is a no-op (same path).
- On macOS (`/opt/datadog-agent/logs`) and Windows (`C:\ProgramData\Datadog\logs`, registry-derived), the pre-cutover code fell back to the platform path when the key was unset; it now writes the Linux path instead. That's a default regression.

Note the asymmetry with `syslog_uri`, whose schema default is empty (`''`), so *its* platform fallback still works correctly. The Core Agent's own `log_file` also defaults to `''`. The cleanest fix is probably to give `data_plane.log_file` an empty default too (via the overlay/vendored schema), so the config layer keeps its "empty means use the platform default" contract and the existing `translate` fallback governs — rather than baking a single-platform literal into the model.

(Both Copilot and Codex flagged this independently. Copilot's comment references an `adp-logging-windows-default-path` integration test that doesn't exist in the tree, so ignore that specific citation — the underlying issue is real regardless.)

## 2. Dropped permissive-boolean coercion

The old component ran the logging booleans (`log_format_json`, `log_to_console`, `log_to_syslog`, `syslog_rfc`, `disable_file_logging`, etc.) through `PermissiveBool`; the typed model deserializes plain `bool`. Plain `true`/`false` still work (figment coerces those from env), but the forms `PermissiveBool` explicitly accepted — the short forms `1`/`0`/`t`/`T`/`f`/`F`, and quoted-YAML strings like `log_to_console: "false"` — no longer deserialize, and at the bootstrap sites they'll fail `DatadogConfiguration` deserialization and abort startup rather than being coerced.

Logging was one of the few places that deliberately reached for `PermissiveBool`, so the question is whether that was load-bearing (mirroring the Agent's viper-style leniency for these keys) or belt-and-suspenders that's fine to drop for consistency with the other migrated bool keys. If it should be preserved, it needs hoisting to the source layer (e.g. alongside the existing `normalize_datadog_input_forms` coercions) rather than living in the component. Worth a conscious decision either way.

## Minor notes (not blocking)

- The two bootstrap sites go through the lenient `translate_snapshot`, which turns an invalid `log_file_max_size` into a warning and leaves `file_max_size = 0` (→ `ByteSize::b(0)`). The pre-cutover component returned a hard error that aborted startup. This matches the config system's documented lenient-update contract, so it's likely intentional — just flagging that bootstrap no longer rejects a malformed byte size.
- `consume_log_file_max_rolls` clamps negatives to `0` (`value.max(0)`) where the old `try_get_typed::<usize>` would have errored. Benign, and consistent with the project's integer handling.

The `LoggingConfiguration` field-by-field construction, syslog gating, and disable-file short-circuit all match the prior behavior. #1 is the one I'd want addressed before merge; #2 deserves an explicit decision.

webern:
Deferred

PR #2022, opened 2026-07-06
fix: honor site when dd_url is only default-sourced

### LINKED ISSUE #1965: Investigate whether `dd_url`'s schema default makes `site` unreachable in ADP.

jszwedko:
## Summary

When the Core Agent's config stream sends `dd_url` at its schema default (https://app.datadoghq.com) with `source="default"`, ADP was treating it as explicitly set and ignoring the operator's `site` configuration. This caused non-US1 site configurations to misroute to the US intake.

The fix filters out settings whose `source` equals `"default"` at config snapshot ingest so that ADP's own serde defaults apply, allowing the `dd_url` → `site` fallback to work correctly when only `site` is configured.

## Test plan

- Three new unit tests added covering:
  - Explicitly-set keys (any source except `"default"`) are included in the map
  - Default-sourced keys are excluded (even if the value is present)
  - Explicitly-set `dd_url` overrides still take precedence

Resolves #1965.

### DIFF
--- a/bin/agent-data-plane/src/internal/remote_agent.rs
+++ b/bin/agent-data-plane/src/internal/remote_agent.rs
@@ -35,7 +35,15 @@ use tokio::{
     time::{interval, MissedTickBehavior},
 };
 use tonic::{server::NamedService, Status};
-use tracing::{debug, error, info, warn};
+use tracing::{debug, error, info, trace, warn};
+
+/// The source identifier used by the Core Agent for settings that are at their schema default.
+///
+/// The Core Agent includes all known settings in its config snapshot regardless of whether they were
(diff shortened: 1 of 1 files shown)


jszwedko:
Closing in-lieu of a less risky fix in https://github.com/DataDog/saluki/pull/2028

PR #1942, opened 2026-06-29
chore(core): integrate component tasks via supervision trees

tobz:
## Summary

This PR moves the management of component tasks over to supervision trees.

After adding support for managing dynamic child processes in #1874, we're following up with moving the management of component tasks themselves over to supervision trees to work towards full alignment of all asynchronous tasks being managed by a supervisor.

The main change here is that all components run under supervision (duh!) but it's _how_ we do it that matters most. We've designed the topology supervisor construction such that every component runs under its own supervisor, _and_ it gets access to a handle to spawn dynamic child processes on that supervisor. Coupled together with the support we added for significant children in #1874, and this gives us a mechanism to dynamically spawn child processes on a per-component basis that functionally are scoped to the component such that when the component dies, the dynamic child processes also die to. This is poor man's structured concurrency.

None of this replaces (yet) stuff like the global thread pool, which is specifically about compute-bound tasks rather than structured concurrency by itself... but we'll approach that in a future PR.

Overall, I'm not entirely thrilled with the boilerplate traded into here, and I also plan to get rid of most of it in a follow-up PR, but it's deeply tied to the patterns we use for defining component bounds, health check names, and so on... it requires a holistic refactoring that unifies everything in a single go, and it would have just been too much of a change for a single PR. 😅 

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Non-functional (chore, refactoring, docs)
- [ ] Performance

## How did you test this PR?

Existing tests.

## References

DADP-2

PR #2029, opened 2026-07-06
chore(core): unify how we define identifiers for subsystems/components

tobz:
## Summary

This PR attempts to lay the groundwork to unify how we define the unique identifier for a given component/subsystem within the entire process to drive towards avoiding mismatched identifiers for the same component between different registries.

In Saluki, we have a number of "components" -- topology components, yes, but also more "generic" components like standalone subsystems... think the environment provider, specific background tasks, and so on -- that all end up using some sort of unique name within the process to identify themselves for the purpose of health checking or resource accounting and so on. Ideally, these identifiers would always be the same across every possible registry that wants such an identifier _for_ a single component... but currently that's not the case. This is a problem because it makes it hard to aggregate things across registries: imagine trying to get the live usage of a component (resource accounting) while figuring out how much of its memory bounds that usage represents. Without a stable/consistent identifier between both registries, it becomes hard to do and, ultimately, ends up becoming very fragile.

This PR introduces the concept of `SubsystemIdentifier` (don't loveeee the name, might change it) which is meant to represent this shared identifier that a component uses across various registries. It's designed to be built in a hierarchical fashion, so that we define base prefixes that can then be used consistently to build the version of the identifier for each child component and so on.

With this, every "component" -- a specific metadata collector in the environment provider, the Aggregate transform in the metrics portion of the topology, etc -- gets a unique name/identifier that is used everywhere: memory bounds, resource accounting, health registry, supervisor worker names, and so on. Most importantly, we accomplish this through a specialized type that ensures the name/identifier is sanitized and normalized, and acts as a way to gate the usages of this name/identifier by requiring the new type at the API boundary instead of any old plain string

This PR _doesn't_ move everything over to using `SubsystemIdentifier` just yet, only the topology components. The rest will happen in a follow-up PR. However, we did a large amount of simplification related to `ComponentContext`: as it now carries the identifier from the topology, it's somewhat easier to generate it once and use it in places that previously used `ComponentId`, which was the case in `TopologyBlueprint`, `BuiltTopology`, and `ComponentInterconnects`, which have now all been simplified to taken advantage of these changes. Like `SubsystemIdentifier`, I have more changes planned in this area in the future.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

- [x] Existing unit/integration/correctness tests.
- [x] (manual) Ensure there are no straggler references to hand-written topology component identifiers.
- [x] (manual) Ensure resource accounting data is still present for topology components _and_ non-topology components.

## References

DADP-2

### DIFF
--- a/bin/agent-data-plane/src/components/ottl_filter_processor/mod.rs
+++ b/bin/agent-data-plane/src/components/ottl_filter_processor/mod.rs
@@ -160,6 +160,7 @@ mod tests {
             trace::{AttributeValue, Span, Trace},
             Event,
         },
+        support::SubsystemIdentifier,
         topology::{ComponentId, EventsBuffer},
     };
     use stringtheory::MetaString;
@@ -198,12 +199,19 @@ mod tests {
             .sum()
     }
 
+    fn test_component_context() -> ComponentContext {
+        ComponentContext::transform(
+            &SubsystemIdentifier::from_segments(["test"]),
+            ComponentId::try_from("ottl_filter").unwrap(),
+        )
+    }
+
     /// When `ottl_filter_config` is absent, config defaults to empty conditions and no spans are dropped.
     #[tokio::test]
     async fn from_configuration_absent_key_returns_default() {
         let (config, _) = ConfigurationLoader::for_tests(None, None, false).await;
         let ottl_config = OttlFilterConfiguration::from_configuration(&config).expect("should succeed");
-        let ctx = ComponentContext::transform("test", ComponentId::try_from("ottl_filter").unwrap());
+        let ctx = test_component_context();
         let mut transform = ottl_config.build(ctx).await.expect("build should succeed");
         let span = make_span(1, 1, HashMap::from([("a".into(), "b".into())]));
         let trace = make_trace(vec![span], None);
@@ -236,7 +244,7 @@ mod tests {
         });
         let (config, _) = ConfigurationLoader::for_tests(Some(cfg_json), None, false).await;
         let ottl_config = OttlFilterConfiguration::from_configuration(&config).expect("config is valid");
-        let ctx = ComponentContext::transform("test", ComponentId::try_from("ottl_filter").unwrap());
+        let ctx = test_component_context();
         let result = ottl_config.build(ctx).await;
         assert!(result.is_err(), "invalid OTTL condition must make build fail");
     }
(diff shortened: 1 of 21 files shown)


tobz:
* **#2029** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2029?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2029?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1942** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1942?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1874** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1874?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1873** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1873?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1860** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1860?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1856** <a href="https://app.graphite.com/github/pr/DataDog/saluki/1856?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

aqian01:
Similar to the [codex comment](https://github.com/DataDog/saluki/pull/2029/s#discussion_r3537258127)

I think two different topology names can turn into the same internal ID,
Example:
  - primary-a becomes topology.primary_a
  - primary_a also becomes topology.primary_a
  
  Had codex make a quick test to check for this
  ```
  #[cfg(test)]
    61 +mod tests {
    62 +    use super::topology_identifier;
    63 +
    64 +    #[test]
    65 +    fn topology_identifier_keeps_distinct_raw_names_distinct() {
    66 +        let hyphenated = topology_identifier("primary-a");
    67 +        let underscored = topology_identifier("primary_a");
    68 +
    69 +        assert_ne!(
    70 +            hyphenated.to_string(),
    71 +            underscored.to_string(),
    72 +            "distinct topology names must not collapse to the same identifier"
    73 +        );
    74 +    }
    75 +}
  ```

tobz:
> Similar to the [codex comment](https://github.com/DataDog/saluki/pull/2029/s#discussion_r3537258127)
> 
> I think two different topology names can turn into the same internal ID, Example:

@aqian01 Mmmm, yes. That's suboptimal, although luckily not an issue right now since we never create more than one topology at a time.

I'll approach this as follow-up PR since we'll need to make `TopologyBlueprint::new` fallible so that we can do validation on the name... or maybe I can figure out a better solution to this problem that doesn't involve having to always validate at every individual point. 😅

aqian01 (approved):

PR #2031, opened 2026-07-06
fix(process-memory): support RSS queries on AIX

thieman:
## Summary

Adds AIX support to the `process-memory` crate by reading `/proc/<pid>/psinfo` and extracting the resident set size from the AIX `psinfo_t` layout.

This is one of the Saluki-side blockers from DADP-142 / the AIX Confluence notes. With only the temporary `aws-lc-rs` AIX branch applied, the AIX ADP build previously stopped at `process-memory` because the crate rejected all non-Linux/macOS/Windows targets.

## Key changes

- Add an AIX `Querier` backend for resident set size queries.
- Include AIX in the crate's supported target list.
- Parse `psinfo_t.pr_rssize` as a native-endian `uint64_t` value in KiB units, then convert to bytes with overflow checks.
- Add AIX parser coverage for successful parsing, zero RSS, truncated `psinfo`, and overflow.
- Add AIX-only integration coverage that compares `Querier` against the raw `/proc/<pid>/psinfo` RSS field.
- Keep the `dhat` allocation test active on non-AIX targets and make the harness-free test binary a no-op on AIX, where `dhat` is intentionally not enabled.

## AIX implementation notes

- AIX on `soaix499` does not expose `/proc/self/psinfo`; it exposes `/proc/<pid>/psinfo`.
- IBM's AIX `/proc` documentation states that `/proc` files provide 64-bit mode-invariant data to observers and that future structure growth appends fields.
- IBM documents `psinfo_t.pr_rssize` as `uint64_t pr_rssize`, resident set size in KiB (1024) units.
- I verified the AIX 7.3 `psinfo_t` layout on `soaix499` with a small C program using `<sys/procfs.h>`:
  - `sizeof(psinfo_t) = 448`
  - `offsetof(psinfo_t, pr_size) = 96`
  - `offsetof(psinfo_t, pr_rssize) = 104`
  - `offsetof(psinfo_t, pr_start) = 112`
- Manual magnitude validation on `soaix499` showed `Querier`, raw `pr_rssize * 1024`, and `ps v <pid>` report the same RSS magnitude. Treating `pr_rssize` as pages would report roughly 4x too high on the 4 KiB page-size test host.
- The implementation is fail-closed: open/read/truncation/overflow failures return `None` rather than panicking.

## Continuation context

This PR is part of a three-PR Saluki-side AIX enablement stack:

1. This PR: `process-memory` AIX RSS support.
2. #2032: `stringtheory` big-endian `MetaString` fallback.
3. #2034: `datadog-agent-commons` AIX platform settings.

A separate upstream/fork fix is still needed for `aws-lc-rs`; during validation I used the local checkout of Travis's fork/branch on the AIX host. That temporary override is **not** included in this PR.

## Test plan

Local:

- `make fmt`
- `cargo check -p process-memory`
- `cargo check -p process-memory --tests`
- `cargo test -p process-memory`
- Direct parser test compile/run for `lib/process-memory/src/aix.rs`.

AIX 7.3 (`soaix499`):

- `cargo test -p process-memory`
- Result:
  - AIX unit tests: `4 passed`
  - AIX integration tests: `1 passed`
  - `tests/no_runtime_allocations.rs` runs successfully as a no-op on AIX
  - Doc tests: `0 failed`
- Manual RSS magnitude check:
  - `Querier` and raw `pr_rssize * 1024` reported about `5.9-6.1 MiB` for the test process.
  - `ps v <pid>` reported the same RSS magnitude.
  - Treating `pr_rssize` as a page count would have reported about `23.7-24.4 MiB`.

## Notes

- This PR does not include the temporary local `aws-lc-rs` patch used for broader AIX validation.
- The latest commits were made with the normal pre-commit hook path; the hook completed successfully, including formatting, clippy, license/advisory checks, prose checks, and API docs build.

### DIFF
--- a/lib/process-memory/src/aix.rs
+++ b/lib/process-memory/src/aix.rs
@@ -1,0 +1,83 @@
+use std::{fs::File, io::Read, path::PathBuf};
+
+// Offset of `pr_rssize` in the 64-bit AIX 7.3 `psinfo_t` structure.
+const PR_RSSIZE_OFFSET: usize = 104;
+const PR_RSSIZE_SIZE: usize = std::mem::size_of::<i64>();
+
+/// A memory usage querier.
+pub struct Querier {
+    psinfo_path: PathBuf,
+}
+
+impl Querier {
+    /// Gets the resident set size of this process, in bytes.
+    ///
+    /// If the resident set size can't be determined, `None` is returned. This could be for a number of underlying
+    /// reasons, but should generally be considered an incredibly rare/unlikely event.
+    pub fn resident_set_size(&mut self) -> Option<usize> {
+        let mut file = File::open(&self.psinfo_path).ok()?;
+        let mut buf = [0; PR_RSSIZE_OFFSET + PR_RSSIZE_SIZE];
+        file.read_exact(&mut buf).ok()?;
+
+        resident_set_size_from_psinfo(&buf, page_size()?)
+    }
+}
+
+impl Default for Querier {
+    fn default() -> Self {
+        Self {
+            psinfo_path: PathBuf::from(format!("/proc/{}/psinfo", std::process::id())),
+        }
+    }
+}
+
+fn page_size() -> Option<usize> {
+    let page_size = unsafe { libc::sysconf(libc::_SC_PAGESIZE) };
+    if page_size <= 0 {
+        None
+    } else {
+        Some(page_size as usize)
+    }
+}
+
+fn resident_set_size_from_psinfo(psinfo: &[u8], page_size: usize) -> Option<usize> {
+    let raw_rss_pages = psinfo.get(PR_RSSIZE_OFFSET..PR_RSSIZE_OFFSET + PR_RSSIZE_SIZE)?;
+    let rss_pages = i64::from_ne_bytes(raw_rss_pages.try_into().ok()?);
+    let rss_pages = usize::try_from(rss_pages).ok()?;
+
+    rss_pages.checked_mul(page_size)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{resident_set_size_from_psinfo, Querier, PR_RSSIZE_OFFSET, PR_RSSIZE_SIZE};
+
+    #[test]
+    fn basic() {
+        let mut querier = Querier::default();
+        assert!(querier.resident_set_size().is_some());
+    }
+
+    #[test]
+    fn parses_resident_set_size_from_psinfo() {
+        let page_size = 4096;
+        let rss_pages = 7i64;
+        let mut psinfo = [0; PR_RSSIZE_OFFSET + PR_RSSIZE_SIZE];
+        psinfo[PR_RSSIZE_OFFSET..PR_RSSIZE_OFFSET + PR_RSSIZE_SIZE].copy_from_slice(&rss_pages.to_ne_bytes());
+
+        assert_eq!(resident_set_size_from_psinfo(&psinfo, page_size), Some(28672));
+    }
+
+    #[test]
+    fn rejects_truncated_psinfo() {
+        assert_eq!(resident_set_size_from_psinfo(&[], 4096), None);
+    }
+
+    #[test]
+    fn rejects_negative_resident_set_size() {
+        let mut psinfo = [0; PR_RSSIZE_OFFSET + PR_RSSIZE_SIZE];
+        psinfo[PR_RSSIZE_OFFSET..PR_RSSIZE_OFFSET + PR_RSSIZE_SIZE].copy_from_slice(&(-1i64).to_ne_bytes());
+
+        assert_eq!(resident_set_size_from_psinfo(&psinfo, 4096), None);
+    }
+}
--- a/lib/process-memory/src/aix.rs
+++ b/lib/process-memory/src/aix.rs
@@ -43,13 +43,7 @@ fn resident_set_size_from_psinfo(psinfo: &[u8]) -> Option<usize> {
 
 #[cfg(test)]
 mod tests {
-    use super::{resident_set_size_from_psinfo, Querier, KIB, PR_RSSIZE_OFFSET, PR_RSSIZE_SIZE};
-
-    #[test]
-    fn basic() {
-        let mut querier = Querier::default();
-        assert!(querier.resident_set_size().is_some());
-    }
+    use super::{resident_set_size_from_psinfo, KIB, PR_RSSIZE_OFFSET, PR_RSSIZE_SIZE};
 
     #[test]
     fn parses_resident_set_size_from_psinfo() {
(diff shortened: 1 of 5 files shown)


lib/process-memory/src/aix.rs
```diff
@@ -0,0 +1,123 @@
+use std::{fs::File, io::Read, path::PathBuf};
+
+const KIB: u64 = 1024;
+// Verified against AIX 7.3 <sys/procfs.h>: offsetof(psinfo_t, pr_rssize) == 104.
```
thieman, line 5: Is there a risk that this is going to vary based on the specific AIX version? Can you check docs or anything here?

lib/process-memory/tests/no_runtime_allocations.rs
```diff
@@ -8,29 +8,35 @@
...
 #[global_allocator]
 static ALLOC: dhat::Alloc = dhat::Alloc;
 
 fn main() {
-    // This test ensures that after initially creating `Querier`, there are _no_ runtime allocations made when calling
-    // `resident_set_size`.
-    //
-    // This invariant should always hold, and should do so for all supported platforms.
-    let mut querier = Querier::default();
+    #[cfg(not(target_os = "aix"))]
```
thieman, line 21: Can we gate this entire file off from running on AIX? Or do we need it here for some reason

lib/process-memory/src/aix.rs
thieman: [OpenAI API model] Checked IBM AIX /proc docs and the test box. The docs state that /proc files provide 64-bit mode-invariant data to both 32-bit and 64-bit observers, and that future structure growth is by appending fields at the end. `pr_rssize` is documented as `uint64_t` in KiB units. I also verified on `soaix499` with `offsetof(psinfo_t, pr_rssize) == 104`, and the live AIX test plus manual `ps v` comparison match the expected magnitude. Added a code comment for the documented stability guarantee in eb0accaa30.

lib/process-memory/tests/no_runtime_allocations.rs
thieman: [OpenAI API model] We need the file to still compile as a test target on AIX because it is declared as a `harness = false` `[[test]]` in Cargo.toml, and Cargo does not give us a target-specific way to omit that test target only on AIX. If the whole file is crate-gated away on AIX, the harness-free test binary has no `main` and fails to build. The current shape keeps `fn main()` available as a no-op on AIX while gating the `dhat` imports/global allocator/assertion body to non-AIX targets. Verified on `soaix499`: `cargo test -p process-memory` runs `tests/no_runtime_allocations.rs` successfully as a no-op on AIX.

lib/process-memory/src/aix.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    #[test]
+    #[cfg(target_endian = "little")]
+    fn parses_little_endian_resident_set_size() {
+        let mut psinfo = [0; PR_RSSIZE_OFFSET + PR_RSSIZE_SIZE];
+        psinfo[PR_RSSIZE_OFFSET..PR_RSSIZE_OFFSET + PR_RSSIZE_SIZE].copy_from_slice(&[7, 0, 0, 0, 0, 0, 0, 0]);
+
+        assert_eq!(resident_set_size_from_psinfo(&psinfo), Some(7168));
+    }
```
tobz: These seem unnecessary?

The little endian one is never going to run on AIX, but more importantly, the actual parsing code uses `from_ne_bytes`, and above tests cover that when using `to_ne_bytes`.

lib/process-memory/src/aix.rs
```diff
@@ -0,0 +1,124 @@
...
+        const PR_START_OFFSET: usize = 112;
+
+        let rss_kib = 7u64;
+        let mut psinfo = [0; PR_START_OFFSET + PR_RSSIZE_SIZE];
+        psinfo[PR_SIZE_OFFSET..PR_SIZE_OFFSET + PR_RSSIZE_SIZE].copy_from_slice(&3u64.to_ne_bytes());
+        psinfo[PR_RSSIZE_OFFSET..PR_RSSIZE_OFFSET + PR_RSSIZE_SIZE].copy_from_slice(&rss_kib.to_ne_bytes());
+        psinfo[PR_START_OFFSET..PR_START_OFFSET + PR_RSSIZE_SIZE].copy_from_slice(&99u64.to_ne_bytes());
+
+        assert_eq!(resident_set_size_from_psinfo(&psinfo), Some((rss_kib * KIB) as usize));
+    }
```
tobz: This one also seems unnecessary. `parses_resident_set_size_from_psinfo` already asserts that we specifically read at the desired offset/length, so testing that we _don't_ read from adjacent fields doesn't really ... tell us anything we don't already know, I don't think?

lib/process-memory/src/aix.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    #[test]
+    fn converts_resident_set_size_from_kib_not_pages() {
+        let rss_kib = 7u64;
+        let mut psinfo = [0; PR_RSSIZE_OFFSET + PR_RSSIZE_SIZE];
+        psinfo[PR_RSSIZE_OFFSET..PR_RSSIZE_OFFSET + PR_RSSIZE_SIZE].copy_from_slice(&rss_kib.to_ne_bytes());
+
+        assert_eq!(resident_set_size_from_psinfo(&psinfo), Some(7168));
+    }
```
tobz: Similarly: we're already inherently asserting this in the first very test by checking for `rss_kib * KIB`.

tobz (changes requested):

lib/process-memory/src/aix.rs
thieman: [OpenAI API model] Removed these in 6c6d737ccb. Agreed: they were testing `from_ne_bytes`/`to_ne_bytes` rather than process-memory behavior, and the parser happy-path test already covers native-endian parsing through the actual code path.

lib/process-memory/src/aix.rs
thieman: [OpenAI API model] Removed in 6c6d737ccb. Agreed that `parses_resident_set_size_from_psinfo` already proves the selected offset/length because it places the fixture value at `PR_RSSIZE_OFFSET` and expects the parsed result.

lib/process-memory/src/aix.rs
thieman: [OpenAI API model] Removed in 6c6d737ccb. Agreed that the first parser test already asserts the KiB conversion via `rss_kib * KIB`, so this was duplicative.

thieman:
[GPT 5.5] Test cleanup note: in addition to the parser tests Tobz explicitly called out, I also removed two duplicate smoke tests while self-reviewing the test set:

- `aix::tests::basic` in `lib/process-memory/src/aix.rs`
- `resident_set_size_returns_nonzero_value` in `lib/process-memory/tests/aix.rs`

Both only asserted that `Querier::resident_set_size()` returned `Some` / nonzero on AIX. The remaining AIX integration test, `resident_set_size_is_close_to_proc_psinfo_rss_field`, covers that same live behavior and is stronger because it also compares the queried RSS against an independent parse of `/proc/<pid>/psinfo`. The remaining unit tests now cover distinct parser behavior: successful parse/conversion, zero RSS, truncation, and overflow.

tobz (approved):

PR #2032, opened 2026-07-06
fix(stringtheory): support big-endian platforms for `MetaString`

thieman:
## Summary

Adds big-endian support to `stringtheory::MetaString` while keeping the existing optimized three-word union representation.

This is one of the Saluki-side blockers from DADP-142 / the AIX Confluence notes. After getting past `process-memory`, the AIX ADP build stopped at `stringtheory` because the previous `MetaString` implementation explicitly required 64-bit little-endian platforms.

## Key changes

- Keep `MetaString` on the existing compact three-word representation for both little- and big-endian 64-bit targets.
- Make the discriminant/tag byte endian-aware:
  - little-endian uses byte `23`, preserving the existing layout and 23-byte inline capacity.
  - big-endian uses byte `16`, the high byte of the third machine word, with a 16-byte inline capacity so inline string bytes remain contiguous.
- Preserve owned/static/interned/shared tagging through the high byte of the third `usize` across endianness.
- Remove the separate big-endian fallback implementation in favor of the unified layout in `lib/stringtheory/src/lib.rs`.
- Add tests for endian-specific layout constants, discriminant byte values, inline capacity boundaries, contiguous inline storage, and the public big-endian API surface.

## Design notes

- The optimized `MetaString` layout stores all variants in three machine words and determines the active variant from a single tag byte.
- On little-endian targets, that tag byte is the final byte of the 24-byte layout. On big-endian targets, the high byte of the third `usize` is instead the first byte of the third machine word, so the tag byte index must change.
- Big-endian inline capacity is intentionally reduced from 23 bytes to 16 bytes. This keeps inline string contents as a contiguous prefix, so `Deref<Target = str>` can still return a borrowed `&str` without allocating.
- Owned strings remain non-cheap to clone on both endiannesses because `From<String>` now uses the same owned-string variant everywhere.

## Continuation context

This PR is part of a three-PR Saluki-side AIX enablement stack:

1. #2031: `process-memory` AIX RSS support.
2. This PR: `stringtheory` big-endian `MetaString` support.
3. #2034: `datadog-agent-commons` AIX platform settings.

A separate upstream/fork fix is still needed for `aws-lc-rs`; during validation I used the local checkout of Travis's fork/branch on the AIX host. That temporary override is **not** included in this PR.

Temporary AIX validation setup used on `soaix499`:

```sh
export PATH=/opt/freeware/lib/RustSDK/1.92/bin:/opt/freeware/bin:/usr/sbin:/usr/bin:/bin
export CC=/opt/freeware/bin/gcc-8
export CXX=/opt/freeware/bin/g++-8
export AR=/usr/bin/ar
export RANLIB=/usr/bin/ranlib
export CARGO_HOME=$HOME/dd/cargo-home
export CARGO_TARGET_DIR=$HOME/dd/saluki-target
export CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
```

## Test plan

- RED on AIX before implementation:
  - Added temporary big-endian coverage while developing the initial fix.
  - `cargo test -p stringtheory --test big_endian --no-run` failed at the existing little-endian compile gate.
  - The final PR folds that coverage into the normal crate unit tests and removes the temporary integration test.
- Local/native little-endian:
  - `make fmt`
  - `cargo nextest run -p stringtheory`
  - Focused inline-boundary tests for the tag-byte off-by-one case:
    - `cargo test -p stringtheory inlined_string_uses_every_byte_before_tag_byte -- --exact`
    - `cargo test -p stringtheory inlined_string_rejects_one_byte_past_tag_byte -- --exact`
  - `cargo check --workspace`
  - `make check-deny`
  - `git diff --check`
- Big-endian compile target:
  - `rustup target add s390x-unknown-linux-gnu`
  - `cargo check -p stringtheory --target s390x-unknown-linux-gnu`
  - `cargo check -p stringtheory --tests --target s390x-unknown-linux-gnu`
- AIX 7.3 (`soaix499`) on latest PR head `90a2aa6560` archived to `~/dd/saluki-pr2032-latest`:
  - `cargo test -p stringtheory --lib --ignore-rust-version`
  - Result: `45 passed; 0 failed`, including the endian-specific layout and public API unit tests.
- AIX 7.3 (`soaix499`) with temporary local `aws-lc-rs` patch and the process-memory fix synced:
  - `cargo build -p agent-data-plane` got past `stringtheory` and failed at the next known blocker, AIX platform constants in `datadog-agent-commons`.

## Notes

- This PR does not include the temporary local `aws-lc-rs` patch used for AIX validation.
- The current branch commits were created with normal pre-commit checks enabled. Earlier notes about bypassing local hooks are obsolete after the rebase and follow-up commits.


lib/stringtheory/src/lib.rs
```diff
@@ -52,9 +58,9 @@ const fn get_scaled_union_tag(tag: u8) -> usize {
 }
 
 // High-level invariant checks to ensure `stringtheory` isn't being used on an unsupported platform.
-#[cfg(not(all(target_pointer_width = "64", target_endian = "little")))]
+#[cfg(not(all(target_pointer_width = "64", any(target_endian = "little", target_endian = "big"))))]
```
thieman: any(little, big) seems extraneous, can you simplify

thieman: [GPT-5] Simplified to `#[cfg(not(target_pointer_width = "64"))]`. Also added focused layout tests for the endian-specific tag byte, inline capacity, and contiguous inline storage boundary.

lib/stringtheory/src/lib.rs
```diff
@@ -26,7 +27,12 @@ use self::interning::{InternedString, InternedStringState, Interner};
 const ZERO_VALUE: usize = 0;
 const TOP_MOST_BIT: usize = usize::MAX & !(isize::MAX as usize);
 const INLINED_STR_DATA_BUF_LEN: usize = std::mem::size_of::<usize>() * 3;
-const INLINED_STR_MAX_LEN: usize = INLINED_STR_DATA_BUF_LEN - 1;
+const INLINED_STR_TAG_INDEX: usize = if cfg!(target_endian = "little") {
+    INLINED_STR_DATA_BUF_LEN - 1
+} else {
+    std::mem::size_of::<usize>() * 2
+};
+const INLINED_STR_MAX_LEN: usize = INLINED_STR_TAG_INDEX;
```
tobz, line 38: I think this is wrong?

Naively, I would expect the maximum inlined tag to be `tag_byte_index - 1`, since the implication is that the tag byte is unusable for string data.

tobz (changes requested):

lib/stringtheory/src/lib.rs
thieman: Clanker thinks it's right because the index already has the implicit +1 in it, so if you can build up to index 16 then you can use bytes 0..15 which is len 16 (len == index)

Will have it add some additional tests on boundary checking here

thieman: [GPT-5] I checked this carefully: `INLINED_STR_TAG_INDEX` is a byte index, while `INLINED_STR_MAX_LEN` is a byte count. If the tag byte is at index 16, the usable contiguous prefix is `0..16`, which is 16 bytes. Using `tag_byte_index - 1` would leave the byte immediately before the tag unused and would also regress little-endian capacity from 23 to 22. I added focused tests for this boundary: one uses every byte before the tag byte, and one rejects `tag_byte_index + 1` bytes.

tobz: Yeah, looking at this now, I'm reminded why I always hate index-vs-length stuff.

The explanation makes sense... it was really the difference in how we calculate the index (one side doing foo - 1, the other just doing blah() * 2) that was making me second guess things.

👍🏻

lib/stringtheory/src/lib.rs
```diff
@@ -952,9 +965,80 @@ mod tests {
...
+        assert_eq!(generic_map.inner.get_union_type(), UnionType::InternedGenericMap);
+        assert_eq!(
+            discriminant_byte(&generic_map.inner),
+            UNION_TYPE_TAG_VALUE_INTERNED_GENERIC_MAP
+        );
+    }
+
+    #[test]
+    fn inlined_string_uses_every_byte_before_tag_byte() {
+        let input = "a".repeat(INLINED_STR_TAG_INDEX);
```
tobz: Maybe use the length constant for clarity? (I recognize that index == length.)

lib/stringtheory/src/lib.rs
```diff
@@ -1154,8 +1238,17 @@ mod tests {
...
     }
 
+    #[test]
+    fn inlined_string_respects_endian_capacity() {
+        let max_len = "a".repeat(INLINED_STR_MAX_LEN);
+        let too_long = "a".repeat(INLINED_STR_MAX_LEN + 1);
+
+        assert!(MetaString::try_inline(&max_len).is_some());
+        assert!(MetaString::try_inline(&too_long).is_none());
+    }
```
tobz: Seems duplicative: we're already testing this above with the two individual inlining tests.

lib/stringtheory/tests/big_endian.rs
```diff
@@ -0,0 +1,55 @@
+#![cfg(target_endian = "big")]
```
tobz: It's ... unclear to me why this has to be a separate test. Are we somehow not testing the public API in the normal unit tests when run on a big-endian platform?

lib/stringtheory/src/lib.rs
```diff
@@ -952,9 +965,80 @@ mod tests {
...
+            std::mem::size_of::<usize>() - 1
+        } else {
+            0
+        };
+        let expected_tag_index = std::mem::size_of::<usize>() * 2 + expected_top_byte_index;
+
+        assert_eq!(TOP_BYTE_INDEX, expected_top_byte_index);
+        assert_eq!(INLINED_STR_TAG_INDEX, expected_tag_index);
+        assert_eq!(INLINED_STR_MAX_LEN, expected_tag_index);
+    }
```
tobz: I think we can drop this? We're just restating the const definitions, and other tests should break if we mess those up somehow.

tobz (changes requested):

lib/stringtheory/tests/big_endian.rs
thieman: Confirm these public methods are all covered in the module unit tests and if so delete this file

lib/stringtheory/src/lib.rs
thieman: [GPT-5] Updated to use `INLINED_STR_MAX_LEN` in the boundary tests. The tests still validate the off-by-one behavior, but now read in terms of length rather than the underlying tag-byte index.

lib/stringtheory/src/lib.rs
thieman: [GPT-5] Removed this duplicate test. The two explicit boundary tests cover the max accepted length and the first rejected length.

lib/stringtheory/src/lib.rs
thieman: [GPT-5] Removed this test. It was restating the constant definitions rather than testing behavior; the inline capacity and raw discriminant tests cover the behavior those constants drive.

lib/stringtheory/tests/big_endian.rs
thieman: [GPT-5] Confirmed the normal unit tests cover the public API on big-endian now that the implementation is unified. I moved the remaining coverage from this integration test into `src/lib.rs` (ordering, protobuf `Chars` conversions, and `CheapMetaString::try_cheap_clone`) and deleted this cfg-gated test file.

tobz (approved):

PR #2035, opened 2026-07-06
chore(ci): pin Rust MSRV to 1.94

thieman:
## Summary

Pin Saluki's Rust MSRV to 1.94 and align the Rust toolchain pins used by local development, Cargo metadata, CI images, Docker build defaults, and nightly-only tooling.

## Changes

- Set `workspace.package.rust-version = "1.94"` in the root `Cargo.toml`.
- Keep `rust-version = "1.94"` in the standalone `lib/saluki-io/fuzz/Cargo.toml`, which cannot inherit from the root workspace.
- Pin `rust-toolchain.toml` to `1.94.0`.
- Update Rust defaults in CI/build Dockerfiles from `1.93.0`/`stable` to `1.94.0`.
- Pin nightly-only tooling to `nightly-2026-01-18`, which reports `rustc 1.94.0-nightly`.
- Replace unqualified `cargo +nightly` commands in Makefile, GitLab fuzzing, docs, and the fuzz infrastructure script with the pinned nightly version.
- Replace uses of unstable `Atomic*::try_update` with stable `fetch_update` so the workspace builds on Rust 1.94.
- Pass an explicit cwd to `cargo sort --workspace` in Makefile fmt targets so formatting checks work from linked worktrees.
- Make the `host_exec_times_out` unit test use a Windows-native sleep command so it exercises timeout behavior on Windows CI.
- Add retrying cleanup for macOS Datadog Agent DMG mounts in both provisioning and CI before/after hooks for persistent runners.
- Bump `crossbeam-epoch` in `Cargo.lock` to `0.9.20` to resolve `RUSTSEC-2026-0204` reported by `cargo-deny` during pre-commit.

## Test plan

- `cargo +1.94.0 nextest run -p panoramic actions::target_exec::tests::host_exec_times_out`
- `cargo +1.94.0 check --workspace --all-targets`
- `make check-fmt`
- `make generate-api-docs`
- `make check-deny`
- `make check-docs`
- `python3 -m py_compile ci/tooling/fuzz_infra.py`
- `bash -n ci/tooling/cleanup-macos-dda-mounts.sh`
- `make -n provision-macos-test-env`
- `git diff --check`
- Pre-commit hook completed successfully, including fmt, clippy, license check, cargo-deny, Vale, and API docs check.


lib/stringtheory/src/interning/fixed_size.rs
```diff
@@ -225,17 +225,9 @@ impl EntryHeader {
...
-            .is_ok();
-        #[cfg(feature = "loom")]
-        return self
-            .refs
+        self.refs
             .fetch_update(AcqRel, Acquire, |refs| (refs != 0).then_some(refs + 1))
-            .is_ok();
+            .is_ok()
     }
 }
```
jszwedko, line 232: This may break the `make generate-api-docs` again since that uses nightly where `fetch_update` is deprecated.

Makefile
```diff
@@ -525,7 +525,7 @@ check-fmt: ## Check that all Rust source files are formatted properly
 	@echo "[*] Checking Rust source code formatting..."
 	@cargo +nightly fmt -- --check
 	@echo "[*] Checking Cargo.toml formatting..."
-	@cargo sort --workspace --check >/dev/null
+	@cargo sort --workspace . --check >/dev/null
```
jszwedko, line 528: Curious about why this change is needed but 🤷

jszwedko (approved):

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.96.0"
+channel = "1.94.0"
```
tobz, line 2: Why are we downgrading?

thieman: AIX is ~3 months behind and so its current latest is 1.94, see https://dd.slack.com/archives/C070PQKLLP5/p1783368960098579

.gitlab/fuzz.yml
```diff
@@ -11,7 +11,7 @@ fuzz:infra:
     - when: manual
       allow_failure: true
   before_script:
-    - cargo +nightly install cargo-fuzz
+    - cargo +nightly-2026-01-18 install cargo-fuzz
```
thieman: This is a 1.94 nightly

Cargo.lock
```diff
@@ -1058,9 +1058,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
```
thieman, line 1061: Why did this get upgraded as part of this?

lib/stringtheory/src/interning/fixed_size.rs
thieman: [GPT-5] I checked this after aligning the nightly pin. The PR now uses `nightly-2026-01-18`, which reports `rustc 1.94.0-nightly`, and `make generate-api-docs` passes with `fetch_update`. The earlier `nightly-2026-01-15` also reported 1.94 but ICE'd in `chumsky`; `2026-01-18` avoids that while staying on the 1.94 nightly train.

Makefile
thieman: [GPT-5] This is needed for linked worktrees with the current `cargo-sort` behavior. I rechecked: `cargo sort --workspace` fails with `no file found at: .../.worktrees/pin-msrv-1.94`, while `cargo sort --workspace . --check` succeeds. Passing `.` makes the cwd explicit and lets the hook/Makefile checks run from this worktree.

.gitlab/fuzz.yml
thieman: [GPT-5] Confirmed. I updated this to `nightly-2026-01-18`; `rustc +nightly-2026-01-18 --version` reports `rustc 1.94.0-nightly (fe98ddcfc 2026-01-17)`. I also switched the Makefile's unqualified `+nightly` fmt invocations and the fuzz README examples to the same pinned nightly.

Cargo.lock
thieman: [GPT-5] The pre-commit hook started failing in `make check-deny` because `crossbeam-epoch 0.9.18` is now flagged by `RUSTSEC-2026-0204`. I bumped only that lockfile entry to `0.9.20`, the fixed version, so the existing cargo-deny gate passes again. No manifest dependency changed.

.gitlab/fuzz.yml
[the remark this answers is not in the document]
thieman: [GPT-5] Fixed in `bedacc39f3`. `fuzz_infra.py` now reads `RUST_NIGHTLY_VERSION` and uses it for both `cargo fuzz list` and `cargo fuzz build`; the GitLab job defines that variable as `nightly-2026-01-18` and uses the same value when installing `cargo-fuzz`. I also verified there are no remaining unqualified `+nightly` fuzz invocations and `python3 -m py_compile ci/tooling/fuzz_infra.py` passes.

PR #2049, opened 2026-07-09
feat(metrics): Make V3 encoder no_std

atanzu:
## Summary
Handroll the protobuf encoder to make our V3 protocol encoder usable for no_std environments.

We already did some manual work wrt the representation of the payload structure, so the protobuf dependency we only needed to serialize the data. Since the serialization layer is thin, now we do that ourselves, which allows us to remove the dependency from protobuf.

Among other changes, we replace the default HashMaps with hashbrown, do std->core where possible, and add a no_std header.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?
To make sure that the handrolled serializer works as it should, we extend the tests with `stele` which allows to verify the serialization against a good known encoder.

## References

AGTMETRICS-535

### DIFF
--- a/lib/datadog-agent/metrics-v3/Cargo.toml
+++ b/lib/datadog-agent/metrics-v3/Cargo.toml
@@ -6,6 +6,11 @@ edition = { workspace = true }
 license = { workspace = true }
 repository = { workspace = true }
 
+[lints]
+workspace = true
+
 [dependencies]
-foldhash = { workspace = true }
-protobuf = { workspace = true }
+# Not inherited from the workspace: the workspace default enables foldhash's "std" feature, which
+# would pull `std` back into this otherwise `no_std` crate.
+foldhash = { version = "0.2", default-features = false }
+hashbrown = { workspace = true }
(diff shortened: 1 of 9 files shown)


lib/datadog-agent/metrics-v3/Cargo.toml
```diff
@@ -6,6 +6,11 @@ edition = { workspace = true }
...
 repository = { workspace = true }
 
+[lints]
+workspace = true
+
 [dependencies]
-foldhash = { workspace = true }
-protobuf = { workspace = true }
+# Not inherited from the workspace: the workspace default enables foldhash's "std" feature, which
+# would pull `std` back into this otherwise `no_std` crate.
```
webern, line 14: Does this actually work during a workspace build?

webern (commented):
Is this crate going to be published into an environment that requires nostd?

atanzu:
Closing this PR because we will make V3 encoder part of libdatadog instead, and then import it from there.

PR #2048, opened 2026-07-09
chore(core): unify how identifiers are created for health registration and environment providers

tobz:
## Summary

This PR is a follow-up to #2029 to move over other areas of the codebase to defining component identifiers via `SubsystemIdentifier` for use with the health registry and in environment providers and so on.

We've focused on two main areas here:

- cleaning up how we establish our component/subsystem identifiers in our primary environment provider implementation (in `bin/agent-data-plane`)
- augmenting the health registry itself to switch over using `SubsystemIdentifier` more directly to avoid pitfalls at the API boundary between `SubsystemIdentifier` and `&str`/`String`

Overall, we've unified the identifiers that get generated for components in the environment provider in terms of what exists in the health registry and what exists in the resource accounting registry.

Through this PR, it's become apparent that we'll need to think about moving `SubsystemIdentifier` somewhere more central so other non-Saluki crates can potentially depend on it, including and up to potentially just establishing a centralized registry for identifiers themselves... in order to solve the whole issue of "how do I know I have a unique identifier?". I plan to tackle _that_, and the update to switch `resource_accounting::ComponentRegistry` over to depending on `SubsystemIdentifier`, in a follow-up PR.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Non-functional (chore, refactoring, docs)
- [ ] Performance

## How did you test this PR?

Existing and new unit tests.

## References

DADP-2

PR #2050, opened 2026-07-09
chore(core): split `resource-accounting` crate into resource tracking and bounds accounting

tobz:
## Summary

This PR splits the `resource-accounting` crate into its two primary parts -- resource tracking and bounds accounting -- and moves them to `saluki-common` and `saluki-core`, respectively.

As part of the refactoring work around `SubsystemIdentifier` (#2029, #2048), we want to utilize `SubsystemIdentifier` for all things that generate structured identifiers that are attached to "components", like topology components or the environment provider and so on. Part of this is utilizing `SubsystemIdentifier` instead of utilizing opaque string types so that we can enforce invariants -- only certain characters in identifiers, etc -- at the type system. However, to do this for the resource accounting component registry (`resource_accounting::ComponentRegistry`) means depending on `saluki-core` where `SubsystemIdentifier` is defined... which ends up as a circular dependency. Doh!

In preparation of moving `SubsystemIdentifier` to `saluki-common`, we need to split out the parts of `resource-accounting` that `saluki-common` depends on, and as it so happens, there's already a natural "seam", as Claude would call it, between the _tracking_ of resource usage (allocations, CPU) and the _accounting_ of expected usage (memory bounds, bounds verification, etc).

This PR moves all resource tracking-related bits into `saluki-common` to satisfy not only its own usages of those types, but also acts as the non-circular dependency on which other crates can depend to utilize those types as well. Since memory bounds-related bits are (at least as of right now) only utilized in higher-level crates where components and such are defined, we've moved those to `saluki-core` as all crates that have types with defined bounds already import `saluki-core`. This resolves all of our dependency tensions and unlocks the ability for us to move `ComponentRegistry` to utilize `SubsystemIdentifier`.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Existing tests.

## References

DADP-2


lucastemb (approved):
LGTM :shipit:

PR #2057, opened 2026-07-09
chore(core): refactor `ComponentRegistry` to use unified identifiers

tobz:
## Summary

This PR continues with our unification work from #2048, adjusting `ComponentRegistry` to also use `SubsystemIdentifier`.

We've accomplished this by actually refactoring `ComponentRegistry` such that it is no longer recursive: that is, callers don't hold on to an owned `ComponentRegistry` that may or may not be scoped/nested to some arbitrary depth within the overall hierarchy. In doing so, we've adjusted the relevant methods on `ComponentRegistry` to take `SubsystemIdentifier` and callers are expected to pass absolute identifiers, meaning that we move one step closer to everyone using the exact same identifier in all of the different places that an identifier is required.

Beyond that, we've just done some tidying of how we construct identifiers in ADP's environment provider code to utilize more of that nested pattern, where we derive child identifiers incrementally rather than recreating them completely everytime.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Existing unit tests.

## References

DADP-2

### DIFF
--- a/bin/agent-data-plane/src/internal/env/mod.rs
+++ b/bin/agent-data-plane/src/internal/env/mod.rs
@@ -4,6 +4,7 @@ use saluki_config::GenericConfiguration;
 use saluki_core::accounting::ComponentRegistry;
 use saluki_core::health::HealthRegistry;
 use saluki_core::runtime::Supervisor;
+use saluki_core::support::SubsystemIdentifier;
 use saluki_env::{
     autodiscovery::providers::BoxedAutodiscoveryProvider,
     host::providers::{BoxedHostProvider, FixedHostProvider},
@@ -68,17 +69,21 @@ impl ADPEnvironmentProvider {
         }
 
         // Otherwise, construct our real providers that will interact directly with the Datadog Agent.
-        let mut provider_component = component_registry.get_or_create("env_provider");
+        //
+        // Each provider's component registry node is created in a single call from the root registry using its
+        // canonical subsystem identifier, rather than by chaining relative `get_or_create` calls to synthesize the
+        // intermediate scopes. The tree de-duplicates the shared `env_provider` prefix.
         let mut env_supervisor = Supervisor::new("env-provider")?;
 
         let host_provider = RemoteAgentHostProvider::from_configuration(config).await?;
-        provider_component
+        component_registry
+            .get_or_create(&SubsystemIdentifier::from_segments(["env_provider"]))
             .bounds_builder()
             .with_subcomponent("host", &host_provider);
 
         let (workload_provider, workload_supervisor) = RemoteAgentWorkloadProvider::from_configuration(
             config,
-            provider_component.get_or_create("workload"),
+            component_registry.get_or_create(&workload::workload_root()),
             health_registry,
         )
         .await?;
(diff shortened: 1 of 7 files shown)


tobz:
* **#2057** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2057?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2057?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2048** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2048?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lucastemb (approved):
Everything looks good me. I think this is a really neat initiative 👍

bin/agent-data-plane/src/internal/env/mod.rs
```diff
@@ -106,10 +103,9 @@ impl ADPEnvironmentProvider {
     pub fn wait_for_ready(&self) -> impl Future<Output = ()> + Send + 'static {
         let health_registry = self.health_registry.clone();
         let has_workload_provider = self.workload_provider.is_some();
-        let workload_root = workload::workload_root();
         async move {
             if has_workload_provider {
-                health_registry.all_ready_under(workload_root).await;
+                health_registry.all_ready_under(root_provider_id()).await;
```
thieman, line 108: [GPT 5.6 Sol] `wait_for_ready` is documented to wait specifically for the workload provider’s metadata aggregator and collectors, but this broadens the readiness boundary from `env_provider.workload` to all strict descendants of `env_provider`. There are no other health-registered environment-provider siblings today, so this does not change current behavior, but a future host or autodiscovery health component under the same root would unexpectedly delay topology startup. Could we preserve the workload-only boundary here with `root_provider_id().child("workload")` (and ideally cover sibling exclusion in a focused test)?

tobz: This was intentional because when I was looking at it... we actually _do_ want the environment provider to be ready before starting the topology. Even with the workload provider, we're not waiting until we get all the tagger entities or whatever, we're just waiting until all of the relevant workers have deterministically started so that we know things _should_ be flowing.

It _should_ be the same across the entire environment provider and anything it spawns: make sure all relevant workers have started and are theoretically ready to do their thang before we proceed.

lib/saluki-core/src/accounting/registry.rs
```diff
@@ -154,75 +155,46 @@ impl ComponentMetadata {
 /// Every component is also able to be registered with its own resource group when using the tracking allocator
 /// implementation. This is done on demand when the component's token is requested, which avoids polluting the tracking
 /// allocator with components that are never actually used, such as those used for organizational/aesthetic purposes.
+#[derive(Clone)]
```
thieman, line 158: [GPT 5.6 Sol] Adding `Clone` here makes the later `ComponentRegistryHandle` documentation stale: it currently says that, unlike `ComponentRegistry`, the handle can be freely cloned and shared, but both types now share the same `Arc<Mutex<ComponentMetadata>>`. The meaningful distinction is capability—`ComponentRegistryHandle` exposes only read/report/verification operations—so could we reword that paragraph accordingly?

thieman (approved):

webern (approved):

PR #2079, opened 2026-07-10
fix(dogstatsd): decouple datagram reads from decoding and increase buffer limit to 256MiB

aqian01:
## Summary
[Issue](https://datadoghq.atlassian.net/browse/DADP-150)
Improve ADP's UDP and UDS datagram burst handling. Socket reads were coupled to decoding, and ADP's default elastic buffer ceiling was about 2 MiB and raised to match the Core Agent's default of roughly 256 MiB.

This PR:

- Queues connectionless socket reads in a bounded global buffer pool and decodes them in parallel.
- Uses a one-item acknowledged channel for connected streams, preserving order without building a backlog.
- Supports `dogstatsd_workers_count`. When unset or zero, ADP uses `max(vCPUs - 2, 2)`, matching the Core Agent's default formula. This count configures one global decoder worker pool shared by all connectionless UDP and UDS listeners. Connected streams continue to use one decoder per connection.
- Raises the elastic ceiling from 256 to 32,768 buffers of about 8 KiB each—approximately 2 MiB to 256 MiB—while keeping the initial allocation at 128 buffers.
- Shares the elastic buffer pool across connectionless listeners and connected readers. Connectionless datagrams wait in the bounded global queue, while connected streams use one-item acknowledged channels.
- Reclaims excess idle buffers after bursts while retaining a small warm pool.

### Original benchmark: baseline ADP

The original DADP-150 benchmark used the pre-PR ADP behavior: sequential decoding with the original ~2 MiB buffer ceiling. It ran for 180 seconds with 60 clients, estimated 32 MiB bursts, client aggregation, and background sending. It reported `bytes_dropped_writer / (bytes_dropped_writer + bytes_sent)`.

| Burst interval | Core datagram | Core stream | ADP datagram | ADP stream |
|---|---:|---:|---:|---:|
| 15s | 3.9% | 0.5% | 13.9% | 3.9% |
| 30s | 5.1% | 0.5% | 9.7% | 2.2% |
| 60s | 4.4% | 0.6% | 17.9% | 6.2% |

### Original benchmark: decoupled reads from decoding

The original benchmark was rerun on four CPUs against the Core Agent, baseline ADP, and an early PR build that decoupled socket reads from decoding but retained one decoder and the original ~2 MiB buffer ceiling. Datagram results were:

| Burst interval | Core Agent | Baseline ADP | Decoupled reads from decoding (~2 MiB) |
|---|---:|---:|---:|
| 15s | 11.4% | 59.8% | 52.7% |
| 30s | 9.2% | 51.5% | 48.5% |
| 60s | 14.2% | 52.8% | 51.6% |
| Unweighted mean | 11.6% | 54.7% | 50.9% |

So decoupling socket reads from decoding was an incremental change but did not close the gap between ADP and the agent

### Original benchmark: parallel decoding with larger buffers

The original benchmark was rerun on four CPUs against the Core Agent, baseline ADP, and a PR build that combined parallel decoding with the increased ~256 MiB buffer ceiling. Datagram results were:

| Burst interval | Core Agent | Baseline ADP | Parallel decoding + ~256 MiB |
|---|---:|---:|---:|
| 15s | 41.8% | 82.5% | 6.8% |
| 30s | 20.9% | 48.5% | 4.1% |
| 60s | 28.3% | 48.0% | 7.7% |
| Unweighted mean | 30.3% | 59.7% | 6.2% |

The larger buffer produced a substantial improvement in this run. However, five of the nine final counter snapshots were mathematically inconsistent, and the 15s PR result required one retry after the first run produced no telemetry. These results are directional evidence only.

### Why the benchmark changed

The benchmark was changed because:

- It was time-based, so contenders could complete different amounts of work. In the 30s case, [the original logs](https://github.com/user-attachments/files/30310804/dadp150-pr2079-benchmark-bundle.zip) captured 45,982,301 attempted bytes for ADP versus 36,695,095 for Core—25.3% more traffic during the same 180-second window. Attempted bytes are calculated as `bytes_sent + bytes_dropped_writer`.
- The [original collector and logs](https://github.com/user-attachments/files/30310804/dadp150-pr2079-benchmark-bundle.zip) queried counters sequentially while traffic continued, producing non-atomic snapshots. One sample reported `bytes_dropped=2,541,636` below `bytes_dropped_writer=2,762,874`, which violates [the counter definition](https://github.com/DataDog/datadogpy/blob/v0.52.1/datadog/dogstatsd/base.py#L1275-L1277).
- The [original collector](https://github.com/user-attachments/files/30310804/dadp150-pr2079-benchmark-bundle.zip) forcibly terminated workers without a drain step or verifying that their background queues were empty.
- Each case ran once, so variance was unknown.

### Fixed workload: parallel decoding with larger buffers

The revised benchmark tested parallel decoding with the buffer ceiling increased from ~2 MiB to ~256 MiB. It used a fixed workload instead of a fixed duration. Each contender received the same load: 60 clients sending three 32 MiB bursts over UDS datagrams, with four CPUs, client aggregation, and background sending. After the workload completed, client queues were drained before lifetime counters were collected.

The reported percentage is the share of attempted packet bytes that clients failed to send. These failures are consistent with receiver backpressure, although the harness does not capture the specific socket errors.

| Burst interval | Core Agent | Parallel decoding + ~256 MiB | Core elapsed | ADP elapsed |
|---|---:|---:|---:|---:|
| 15s | 18.1% | 8.7% | 43.7s | 42.6s |
| 30s | 8.6% | 2.5% | 70.7s | 70.6s |
| 60s | 8.7% | 2.6% | 130.5s | 130.2s |
| Unweighted mean | 11.8% | 4.6% | — | — |

With the fixed-workload harness, parallel decoding with a ~32 MiB ceiling averaged 39.8%. Increasing the ceiling to ~256 MiB reduced the average to 4.6%. Each case was run once, so variance remains unknown.

These fixed-workload results were produced at commit [`b0a95a65`](https://github.com/DataDog/saluki/commit/b0a95a65f253932e5f7f4741bd79d723e4424e05), before the [idle-buffer reclamation change](https://github.com/DataDog/saluki/commit/c04f48fb4ff74a8d51bb99da5f0e2660e69bfd6a). They are directional evidence, not direct validation of the current head.

[Download the benchmark scripts, configuration, pinned dependencies, logs, and JSON summaries](https://github.com/user-attachments/files/30310804/dadp150-pr2079-benchmark-bundle.zip).

## Memory trade-off

Decoupling socket reads from decoding lets ADP continue draining the socket instead of making senders wait for capacity. This moves more of the backlog into ADP's userspace buffer pool, trading higher memory usage for less backpressure and fewer packet drops during bursts.

The pool allocates 128 buffers initially and can grow on demand to 32,768 buffers, about 256 MiB of payload backing capacity. After five seconds without growth, it shrinks toward 256 idle buffers, so the full ceiling is not normally retained. The [performance report](https://github.com/DataDog/saluki/pull/2079#issuecomment-4937491711) shows the expected trade-off: medium-load RSS increased 22.31% to 83.3 MiB, above its 75 MiB bound, while low and idle loads remained within bounds.

To test whether reduced backpressure explains the higher RSS, three isolated three-minute `quality_gates_rss_dsd_medium` runs measured Lading's cumulative [`send().await` duration](https://github.com/DataDog/lading/blob/ed99f5f1329ce90511b579129afa15dd89570544/lading/src/generator/unix_datagram.rs#L283-L294). Lading waits for the UDS socket to accept the datagram, while the [original benchmark client](https://github.com/user-attachments/files/30310804/dadp150-pr2079-benchmark-bundle.zip) used 10 ms socket and queue timeouts that could turn backpressure into counted drops. Longer send duration therefore indicates more sender backpressure. Each run sent about 487,000 packets.

| Case | Buffer ceiling | Send duration | Average RSS | Average pool size |
|---|---:|---:|---:|---:|
| Baseline ADP (sequential decoding) | ~2 MiB | 5.91s | 58.9 MiB | ~1.0 MiB |
| Parallel decoding only | ~2 MiB | 5.34s | 61.3 MiB | ~2.0 MiB |
| Parallel decoding + larger buffer | ~256 MiB | 4.78s | 67.0 MiB | ~5.1 MiB |

Compared with baseline ADP, parallel decoding with the ~256 MiB ceiling spent 1.13 seconds less in `send().await` and used 8.0 MiB more RSS. Running the same parallel-decoding build with the original ~2 MiB ceiling increased send duration by 0.57 seconds and reduced RSS by 5.7 MiB. This supports the explanation that reduced socket backpressure shifts backlog into ADP memory. Parallel decoding alone remained 2.3 MiB above baseline, so other factors also contribute. Each case was run once, so variance is unknown.

## Correctness test harness

Decoupled datagram processing can leave packets [queued for separate decoder workers](https://github.com/DataDog/saluki/blob/278ff0c470a7c21773563e98bfdbb2299423e378/lib/saluki-components/src/sources/dogstatsd/mod.rs#L1455-L1472). Origin enrichment [uses the sender PID from UDS credentials](https://github.com/DataDog/saluki/blob/278ff0c470a7c21773563e98bfdbb2299423e378/lib/saluki-components/src/sources/dogstatsd/mod.rs#L2111-L2120), but the harness previously [exited Millstone before flushing and collecting](https://github.com/DataDog/saluki/blob/c04f48fb4ff74a8d51bb99da5f0e2660e69bfd6a/bin/correctness/panoramic/src/correctness/runner.rs#L438-L469). It now [reports completion while keeping Millstone alive](https://github.com/DataDog/saluki/blob/278ff0c470a7c21773563e98bfdbb2299423e378/bin/correctness/millstone/src/main.rs#L75-L105) until collection finishes. This prevents the sender-PID lifecycle race without changing the workload or assertions.

## Validation

- `make fmt`
- 139 DogStatsD tests passed
- `cargo check --workspace`
- `make check-all`


webern (commented):
This stood up to scrutiny...

Some interesting findings:

```
 Potential future performance work, only if profiling justifies it:
 - recvmmsg to receive multiple datagrams per syscall,
```

> Why run the new reader/channel/oneshot architecture for connection-oriented streams when it provides no pipelining? Is the added unbounded per-connection task/channel overhead acceptable, or should connection-oriented streams retain a specialized inline path?

aqian01:
[dadp150-pr2079-benchmark-bundle.zip](https://github.com/user-attachments/files/30310804/dadp150-pr2079-benchmark-bundle.zip)

Modified benchmark script

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1353,15 +1484,359 @@ fn origin_detection_failed_for_telemetry(
...
+                Err(error)
+            }
+        };
+
+        let receive_failed = result.is_err();
+        let queued = QueuedDatagram {
+            result,
+            socket_context: socket_context.clone(),
+        };
+        if datagram_sender.send(queued).await.is_err() {
```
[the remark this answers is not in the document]
aqian01: Resolved. The socket reader now resolves and pins the process origin before enqueueing, and decoder/capture paths use that pinned entity. Added a PID-reuse regression test.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1353,15 +1484,359 @@ fn origin_detection_failed_for_telemetry(
...
+    let mut buffer_manager = IoBufferManager::new(&io_buffer_pool);
+    loop {
+        select! {
+            _ = datagram_sender.closed() => break,
+            _ = memory_limiter.wait_for_capacity() => {}
+        }
+
+        let mut buffer = select! {
+            _ = datagram_sender.closed() => break,
+            buffer = buffer_manager.take_buffer() => buffer,
```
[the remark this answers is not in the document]
aqian01: With the default 32,768 buffers, this would require roughly 32,000 stalled connections so it's unlikely enough that I don’t consider it a blocker.

lib/datadog-agent/config-overlay-model/src/saluki_keys.rs
```diff
@@ -173,11 +173,11 @@ pub static SALUKI_KEYS: &[SalukiKey] = &[
     },
     SalukiKey {
         yaml_path: "dogstatsd_buffer_count_max",
-        description: "Max receive buffers",
-        default: "256",
+        description: "Max receive buffers (burst testing)",
```
tobz: I don't think we want to say `Burst testing` here... feels like a work-in-progress sort of comment?

lib/saluki-core/src/pooling/elastic.rs
```diff
@@ -17,28 +17,31 @@ use pin_project::pin_project;
...
 /// item is requested and the pool is empty, but hasn't yet reached its maximum size, it will allocate the item on
 /// demand.
 ///
-/// Periodically, a background task will evaluate the utilization of the pool and shrink the pool size in order to
-/// attempt to size it more closely to the recent demand. The frequency of this shrinking, as well as how usage demand
-/// is captured and rolled off, is configurable.
+/// After a grace period without growth, a background task shrinks idle capacity toward a retained-idle ceiling.
+/// Shrinking removes a fraction of the excess idle items each interval, and returned items above the ceiling are
+/// dropped immediately while shrinking is active. A new on-demand allocation disables shrinking for another grace
+/// period so nearby bursts can reuse the existing capacity.
```
tobz: I think the concept of the "retained idle capacity" being used here is more fragile/confusing than it needs to be? Conceptually, the main improvement here is that we're doing geometric decay on the excess capacity (to shrink more quickly than removing one item per shrinker loop tick) and that we pause shrinking when we think there might be additional acquisitions coming that would likely cause on-demand allocations.

We can do both of those things, from what I can tell, without this "retained idle capacity" setting... but I'm open to hearing on why it needs to exist.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -122,6 +121,16 @@ enum Error {
...
 const INTERNER_BASELINE_BYTES_PER_ENTRY: u64 = 512;
+const DOGSTATSD_LISTENER_WORKER_COUNT: usize = 1;
+const DOGSTATSD_PIPELINE_COUNT: usize = 1;
+const MIN_DOGSTATSD_WORKER_COUNT: usize = 2;
+const DOGSTATSD_RETAINED_IDLE_BUFFER_COUNT: usize = 256;
+
+fn default_decoder_worker_count(vcpus: usize) -> usize {
+    vcpus
+        .saturating_sub(DOGSTATSD_LISTENER_WORKER_COUNT + DOGSTATSD_PIPELINE_COUNT)
+        .max(MIN_DOGSTATSD_WORKER_COUNT)
```
tobz, line 129: Can we just add `.min(1)` at the end here?

I know we call it in a way that the value should never be zero, but it's an easy thing to do here just to be super sure.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1353,32 +1484,404 @@ fn origin_detection_failed_for_telemetry(
...
+
+struct BufferReturn(Option<oneshot::Sender<BytesBuffer>>);
+
+impl BufferReturn {
+    fn return_buffer(self, buffer: BytesBuffer) {
+        if let Some(sender) = self.0 {
+            let _ = sender.send(buffer);
+        }
+    }
+}
```
tobz: I feel like I'm not understanding _why_ we have this separate mechanism for returning buffers to the pool... the whole point of our object pooling design is that you _shouldn't_ need to roll your own mechanism for returning objects to object pools. 😅

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1353,32 +1484,404 @@ fn origin_detection_failed_for_telemetry(
...
+    capture_entity_resolver: Option<Arc<dyn CaptureEntityResolver + Send + Sync>>,
+    packets_tx: mpsc::Sender<io::Result<ReceivedBuffer>>,
+) {
+    debug!("Stream reader started.");
+
+    let mut buffer_manager = IoBufferManager::new(&io_buffer_pool);
+
+    loop {
+        select! {
+            _ = packets_tx.closed() => break,
```
tobz: These multiple `select!` blocks -- both here and down in `receive_connectionless_stream` -- are a code smell to me because it makes the code difficult to read through in a straight line.

It's totally OK if we accept a packet even though the downstream decoder task has gone away (`packets_tx.closed() == true`) because we can just check this when we send, and do our `break` when that happens.

One way or another, we either don't accept the packet (we bail out as soon as `packets_tx.closed()` returns `true`) and it gets dropped in the socket buffers with zero visibility... or we accept it even though we're going to fail to forward it to the decoder tasks, and we can at least log or otherwise track the fact that it's been dropped.

tobz (changes requested):

lib/datadog-agent/config-overlay-model/src/saluki_keys.rs
aqian01: Agreed. Updated it to `Maximum receive buffer count`.

lib/saluki-components/src/sources/dogstatsd/mod.rs
aqian01: I think you mean `.max(1)`?

```
6.min(1) = 1
0.min(1) = 0
6.max(1) = 6
0.max(1) = 1
```

So `const MIN_DOGSTATSD_WORKER_COUNT: usize = 2;` and
`.max(MIN_DOGSTATSD_WORKER_COUNT)` should prevent it from being 0

lib/saluki-components/src/sources/dogstatsd/mod.rs
aqian01: Oh, this [hands the buffer back from the decoder](https://github.com/DataDog/saluki/blob/07b2c2e131987b872b0a0c4a9155473f217b34c6/lib/saluki-components/src/sources/dogstatsd/mod.rs#L2024-L2026) to the [stream reader](https://github.com/DataDog/saluki/blob/07b2c2e131987b872b0a0c4a9155473f217b34c6/lib/saluki-components/src/sources/dogstatsd/mod.rs#L1590-L1593) to preserve partial data; it doesn’t return it to the pool. I renamed it to make that clear.

lib/saluki-components/src/sources/dogstatsd/mod.rs
tobz: Duh, yes, you're right and I was not thinking clearly. 😅

bin/correctness/millstone/src/main.rs
```diff
@@ -67,7 +72,39 @@ fn run() -> Result<(), GenericError> {
...
+            )
+        })?;
+        info!(
+            completion_file = %completion_file.display(),
+            "Millstone finished sending. Waiting for shutdown."
+        );
+
+        loop {
+            thread::park();
+        }
```
tobz: Maybe just do `thread::sleep(Duration::MAX)` instead to make the intent clearer?

Parking usually implies wanting to be able to be woken up at some later point, and it seems like we're using it here as a means to stay running until Panoramic intentionally shuts us down?

bin/correctness/panoramic/src/correctness/runner.rs
```diff
@@ -433,40 +467,47 @@ impl CorrectnessRunner {
...
-            )
-            .await;
-            return Err(generic_error!(
-                "Shared millstone exited with non-zero exit code ({}). Error: {}",
-                code,
-                error
-            ));
-        }
+        let mut millstone_exit = Box::pin(millstone_guard.millstone_handle.wait());
+        info!("Both Millstone senders completed. Keeping them alive while agents flush...");
```
tobz: Can we drop this log line?

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1353,343 +1476,699 @@ fn origin_detection_failed_for_telemetry(
...
+    } = handler_context;
+    let mut decoder = DogStatsDDecoder::new(source_context, decoder_context);
+    let mut framer = get_framer(&listen_addr, eol_required);
+    let mut stream_capture = StreamCaptureState::new();
+    // Set a buffer flush interval of 100ms, which will ensure we always flush buffered events at least every 100ms if
+    // we're otherwise idle and not receiving packets from the client.
+    let mut buffer_flush = interval(Duration::from_millis(100));
+    buffer_flush.set_missed_tick_behavior(MissedTickBehavior::Delay);
+
+    let mut last_process_origin = None;
```
tobz: I think you can just use `process_origin` directly in the below code... you never use `last_process_origin` for anything other than feeding into the buffer decode context.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1353,343 +1476,699 @@ fn origin_detection_failed_for_telemetry(
...
+                        last_process_origin = process_origin;
+                    }
+
+                    let outcome = decoder
+                        .decode_buffer(
+                            &mut framer,
+                            &mut stream_capture,
+                            &mut buffer,
+                            bytes_read,
+                            BufferDecodeContext {
```
tobz: I might try and rework `BufferDecodeContext` to hold the things that _don't_ change per-received buffer, so that you can change the call to `decode_buffer` to be more like `decode_buffer(&buf_decode_context, received_buffer)` which would allow for a lot of cleanup in terms of having to destructure everything here.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1353,343 +1476,699 @@ fn origin_detection_failed_for_telemetry(
...
+struct ReadBufferHandoff(Option<oneshot::Sender<BytesBuffer>>);
 
-    if !stream.is_connectionless() {
-        metrics.connections_active().increment(1);
+impl ReadBufferHandoff {
+    fn return_to_reader(self, buffer: BytesBuffer) {
+        if let Some(sender) = self.0 {
+            let _ = sender.send(buffer);
+        }
     }
```
tobz: You _could_ do something here where you implement `Drop` to automatically return the buffer to the reader if there's a handoff object provided.

You'd have to do something like:

- make the buffer field optional, and hold the handoff sender directly (still `Option`, as it is in `ReadBufferHandoff`)
- add a getter that returns a mutable reference to the buffer (and use `expect` since buffer should always be present; we just need `Option` to be able to move it during drop)
- add a `Drop` impl to `ReceivedBuffer` that checks if there's a handoff sender present, and if so, it `take`s the buffer and sends it

This gets you free returning of the buffer to the reader if the sender is present as soon as `ReceivedBuffer` goes out of scope, and _sending_ via a oneshot is synchronous/fast so we don't have to worry about any async-related behaviors.

This would fit in with my recommendation further down in the decoder logic that we could just pass `ReceivedBuffer` directly to `decode_buffer` instead of destructing all of the fields.

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1353,343 +1476,699 @@ fn origin_detection_failed_for_telemetry(
...
+        let process_origin = resolve_process_origin(capture_entity_resolver.as_deref(), &peer_addr);
+
+        let (buffer_sender, returned_buffer) = oneshot::channel();
+        let received = ReceivedBuffer {
+            buffer,
+            bytes_read,
+            peer_addr,
+            process_origin,
+            buffer_handoff: ReadBufferHandoff(Some(buffer_sender)),
+        };
```
tobz: Related to my above comment about about doing returning of buffers via a `Drop` impl on `ReceivedBuffer`, you could also simplify this code here by wrapping it up in a method like `ReceivedBuffer::with_return(buffer, bytes_read, peer_addr, process_origin)` which creates the oneshot sender/receiver pair for you, and gives you back `(received, returned_buffer)` here... and then add another method in the same style but named `without_return` that leaves `buffer_hander` empty, and just returns `ReceivedBuffer`, etc.

tobz (changes requested):

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -1112,36 +1145,119 @@ pub struct DogStatsD {
...
-    codec: DogStatsDCodec,
-    io_buffer_pool: ElasticObjectPool<BytesBuffer>,
+    eol_required: bool,
     metrics: Metrics,
+    packet_forwarder: Option<PacketForwarder>,
+}
+
+struct QueuedDatagram {
+    result: io::Result<ReceivedBuffer>,
+    socket_context: Arc<DatagramSocketContext>,
```
tobz, line 1193: I'm sort of wondering... is there any reason we can't just provide this up front to the decoder task?

From what I can see, I believe this only carries things that we know up front -- the listen address, metrics struct, etc -- at the time of creating the decoder tasks for a given listener... and so it's just costing us struct size and atomic operations to clone the `Arc<..>` by including it here in `QueuedDatagram`.

tobz: Answered out-of-band: decoder tasks service packets from potentially _all_ listeners, so they aren't actually fixed in the way I was thinking they are.

tobz (approved):

PR #2034, opened 2026-07-06
fix(datadog-agent-commons): add AIX platform settings

thieman:
## Summary

Adds AIX platform defaults to `datadog-agent-commons` so platform settings resolve when building ADP on AIX.

This is one of the Saluki-side blockers from DADP-142 / the AIX Confluence notes. After getting past `process-memory` and `stringtheory`, the AIX ADP build stopped in `datadog-agent-commons` because `platform/mod.rs` referenced platform constants/functions that are not defined for AIX.

## Key changes

- Add `platform/aix_impl.rs` with AIX config, log, and syslog defaults.
- Wire the AIX platform module into `platform/mod.rs`.
- Add an AIX-only unit test for the expected platform paths.

## AIX defaults

The AIX defaults match the standard Datadog Agent paths used in the AIX build notes:

- Config directory: `/etc/datadog-agent`
- Log directory: `/var/log/datadog`
- Local syslog URI: `unixgram:///dev/log`

## Continuation context

This PR is part of a three-PR Saluki-side AIX enablement stack:

1. #2031: `process-memory` AIX RSS support.
2. #2032: `stringtheory` big-endian `MetaString` fallback.
3. This PR: `datadog-agent-commons` AIX platform settings.

A separate upstream/fork fix is still needed for `aws-lc-rs`; during validation I used the local checkout of Travis's fork/branch on the AIX host. That temporary override is **not** included in this PR.

Temporary AIX validation setup used on `soaix499`:

```sh
export PATH=/opt/freeware/lib/RustSDK/1.92/bin:/opt/freeware/bin:/usr/sbin:/usr/bin:/bin
export CC=/opt/freeware/bin/gcc-8
export CXX=/opt/freeware/bin/g++-8
export AR=/usr/bin/ar
export RANLIB=/usr/bin/ranlib
export CARGO_HOME=$HOME/dd/cargo-home
export CARGO_TARGET_DIR=$HOME/dd/saluki-target
export CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
```

## Test plan

- RED on AIX before implementation:
  - Added the AIX-specific platform test.
  - With the proper AIX build env, `cargo test -p datadog-agent-commons platform::tests::aix_uses_standard_agent_paths --no-run` failed because AIX platform constants/functions were missing.
- Local:
  - `cargo fmt --package datadog-agent-commons --all`
  - `cargo check -p datadog-agent-commons`
  - `cargo test -p datadog-agent-commons platform::tests`
[...]

PR #2084, opened 2026-07-10
chore(agent-data-plane): add AIX build helper

thieman:
## Human Summary

Adds a script and Makefile target to build a release ADP binary on AIX. This comes with two intentional divergences from how we build on the other platforms:

- We can't use `cargo auditable` because it does not support XCOFF. Clanker tried to get it to work with some manual patching but wasn't able to. Seems like this will be a more significant contribution to upstream if we want this to work.
- We can't use link-time optimization (LTO) because it doesn't seem to work on AIX at all, even for a hello-world crate. This is actually called out in IBM's docs [here](https://www.ibm.com/docs/en/osfroa/1.90.0?topic=started-introducing-rustc-compiler).

Neither of these should be showstoppers, but please comment if they are. The plan will be to call this script from within the Agent's `packaging/aix` machinery when we do AIX builds. Working on that follow-up now.

## Summary

- Adds `make build-adp-aix` for native AIX ADP builds using the IBM Rust SDK and AIX Toolbox GCC toolchain.
- Adds `ci/tooling/build-adp-aix.sh` to record the AIX build environment and run the native Cargo build.
- Adds an `aix-optimized-release` Cargo profile that keeps the optimized-release binary settings that work on AIX while disabling LTO.
- Checks the major toolchain versions for cargo/rustc/gcc/g++ so AIX builds fail loudly if the toolchain drifts unexpectedly.

## Notes

- The AIX enablement and AWS-LC dependency changes from #2031, #2032, #2034, #2060, and #2082 are now on `main`; this PR is rebased on top of them and only carries the build helper/profile changes.
- `aws-lc-sys` is built from the bumped dependency without patching Cargo's registry cache.
- On the AIX test host, `gcc-8` fails in an AWS-LC compiler feature probe; `/opt/freeware/bin/gcc` (`gcc 13.3`) succeeds, so the helper defaults to that compiler.
- AIX currently does not use `cargo auditable`: cargo-auditable emits ELF/generic Unix linker inputs and flags that AIX's XCOFF linker rejects.
- AIX currently uses `aix-optimized-release` instead of `optimized-release`: the IBM Rust SDK 1.92 fails LTO even for a tiny hello-world crate with `failed to get bitcode from object file for LTO (Can't find section .ipa)`.
- The helper still uses the same metadata inputs that affect the ADP binary (`APP_FULL_NAME`, `APP_SHORT_NAME`, `APP_IDENTIFIER`, `APP_VERSION`, `APP_GIT_HASH`, `APP_BUILD_TIME`, `APP_DEV_BUILD`) and passes `BUILD_FEATURES` through to Cargo.
- Toolchain version prefixes are overrideable through `ADP_AIX_EXPECTED_*_PREFIX` variables when an intentional toolchain update occurs.

## Test Plan

- [x] `bash -n ci/tooling/build-adp-aix.sh`
- [x] `ADP_AIX_BUILD_DRY_RUN=true make build-adp-aix`
- [x] `cargo metadata --no-deps --format-version 1`
- [x] AIX: `CARGO_HOME=/opt/cargo-home-aix-adp-pr-stack CARGO_TARGET_DIR=/opt/saluki-aix-adp-pr-stack-target make build-adp-aix`
- [x] AIX: `/opt/saluki-aix-adp-pr-stack-target/aix-optimized-release/agent-data-plane --help`
- [x] AIX: `/opt/saluki-aix-adp-pr-stack-target/aix-optimized-release/agent-data-plane version`
- [x] Pre-commit hook: `fmt`, `check-clippy`, `check-licenses`, `check-deny`, `check-docs`, `generate-api-docs`

AIX toolchain observed:

- `cargo 1.92.0 (IBM Open SDK for Rust on AIX 1.92.0.0)`
- `rustc 1.92.0 (IBM Open SDK for Rust on AIX 1.92.0.0)`
- `gcc (GCC) 13.3.0`
- `g++ (GCC) 13.3.0`

AIX build result:

- `Finished aix-optimized-release profile [optimized + debuginfo] target(s) in 26m 22s`
- Binary: `64-bit XCOFF executable or object module not stripped`


ci/tooling/build-adp-aix.sh
```diff
@@ -0,0 +1,94 @@
...
+    echo "CARGO_HOME=${CARGO_HOME}"
+    echo "CARGO_TARGET_DIR=${CARGO_TARGET_DIR}"
+    echo "BUILD_PROFILE=${BUILD_PROFILE}"
+    echo "cargo build --profile ${BUILD_PROFILE} --bin agent-data-plane"
+}
+
+if [[ "${ADP_AIX_BUILD_DRY_RUN}" == "true" ]]; then
+    print_environment
+    exit 0
+fi
```
webern, line 109: This is definitely a nit, but I think a `--dry-run` arg is a little bit more canonical and visible, i.e. in the Makefile.

webern (approved):
Nice trailblazing.

ci/tooling/build-adp-aix.sh
thieman: Since this is mainly intended for use via `make` the CLI flag ends up being less ergonomic than env vars since it'd require a separate Make target or a way to forward around arguments. I don't expect anyone will be running this directly, should be via Make.

ci/tooling/test-build-adp-aix.sh
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env bash
```
thieman: This script doesn't seem like it's actually giving us much, can you remove it and its reference in the Makefile?

ci/tooling/build-adp-aix.sh
```diff
@@ -0,0 +1,141 @@
...
+
+cd "${repo_root}"
+
+print_environment
+
+echo "[*] Fetching Cargo dependencies..."
+cargo fetch
+
+echo "[*] Building agent-data-plane for AIX..."
+cargo build --profile "${BUILD_PROFILE}" --bin agent-data-plane --features "${BUILD_FEATURES}"
```
thieman, line 141: Add an explanatory comment above this line explaining why we aren't using `cargo auditable` the way the other builds are (doesn't support XCOFF)

Cargo.toml
```diff
@@ -277,6 +277,14 @@ overflow-checks = false
 lto = "fat"
 codegen-units = 1
 
+[profile.aix-optimized-release]
+inherits = "release"
+debug-assertions = false
+overflow-checks = false
+# IBM Open SDK for Rust 1.92 on AIX fails LTO during final linking with `Can't find section .ipa`.
```
thieman: Add a bit more information in this comment based on our investigation into these build failures

Cargo.toml
```diff
@@ -277,6 +277,15 @@ overflow-checks = false
 lto = "fat"
 codegen-units = 1
 
+[profile.aix-optimized-release]
+inherits = "release"
```
thieman: Make this inherit from optimized-release and then just keep the LTO override to false, if that works

thieman: [GPT-5.6 Sol] Updated in 96e35a238c. The AIX profile now inherits from `optimized-release` and only overrides `lto = false`, which keeps the shared optimized-release settings centralized while preserving the AIX LTO workaround. I validated the profile parses with `cargo metadata --no-deps --format-version 1`, and the pre-commit checks passed.

ci/tooling/test-build-adp-aix.sh
thieman: [GPT-5.6 Sol] Fixed in a3156d326b. Removed `ci/tooling/test-build-adp-aix.sh` and the `test-build-adp-aix-script` Makefile target; the dry-run path remains available manually with `ADP_AIX_BUILD_DRY_RUN=true make build-adp-aix`.

ci/tooling/build-adp-aix.sh
thieman: [GPT-5.6 Sol] Fixed in a3156d326b. Added the XCOFF/cargo-auditable comment above the AIX cargo build invocation to explain why this path uses plain `cargo build`.

Cargo.toml
thieman: [GPT-5.6 Sol] Fixed in a3156d326b. Expanded the LTO comment to include the IBM Rust SDK 1.92 failure mode and the minimal hello-world repro result (`Can't find section .ipa`).

PR #2086, opened 2026-07-10
feat(agent-data-plane): tag COAT telemetry with emitter

jszwedko:
## Summary

Ensure all continuity-of-agent-telemetry (COAT) that ADP exposes to the Datadog Agent carries an `emitter` tag identifying ADP as the source. The Agent's Remote Agent Registry only auto-attributes forwarded telemetry via `emitter` as a fallback (using the sanitized display name), and its merge logic that folds ADP's overlapping metrics into customer-facing `datadog.agent.*` telemetry requires the label to be present. Emitting the tag explicitly at the source gives ADP telemetry a stable, self-declared identity rather than depending on that fallback, and the value is derived from the app's full name using the same sanitization the Agent applies (lowercased, spaces → hyphens) so the two always agree.

The tag is injected at render time in the `saluki-core` telemetry processor (after grouping/dedup, so it does not affect how series merge) and applied to both COAT surfaces: the Remote Agent Registry `GetTelemetry` output and the `/compat/metrics` endpoint.

## Change Type

## How did you test this PR?

New unit tests: injected tags appear on both raw and remapped series (`saluki-core`); the `emitter_tag()` value matches the Agent-side sanitization; and end-to-end the RAR and compat COAT outputs both carry the emitter label. All pass.

### DIFF
--- a/bin/agent-data-plane/src/internal/remote_agent.rs
+++ b/bin/agent-data-plane/src/internal/remote_agent.rs
@@ -38,7 +38,7 @@ use tonic::{server::NamedService, Status};
 use tracing::{debug, error, info, warn};
 
 use crate::config::DataPlaneConfiguration;
-use crate::state::metrics::get_datadog_agent_remappings;
+use crate::state::metrics::{emitter_tag, get_datadog_agent_remappings};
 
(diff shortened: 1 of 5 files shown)


jszwedko:
Clanker went off the rails on this one. Meant to target the datadog-agent repo.

PR #1995, opened 2026-07-05
chore(config): cutover adp bin components to typed config system

webern:
## Human Summary

Cut over some of the `agent-data-plane` binary components to the new typed config system.

As it turns out, many of the `GenericConfiguration`-reading components and sites in `agent-data-plane` are difficult to cut over without further work. Many are involved in bootstrapping and remote agent set-up. Some are pretty complicated (#1998). But for those that seemed pretty reasonable to cut over to typed config, this PR does so.

## AI Summary

Continues the `agent-data-plane` cutover from the raw `GenericConfiguration` map to the typed
`SalukiConfiguration` model. Two components move over, plus a supporting consolidation of duration
parsing.

- **OTTL transform processor.** `OttlTransformConfiguration::from_configuration` now takes the
  resolved `domains::traces::OttlTransform` settings and copies them, instead of deserializing an
  `OttlTransformConfig` out of `GenericConfiguration`. The component-local config module (its
  `ErrorMode` enum and `OttlTransformConfig` struct) is removed in favor of the model's shared
  `OttlErrorMode`. This was the last raw-config consumer in
  `add_baseline_traces_pipeline_to_blueprint`, so its now-unused `GenericConfiguration` parameter is
  dropped. The `ottl_transform_config` seed source in `saluki_only.rs` is aligned with the OTTL
  filter's: `error_mode` deserializes directly into `OttlErrorMode` (replacing a loose
  `Option<String>` and a hand-rolled parse helper) and it rejects unknown fields.

- **Host tags.** `expected_tags_duration` is witnessed and already modeled at
  `shared.tags.expected_tags_duration`, so `HostTagsConfiguration` reads it directly from the typed
  slice rather than parsing a `DurationString` off `GenericConfiguration`; the hand-written zero
  default is dropped in favor of the generated schema default. `RemoteAgentClientConfiguration` is
  still shared by not-yet-migrated remote-agent components, so it stays on `GenericConfiguration`,
  built at the call site and passed in. This removes `GenericConfiguration` from the component
  entirely.

- **Duration parsing consolidation.** The viper/cast coercion that `saluki_config::DurationString`
  performed (trim whitespace, parse the Go duration grammar, else treat a bare integer as
  nanoseconds; reject negatives and values past the `time.Duration` `i64::MAX`-nanosecond bound) now
  lives once in the `go-duration` crate as `parse_viper_duration` plus checked numeric conversions.
  `DurationString` and the typed `DatadogConfiguration` duration deserializer both build on these, so
[...]

PR #2094, opened 2026-07-13
chore(config): integrate and test the config system

webern:
## Human Summary

Integrates the typed config system into `agent-data-plane` even though no components are yet reading from it. We decided to take a migration approach to this, so components will be cut over to using typed config in isolated PRs where we can focus on correctness for each component.

This PR introduces a privileged API endpoint `/config/internal` which is a serialization of `SalukiConfig`. This is going to be important for end-to-end integration testing, and the first such test is added in this PR.

Also noteworthy is that all integration and correctness tests are passing here, which was not the case prior to #2093's improvement of how the type config system handles environment variables.

## AI Summary

Integrate the typed configuration system into the Agent Data Plane runtime and expose its current value through the internal configuration endpoint.

- Load the configuration system after the initial configuration snapshot is available.
- Thread the configuration-system handle through runtime setup and the internal supervisor.
- Serve the current configuration through `/config/internal`.
- Use fallback environment-overlay behavior so Agent-provided values remain authoritative.
- Add integration coverage for runtime configuration updates through the internal endpoint.

## Change Type

## How did you test this PR?

- `make build-schema-overlay`
- `make fmt`
- `make check-docs`
- `make check-all`
- `make test`

All checks passed; 1,830 tests passed and 31 were skipped.

## References

Stacked on [#2093](https://github.com/DataDog/saluki/pull/2093).

### DIFF
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -88,7 +88,7 @@ pub async fn handle_run_command(
 
     // A static snapshot of the local sources, used only for decisions the Agent stream can never
     // change: whether we run standalone, and the remote-agent registration parameters.
-    let bootstrap = loaded.bootstrap();
+    let bootstrap = loaded.raw_config();
     let bootstrap_dp_config = DataPlaneConfiguration::from_configuration(&bootstrap)
         .error_context("Failed to load data plane configuration.")?;
     let standalone = bootstrap_dp_config.standalone_mode();
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -88,8 +88,7 @@ pub async fn handle_run_command(
 
     // A static snapshot of the local sources, used only for decisions the Agent stream can never
     // change: whether we run standalone, and the remote-agent registration parameters.
-    let bootstrap = loaded.raw_config();
-    let bootstrap_dp_config = DataPlaneConfiguration::from_configuration(&bootstrap)
+    let bootstrap_dp_config = DataPlaneConfiguration::from_configuration(&loaded.raw_config())
         .error_context("Failed to load data plane configuration.")?;
     let standalone = bootstrap_dp_config.standalone_mode();
 
@@ -105,7 +104,7 @@ pub async fn handle_run_command(
         (config_sys, None)
     } else {
         // Blocks until the Core Agent acknowledges registration.
-        let ra_bootstrap = RemoteAgentBootstrap::from_configuration(&bootstrap, &bootstrap_dp_config)
+        let ra_bootstrap = RemoteAgentBootstrap::from_configuration(&loaded.raw_config(), &bootstrap_dp_config)
             .await
             .error_context("Failed to bootstrap remote agent state.")?;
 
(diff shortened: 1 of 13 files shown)


webern:
* **#2094** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2094?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2094?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2093** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2093?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

webern:
> Static analysis only (build unavailable — `hyper-http-proxy` git dependency unreachable in sandbox). `DD_MEMORY_LIMIT=10485760` arrives at `ByteSize::from_str` as the bare string `"10485760"` (no unit suffix) via `EnvDecode::RawString`; if bytesize 2.4.0 rejects unitless integer strings, ADP startup fails whenever `memory_limit` is set as a plain byte count.
> 
> ADP startup aborts for any deployment that sets `DD_MEMORY_LIMIT` to a plain byte count rather than a suffixed string like `"10MB"`, which is a common scripting pattern.

@datadog-datadog-prod-us1 I'm 99.9% percent sure I fixed this in the preceding PR, #2093, which has been merged and this has been rebased to pick it up.

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -80,76 +78,78 @@ pub async fn handle_run_command(
...
+    // Load the local configuration sources (file + environment) once. The configuration system owns
+    // the loader wiring; we supply the file path and the environment precedence. Environment
+    // variables sit above the file so ADP-specific overrides (log level, etc.) win.
+    let loaded = LoadedConfiguration::load(&config_path, EnvPrecedence::AfterFile)
+        .await
+        .error_context("Failed to load configuration.")?;
+
+    // A static snapshot of the local sources, used only for decisions the Agent stream can never
+    // change: whether we run standalone, and the remote-agent registration parameters.
+    let bootstrap = loaded.bootstrap();
```
tobz: What's the difference between this method and `raw_map` that gets used down on line 129?

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -80,76 +78,78 @@ pub async fn handle_run_command(
...
-                    if let Err(e) = bootstrap_guard.logging_mut().reload(logging_config).await {
-                        warn!(
-                            error = %e,
-                            "Failed to reload logging from Agent configuration; continuing with bootstrap logging settings."
-                        );
-                    }
+    // Hand un-migrated components the resolved configuration as the legacy `GenericConfiguration`
+    // they still read from. Raw reads go through `config`; typed/live reads go through `config_sys`.
+    let config = config_sys.raw_map();
+    let dp_config = DataPlaneConfiguration::from_configuration(&config)
```
tobz: It might be advantageous to, both here and in the other spots we're touching in this file, change function signatures to take `&ConfigurationSystem` and then get a `GenericConfiguration` if needed, that way we start enforcing things go through `&ConfigurationSystem` from the get go, even if they're still on the compatibility path.

It would also clean up some of these signatures which are now taking 2-3 different "config" objects, which is getting somewhat clunky.

bin/agent-data-plane/src/cli/run.rs
webern: They are actually on two different types. I modeled the phases with two types `LoadedConfiguration` for the bootstrap phase and `ConfigurationSystem` for once it's running with the remote agent stream.

This renames the function to `raw_map` to match the other function so it's more clear what it is: https://github.com/DataDog/saluki/pull/2094/commits/f1ac5de8c5655267893212b84d91f27453c71e6e

bin/agent-data-plane/src/cli/run.rs
webern: I elimited a couple of the held `GenericConfiguration` variables and pushed `ConfigurationSystem` downward in a few places, but it didn't de-obfuscate things as much as we would like. Cutting this part of the code over to typed config is actually the hardest part of the whole project, I think.

I started trying to actually cut it over, because `DataPlaneConfiguration::from_configuration` would have to take a `LoadedConfiguration` in one phase and a `ConfigurationSystem` in the next, and then I back-tracked. I'm not quite ready to tackle that yet.

Anyway, this commit at least gets rid of the held/named `GenericConfiguration` variables: https://github.com/DataDog/saluki/pull/2094/changes/f8aabb772e11213c34bcb7cbe8911e80406df665

I will clean-up `run.rs` with typed config soon.

tobz (approved):

PR #2095, opened 2026-07-13
enhancement(core): add support for transient messages to dataspaces

tobz:
## Summary

This PR adds support for emitting "transient" messages to dataspaces.

Currently, dataspaces (aka `DataspaceRegistry`) allow for assertion/retraction of "facts": values that are set (and sometimes updated) over the lifecycle of a process but are always attached to that particular process from which they were asserted, automatically removed when the process stops. In some cases, however, we want to communicate values that can live beyond the lifecycle of a process without abusing the assertion/retraction pattern... otherwise, callers would have to try and figure out when they should pay attention to retractions vs when they don't need to, which is something we'd rather not do at all.

This PR adds support for emitting "transient" messages to dataspaces to support this lifecycle-less model of communication. In Syndicated Actors, where the concept of dataspaces originates from, this is actually an officially supported pattern, so really we're just bringing _our_ implementation of dataspaces up to parity. 😅 

In no specific order:

- we changed `AssertionUpdate` to `DataspaceUpdate` since there's now a new enum variant for messages that is unrelated to assertions
- there's a new method, `DataspaceRegistry::send`, for sending transient messages
- transient messages have the same properties -- strongly typed, carries an identifier -- as assertions, but aren't tied to a process
- messages come in alongside assertions/retractions on a given subscription
- messages are _only_ sent to active subscribers: they're never stored or replayed (hence "transient")

We're not optimizing for performance here since there's no use case (yet?) that requires this to act as a high-performance event bus.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

New and existing tests.

## References

DADP-2

### DIFF
--- a/lib/saluki-app/src/dynamic_api.rs
+++ b/lib/saluki-app/src/dynamic_api.rs
@@ -25,7 +25,7 @@ use rustls_pki_types::PrivatePkcs8KeyDer;
 use saluki_api::{APIHandler, DynamicRoute, EndpointProtocol, EndpointType};
 use saluki_common::{collections::FastIndexMap, sync::shutdown::ShutdownHandle};
 use saluki_core::runtime::{
-    state::{AssertionUpdate, DataspaceRegistry, Identifier, IdentifierFilter, Subscription},
+    state::{DataspaceRegistry, DataspaceUpdate, Identifier, IdentifierFilter, Subscription},
     InitializationError, Supervisable, SupervisorFuture,
 };
 use saluki_error::{generic_error, GenericError};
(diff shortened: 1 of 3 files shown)


tobz:
* **#2096** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2096?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2095** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2095?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2095?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

webern (approved):
Makes sense to me. 👍

PR #2098, opened 2026-07-13
fix(ci): handle patch releases in bump-adp-version workflow

jszwedko:
## Summary

The bump-adp-version workflow always bumped the minor version and targeted `main`, so patch releases cut from a `release/x.y.x` branch had no way to advance their own version. This teaches the workflow to distinguish the two cases from the triggering tag: minor releases (tag `x.y.0`) still bump the minor on `main`, while patch releases (tag `x.y.z` with `z > 0`) check out `release/x.y.x` and bump the patch segment there, opening the PR against that same release branch.

## Behavior

| Tag pushed | Target branch | Version bump |
| --- | --- | --- |
| `1.2.0` | `main` | `1.3.0` (minor) |
| `1.2.4` | `release/1.2.x` | `1.2.5` (patch) |

The obsolete-PR cleanup is now scoped by base branch so minor bumps on `main` and patch bumps on `release/*` no longer close each other. `workflow_dispatch` (no tag) keeps the prior minor-on-`main` behavior.

## Test plan

- [ ] Push a `x.y.0` tag and confirm a minor-bump PR opens against `main`
- [ ] Push a `x.y.z` (z>0) tag and confirm a patch-bump PR opens against `release/x.y.x` with the patch incremented
- [ ] Confirm concurrent minor and patch bump PRs do not close one another

### DIFF
--- a/.github/workflows/bump-adp-version.yml
+++ b/.github/workflows/bump-adp-version.yml
@@ -19,11 +19,38 @@ jobs:
           scope: DataDog/saluki
           policy: self.bump-adp-version.create-pr
 
+      - name: Determine release type and target branch
+        id: release-info
+        run: |
+          TAG_NAME="${{ github.ref_name }}"
+
+          if [[ "$TAG_NAME" =~ ^([0-9]+)\.([0-9]+)\.([0-9]+)$ ]]; then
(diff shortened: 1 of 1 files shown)


webern (approved):
cool

PR #2102, opened 2026-07-13
chore(ci): build ADP against glibc once again

tobz:
## Summary

This PR switches ADP to build against glibc instead of MUSL.

A long time ago, we made the decision to link against MUSL, instead of glibc, as our "libc" library. We did this because cleanly and repeatably building against an _old_ enough version of glibc (in keeping with the Datadog Agent's oldest supported operating system versions) was very difficult and fragile. Linking against MUSL, and statically linking in general, let us control our own destiny by not having to worry about using an insanely old build toolchain.

However, as we've started working ADP into the Datadog Agent releases, small issues have come up around our lack of glibc usage.... things like servers that do all of their DNS resolution through glibc NSS/SSSD magic, which totally bypasses normal DNS resolution semantics relied upon by MUSL or other userland DNS library implementations. Similarly, other concerns around the lack of dynamic linking at all have come up over time, namely over what it might mean to try and integrate non-Rust code into ADP.

Rather than continue to step on more landmines around glibc incompatibilities, or a lack of dynamic linking, we've taken another look at once again building ADP against glibc, and the _right_ version of glibc... and things look better these days, so we're switching back.

This PR, in a nutshell, does the following:

- updates our `build-ci` container image to be based off of `datadog-agent-buildimages/linux`, the container image used by the actual DDA CI for building the Agent on Linux
- removes all of our explicit usages of MUSL
- updated our build scripts to avoid the MUSL-specific hacks that were in place (kernel header copying, AArch64 "no outline atomics", etc)

There's still some follow-up cleanup we'll need to do, but in general, this is mostly solved by using the container image that Agent uses and just setting the right settings to tell Cargo to invoke things with the crosstool-NG toolchain so that the right glibc version gets used.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Ensured that the existing CI build jobs (AMD64/ARM64, FIPS vs non-FIPS) still built correctly and that they built against glibc, and the right version of glibc.

## References

DADP-2


ci/images/definitions/build/Dockerfile
```diff
@@ -1,4 +1,9 @@
-ARG BASE_IMAGE=registry.ddbuild.io/images/base/gbi-ubuntu_2204:release
+# We base our build image on the Datadog Agent's own Linux build image. It is Ubuntu-based
+# (`ubuntu:noble`) but ships crosstool-NG cross-toolchains under `/opt/toolchains/<arch>/bin` that
+# target an old glibc (x86_64 -> glibc 2.17, aarch64 -> glibc 2.23). ADP is built against those
+# toolchains (see docker/scripts/agent-data-plane/build/targets/*-unknown-linux-gnu.sh) so its glibc
+# floor matches the Agent's supported-OS matrix by construction, and tracks it automatically over time.
+ARG BASE_IMAGE=registry.ddbuild.io/ci/datadog-agent-buildimages/linux:v120029388-43933fee
```
jszwedko: Do we need new renovate config to keep this in-sync? Am I correct that https://github.com/DataDog/datadog-agent/blob/c2775884e0df412ab2c1086e34ca671dc34e54d2/.gitlab-ci.yml#L294-L316 is the source of truth?

tobz: That does seem to be the source of truth, yeah.

I'll update it to the latest version here, since clearly the clanker didn't grab the latest. 😢

.gitlab/benchmark.yml
```diff
@@ -45,7 +45,7 @@ build-adp-baseline-image:
       --file ./docker/Dockerfile.agent-data-plane
       --tag ${BASELINE_ADP_IMG}
       --build-arg BUILD_PROFILE=release
-      --build-arg BUILD_TARGET=x86_64-unknown-linux-musl
+      --build-arg BUILD_TARGET=x86_64-unknown-linux-gnu
```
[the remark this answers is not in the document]
tobz: Not wrong, but unfortunately nothing we can do about that here without CI getting very, very messy.

jszwedko:
Ran a manual verification to check the min glibc version:

amd64
```
root@f9a719680cf1:/#  objdump -T /usr/local/bin/agent-data-plane | grep GLIBC | sed 's/.*GLIBC_\([.0-9]*\).*/\1/g' | sort -Vu
2.2.5
2.3
2.3.2
2.3.4
2.4
2.6
2.7
2.9
2.10
2.12
2.14
2.16
2.17
```

arm64

```
root@594c77fe9a2e:/# objdump -T /usr/local/bin/agent-data-plane | grep GLIBC | sed 's/.*GLIBC_\([.0-9]*\).*/\1/g' | sort -Vu
2.17
2.18
```

docker/scripts/agent-data-plane/build/20-check-glibc-floor.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env sh
```
jszwedko, line 1: Big ➕  to this check. I remember we often accidentally broke glibc compatibility on Vector.

docker/scripts/agent-data-plane/build/targets/aarch64-unknown-linux-gnu.sh
```diff
@@ -0,0 +1,22 @@
...
+# the cross-compilation toolchain.
+#
+# This ensures that we use the right tool versions, glibc versions, and so on.
+if [ -d /opt/toolchains/aarch64 ]; then
+    _ctng=/opt/toolchains/aarch64/bin/aarch64-unknown-linux-gnu
+    export CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="${_ctng}-gcc"
+    export CC_aarch64_unknown_linux_gnu="${_ctng}-gcc"
+    export CXX_aarch64_unknown_linux_gnu="${_ctng}-g++"
+    export AR_aarch64_unknown_linux_gnu="${_ctng}-ar"
+    TARGET_MAX_GLIBC="2.23"
```
jszwedko: Is this the right glibc max? Looking at the Agent docker image, at least, the Agent seems to only depend on glibc 2.17. Also, ADP seems to only use 2.18:

```
root@594c77fe9a2e:/# objdump -T /usr/local/bin/agent-data-plane | grep GLIBC | sed 's/.*GLIBC_\([.0-9]*\).*/\1/g' | sort -Vu
2.17
2.18
```

jszwedko (commented):
One comment about the floor for aarch64, otherwise 👍

docker/scripts/agent-data-plane/build/targets/aarch64-unknown-linux-gnu.sh
tobz: This is the right one I believe, yes.

Based on our [supported platforms matrix for ARMv8](https://docs.datadoghq.com/agent/supported_platforms/?tab=linux#64-bit-arm-v8) (aka ARM64), the minimum glibc version jumps up to 2.23.

jszwedko: Tested with the aarch64 deb package too: the Agent seems to only depend on 2.17.

tobz: As far as the 2.18 thing... you have to separate out weak symbols. The "floor" is the highest glibc version detected for non-weak symbols, which are the symbols that, when missing, will cause the process to panic.

Weak symbols are basically the spiritual equivalent of `Option<T>`.

jszwedko (approved):

docker/scripts/agent-data-plane/build/targets/aarch64-unknown-linux-gnu.sh
tobz: For posterity: we opted to set the floor at 2.17 for both AMD64 and ARM64 since they're actually _building_ that way currently. We'll opt to raise the ARM64 side back up to 2.23 after chatting with folks internally: the supported platforms aspect is correct, but we might actually want to expand it further back if possible.

PR #2104, opened 2026-07-13
fix(otlp): align cumulative histogram sketch handling

lucastemb:
…nd normalized buckets to reconcile u16 and u32 bucket size difference

## Summary

Implemented fixes to match Core Agent behavior for OTLP metrics. 

Previously, cumulative aggregation temporality incorrectly calculated local minima for histograms. 

Additionally, DDSketch bins in Saluki are of size `u32`, whereas in the Core Agent, they are `u16`. To reconcile differences that emerge during testing, there needed to be a way to calculate the bins to ensure that the data was being processed _effectively_ the same. The Core Agent, breaks up data overflow into separate buckets, so to rectify this discrepancy, buckets needed to be combined before comparison. 

The `.yaml` config files we use for OTLP metrics testing also needed to be adjusted since it was recently discovered that the batching mechanism in the Core Agent, makes it such that the payloads received from millstone are not processed sequentially, which led to potential cache entries being erroneously dropped. 

The fix was to increase the minimum batch size so that all payloads would all be processed in a singular batch, and thus, sequentially. 

Lastly, the millstone volume needed to be reduced to 1000 to prevent the Core Agent from dropping payloads, which would affect direct comparison, since the two systems handle backpressure differently. 

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Integration tests and unit tests

## References

- Related Lading PR: https://github.com/DataDog/lading/pull/1908


lib/saluki-components/src/encoders/datadog/metrics/v2/mod.rs
```diff
@@ -485,8 +507,7 @@ fn write_dogsketch(
     output_stream: &mut CodedOutputStream<'_>, scratch_buf: &mut Vec<u8>, packed_scratch_buf: &mut Vec<u8>,
     timestamp: Option<NonZeroU64>, sketch: &DDSketch,
 ) -> Result<(), protobuf::Error> {
-    // If the sketch is empty, we don't write it out.
-    if sketch.is_empty() {
+    if !sketch_has_emittable_values(sketch) {
```
webern, line 488: I noticed v2 in the path and wondered if v3 needs the same treatment. Also, if it does, what is the code reuse story between v2 and v3 for something like this?

AI writes:

> Should we apply this same emittability and stored-summary handling to the v3 serializer as well? `write_metric_to_v3` in `encoders/datadog/metrics/mod.rs` still gates distributions on `sketch.is_empty()`, so a zero-count sketch that retains bins—the cumulative-to-delta case addressed here—will be emitted by v2 but dropped when v3 sketches are enabled. If that guard is relaxed in v3, it should also use `stored_min()`, `stored_max()`, and `stored_sum()` rather than the `unwrap_or(0.0)` accessors, so the two serializers preserve the same sketch data. A v2/v3 parity test for a zero-count sketch with retained bins would help lock this down.

lib/saluki-components/src/encoders/datadog/metrics/v2/mod.rs
```diff
@@ -254,6 +254,11 @@ fn encode_single_metric(
     secondary_scratch_buf: &mut Vec<u8>, packed_scratch_buf: &mut Vec<u8>,
     tags_deduplicator: &mut ReusableDeduplicator<Tag>,
 ) -> Result<(), protobuf::Error> {
+    if !metric_has_emittable_values(metric) {
```
webern: I was probing for performance impacts and came up with this, though I believe it is a low priority finding:

> Could we consolidate this check with the existing `v1::has_emittable_point` encoder gate instead of scanning here? That gate already filters scalar series metrics, so this adds a second `.any()` traversal for every series metric; only the distribution/histogram arms need the new retained-bin/sample semantics. Extending the shared gate to those arms and keeping the per-sketch guard would preserve the correctness behavior while removing the redundant work.
>
> Also, could this use `debug!` and increment `events_dropped_encoder()` like the existing drop path? A producer emitting many non-emittable metrics could flood warning logs, and these drops currently aren’t reflected in telemetry.

webern (commented):
The v3 finding seems important. The performance finding seems like a bit of a nit. Otherwise this is looking good and correct based on my assisted perusal.

The title, which becomes a commit message in the repo could use a tweak.

lib/ddsketch/src/agent/sketch.rs
```diff
@@ -478,15 +518,21 @@ impl DDSketch {
...
             let mut upper = bucket.upper_limit;
             if upper.is_sign_positive() && upper.is_infinite() {
                 upper = lower;
             } else if lower.is_sign_negative() && lower.is_infinite() {
                 lower = upper;
+            } else if lower == 0.0 && upper > 0.0 {
+                // OpenTelemetry explicit buckets use (lower, upper]. After a zero boundary, the
+                // next bucket is (0, upper] and excludes zero. Collapse interpolation to `upper`
+                // so it does not add samples at zero.
+                lower = upper;
```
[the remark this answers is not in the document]
lucastemb: This code mimics Core Agent behavior.

lib/saluki-components/src/encoders/datadog/metrics/v2/mod.rs
lucastemb: Just added proper V3 handling. I wrote it down as something that I needed to address, but it must've slipped my mind, nice catch 🫡

webern (approved):

PR #2106, opened 2026-07-13
feat(dev): introduce Agent version detection for version-gated behavior

jszwedko:
## Summary

Introduces a minimal Agent version gate in `datadog-agent-commons` so ADP can conditionally reproduce version-specific Agent behavior matching the Agent it is bundled with. This is the infrastructure PR; the first consumer is in the stacked OTLP PR.

`DD_AGENT_VERSION` is baked into the converged image at build time (release tag for regular CI, `"nightly"` for nightly CI) and read by ADP at startup via `agent_version::version()`. Any component calls `agent_version::meets(major, minor, patch)` to gate behavior. When the variable is unset (standalone / local dev), gates default to the most recent behavior.

See example usage in the stacked PR: https://github.com/DataDog/saluki/pull/2099

## Test plan

- [x] 7 unit tests for `AgentVersion::parse` / `meets` covering release tags, dev builds, bare dev markers, and edge cases


lib/datadog-agent/commons/src/agent_version.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+}
+
+/// Returns the Agent version, if known.
+///
+/// The value is read once from the [`DD_AGENT_VERSION_ENV_VAR`] environment variable and cached. Returns `None` if the
+/// variable is unset or cannot be parsed.
+pub fn version() -> Option<AgentVersion> {
+    static VERSION: OnceLock<Option<AgentVersion>> = OnceLock::new();
+    *VERSION.get_or_init(|| {
```
thieman: I'm not understanding how this gets baked in during the build, is that existing code somewhere else?

jszwedko: Doh, yes, this isn't right. It should be happening at build time, not runtime.

jszwedko: Moved to build time in [8044809](https://github.com/DataDog/saluki/pull/2106/commits/804480938bdc5609f030d6ec659d9afb251e98bb).

thieman (approved):

webern (approved):
I'm concerned about complexity as we use this in our code, but detecting this at build time does seem like a necessary step for us.

jszwedko:
> I'm concerned about complexity as we use this in our code, but detecting this at build time does seem like a necessary step for us.

@webern Yeah, I'm also a bit concerned about the complexity. We can remove the usages, at least, as each Agent release is published (e.g. we can remove the usages here once 7.82.0 is published and ADP's CI updated).

PR #2121, opened 2026-07-14
chore(io): remove Hickory DNS in favor of glibc/GAI

### LINKED ISSUE #2115: hickory DNS resolver doesn't respect NSS/etc/hosts for intake failover
## Problem

When testing ADP forwarder behavior with DNS failover scenarios, the agent's core forwarder correctly honors NSS-based hostname resolution (via `/etc/nsswitch.conf` and `/etc/hosts`), but **ADP's forwarder does not** because it uses hickory (a pure-Rust DNS resolver) which bypasses the system NSS stack entirely.

## Observed Behavior

Test: `test/new-e2e/tests/agent-runtimes/forwarder_nss_failover_test.go::TestNSSFailover`

1. Agent configured to forward to `ddintake` (aliased in `/etc/hosts` to fakeintake1's IP).
2. Agent successfully sends metrics/logs/traces to fakeintake1.
3. `/etc/hosts` entry is updated: `ddintake` now points to fakeintake2's IP.
4. Core agent's forwarder detects the change (via `connection_reset_interval`) and switches to fakeintake2. ✅
5. **ADP's forwarder keeps resolving `ddintake` to fakeintake1's IP and continues sending metrics there.** ❌

Result: Old intake continues receiving metrics after failover, breaking assumptions about deterministic failover behavior.

## Root Cause

ADP's forwarder uses hickory for DNS resolution, which:
- Performs its own pure-Rust DNS queries
- Does **not** consult `/etc/nsswitch.conf` 
- Does **not** respect `/etc/hosts` entries
- Caches resolved IPs independently from system NSS

## Impact

Deployments relying on NSS-based DNS failover (e.g., changing `/etc/hosts` or nsswitch config to move intake traffic) will have incomplete failover with ADP enabled — core agent metrics move to the new intake, but ADP metrics lag behind or stay on the stale endpoint.

## Fix Direction

Either:
1. Have hickory query the system NSS stack (if available as a resolver option), or
2. Implement a custom hickory resolver provider that checks `/etc/hosts` before DNS queries, or
3. Document that NSS-based failover is not supported with ADP and recommend DNS CNAME/A record updates instead

Ref: DADP-72 (full e2e suite run with ADP enabled discovered this incompatibility)

tobz:
## Summary

This PR removes Hickory DNS in favor of glibc/GAI.

After cutting back to glibc over MUSL in #2102, we want to follow-up by also removing Hickory DNS such that we revert back completely to relying on glibc for DNS resolution. This ensures that any glibc/NSS black magic that is configured on a given system works properly when ADP is used the same as it would when operations are happening in the Datadog Agent directly.

We've kept our own resolver implementation but simply forward to `hyper`'s `GaiResolver` under the hood. This leaves the necessary scaffolding in place to, in the future, change how we spawn the underlying tasks that run the actual resolve operations.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Existing unit and integration tests.

## References

DADP-2

Closes #2115

### DIFF
--- a/bin/agent-data-plane/Cargo.toml
+++ b/bin/agent-data-plane/Cargo.toml
@@ -63,7 +63,7 @@ tokio = { workspace = true, features = [
 tokio-util = { workspace = true }
 tonic = { workspace = true }
 tracing = { workspace = true }
-uuid = { workspace = true }
+uuid = { workspace = true, features = ["v4"] }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 tokio = { workspace = true, features = ["taskdump"] }
--- a/lib/saluki-io/src/net/client/http/conn.rs
+++ b/lib/saluki-io/src/net/client/http/conn.rs
@@ -8,7 +8,6 @@ use std::{
 #[cfg(unix)]
 use std::{path::PathBuf, sync::Arc};
 
-use hickory_resolver::net::NetError;
 use http::{Extensions, Uri};
 use hyper_rustls::{HttpsConnector, HttpsConnectorBuilder, MaybeHttpsStream};
 use hyper_util::{
@@ -18,7 +17,7 @@ use hyper_util::{
 use metrics::Counter;
 use pin_project_lite::pin_project;
 use rustls::ClientConfig;
-use saluki_error::{ErrorContext as _, GenericError};
+use saluki_error::GenericError;
 use tokio::net::TcpStream;
 #[cfg(target_os = "linux")]
 use tokio_vsock::{VsockAddr, VsockStream};
@@ -26,7 +25,7 @@ use tower::{BoxError, Service};
 use tracing::debug;
 
 use super::telemetry::HttpTransactionErrorTelemetry;
-use crate::net::dns::{HickoryHttpConnector, HickoryResolver};
+use crate::net::dns::{DnsError, SystemHttpConnector, SystemResolver};
 
 /// Imposes a limit on the age of a connection.
 ///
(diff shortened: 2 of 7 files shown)


lib/saluki-io/src/net/client/http/conn.rs
```diff
@@ -506,28 +502,9 @@ impl HttpsCapableConnectorBuilder {
...
-        #[cfg(unix)]
-        let unix_socket_only = self.unix_socket_path.is_some();
-        #[cfg(not(unix))]
-        let unix_socket_only = false;
-
-        let hickory_resolver = if vsock_only || unix_socket_only {
-            HickoryResolver::noop()
-        } else {
-            build_dns_resolver(&self.error_telemetry)?
-        };
```
jszwedko, line 526: 👍 I see that `build_dns_resolver` no longer returns an error so I'm assuming there is no potential error at construction time but only at runtime.

bin/agent-data-plane/Cargo.toml
```diff
@@ -63,7 +63,7 @@ tokio = { workspace = true, features = [
 tokio-util = { workspace = true }
 tonic = { workspace = true }
 tracing = { workspace = true }
-uuid = { workspace = true }
+uuid = { workspace = true, features = ["v4"] }
```
jszwedko, line 66: Why is this required? Was hickory pulling in v4 for us through feature unification?

jszwedko (approved):

lib/saluki-io/src/net/client/http/conn.rs
tobz: Correct.

bin/agent-data-plane/Cargo.toml
tobz: Exactly, yep.

PR #2175, opened 2026-07-21
fix(millstone): retry initial gRPC connections

thieman:
## Summary

- retry Millstone's initial gRPC channel connection with bounded exponential backoff
- use the workspace-standard `backon` retry executor for both initial connections and transient payload failures
- use six total attempts with exact delays of 100, 200, 400, 800, and 1600 milliseconds
- preserve the final Tonic transport error source chain when retries are exhausted
- add regression coverage for the bounded retry schedule and final error propagation

## Why

The correctness harness can consider the Agent container healthy before its DDOT gRPC receiver begins listening. Millstone previously made one eager connection attempt and exited immediately when it lost that startup race, as seen in [job 1876833566](https://gitlab.ddbuild.io/DataDog/saluki/-/jobs/1876833566).

## Test plan

- `cargo test -p millstone`
- `cargo clippy -p millstone --all-targets -- -D warnings`
- `cargo check --workspace`
- `make check-fmt`
- signed-commit pre-commit checks

`make check-all` reached `check-features` but could not build the AWS-LC FIPS variants because `cmake` is not installed in the local environment.

### DIFF
--- a/bin/correctness/millstone/src/target.rs
+++ b/bin/correctness/millstone/src/target.rs
@@ -2,6 +2,7 @@
 use std::os::unix::net::{UnixDatagram, UnixStream};
 use std::{
     fs::File,
+    future::Future,
     io::Write as _,
     net::{Ipv4Addr, TcpStream, UdpSocket},
     path::Path,
@@ -14,6 +15,9 @@ use tonic::{client::Grpc, transport::Channel};
 use tracing::warn;
 
 use crate::config::{Config, TargetAddress};
+
+const MAX_RETRIES: u32 = 5;
+const BASE_RETRY_DELAY_MS: u64 = 100;
 
 enum TargetBackend {
     Tcp(TcpStream),
@@ -186,14 +190,11 @@ async fn try_grpc_unary(channel: Channel, service_method_path: &str, payload: &[
 fn send_grpc_payload(
     runtime: &tokio::runtime::Runtime, channel: Channel, service_method_path: &str, payload: &[u8],
 ) -> Result<(), GenericError> {
-    const MAX_RETRIES: u32 = 5;
-    const BASE_DELAY_MS: u64 = 100;
-
     let mut last_status: Option<tonic::Status> = None;
 
     for attempt in 0..=MAX_RETRIES {
         if let Some(ref status) = last_status {
-            let delay_ms = BASE_DELAY_MS * (1u64 << attempt.saturating_sub(1));
+            let delay_ms = retry_delay_ms(attempt);
             warn!(
                 attempt,
                 delay_ms,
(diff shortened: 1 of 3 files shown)


bin/correctness/millstone/src/target.rs
```diff
@@ -221,6 +222,40 @@ fn send_grpc_payload(
     ))
 }
 
+const fn retry_delay_ms(attempt: u32) -> u64 {
+    BASE_RETRY_DELAY_MS * (1u64 << attempt.saturating_sub(1))
+}
+
+async fn retry_with_backoff<T, E, F, Fut>(operation_name: &str, mut operation: F) -> Result<T, E>
```
thieman: Do we need to add this machinery? Is there not an existing retry lib or helper we can use?

thieman: [GPT-5.6 Sol (OpenAI)] Replaced the custom retry helper and the existing payload retry loop with the workspace-standard `backon` executor in b50fed89a5. Both paths now share the same exact bounded delay schedule.

bin/correctness/millstone/src/target.rs
```diff
@@ -15,6 +16,9 @@ use tracing::warn;
 
 use crate::config::{Config, TargetAddress};
 
+const MAX_RETRIES: u32 = 5;
+const BASE_RETRY_DELAY_MS: u64 = 100;
```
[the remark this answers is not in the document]
thieman: [GPT-5.6 Sol (OpenAI)] Addressed in 496f1f9140. The production wrapper still uses the full bounded schedule, while the final-error test injects an empty `backon` backoff and now completes in 0.00s instead of roughly 3.1s.

bin/correctness/millstone/src/target.rs
```diff
@@ -306,3 +341,44 @@ impl tonic::codec::Decoder for NoopDecoder {
...
+        let error = match create_grpc_client("127.0.0.1:0/test.Service/Call") {
+            Ok(_) => panic!("connection should fail"),
+            Err(error) => error,
+        };
+        let error_chain = error.chain().map(ToString::to_string).collect::<Vec<_>>();
+
+        assert!(
+            error_chain.iter().any(|cause| cause == "tcp connect error"),
+            "expected TCP connection failure in error chain, got: {error_chain:?}"
+        );
```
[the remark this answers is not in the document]
thieman: [GPT-5.6 Sol (OpenAI)] Addressed in 496f1f9140. The assertion now uses substring matching for `tcp connect error`, preserving the intended source-chain check without depending on exact platform or Tonic formatting.

bin/correctness/millstone/src/target.rs
```diff
@@ -8,13 +8,23 @@ use std::{
...
 
 use crate::config::{Config, TargetAddress};
 
+const GRPC_RETRY_DELAYS: [Duration; 5] = [
+    Duration::from_millis(100),
+    Duration::from_millis(200),
+    Duration::from_millis(400),
+    Duration::from_millis(800),
+    Duration::from_millis(1600),
+];
```
thieman: Does backon give us a less verbose way to define these two constants?

thieman: [GPT-5.6 Sol (OpenAI)] Yes. Simplified in f721e17624 using `backon::ExponentialBuilder` with a 100ms minimum, 1600ms maximum, and five retries. This removes the explicit delay array and retains the intended schedule at millisecond precision.

lucastemb (approved):

PR #2178, opened 2026-07-21
feat(agent-data-plane): emit forwarder retry COAT metrics

thieman:
## Summary

- emit INFO-tier forwarder retry and requeue counters with Core Agent-compatible queue-dispatch semantics
- preserve retry-queue provenance across overflow and persisted entries, while distinguishing actual dispatch from blocked requeue
- export both counters through RAR as `transactions.retries` and `transactions.requeued` with matching HELP text
- keep Cluster Agent logical endpoint telemetry stable across its request URI rewrite
- consolidate retry tests around observable behavior, retaining distinct overflow, blocked-requeue, and end-to-end contracts
- avoid PR-added steady-state allocations by using borrowed endpoint lookup and cached counter handles instead of request extensions

## Semantics

- `transactions.retries` increments when an item popped from the retry queue completes an attempt that crossed the circuit breaker into the HTTP service
- `transactions.requeued` increments when such an item is rejected before inward dispatch and is successfully returned to the retry queue
- initial failed sends do not increment either counter; high-priority overflow counts when later dispatched from the retry queue

## Allocation behavior

- warmed incoming paths use borrowed endpoint lookups and existing metric handles
- warmed retry paths clone cached `Counter` handles, which only update `Arc` reference counts
- metric labels and cache keys allocate only on first observation of a logical endpoint
- existing request cloning, retry-queue storage, Tokio task, and boxed backoff allocations are unchanged

## Test plan

- [x] `make fmt`
- [x] `cargo nextest run -p saluki-io retry_circuit_breaker` (4 passed)
- [x] `cargo nextest run -p saluki-components 'common::datadog::io::tests'` (23 passed after merging current `main`)
- [x] `cargo nextest run -p saluki-components 'common::datadog::telemetry::tests'` (4 passed)
- [x] `cargo nextest run -p saluki-components 'forwarders::cluster_agent::tests'` (3 passed)
- [x] `cargo nextest run -p agent-data-plane 'state::metrics::tests'` (8 passed)
- [ ] `make check-all` — local FIPS feature checks require `cmake`, which is not installed in this environment; signed-commit pre-commit checks passed

## Follow-up dependency

The Datadog Agent profile change is intentionally separate. Current Agent `main` does not yet inject `emitter:agent` for local metrics that lack an incoming emitter, so adding `preserve_tags: [emitter]` now would filter out the Core Agent's own retry series. Equivalent emitter-default behavior must land before the profile is updated.

## References

- DADP-155


lib/saluki-components/src/forwarders/cluster_agent/mod.rs
```diff
@@ -191,11 +191,8 @@ fn cluster_agent_request_mapper<B>(
     })
 }
 
-fn get_cluster_agent_endpoint_name(uri: &Uri) -> Option<MetaString> {
-    match uri.path() {
-        CLUSTER_AGENT_SERIES_PATH => Some(MetaString::from_static("cluster_agent_series")),
-        _ => None,
-    }
+fn get_cluster_agent_endpoint_name(_uri: &Uri) -> Option<MetaString> {
+    Some(MetaString::from_static("cluster_agent_series"))
```
thieman, line 195: Explain this change?

lib/saluki-components/src/common/datadog/telemetry.rs
```diff
@@ -26,6 +28,60 @@ impl TransactionInputTelemetry {
     }
 }
 
+#[derive(Clone)]
+pub(super) struct TransactionRetryTelemetry {
+    builder: MetricsBuilder,
+    domain: MetaString,
+    counters_by_endpoint: Arc<Mutex<FastHashMap<MetaString, TransactionRetryCounters>>>,
```
thieman: It seems ilke `TransactionInputTelemetry` directly above isn't synchronized, is the Mutex in this one necessary? If yes, could we get rid of it by splitting into two struct members rather than having a single map?

lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -567,15 +656,13 @@ async fn run_endpoint_io_loop<B>(
                         strip_metrics_validation_headers(txn)
                     };
                     let transaction_size = txn.size_bytes();
-                    let transaction_endpoint_name = endpoint_name(txn.request_uri());
-                    let transaction_endpoint_name = transaction_endpoint_name
-                        .as_deref()
-                        .unwrap_or_else(|| txn.request_uri().path());
+                    let transaction_endpoint_name =
+                        resolve_logical_endpoint(endpoint_name.as_ref(), txn.request_uri());
```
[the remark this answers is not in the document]
thieman: [GPT-5.6 Sol (OpenAI)] Fixed in c9e510abf4. The input path now keeps the resolver result local and falls back to the borrowed `uri.path()` value. `MetaString` is materialized only when the per-endpoint cache misses, matching the prior hot-path behavior.

lib/saluki-components/src/forwarders/cluster_agent/mod.rs
thieman: [GPT-5.6 Sol (OpenAI)] Input telemetry resolves the logical endpoint before the Cluster Agent request mapper runs, while retry telemetry resolves it after the mapper rewrites the URI to `/series`. The previous path-based resolver could therefore label one transaction differently across those stages. This forwarder sends every accepted request to its single Cluster Agent series endpoint, so returning `cluster_agent_series` unconditionally keeps the identity stable before and after mapping. The mapper test now covers both stages.

lib/saluki-components/src/common/datadog/telemetry.rs
thieman: [GPT-5.6 Sol (OpenAI)] Fixed in c9e510abf4. The retry telemetry cache is now owned directly by the endpoint I/O loop as a plain `FastHashMap`; the `Arc<Mutex<...>>` is gone. Cached `Counter` handles are cloned into the in-flight state, which only updates their existing Arc reference counts and does not require shared map access.

lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -607,35 +688,38 @@ async fn run_endpoint_io_loop<B>(
...
-                                Ok(push_result) => track_queue_drops(&telemetry, &endpoint_domain, push_result),
-                                Err(e) => error!(endpoint_url, error = %e, "Failed to re-enqueue failed transaction. Events may be permanently lost."),
-                            }
-                        },
+                    Ok(InFlightTransaction { metadata, retry_counters, result }) => {
+                        let requeued_counters = account_retry_result(retry_counters.as_ref(), &result);
+                        match result {
+                            // We got a response -- maybe successful, maybe not -- so just process that.
+                            Ok(http_response) => process_http_response(http_response, metadata, &telemetry, &endpoint_url, &endpoint_domain).await,
+
```
tobz: I'm guessing that the clanker decided to encapsulate the logic this way, behind a free function, so it could observe the metrics being updated in a unit test.

Could we actually just extract this whole match block (`match task_result`) into its own function instead? It's a lot cleaner than this, where we're interleaving inline logic with encapsulated logic.

lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -607,35 +688,38 @@ async fn run_endpoint_io_loop<B>(
...
-                        // later.
-                        Err(RetryCircuitBreakerError::Open(req)) => {
-                            let reassembled_txn = Transaction::reassemble(metadata, req);
-                            match pending_txns.push_low_priority(reassembled_txn).await {
-                                Ok(push_result) => track_queue_drops(&telemetry, &endpoint_domain, push_result),
-                                Err(e) => error!(endpoint_url, error = %e, "Failed to re-enqueue failed transaction. Events may be permanently lost."),
-                            }
-                        },
+                    Ok(InFlightTransaction { metadata, retry_counters, result }) => {
+                        let requeued_counters = account_retry_result(retry_counters.as_ref(), &result);
```
tobz: It seems like all this does is pass back a reference to `retry_counters`... why can't we just use that again instead of rebinding? 😅

lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -595,9 +668,17 @@ async fn run_endpoint_io_loop<B>(
             // While we're not done and there are pending transactions, wait for the service to become ready and then
             // next the next available pending transaction.
             svc = service.ready(), if !done && !pending_txns.is_empty() => match svc {
-                Ok(svc) => if let Some(txn) = pending_txns.pop().await {
-                    let (metadata, request) = txn.into_parts();
-                    in_flight.spawn(svc.call(request).map(move |result| (metadata, result)));
+                Ok(svc) => if let Some(pending_txn) = pending_txns.pop().await {
+                    let dispatch = prepare_transaction_dispatch(
```
tobz: Why do we construct this brand new `TransactionDispatch` type only to then immediately destructure it? Why can't `prepare_transaction_dispatch` return the three parts directly? They're all distinctly typed, so it's not the classic risk of returning multiple values and ending up with potential for misbinding them.

Also, maybe call it `prepare_transaction_for_dispatch` after making those changes. 😅

lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -57,6 +58,77 @@ use super::{
...
+        request,
+        retry_counters,
+    }
+}
+
+fn account_retry_result<'a, T, E, R>(
+    retry_counters: Option<&'a TransactionRetryCounters>, result: &Result<T, RetryCircuitBreakerError<E, R>>,
+) -> Option<&'a TransactionRetryCounters> {
+    match result {
+        Ok(_) | Err(RetryCircuitBreakerError::Service(_) | RetryCircuitBreakerError::Retry(_)) => {
```
tobz: This is incrementing the number of retries if retry is indicated _or_ if everything was alright _or_ if there was a service error (which might not be retryable.)

What am I missing? 🧐

lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -888,7 +978,7 @@ impl<T: Retryable> PendingTransactions<T> {
                         low_prio_queue_len = self.low_priority.len(),
                         "Dequeued pending transaction from low-priority queue."
                     );
-                    return Some(transaction);
+                    return Some(PendingTransaction::Retry(transaction));
```
tobz: I'm not sure it's actually this simple, because we theoretically would buffer incoming requests in the low-priority queue if we overflow the high-priority queue, including new requests that we haven't actually retried yet. 😅

tobz (changes requested):

lib/saluki-components/src/common/datadog/io.rs
thieman: @\clanker confirm that the datadog-agent equivalents for this telemetry also emit in the case of non-retry transactions such as high priority queue overflow. provide links to datadog-agent sources if you find them

lib/saluki-components/src/common/datadog/io.rs
thieman: [GPT-5.6 Sol (OpenAI)] Implemented in [87ef5415](https://github.com/DataDog/saluki/commit/87ef54150208d4be88f5a954d6ff64b92ec1be58). The full `JoinSet` task result, including `JoinError` handling, now goes through `handle_in_flight_transaction_result`, leaving the select branch as orchestration only. [bbafc2d3](https://github.com/DataDog/saluki/commit/bbafc2d3202d0a438f353efdc9183501435884a9) updates the blocked-retry test to exercise that extracted handler directly.

lib/saluki-components/src/common/datadog/io.rs
thieman: [GPT-5.6 Sol (OpenAI)] Implemented in [87ef5415](https://github.com/DataDog/saluki/commit/87ef54150208d4be88f5a954d6ff64b92ec1be58). `TransactionDispatch` is removed, and `prepare_transaction_for_dispatch` now returns the distinctly typed `(Metadata, Request<TransactionBody<B>>, Option<TransactionRetryCounters>)` tuple directly.

lib/saluki-components/src/common/datadog/io.rs
thieman: [GPT-5.6 Sol (OpenAI)] Clarified in [87ef5415](https://github.com/DataDog/saluki/commit/87ef54150208d4be88f5a954d6ff64b92ec1be58). The optional counters identify a transaction dispatched from the low-priority/retry queue; the result identifies what happened after dispatch. `Ok`, final `Service`, and `Retry` all mean an actual retry-queue dispatch occurred, so each increments `transactions.retries`. `Open` is the only case where the breaker rejected the request before inward dispatch, so it does not increment retries and instead increments requeued after successful queue insertion.

lib/saluki-components/src/common/datadog/io.rs
thieman: [GPT-5.6 Sol (OpenAI)] Implemented in [87ef5415](https://github.com/DataDog/saluki/commit/87ef54150208d4be88f5a954d6ff64b92ec1be58). `account_retry_result` and the `requeued_counters` rebinding are gone. Retry accounting now sits directly beside each result path in the extracted handler, and only the `Open` branch passes the existing counters to requeue accounting.

lib/saluki-components/src/common/datadog/io.rs
thieman: [GPT-5.6 Sol (OpenAI)] The observation is correct. The low-priority queue also contains fresh requests displaced by high-priority overflow. This matches the Core Agent: overflow is inserted into its retry queue, and a later dispatch from that queue increments `transactions.retries`. [0b7e723a](https://github.com/DataDog/saluki/commit/0b7e723a31d6340487e335c86b239ea7082d0ef9) renames the private provenance variant from `Retry` to `LowPriority` and documents that overflow is intentionally included in Core Agent-compatible retry accounting.

thieman: [GPT-5.6 Sol (OpenAI)] Core Agent source references: [`sendHTTPTransactions`](https://github.com/DataDog/datadog-agent/blob/9de2a8371cf8d95794f238939709491907893288/comp/forwarder/defaultforwarder/impl/domain_forwarder.go#L342-L355) sends a fresh transaction to `retryQueue` when `highPrio` is full. [`retryTransactions`](https://github.com/DataDog/datadog-agent/blob/9de2a8371cf8d95794f238939709491907893288/comp/forwarder/defaultforwarder/impl/domain_forwarder.go#L101-L121) extracts every transaction from that queue and increments `tlmTxRetried` when it dispatches one to `lowPrio`. So high-priority overflow is included in `transactions.retries` once it is later dispatched from the retry queue.

webern (approved):
To the best of my ability in unfamiliar code:

- verified independently that tobz's comments were addressed
- read through some critical areas visually
   - lib/saluki-io/src/net/util/middleware/retry_circuit_breaker.rs
       - Understand the distinction between Retry and Open.
   - lib/saluki-components/src/common/datadog/io.rs
       - PendingTransaction
       - prepare_transaction_for_dispatch
       - handle_in_flight_transaction_result
       - PendingTransactions::pop

Verified these desired invariants hold using AI:
1. Can retries increment before the inner service is called? — No.
  Retry, Service, and successful responses are produced only after polling the inner service. Open is the sole pre-dispatch
  result, and the handler explicitly excludes it before incrementing retries.

2. Can requeued increment if insertion fails? — No.
  requeue_transaction performs push_low_priority(...).await? before incrementing. An insertion error returns immediately. There is
  no direct test of this failure branch, but the control flow is unambiguous.

3. Is provenance preserved for overflowed and persisted transactions? — Yes, by queue location.
  High-priority overflow enters the low-priority queue and later pops as PendingTransaction::LowPriority. Before shutdown
  persistence, remaining high-priority entries are moved into that same persistent low-priority queue, so rehydrated entries also
  retain low-priority provenance. Overflow accounting is directly tested; persistence-plus-counter behavior is primarily
  guaranteed by construction rather than a focused telemetry test.

4. Does JoinError alter counters or lose transactions differently from before? — No.
  The extracted handler returns immediately on JoinError, before updating counters. The transaction is dropped and an error is
  logged, exactly as on the PR base; the refactor only moved this behavior into the helper. There is no panic/JoinError test.

5. Do tests verify observable behavior rather than only helper internals? — Mostly yes.
  The server-error test exercises the full forwarder and asserts emitted retry metrics. The blocked-retry test calls the extracted
  handler directly, but uses a real circuit breaker, verifies that the inner service was not called, and asserts queue and metric
  outcomes. The main gap is that Open/requeued lacks full-I/O-loop coverage.

6. Are Cluster Agent labels stable across URI rewriting? — Yes.
  Input telemetry resolves the endpoint before mapping, while retry telemetry may see the rewritten /series URI. Returning the
  constant cluster_agent_series is appropriate because this forwarder has only that logical endpoint, and the updated test checks
  endpoint identity both before and after mapping.

tobz (approved):

PR #2180, opened 2026-07-22
feat(io): tag successful HTTP transactions by protocol

thieman:
## Summary

- tag successful ADP HTTP transaction counts with the actual response protocol version
- retain `proto_version` when exposing `transactions.success` through Remote Agent Registry telemetry
- keep `transactions.success_bytes` aggregated by domain and endpoint to match the Core Agent
- use fixed lazy `OnceLock` slots for known protocols, avoiding per-success allocation, locking, hashing, formatting, and handle cloning
- group unfamiliar future HTTP versions in one lazy `proto_version:unknown` counter instead of panicking

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

- `make fmt`
- `cargo check --workspace`
- `cargo nextest run -p saluki-io` (113 passed)
- `cargo nextest run -p agent-data-plane state::metrics` (9 passed)
- `make test-all`
- Commit hooks: formatting, Clippy, dependency checks, documentation, and API documentation

`make check-all` passed its normal checks and reached the feature compatibility matrix. The local host could not build the AWS-LC FIPS variants because `cmake` is not installed.

## References

- [DADP-156](https://datadoghq.atlassian.net/browse/DADP-156)

[DADP-156]: https://datadoghq.atlassian.net/browse/DADP-156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/saluki-io/src/net/client/http/telemetry.rs
+++ b/lib/saluki-io/src/net/client/http/telemetry.rs
@@ -185,18 +185,33 @@ struct SuccessCounters {
     http_11: OnceLock<Counter>,
     http_2: OnceLock<Counter>,
     http_3: OnceLock<Counter>,
+    fallback: OnceLock<Mutex<HashMap<Version, Counter>>>,
 }
 
 impl SuccessCounters {
-    fn slot_for_version(&self, version: Version) -> (&OnceLock<Counter>, &'static str) {
+    fn slot_for_version(&self, version: Version) -> Option<(&OnceLock<Counter>, &'static str)> {
         match version {
-            Version::HTTP_09 => (&self.http_09, "HTTP/0.9"),
-            Version::HTTP_10 => (&self.http_10, "HTTP/1.0"),
-            Version::HTTP_11 => (&self.http_11, "HTTP/1.1"),
-            Version::HTTP_2 => (&self.http_2, "HTTP/2.0"),
-            Version::HTTP_3 => (&self.http_3, "HTTP/3.0"),
-            _ => unreachable!("unsupported HTTP version"),
+            Version::HTTP_09 => Some((&self.http_09, "HTTP/0.9")),
+            Version::HTTP_10 => Some((&self.http_10, "HTTP/1.0")),
+            Version::HTTP_11 => Some((&self.http_11, "HTTP/1.1")),
+            Version::HTTP_2 => Some((&self.http_2, "HTTP/2.0")),
+            Version::HTTP_3 => Some((&self.http_3, "HTTP/3.0")),
+            _ => None,
         }
+    }
+
+    fn fallback_counter(&self, builder: &MetricsBuilder, version: Version) -> Counter {
+        let fallback = self.fallback.get_or_init(|| Mutex::new(HashMap::new()));
+        let mut counters = fallback.lock().unwrap();
+        counters
+            .entry(version)
+            .or_insert_with(|| {
+                builder.register_counter_with_tags(
+                    "network_http_requests_success_total",
+                    [("proto_version", format!("{version:?}"))],
+                )
+            })
+            .clone()
     }
 }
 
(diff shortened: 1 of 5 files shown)


lib/saluki-io/src/net/client/http/telemetry.rs
```diff
@@ -175,10 +178,32 @@ impl<S> Layer<S> for EndpointTelemetryLayer {
...
 }
 
+#[derive(Default)]
+struct SuccessCounters {
+    http_09: OnceLock<Counter>,
+    http_10: OnceLock<Counter>,
+    http_11: OnceLock<Counter>,
+    http_2: OnceLock<Counter>,
+    http_3: OnceLock<Counter>,
+}
```
thieman, line 190: These get us around needing to have a mutex-protected hashmap for differentiating the counter by protocol

tobz: Damn, this is crazy. 😭

Has me thinking, though, that I bet we could whip up a macro in `saluki-metrics` to allow for generating value-mapped metrics like this by feeding it an enum and having it generate a metric per variant so that you just have to provide the enum (and maybe a way to map the enum value to the tag value you want or whatever?) and then it does it for you like this -- hides all the boilerplate codegen, but it's optimal at least -- and all that.

Would actually help us out in a few different areas where we do this sort of thing.

webern (approved):
Your version is probably easier to understand than the suggestion below, but you don't necessarily need a mutex-protected map since the quantity of buckets is known.

The `unimplemented!` panic bothers me, but the version consts `Version::HTTP_09`, etc., are pre-existing. I probably would have modeled those as an enum for exhaustive matching.

### Alternative using an array

```rust
   const VERSION_COUNT: usize = 5;
   const VERSION_LABELS: [&str; VERSION_COUNT] =
       ["HTTP/0.9", "HTTP/1.0", "HTTP/1.1", "HTTP/2.0", "HTTP/3.0"];

   #[derive(Default)]
   struct SuccessCounters {
       slots: [OnceLock<Counter>; VERSION_COUNT],
   }

   fn version_index(version: Version) -> usize {
       match version {
           Version::HTTP_09 => 0,
           Version::HTTP_10 => 1,
           Version::HTTP_11 => 2,
           Version::HTTP_2 => 3,
           Version::HTTP_3 => 4,
           _ => unreachable!("unsupported HTTP version"),
       }
   }
 ```

 Then:

 ```rust
   fn increment_success(&self, version: Version) {
       let index = version_index(version);

       self.success.slots[index]
           .get_or_init(|| {
               self.builder.register_counter_with_tags(
                   "network_http_requests_success_total",
                   [("proto_version", VERSION_LABELS[index])],
               )
           })
           .increment(1);
   }
 ```

To eliminate OnceLock, you could eagerly create an array:

 ```rust
   struct SuccessCounters {
       counters: [Counter; VERSION_COUNT],
   }
 ```

 and register all five counters in PerEndpointTelemetry::new(). Then success recording is simply:

 ```rust
   self.success.counters[version_index(version)].increment(1);
 ```

lib/saluki-io/src/net/client/http/telemetry.rs
thieman: Unfortunately it's not actually an enum which causes other problems highlighted by Matt below. I'm changing it up so the wildcard case goes to a different Counter rather than causing a panic

thieman:
@webern got rid of the panic by having the wildcard case go to a fallback bucket that will emit the proto_version as "unknown". that turns it into a telemetry papercut rather than a panic, if HTTP/4 ever happens then we can fix it then

PR #2183, opened 2026-07-22
enhancement(agent-data-plane): add reporter to forward diagnostic events to Core Agent as Remote Agent events

tobz:
## Summary

This PR adds the ability to collect internal "diagnostic events" and forward them to the Core Agent as Remote Agent events.

Following on the work done in #2096 to add an emitter for diagnostic events, we're now at the stage of wanting to actually send them to the Core Agent so they can be used to drive behavior, such as refreshing secrets when an invalid API key is detected.

This PR adds a new worker that collects these diagnostic events, which converts them to their respective Remote Agent event types and sends them to the Core Agent, as well as adds handling of 403 errors in the Datadog forwarder/API key validator to trigger emitting these diagnostic events. We've made a few small changes around how we create the emitter to try and bake it more into the forwarder internals so it's not created at the top and then forced to be threaded through a zillion different places.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

- [x] New unit tests for the behavior of emitting the diagnostic events themselves.
- [x] Manually ran ADP alongside the Core Agent and observed debug log messages when the Core Agent received the Remote Agent events.

## References

DADP-2

### DIFF
--- a/bin/agent-data-plane/src/internal/remote_agent.rs
+++ b/bin/agent-data-plane/src/internal/remote_agent.rs
@@ -6,6 +6,10 @@ use async_trait::async_trait;
 use chrono::{DateTime, Utc};
 use datadog_agent_commons::ipc::client::RemoteAgentClient;
 use datadog_agent_commons::ipc::session::{SessionId, SessionIdHandle};
+use datadog_protos::agent::v1::{
+    event::Details as RemoteAgentEventDetails, Event as RemoteAgentEvent, InvalidApiKeyEvent,
+    ReportRemoteAgentEventRequest,
+};
 use datadog_protos::agent::{
     config_event,
     flare::v1::{flare_provider_server::*, *},
@@ -20,10 +24,10 @@ use saluki_common::sync::shutdown::ShutdownHandle;
 use saluki_common::task::spawn_traced_named;
 use saluki_config::{dynamic::ConfigUpdate, upsert, GenericConfiguration};
 use saluki_core::{
-    diagnostic::DiagnosticCollector,
+    diagnostic::{subscribe_events, DiagnosticCollector, DiagnosticDetails, DiagnosticEvent},
     observability::metrics::{get_shared_metrics_state, AggregatedMetricsProcessor, Reflector, TelemetryProcessor},
     runtime::{
-        state::{DataspaceRegistry, IdentifierFilter},
+        state::{DataspaceRegistry, DataspaceUpdate, IdentifierFilter, Subscription},
         InitializationError, Supervisable, SupervisorFuture,
     },
 };
@@ -33,6 +37,7 @@ use serde_json::{Map, Value};
 use tokio::task::spawn_blocking;
 use tokio::time::{timeout, Instant};
 use tokio::{
+    select,
     sync::{mpsc, oneshot, Mutex},
     time::{interval, MissedTickBehavior},
 };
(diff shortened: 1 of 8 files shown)


lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -623,7 +656,15 @@ async fn run_endpoint_io_loop<B>(
                 match task_result {
                     Ok((metadata, result)) => match result {
                         // We got a response -- maybe successful, maybe not -- so just process that.
-                        Ok(http_response) => process_http_response(http_response, metadata, &telemetry, &endpoint_url, &endpoint_domain).await,
+                        Ok(http_response) => {
+                            let credentials_rejected = process_http_response(http_response, metadata, &telemetry, &endpoint_url, &endpoint_domain).await;
```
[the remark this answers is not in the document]
tobz: Hmm, this does make sense, yeah.

I'm going to work on adding a simple "inspection" Tower middleware layer that lets us introspect the raw responses before retry/concurrency/timeout/etc so that we can observe the 403s in the client service stack, and do the diagnostic event emission _there_ instead.

bin/agent-data-plane/src/internal/remote_agent.rs
```diff
@@ -701,6 +713,89 @@ impl Supervisable for DataspaceAnchorWorker {
...
+
+/// Converts a [`DiagnosticEvent`] into the Datadog Agent's remote agent event representation.
+///
+/// Returns `None` for diagnostic details that have no corresponding remote agent event variant, in which case the
+/// event is not reported.
+fn diagnostic_to_remote_agent_event(event: &DiagnosticEvent) -> Option<RemoteAgentEvent> {
+    let details = match event.details() {
+        DiagnosticDetails::InvalidApiKey => RemoteAgentEventDetails::InvalidApiKey(InvalidApiKeyEvent {}),
+        // `DiagnosticDetails` is `#[non_exhaustive]`; any future variant without a remote agent mapping is skipped.
+        _ => return None,
```
webern, line 790: Should we make that visible?

```suggestion
        _ => {
            warn!("Unrecognized diagnostic message {diagnostic:?}");
            return None,
        },
```

lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -763,6 +804,27 @@ async fn process_http_response(
...
+            let mut last_emit = last_emit.lock().expect("invalid API key emit mutex poisoned");
+            if last_emit.is_some_and(|last| now.duration_since(last) < INVALID_API_KEY_EMIT_INTERVAL) {
+                return;
+            }
+            *last_emit = Some(now);
+        }
+
+        emitter.emit(DiagnosticEvent::new(
+            format!("Datadog API key rejected as invalid (HTTP 403) when forwarding to '{endpoint_url}'."),
+            DiagnosticDetails::InvalidApiKey,
```
webern, line 821: I still maintain that converting a 403 into an `InvalidApiKey` event is purporting to know more about what happened than we actually do. I would have preferred to model this as `Intake403` instead. But, we decided and it's a nit.

webern (approved):
Very cool, nice work.

bin/agent-data-plane/src/internal/remote_agent.rs
tobz: Hmm, I don't think so, no. It would be weird to warn that we're not handling a diagnostic event that we ostensibly know about because we added it.

A user can't do anything with that information.

PR #2184, opened 2026-07-22
docs(otlp): document metrics.batch related configuration keys as not planned

### LINKED ISSUE #2061: Support OTLP metrics settings.
> The issue author has used AI to research and write this issue. Please adversarially verify accuracy when working on the issue.

Sub-issue of #2005.

## Background

The Agent exposes `otlp_config.metrics.batch.min_size`, `otlp_config.metrics.batch.max_size`, and
`otlp_config.metrics.batch.flush_timeout`. These settings control the batching stage in the metrics
exporter's sending queue. ADP accepts OTLP metrics and sends them through a shared Datadog metrics
encoder, but it does not expose equivalent OTLP-specific settings. Determine the correct ADP
configuration and topology for this behavior, then implement parity or document the intentional
divergence.

### Agent behavior

The Agent registers the three settings in `pkg/config/setup/otlp_settings.go:48-50` with defaults of
8192, 0, and `200ms`. `comp/otelcol/otlp/config.go:95-117` reads the batch subtree, and
`comp/otelcol/otlp/map_provider.go:68-74` places it under
`exporters.serializer.sending_queue.batch`.

The serializer exporter enables the OpenTelemetry Collector exporter queue in
`comp/otelcol/otlp/components/exporter/serializerexporter/exporter.go:40-46` and attaches it to
metrics consumption in `factory.go:268-272`. The queue's integrated batcher uses item counts by
default. `min_size` is the threshold for a full batch, `max_size` limits or splits an outgoing batch,
and `flush_timeout` flushes accumulated items when traffic does not reach the threshold. The
settings are for the metrics exporter; logs have a separate batch subtree, and traces do not use
this queue by default.

### agent-data-plane behavior

ADP has two different buffering stages, neither of which provides the Agent behavior. The OTLP
source pushes decoded resources through an internal channel and flushes its topology dispatcher on
a 100ms timer in `lib/saluki-components/src/sources/otlp/mod.rs:387-497`. That timer controls
handoff into the topology, not the boundaries of outbound HTTP requests.

OTLP metrics then share this path with DogStatsD and Checks metrics:
`otlp_in.metrics` connects to `metrics_enrich`, then to `dd_metrics_encode` and `dd_out` in
`bin/agent-data-plane/src/cli/run.rs:484-487, 762-769, 854`. The shared encoder already has
`serializer_max_metrics_per_payload` (10,000 items by default), byte-size limits, and
`flush_timeout_secs` (2 seconds) in
`lib/saluki-components/src/encoders/datadog/metrics/mod.rs:71-97, 210-301`. It flushes when a
payload limit is reached or the timeout expires. It has no `min_size` equivalent and the forwarder
only queues, retries, and sends completed payloads.

Adding the three Agent keys directly to the shared encoder would affect non-OTLP metrics. To retain
the Agent's per-OTLP semantics, isolate OTLP metrics in a separately configured encoder or add an
explicit per-source batching stage before the shared path. The implementation must also define how
item-count limits interact with ADP's existing byte and data-point limits.

### Requirements tracking

- OM_CONF_913
- OM_CONF_924
- OM_CONF_935
- OM_RESZ_147

## Definition of done

- [ ] Decide and document whether ADP will provide Agent-compatible, OTLP-specific batching or an intentional architectural divergence.
- [ ] If implementing the settings, add a configuration model with the Agent defaults and define the units and precedence of count, byte, and timeout limits.
- [ ] If implementing the settings, ensure OTLP configuration does not silently change batching for DogStatsD- or Checks-sourced metrics; split or independently configure the OTLP metrics encoder as needed.
- [ ] Add focused behavior tests that observe outbound payload boundaries: low-volume timeout flush, count/size-triggered flush, and maximum payload splitting. Use a component-level test rather than relying only on the OTLP metric correctness harness.
- [ ] Verify configuration defaults, invalid combinations, and the interaction with the existing retry queue and completed-payload boundaries.

lucastemb:
## Summary

The config options `otlp_config.metrics.batch.*` describe batching behavior in the Core Agent before translation to DogSketch; however, ADP batches elsewhere downstream (before enrichment and before forwarding).

The benefits that this posits in the Core Agent does not apply for us and is effectively already taken care of, so this document serves to highlight this choice to intentional diverge. 

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

N/A

## References

- Closes: https://github.com/DataDog/saluki/issues/2061

### DIFF
--- a/lib/datadog-agent/config/schema/schema_overlay.yaml
+++ b/lib/datadog-agent/config/schema/schema_overlay.yaml
@@ -2000,13 +2000,7 @@ inventory:
     description: "OTLP metrics batch flush timeout"
     documentation: |-
       The core Agent batches OTLP metric input before translation through its serializer exporter queue.
-      ADP translates each OTLP resource before batching the resulting metric events for topology dispatch.
-      The shared metrics encoder then batches, limits, and splits encoded payloads before forwarding.
-
-      ADP does not expose a separate pre-translation timeout because its bounded source channel, event
-      dispatcher, and encoder already provide backpressure and batching at their respective stages.
-      The encoder enforces payload byte limits after translation, which is the relevant boundary for
-      avoiding oversized intake requests. Setting this key has no effect in ADP.
+      ADP's architecture does not have the same batching mechanism and is unaffected by this configuration value.
 
   otlp_config.metrics.batch.max_size:
     support: none
@@ -2015,10 +2009,8 @@ inventory:
     pipelines: [otlp]
     description: "Maximum OTLP metrics batch size"
     documentation: |-
-      ADP does not implement the core Agent's pre-translation OTLP metrics batcher. Its source channel
-      bounds in-flight resources, and its shared metrics encoder applies item-count and byte-size limits
-      to the encoded payload that is ultimately forwarded. A raw OTLP item count cannot reliably bound
-      encoded payload size after translation. Setting this key has no effect in ADP.
+      The core Agent batches OTLP metric input before translation through its serializer exporter queue.
+      ADP's architecture does not have the same batching mechanism and is unaffected by this configuration value.
 
   otlp_config.metrics.batch.min_size:
     support: none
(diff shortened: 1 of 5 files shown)


lib/datadog-agent/config/schema/schema_overlay.yaml
webern, line 1: In general the documentation fields here seems little bit over specific and could drift from reality. I think it could be a bit higher level:

> The core Agent batches OTLP metric input before translation through its serializer exporter queue. ADP's architecture does not have the same batching mechanism and is unaffected by this configuration value.

webern (approved):
Seems like the right level of detail now. 👍

Issue #2197, opened 2026-07-23
Clean up run.rs using typed config.


webern:
There are many objects whose lifetime is involved in the bootstrapping of ADP that are constructed from `GenericConfiguration`. These are a little bit harder than components to migrate to typed config due to their position in the bootstrapping process and, in some cases, where they live in the codebase. Here are some of the objects I am talking about that should use typed config (non-exhaustive):

- `DataPlaneConfiguration`
- `DataPlaneDogStatsDConfiguration`
- `DataPlaneOtlpConfiguration`
- `DataPlaneChecksConfiguration`

There is also something a little bit funky about how the "enabled" bool lives on `Control` in the typed config system but is down in these, kinda-sorta-lower objects that include stuff like OtlpProxy configuration items. There may be a little bit of separating to do.

> It might be advantageous to, both here and in the other spots we're touching in this file, change function signatures to take `&ConfigurationSystem` and then get a `GenericConfiguration` if needed, that way we start enforcing things go through `&ConfigurationSystem` from the get go, even if they're still on the compatibility path.
>
> It would also clean up some of these signatures which are now taking 2-3 different "config" objects, which is getting somewhat clunky.

_Originally posted by @tobz in https://github.com/DataDog/saluki/pull/2094#discussion_r3596504262_
            
To which, I promised!

> I will clean-up run.rs with typed config soon.

Issue #2169, opened 2026-07-21
Migrate saluki codebase to typed configuration.


webern:
**Tracking Issue**

This issue an extension of #1788 which was technically done, as written, when `schema_overlay.yaml` and its build processes were merged in #1828 

Since then, a system for strongly typed configuration has been built to replace the use of raw map access. Succinctly, `SalukiConfiguration` replaces `GenericConfiguration`. The foundation of this appeared over this sequence of PRs:

## Current Work

Now we need to migrate components and other sites consuming `GenericConfiguration` to typed config. This will proceed in with OTLP first, since that is being actively worked on (high churn) but is not yet in production (low risk). Next will be traces since that is expected to be high churn prior to release as well. Perhaps in parallel, Metrics/DogStatsD.

- [x] #2195 
- [x] #2192 
- [x] #2193
- [x] #2294
- [x] #2295
- [x] #2296 
- [x] #2297
- [x] #2293 
- [ ] #2196
- [ ] #2197
- [ ] #2200 
- [ ] #2199
- [ ] #2269
- [ ] #2278
- [ ] #2328
- [ ] #2564
- [ ] #2579
- [ ] #2580

### Foundational PRs

- #1863 
- #1957
- #1986 
- #2093 
- #2094

### Issue #1965

These migrations are tracked as sub-issues of #1965 and are also part of this umbrella issue's typed-configuration work:

- [x] #2310
- [x] #2312
- [x] #2313
- [x] #2314
- [x] #2315
- [x] #2316

PR #2205, opened 2026-07-24
chore(config): migrate `DataPlaneConfiguration` to typed config 

webern:
## Human Summary

In this PR we are migrating `DataPlaneConfiguration` to use typed config. This is a bit more involved than migrating a component to typed config because `DataPlaneConfiguration` is involved in system startup.

Several of the related startup-config sub-types proved to be unnecessary and were deleted. However `DataPlaneConfiguration` remains a nice abstraction because it offers some additional logic helpers over the pure configuration values. It is retained but its guts become typed config.

## AI Summary

 - Make LoadedConfiguration the single authoritative configuration snapshot used during startup.
 - Replace the duplicate raw-map DataPlaneConfiguration with a borrowed decision view over the typed configuration.
 - Route startup, pipeline, topology, API, control-plane, and subcommand decisions through the typed model.
 - Keep configuration-layer endpoint values as text and resolve runtime ListenAddress values only at their consumers, removing unnecessary saluki-io coupling.
 - Convert source-level timeout values to Duration at the configuration boundary and remove obsolete defaults and manual implementations.
 - Preserve the plaintext http:// fallback for bare Core Agent OTLP endpoints.

## Change Type

## How did you test this PR?

We are relying on the existing integration and correctness tests not breaking. Unit tests have been added, removed and changed as dictated by the code changes.

## References

- Progresses #2197
- Progresses #2169

### DIFF
--- a/bin/agent-data-plane/src/cli/debug/mod.rs
+++ b/bin/agent-data-plane/src/cli/debug/mod.rs
@@ -1,8 +1,9 @@
+use agent_data_plane_config_system::LoadedConfiguration;
 use argh::FromArgs;
-use saluki_config::GenericConfiguration;
 use tracing::{error, info};
 
-use crate::cli::utils::DataPlaneAPIClient;
+use crate::cli::utils::{api_or_exit, DataPlaneAPIClient};
+use crate::config::DataPlaneConfiguration;
 
 mod workload;
 use self::workload::{handle_workload_command, WorkloadCommand};
@@ -62,14 +63,10 @@ pub struct SetMetricLevelCommand {
 }
 
 /// Entrypoint for the `debug` commands.
-pub async fn handle_debug_command(bootstrap_config: &GenericConfiguration, cmd: DebugCommand) {
-    let mut api_client = match DataPlaneAPIClient::from_config(bootstrap_config).await {
-        Ok(client) => client,
-        Err(e) => {
-            error!("Failed to create data plane API client: {:#}", e);
-            std::process::exit(1);
-        }
-    };
+pub async fn handle_debug_command(local_config: LoadedConfiguration, cmd: DebugCommand) {
+    let config = local_config.local();
+    let dp = DataPlaneConfiguration::from_configuration(config);
+    let mut api_client = api_or_exit(&dp, &local_config.raw_config()).await;
 
     match cmd.subcommand {
         DebugSubcommand::ResetLogLevel(_) => reset_log_level(&mut api_client).await,
--- a/bin/agent-data-plane/src/cli/debug/mod.rs
+++ b/bin/agent-data-plane/src/cli/debug/mod.rs
@@ -2,8 +2,7 @@ use agent_data_plane_config_system::LoadedConfiguration;
 use argh::FromArgs;
 use tracing::{error, info};
 
-use crate::cli::utils::{api_or_exit, DataPlaneAPIClient};
-use crate::config::DataPlaneConfiguration;
+use crate::cli::utils::{get_api_client_or_exit, DataPlaneAPIClient};
 
 mod workload;
 use self::workload::{handle_workload_command, WorkloadCommand};
@@ -64,9 +63,7 @@ pub struct SetMetricLevelCommand {
 
 /// Entrypoint for the `debug` commands.
 pub async fn handle_debug_command(local_config: LoadedConfiguration, cmd: DebugCommand) {
-    let config = local_config.local();
-    let dp = DataPlaneConfiguration::from_configuration(config);
-    let mut api_client = api_or_exit(&dp, &local_config.raw_config()).await;
+    let mut api_client = get_api_client_or_exit(&local_config).await;
 
     match cmd.subcommand {
         DebugSubcommand::ResetLogLevel(_) => reset_log_level(&mut api_client).await,
--- a/bin/agent-data-plane/src/cli/dogstatsd.rs
+++ b/bin/agent-data-plane/src/cli/dogstatsd.rs
@@ -20,7 +20,6 @@ use saluki_components::sources::REPLAY_CREDENTIALS_GID;
 use saluki_config::DurationString;
 // The privileged API client's IPC identity, and the Linux replay path, read bootstrap keys through
 // the by-key view.
-use saluki_config::GenericConfiguration;
 use saluki_error::{generic_error, ErrorContext as _, GenericError};
 #[cfg(target_os = "linux")]
 use saluki_io::net::{unix::uds_sendmsg_with_creds, ProcessCredentials};
@@ -34,8 +33,7 @@ use tokio_util::sync::CancellationToken;
 use tracing::debug;
 use tracing::{error, info};
 
-use crate::cli::utils::DataPlaneAPIClient;
-use crate::config::DataPlaneConfiguration;
+use crate::cli::utils::{get_api_client_or_exit, DataPlaneAPIClient};
 
 mod top;
 use self::top::{handle_dogstatsd_dump_contexts, handle_dogstatsd_top, DumpContextsCommand, TopCommand};
@@ -183,17 +181,15 @@ pub async fn handle_dogstatsd_command(local_config: LoadedConfiguration, cmd: Do
 }
 
 async fn run_dogstatsd_command(local_config: &LoadedConfiguration, cmd: DogstatsdCommand) -> Result<(), GenericError> {
-    let dp = DataPlaneConfiguration::from_configuration(local_config.local());
-    let raw_config = local_config.raw_config();
     match cmd.subcommand {
         DogstatsdSubcommand::Stats(config) => {
-            let mut api_client = data_plane_api_client(&dp, &raw_config).await?;
+            let mut api_client = get_api_client_or_exit(local_config).await;
             handle_dogstatsd_stats(&mut api_client, config)
                 .await
                 .error_context("Failed to run stats subcommand")
         }
         DogstatsdSubcommand::Capture(config) => {
-            let mut api_client = data_plane_api_client(&dp, &raw_config).await?;
+            let mut api_client = get_api_client_or_exit(local_config).await;
             handle_dogstatsd_capture(&mut api_client, config)
                 .await
                 .error_context("Failed to start DogStatsD capture")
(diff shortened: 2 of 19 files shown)


lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -1142,6 +1160,19 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
...
+/// represented as i64 values.
+fn parse_seconds(key: &str, value: i64) -> Result<Duration> {
+    let seconds =
+        u64::try_from(value).map_err(|e| TranslateError::new_with_context(key, "invalid duration seconds value", e))?;
+    Ok(Duration::from_secs(seconds))
+}
+
+/// A helper to parse listen address values in the schema that are defined as strings.
+fn parse_listen_address(key: &str, value: &str) -> Result<ListenAddress> {
+    ListenAddress::try_from(value).map_err(|e| TranslateError::new_with_context(key, "invalid address", e))
```
webern: These listen addresses are difficult. I need to trace this carefully to make sure it's correct. The main issue is that our ListenAddress object that I'm using does not handle a missing protocol, so it doesn't know (or didn't before this PR) what do with `localhost:5100`. And the "default" protocol is different for different address-y fields.

I think I need to introduce a deserialization type for these in the configuration layer, and have it then feed `ListenAddress`.

The problem comes from the fact that I'm trying to resolve listen addresses in the config layer. That might be a bad idea. Maybe I should leave these as strings and say: "Sorry ADP, you have to parse and resolve these"?

---

localhost:1234 will be rejected as an invalid API listen address.

 ListenAddress::try_from interprets the text before the colon as a URL scheme. It therefore sees:

 ```text
   scheme = localhost
 ```

 Since localhost is not a supported listen-address scheme, translation records an error similar to:

 ```text
   data_plane.api_listen_address: invalid address:
   unknown/unsupported address scheme 'localhost'
 ```

 The valid form is:

 ```yaml
   data_plane:
     api_listen_address: tcp://localhost:1234
 ```

 or, typically for a bind address:

 ```yaml
   data_plane:
     api_listen_address: tcp://127.0.0.1:1234
 ```

 With the PR’s new startup path, an invalid local value causes configuration loading to fail before startup
 proceeds. If the invalid value arrives later as a runtime Agent configuration update, the typed
 configuration rejects the update and keeps the last-known-good API address.

 This is not a new rejection introduced by the PR—the old DataPlaneConfiguration also ultimately parsed the
 value through ListenAddress—but the PR moves validation earlier, into typed configuration translation.

lib/agent-data-plane-config-system/src/saluki_env_overlay.rs
```diff
@@ -233,8 +233,12 @@ impl<'de> Deserializer<'de> for PathRecorder<'_, '_> {
...
     record_scalar!(deserialize_f64, visit_f64, 0.0, EnvDecode::Float);
     record_scalar!(deserialize_char, visit_char, '\0', EnvDecode::RawString);
-    record_scalar!(deserialize_str, visit_str, "", EnvDecode::RawString);
-    record_scalar!(deserialize_string, visit_str, "", EnvDecode::RawString);
+    // The dummy string must satisfy every string-typed leaf's own validation so discovery can
+    // finish. `checks_ipc_endpoint` deserializes through `ListenAddress`'s `try_from = "String"`,
+    // which rejects an empty or unparseable value, so feed a well-formed listen address. The value
+    // is discarded; only the recorded leaf path matters.
+    record_scalar!(deserialize_str, visit_str, "tcp://127.0.0.1:1", EnvDecode::RawString);
+    record_scalar!(deserialize_string, visit_str, "tcp://127.0.0.1:1", EnvDecode::RawString);
```
webern: Maybe this would work and be better self documentation?

```suggestion
    record_scalar!(deserialize_str, visit_str, FAKE_LISTEN_ADDRESS.to_string(), EnvDecode::RawString);
    record_scalar!(deserialize_string, visit_str, FAKE_LISTEN_ADDRESS.to_string(), EnvDecode::RawString);
```

webern (commented):
`ListenAddress` was the killer here. I need to find a better way to handle endpoint and listen address configuration.

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
webern: Ok, I had gone down a rabbit hole in the wrong direction and have recovered... now the PR doesn't do anything stupid with `ListenAddress`.

bin/agent-data-plane/src/main.rs
```diff
@@ -140,22 +138,14 @@ fn initialize_antithesis() {
...
-        .and_then(|loader| {
-            // Modeled keys are also contributed at their canonical (potentially nested) paths, so a
-            // consumer reading the Datadog Agent's own configuration shape sees environment values
-            // where it expects them.
-            let provider = EnvironmentProvider::new()
-                .map_err(|message| generic_error!("Failed to read configuration from the environment: {}", message))?;
-            Ok(loader.add_providers([provider]))
+async fn load_bootstrap_config(bootstrap_config_path: &Path) -> Result<LoadedConfiguration, GenericError> {
+    let loaded = LoadedConfiguration::load(bootstrap_config_path, EnvPrecedence::AfterFile)
+        .await
```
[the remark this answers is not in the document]
webern: We *could* do this, but it's a pretty big feature request that amounts to "allow bad config".

bin/agent-data-plane/src/config.rs
```diff
@@ -1,99 +1,109 @@
...
+    pub fn stop_timeout(&self) -> Duration {
+        match self.config.control.stop_timeout {
+            Some(timeout) => timeout,
+            None => self
+                .config
+                .control
+                .aggregator_stop_timeout
+                .checked_add(self.config.shared.endpoints.forwarder.stop_timeout)
+                // arbitrary fallback in case of overflow
+                .unwrap_or(Duration::from_secs(4)),
```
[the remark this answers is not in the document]
webern: If a human reviewer really wants me to return a Result from this function then I will change it.

bin/agent-data-plane/src/cli/debug/mod.rs
```diff
@@ -62,14 +63,10 @@ pub struct SetMetricLevelCommand {
...
-        Ok(client) => client,
-        Err(e) => {
-            error!("Failed to create data plane API client: {:#}", e);
-            std::process::exit(1);
-        }
-    };
+pub async fn handle_debug_command(local_config: LoadedConfiguration, cmd: DebugCommand) {
+    let config = local_config.local();
+    let dp = DataPlaneConfiguration::from_configuration(config);
+    let mut api_client = api_or_exit(&dp);
```
tobz: From what I can see, we're not using `DataPlaneConfiguration` anywhere below except for getting the API client, so I wonder if we just make `api_or_exit` take `&LoadedConfiguration` instead?

Same with other usages of `api_or_exit` below/in other subcommand files.

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -170,34 +175,35 @@ struct StatsResponse<'a> {
...
         error!("{:#}", error);
         std::process::exit(1);
     }
 }
 
-async fn run_dogstatsd_command(
-    bootstrap_config: &GenericConfiguration, cmd: DogstatsdCommand,
-) -> Result<(), GenericError> {
+async fn run_dogstatsd_command(local_config: &LoadedConfiguration, cmd: DogstatsdCommand) -> Result<(), GenericError> {
+    let dp = DataPlaneConfiguration::from_configuration(local_config.local());
```
tobz: I think just update this method to also use `api_or_exit`: while it does return a result, it just logs that result in basically the same way that we do in `api_or_exit`, so I'd rather just get some more code deduplication and deal with method signature unification as a rainy day cleanup when someone is bored.

bin/agent-data-plane/src/cli/utils.rs
```diff
@@ -26,6 +26,17 @@ pub struct DataPlaneAPIClient {
     authority: String,
 }
 
+/// Builds a data plane API client or exits after logging the error.
+pub(super) fn api_or_exit(dp: &DataPlaneConfiguration) -> DataPlaneAPIClient {
```
tobz: Can we just call this `get_api_client_or_exit`?

Just _slightly_ more self-descriptive.

bin/agent-data-plane/src/main.rs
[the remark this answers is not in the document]
tobz: Yeah, I agree: it's extra work, but _barely_. I'd rather have the simple invariant of "we parse the entire config, and it has to be valid post-translation, before we do anything."

bin/agent-data-plane/src/config.rs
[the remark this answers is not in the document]
tobz: Can we just make it a saturating add instead?

tobz (approved):
A bunch of non-blocking nits, but generally looks good to me. 👍🏻

bin/agent-data-plane/src/config.rs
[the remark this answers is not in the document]
webern: Nice, I didn't know about that function. Done. https://github.com/DataDog/saluki/pull/2205/commits/ceedc0a837c4aecc764507ec6682262ba3a5d121

bin/agent-data-plane/src/cli/utils.rs
webern: Done in https://github.com/DataDog/saluki/pull/2205/commits/3d43fa7e728a5bab82e0335bef6446e2cf3ea0aa

bin/agent-data-plane/src/cli/dogstatsd.rs
webern: Deleted the method in favor of using `get_api_client_or_exit` directly.

in Done in https://github.com/DataDog/saluki/pull/2205/commits/3d43fa7e728a5bab82e0335bef6446e2cf3ea0aa

bin/agent-data-plane/src/cli/debug/mod.rs
webern: Done in https://github.com/DataDog/saluki/pull/2205/commits/3d43fa7e728a5bab82e0335bef6446e2cf3ea0aa

PR #2206, opened 2026-07-24
enhancement(metrics): add new attribute macro for defining static metrics

tobz:
## Summary

This PR introduces a new attribute macro-based approach for generating "static metrics" structs in a similar way to the existing `static_metrics!` macro.

Currently, we support the generation of "static" metrics structs through a declarative macro in the `saluki_metrics` crate: `static_metrics!`. This has a somewhat bespoke syntax but ultimately supports generating a struct and struct impl to hold metric handles, register them, and allows for setting the level of each metric as well as defining labels to apply to every metric. It works well enough but is somewhat opaque and very cargo culted: blindly copying existing examples without really knowing _why_ or _how_ the thing works.

This PR introduces a new attribute macro-based approach to (eventually) replace `static_metrics!`. Attribute macros fall under a class of macros called "procedural" macros: macros which are written in Rust themselves and have programmatic access to the actual Abstract Syntax Tree, which allows for more powerful transformations. What this ultimately enables us to do is demystify (in my opinion) what it means to define a static metrics struct:

```rust
// Before:

static_metrics!(
    name => CacheTelemetry,
    prefix => cache,
    labels => [cache_id: String],
    metrics => [
        gauge(current_items),
        debug_counter(miss_fills_total)
    ],
);

// After:

#[static_metrics(prefix = cache, labels(cache_id))]
struct CacheTelemetry {
    current_items: Gauge,

    #[metric(level = debug)]
    miss_fills_total: Counter,
}
```

We can now see the actual struct and the fields/metrics that are present: really, we're just annotating the fields to help drive the code generation for registering them, but the overall structure is closer to if we wrote the struct ourselves, and so is more obvious.

Right now, this attribute macro is doing the same thing -- feature-wise -- as the declarative macro, and so we're just trading code complexity in one place for code complexity in another place. However, the goal is to then build more advanced features on top of the attribute macro that are only reasonably achievable with an attribute macro, such as automatically generating the code for metrics that have variable label values at runtime, or label values that are derived exclusively from enum variants, and so on. Likewise, it will allow us to do additional things like add normal Rust doc comments and then convert those into metric descriptions, suitable for emitting alongside the metric values when appropriate (think Prometheus scrape endpoint).

Currently, we're _just_ adding the attribute macro. We're not actually yet exposing it through `saluki_metrics` because it would shadow the equivalently named declarative macro, and so we're doing that in the subsequent stacked PR that switches all existing usages of the declarative macro over to the attribute macro.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

New tests for the attribute macro.

## References

DADP-2

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4576,6 +4576,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "saluki-metrics-macros"
+version = "0.1.0"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "saluki-tls"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ members = [
   "lib/saluki-io",
   "lib/saluki-metadata",
   "lib/saluki-metrics",
+  "lib/saluki-metrics-macros",
   "lib/saluki-tls",
   "lib/stringtheory",
   "test/antithesis/harness",
@@ -76,6 +77,7 @@ saluki-error = { path = "lib/saluki-error" }
 saluki-io = { path = "lib/saluki-io" }
 saluki-metadata = { path = "lib/saluki-metadata" }
 saluki-metrics = { path = "lib/saluki-metrics" }
+saluki-metrics-macros = { path = "lib/saluki-metrics-macros" }
 saluki-tls = { path = "lib/saluki-tls" }
 stele = { path = "bin/correctness/stele" }
 stringtheory = { path = "lib/stringtheory" }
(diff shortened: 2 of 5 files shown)


tobz:
* **#2208** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2208?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2207** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2207?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2206** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2206?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2206?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

aqian01 (approved):

PR #2207, opened 2026-07-24
chore(metrics): cut over to the `static_metrics` attribute macro and remove the old declarative macro

tobz:
## Summary

This PR is a follow-up to #2206 to actually switch from the `static_metrics!` declarative macro to the `saluki_metrics` attribute macro.

We've adjusted the `saluki-metrics` crate to properly re-export the attribute macro, and switched over all usages of the previous declarative macro. One thing to note is that we now allow struct-wide labels to be any generic type that can be coerced to a `String`, which is why we no longer explicitly define many of these labels as being of type `String`.

Additionally, we've made some changes to `OnDemandPIDResolver` in order to better seperate the Linux vs non-Linux implementations. While we didn't need to do this just to switch to the attribute macro, our refactor highlighted an issue related to the `cfg(target_os = "linux")` stuff that pushed me to do this tangential refactoring to clean things up that were in a bad state without spinning it out into a separate PR.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [x] Non-functional (chore, refactoring, docs)
- [ ] Performance

## How did you test this PR?

Existing tests, and new tests for the attribute macro itself.

## References

DADP-2

PR #2208, opened 2026-07-24
enhancement(metrics): support mapped metric handles in `static_metrics!`

tobz:
## Summary

This PR adds support to `static_metrics!` for declaring a metric handle as being "mapped", generating the necessary code to support dynamic registration of the underlying metric with label values provided at runtime.

In many cases, telemetry structs for components have a fixed set of labels applied to all metrics: the labels don't change or vary at runtime. We support this already through `static_metrics!`. In other cases, however, we have metrics where label values are dynamic, such as when incrementing error metrics related to network failures: we might put the host or operation name as a label value, and so on. Currently, we handle this by manually writing out telemetry structs to include the necessary maps that hold the label-to-metric handle mappings, the functions that take the label values as parameters and then look up the right metric handle, and so on. This is less than optimal because not only is it cumbersome and prone to having mismatched implementations, but it means we have to go _full_ manual: we have to implement all of the other metric fields manually as well.

This PR introduces an enhancement to `static_metrics!` which allows defining a metric field as being "mapped", which generates all of the necessary code for that field to support it having dynamic labels. When a metric field should be mapped, we simply add an additional item to the `#[metric]` attribute that defines the dynamic labels, and the rest is handled for us in a similar fashion to the overall struct labels:

```rust
enum DiscardReason {
  Intentional,
  Unintentional,
}

// impl fmt::Display for DiscardReason { ... }

#[static_metrics(prefix = basic)]
#[derive(Clone)]
struct Telemetry {
    received_total: Counter,

    #[metric(mapped(endpoint))]
    sent_total: Counter,

    #[metric(mapped(reason: DiscardReason))]
    discarded_total: Counter,
}

// We specify the label values when getting the metric handles:
let telemetry = Telemetry::new();
telemetry.received_total().increment(1);
telemetry.sent_total("some_endpoint").increment(1);
telemetry.discarded_total(DiscardReason::Intentional).increment(1);
```

Like struct labels, we can specify multiple dynamic labels for a metric, each of which can be untyped (simply has to be `Stringable`) or typed, and the generated method parameters can accept those label values either by referenced or by value, making it trivial to use whatever value is at hand -- `usize`, `&str`, `DiscardReason`, etc -- so long as it either can be formatted as a string or fits the specified concrete type.

This PR specifically does _not_ convert existing hand-rolled examples of this pattern, and will be done as a follow-up PR.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

New and existing unit tests.

## References

DADP-2

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4575,6 +4575,7 @@ name = "saluki-metrics"
 version = "0.1.0"
 dependencies = [
  "metrics",
+ "papaya",
  "saluki-metrics-macros",
  "trybuild",
 ]
--- a/lib/saluki-metrics-macros/src/lib.rs
+++ b/lib/saluki-metrics-macros/src/lib.rs
@@ -8,14 +8,18 @@
 use proc_macro::TokenStream;
 use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::{format_ident, quote};
-use syn::{parse::Parser, Fields, Ident, ItemStruct, LitStr, Type};
+use syn::{
+    parse::{Parse, ParseStream, Parser},
+    parse_quote, Fields, Ident, ItemStruct, LitStr, Token, Type,
+};
 
 /// Defines a container struct of statically defined metrics.
 ///
 /// Applied to a struct whose fields are metric handles, this generates the registration, accessors, and metadata for
 /// the whole group. It is re-exported (and intended to be used) as `saluki_metrics::static_metrics`.
 ///
-/// Because it rewrites the annotated struct, it must appear **before** any `#[derive(...)]` on the struct.
+/// Because it can rewrite field storage (see "Mapped metrics" below), it must appear **before** any `#[derive(...)]`
+/// on the struct.
 ///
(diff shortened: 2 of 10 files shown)


tobz:
* **#2208** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2208?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2208?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2207** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2207?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2206** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2206?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

aqian01 (approved):

PR #2213, opened 2026-07-27
feat(agent-data-plane): emit liveness signals

thieman:
## Tracking
- Jira: [DADP-157](https://datadoghq.atlassian.net/browse/DADP-157)

## Summary
- Add a periodic Agent Data Plane liveness source that emits `datadog.agent_data_plane.running` and `datadog.agent_data_plane.up` through the existing Datadog encoders and forwarder.
- Attach the ADP hostname to both signals and an ADP version tag to the running gauge. The running gauge uses the existing `metrics_enrich` path, so common host-tag enrichment applies exactly as it does to other metrics.
- Keep liveness emission simple: it has no source-specific Core Agent IPC, host-tag cache, or retry logic; reuse the metric context and service-check payload while stamping each gauge point at emission time. When `basic_telemetry_add_container_tags` is enabled, resolve ADP's self-container identity from local `/proc/self/cgroup` and attach its low-cardinality tags from the existing workload tag store.
- Provision the metrics and service-check output paths for connected topologies with an active data pipeline, including connected OTLP proxy-only topologies, while preserving standalone proxy no-DNS behavior and the established no-data-pipelines exit behavior.

## Validation
- [x] `cargo nextest run -p agent-data-plane --bin agent-data-plane` (160 passed)
- [x] `cargo check -p agent-data-plane --tests`
- [x] `make fmt`
- [x] Independent post-fix review of `origin/main...HEAD` (no blockers or important issues)
- [ ] `make test-integration-macos-run CASE=adp-no-pipelines-exit` — requires provisioned Core Agent plus release ADP and Panoramic binaries, unavailable locally; the exact CI regression is fixed by the restored no-pipeline exit condition and will rerun in CI.
- [ ] `make check-all` — local FIPS feature checks require `cmake`, which is unavailable in this environment.

[DADP-157]: https://datadoghq.atlassian.net/browse/DADP-157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Manual Docker-backed validation

Ran the existing connected correctness case `dsd-added-tags` against the current branch and queried both live `datadog-intake` containers before cleanup. The ADP comparison intake captured the recurrent `datadog.agent_data_plane.running` gauge with value `1`, host `correctness-testing`, common metric tags, and the ADP version tag. It also captured the `datadog.agent_data_plane.up` service check with `OK` status and the same hostname. The full correctness case passed.

## Container-tag integration coverage

The Linux-only `liveness-container-tags` integration case enables `basic_telemetry_add_container_tags`, uses an opt-in host cgroup namespace plus a constrained fake Docker discovery API, and verifies at the fake Datadog intake that both ADP liveness payloads carry deterministic container tags. The case executes a mounted Python regression test that rejects Docker inspect requests for an unknown container ID before checking the liveness payload markers. The test also verifies the running gauge's host, version tag, value, and current timestamp, and the `up` service check's host and `OK` status. A Linux cgroup fixture separately verifies that a non-container cgroup resolves no container identity, preserving the untagged fallback.


thieman:
[GPT-5.6 Terra (OpenAI)] Investigated the automated CI summaries. `adp-no-pipelines-exit` exposed a real regression: liveness incorrectly kept ADP alive with no enabled data pipelines. That was fixed in `cd938ee460`.

The subsequent `adp-ipc-no-dns` failure exposed a second issue: standalone OTLP proxy mode intentionally avoids normal forwarding/DNS, but liveness had forced those output paths. The simplified revision in `41c9389e9a` scopes liveness to connected topologies, removes liveness-specific Core Agent IPC and host-tag logic, and routes the running metric through the common `metrics_enrich` path. Standalone proxy retains its existing no-DNS behavior. The updated commits are pushed and CI will rerun.

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -380,7 +380,8 @@ async fn create_topology(
 
     let mut control_surfaces = TopologyControlSurfaces::default();
 
-    // If no data pipelines are enabled, then there's nothing for us to do.
+    // Without a data pipeline, there is no useful topology to run. In particular, liveness alone must not keep ADP
+    // alive.
```
thieman: This comment was fine before, revert it

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -392,8 +393,8 @@ async fn create_topology(
     // we additionally create metrics- and logs-specific components connected to that forwarder depending on which of
     // the baseline pipelines are required.
     //
-    // Notably, we _don't_ need either of these if all we're doing is running the OTLP pipeline in proxy mode, which
-    // is the only reason we're differentiating here.
+    // Liveness requires the metrics and service-check baseline pipelines for every running topology, including an
+    // OTLP proxy-only topology.
```
thieman: Again, this comment was fine, revert it

bin/agent-data-plane/src/components/liveness/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+
+use crate::internal::env::ADPEnvironmentProvider;
+
+const LIVENESS_INTERVAL: Duration = Duration::from_secs(15);
+const RUNNING_METRIC_NAME: &str = "datadog.agent_data_plane.running";
+const UP_SERVICE_CHECK_NAME: &str = "datadog.agent_data_plane.up";
+
+/// Periodically emits Agent Data Plane liveness signals.
+pub struct LivenessConfiguration {
+    hostname: MetaString,
```
thieman, line 39: Is hostname here necessary? I would have thought this would be added by the downstream metrics/service_checks components

bin/agent-data-plane/src/components/liveness/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+fn create_liveness_events(hostname: MetaString, version: MetaString) -> (Event, Event) {
+    let metric_context = Context::from_static_name(RUNNING_METRIC_NAME)
+        .with_host(hostname.clone())
+        .with_tags(TagSet::from_iter([Tag::from(format!("version:{version}"))]));
+    let metric = Event::Metric(Metric::gauge(metric_context, 1.0));
+
+    let service_check =
+        Event::ServiceCheck(ServiceCheck::new(UP_SERVICE_CHECK_NAME, CheckStatus::Ok).with_hostname(hostname));
+
+    (metric, service_check)
```
thieman: These are going to be around for the lifetime of the process, can we just make them once and then have them be static or something? It's not a lot, but we should minimize allocations here

bin/agent-data-plane/src/components/liveness/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+
+        assert_eq!(configuration.hostname, "host-a");
+        assert_eq!(
+            configuration.version,
+            saluki_metadata::get_app_details().version().raw()
+        );
+    }
+
+    #[test]
+    fn emits_liveness_events_without_direct_system_host_tags() {
```
thieman: Yes this component should not put any host tags on, but that should be happening downstream. Think we want to remove this test.

bin/agent-data-plane/src/cli/run.rs
thieman: [GPT-5.6 Terra (OpenAI)] Restored the original no-data-pipelines comment in `c90a1a2c7b`.

bin/agent-data-plane/src/cli/run.rs
thieman: [GPT-5.6 Terra (OpenAI)] Restored the original baseline explanation in `c90a1a2c7b` and retained a concise connected-liveness exception. Reverting the proxy-mode sentence verbatim would be inaccurate because connected OTLP proxy-only topologies now deliberately provision metric and service-check output paths for liveness.

bin/agent-data-plane/src/components/liveness/mod.rs
thieman: [GPT-5.6 Terra (OpenAI)] Kept the hostname. `metrics_enrich` intentionally does not add metric hosts, and this service check routes directly to `dd_service_checks_encode` without a host-enrichment transform. Removing it would leave the metric without a host dimension and omit `host_name` from the service-check payload.

bin/agent-data-plane/src/components/liveness/mod.rs
thieman: [GPT-5.6 Terra (OpenAI)] Fixed in `c90a1a2c7b`: `Liveness::new` constructs both payloads once, and each timer tick only clones the prebuilt events for dispatch.

bin/agent-data-plane/src/components/liveness/mod.rs
thieman: [GPT-5.6 Terra (OpenAI)] Replaced the implementation-specific assertion about absent source tags in `c90a1a2c7b` with payload-contract tests for metric name/value/host/version and service-check name/status/hostname.

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -891,3 +913,14 @@ fn write_sizing_guide(bounds: ComponentBounds) -> Result<(), GenericError> {
 
     Ok(())
 }
+
+#[cfg(test)]
+mod liveness_topology_tests {
```
thieman: Extraneous test, remove

thieman: [GPT-5.6 Terra (OpenAI)] Removed the constant-only test in `d94bbb262d`.

bin/agent-data-plane/src/components/liveness/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+        debug!("Liveness source stopped.");
+        Ok(())
+    }
+}
+
+fn create_liveness_events(hostname: MetaString, version: MetaString) -> (Event, Event) {
+    let metric_context = Context::from_static_name(RUNNING_METRIC_NAME)
+        .with_host(hostname.clone())
+        .with_tags(TagSet::from_iter([Tag::from(format!("version:{version}"))]));
+    let metric = Event::Metric(Metric::gauge(metric_context, 1.0));
```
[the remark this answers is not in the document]
thieman: [GPT-5.6 Terra (OpenAI)] Fixed in `1d96a16152`: the source now reuses the metric context but constructs the gauge point with the current Unix second on each emission. Added a regression test asserting the supplied emission timestamp is encoded in the gauge point.

aqian01 (approved):

test/integration/cases/liveness-container-tags/run_liveness_test.py
```diff
@@ -0,0 +1,271 @@
...
+
+agent_process = None
+lock = threading.Lock()
+
+
+def remove_artifacts():
+    for path in (RUNNING_MARKER, UP_MARKER, REQUESTS_PATH, ERROR_PATH):
+        try:
+            os.remove(path)
+        except FileNotFoundError:
```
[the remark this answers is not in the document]
thieman: [GPT-5.6 Terra (OpenAI)] Added a permanent comment: absent marker/request artifacts are expected before fixture startup and are intentionally ignored during cleanup.

test/integration/cases/liveness-container-tags/run_liveness_test.py
```diff
@@ -0,0 +1,271 @@
...
+    signal.signal(signal.SIGTERM, handle_signal)
+    signal.signal(signal.SIGINT, handle_signal)
+
+    server = ReusableThreadingHTTPServer(INTAKE_ADDRESS, IntakeHandler)
+    server_thread = threading.Thread(target=server.serve_forever, daemon=True)
+    server_thread.start()
+
+    try:
+        os.remove(DOCKER_SOCKET_PATH)
+    except FileNotFoundError:
```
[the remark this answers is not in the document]
thieman: [GPT-5.6 Terra (OpenAI)] Added a permanent comment: the fake Docker socket may not exist before the fixture binds it, so cleanup intentionally ignores that absence.

test/integration/cases/liveness-container-tags/run_liveness_test.py
```diff
@@ -0,0 +1,271 @@
...
+    try:
+        agent_process = subprocess.Popen(["/bin/entrypoint.sh"])
+        return agent_process.wait()
+    finally:
+        docker_server.shutdown()
+        docker_server.server_close()
+        docker_thread.join(timeout=1)
+        try:
+            os.remove(DOCKER_SOCKET_PATH)
+        except FileNotFoundError:
```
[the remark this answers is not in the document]
thieman: [GPT-5.6 Terra (OpenAI)] Added a permanent comment: teardown tolerates an already-removed fake Docker socket.

PR #2220, opened 2026-07-28
feat(otlp): added tls server config option for http and grpc

### LINKED ISSUE #2014: Add a reusable TLS config for the OTLP source shared between gRPC and HTTP
Sub-issue of #2005.

## Background

The OTLP source currently has no TLS support at all, on either protocol. `OtlpServerBuilder::build()`
(`lib/saluki-components/src/common/otlp/mod.rs:117-200`) builds the gRPC server via a plain
`tonic::transport::Server::builder()...serve_with_incoming(...)` over a raw `TcpListener`
(no TLS acceptor), and the HTTP server via `HttpServer::from_listener(...).listen()`
(lines 195-197), also with no TLS. `GrpcConfig`/`HttpConfig`
(`lib/saluki-components/src/common/otlp/config.rs:53-91`) only carry `endpoint`, `transport`,
and (grpc-only) `max_recv_msg_size_mib` — no cert/key/CA fields whatsoever.

This mirrors the OpenTelemetry Collector's `configtls.ServerConfig`
([docs](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/README.md)),
which both the OTLP receiver's `grpc` and `http` protocol sections embed identically via a shared
`tls` config block.

## Why this needs to be reusable, not two separate implementations

TLS applies identically to both the gRPC and HTTP listeners. Building it twice would duplicate cert-loading,
validation, and reload logic. This is also a new pattern for the codebase: there is currently no
shared, nested TLS config struct reused across two protocol sub-configs — `GrpcConfig` and
`HttpConfig` are siblings under `Protocols` (`config.rs:41-51`), and elsewhere in the codebase
(e.g. `ForwarderConfiguration` in `lib/saluki-components/src/common/datadog/config.rs:284,303-304`)
TLS-related fields are flat, per-consumer fields rather than a shared struct — this issue would
establish that pattern.

There's already a building blocks to build on:
- `saluki-tls` (`lib/saluki-tls/src/lib.rs`) provides `TlsMinimumVersion` (line 52-59) and
  server/client config builders, but no serde-deserializable config struct with cert/key/CA
  paths — callers currently assemble `rustls::ServerConfig`/`ClientConfig` by hand.
- `saluki-io`'s `HttpServerBuilder::with_tls_config(config: ServerConfig)`
  (`lib/saluki-io/src/net/server/http.rs:36,60-61,99-119`) already accepts a raw `rustls::ServerConfig`
  and wraps it in a `TlsAcceptor`, branching TLS/non-TLS per connection — this is the acceptance
  point a parsed config would feed into on the HTTP side. The gRPC side needs the tonic-equivalent
  (`Server::tls_config` / a `rustls`-backed acceptor wrapping the incoming `TcpListener` stream).

## Fields to support (core set, mirroring `configtls.ServerConfig`)

| Field | Type | Default | Description |
|---|---|---|---|
| `cert_file` | path | none (TLS disabled) | Path to the server certificate. Required to enable TLS. |
| `key_file` | path | none | Path to the private key; required alongside `cert_file`. |
| `client_ca_file` | path (optional) | none | Path to a CA bundle used to verify client certificates for mTLS. If set, requires and verifies client certs. |
| `min_tls_version` | string | `"1.2"` | Minimum accepted TLS version; should reuse `saluki_tls::TlsMinimumVersion`. |

## Work

- [ ] Add a new `TlsConfig` struct (name TBD — avoid colliding with anything OTLP-specific) either
  in `saluki-tls` or in `lib/saluki-components/src/common/otlp/config.rs`, following the
  `GrpcConfig`/`HttpConfig` pattern (`Deserialize` + `Facet`, `#[serde(default = "default_...")]`
  per field, explicit `impl Default`), deserializing to a validated `rustls::ServerConfig` (or
  `None` if `cert_file`/`key_file` are unset, meaning TLS stays disabled).
- [ ] Nest one `tls: Option<TlsConfig>` field into both `GrpcConfig` and `HttpConfig`
  (`config.rs:53-91`), sharing the same struct.
- [ ] Wire the resulting `rustls::ServerConfig` into the HTTP path via the existing
  `HttpServerBuilder::with_tls_config` (`lib/saluki-io/src/net/server/http.rs:60-61`).
- [ ] Wire the equivalent into the gRPC path in `OtlpServerBuilder::build()`
  (`lib/saluki-components/src/common/otlp/mod.rs:166-180`) — likely requires wrapping the incoming
  `TcpListener` stream in a `tokio_rustls::TlsAcceptor` before handing it to
  `serve_with_incoming`, since `tonic`'s built-in `tls_config` expects its own certificate types.
- [ ] Reuse `saluki_tls::TlsMinimumVersion` for the `min_tls_version` field rather than
  reintroducing a parallel enum.
- [ ] Document the new config per `AGENTS.md` documentation guidelines (defaults, mTLS behavior,
  who should enable `client_ca_file`).
- [ ] Add unit/config tests validating defaults, deserialization, and that omitting `cert_file`
  leaves TLS disabled on both protocols.

## Requirements
- OM_CONF_954

lucastemb:
## Summary

Adds TLS support for both OTLP HTTP and gRPC servers. 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests

## References

- Closes: https://github.com/DataDog/saluki/issues/2014

### DIFF
--- a/docs/agent-data-plane/configuration/configuration.md
+++ b/docs/agent-data-plane/configuration/configuration.md
@@ -433,16 +433,12 @@ ways that are not yet fully characterized.
 | `otlp_config.receiver.protocols.grpc.keepalive.server_parameters.timeout`                  | gRPC keepalive: ping timeout           |         |
 | `otlp_config.receiver.protocols.grpc.max_concurrent_streams`                               | gRPC max concurrent streams            |         |
 | `otlp_config.receiver.protocols.grpc.read_buffer_size`                                     | gRPC read buffer size                  |         |
-| `otlp_config.receiver.protocols.grpc.tls.ca_file`                                          | gRPC TLS CA file path                  |         |
 | `otlp_config.receiver.protocols.grpc.tls.ca_pem`                                           | gRPC TLS CA PEM                        |         |
-| `otlp_config.receiver.protocols.grpc.tls.cert_file`                                        | gRPC TLS cert file path                |         |
 | `otlp_config.receiver.protocols.grpc.tls.cert_pem`                                         | gRPC TLS cert PEM                      |         |
 | `otlp_config.receiver.protocols.grpc.tls.cipher_suites`                                    | gRPC TLS cipher suites                 |         |
-| `otlp_config.receiver.protocols.grpc.tls.client_ca_file`                                   | gRPC TLS client CA file                |         |
 | `otlp_config.receiver.protocols.grpc.tls.client_ca_file_reload`                            | gRPC TLS client CA reload              |         |
 | `otlp_config.receiver.protocols.grpc.tls.curve_preferences`                                | gRPC TLS curve preferences             |         |
 | `otlp_config.receiver.protocols.grpc.tls.include_system_ca_certs_pool`                     | gRPC TLS include system CA pool        |         |
-| `otlp_config.receiver.protocols.grpc.tls.key_file`                                         | gRPC TLS key file path                 |         |
 | `otlp_config.receiver.protocols.grpc.tls.key_pem`                                          | gRPC TLS key PEM                       |         |
 | `otlp_config.receiver.protocols.grpc.tls.max_version`                                      | gRPC TLS max version                   |         |
 | `otlp_config.receiver.protocols.grpc.tls.min_version`                                      | gRPC TLS min version                   |         |
@@ -466,16 +462,12 @@ ways that are not yet fully characterized.
 | `otlp_config.receiver.protocols.http.read_header_timeout`                                  | HTTP read header timeout               |         |
 | `otlp_config.receiver.protocols.http.read_timeout`                                         | HTTP read timeout                      |         |
 | `otlp_config.receiver.protocols.http.response_headers`                                     | HTTP response headers                  |         |
-| `otlp_config.receiver.protocols.http.tls.ca_file`                                          | HTTP TLS CA file path                  |         |
 | `otlp_config.receiver.protocols.http.tls.ca_pem`                                           | HTTP TLS CA PEM                        |         |
-| `otlp_config.receiver.protocols.http.tls.cert_file`                                        | HTTP TLS cert file path                |         |
 | `otlp_config.receiver.protocols.http.tls.cert_pem`                                         | HTTP TLS cert PEM                      |         |
 | `otlp_config.receiver.protocols.http.tls.cipher_suites`                                    | HTTP TLS cipher suites                 |         |
-| `otlp_config.receiver.protocols.http.tls.client_ca_file`                                   | HTTP TLS client CA file                |         |
 | `otlp_config.receiver.protocols.http.tls.client_ca_file_reload`                            | HTTP TLS client CA reload              |         |
 | `otlp_config.receiver.protocols.http.tls.curve_preferences`                                | HTTP TLS curve preferences             |         |
 | `otlp_config.receiver.protocols.http.tls.include_system_ca_certs_pool`                     | HTTP TLS include system CA pool        |         |
-| `otlp_config.receiver.protocols.http.tls.key_file`                                         | HTTP TLS key file path                 |         |
 | `otlp_config.receiver.protocols.http.tls.key_pem`                                          | HTTP TLS key PEM                       |         |
 | `otlp_config.receiver.protocols.http.tls.max_version`                                      | HTTP TLS max version                   |         |
 | `otlp_config.receiver.protocols.http.tls.min_version`                                      | HTTP TLS min version                   |         |
@@ -823,8 +815,16 @@ Both commands scrub recognized secret values before writing JSON to standard out
 | `otlp_config.metrics.tags`                                     | Comma-separated tags for all OTLP metrics.         |
 | `otlp_config.receiver.protocols.grpc.endpoint`                 | otlp_config.receiver.protocols.grpc.endpoint       |
 | `otlp_config.receiver.protocols.grpc.max_recv_msg_size_mib`    | Max OTLP inbound gRPC message size (MiB)           |
+| `otlp_config.receiver.protocols.grpc.tls.ca_file`              | gRPC TLS CA file path                              |
+| `otlp_config.receiver.protocols.grpc.tls.cert_file`            | gRPC TLS cert file path                            |
+| `otlp_config.receiver.protocols.grpc.tls.client_ca_file`       | gRPC TLS client CA file                            |
+| `otlp_config.receiver.protocols.grpc.tls.key_file`             | gRPC TLS key file path                             |
 | `otlp_config.receiver.protocols.grpc.transport`                | otlp_config.receiver.protocols.grpc.transport      |
 | `otlp_config.receiver.protocols.http.endpoint`                 | otlp_config.receiver.protocols.http.endpoint       |
+| `otlp_config.receiver.protocols.http.tls.ca_file`              | HTTP TLS CA file path                              |
+| `otlp_config.receiver.protocols.http.tls.cert_file`            | HTTP TLS cert file path                            |
+| `otlp_config.receiver.protocols.http.tls.client_ca_file`       | HTTP TLS client CA file                            |
+| `otlp_config.receiver.protocols.http.tls.key_file`             | HTTP TLS key file path                             |
 | `otlp_config.traces.enabled`                                   | otlp_config.traces.enabled                         |
 | `otlp_config.traces.internal_port`                             | otlp_config.traces.internal_port                   |
 | `otlp_config.traces.probabilistic_sampler.sampling_percentage` | OTLP trace sampling percentage                     |
--- a/lib/agent-data-plane-config/src/domains/otlp.rs
+++ b/lib/agent-data-plane-config/src/domains/otlp.rs
@@ -92,6 +92,7 @@ impl FromStr for HistogramMode {
 
 /// TLS settings for an inbound OTLP receiver.
 #[derive(Clone, Debug, Default, Deserialize, PartialEq, Eq, Serialize)]
+#[serde(default)]
 pub struct TlsConfig {
     /// Path to PEM-encoded certificate authorities loaded for Collector configuration compatibility.
     ///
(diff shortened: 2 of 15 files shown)


lib/saluki-components/src/common/otlp/mod.rs
```diff
@@ -176,8 +197,30 @@ impl OtlpServerBuilder {
...
-        thread_pool_handle.spawn_traced_named("otlp-grpc-server", grpc_server.serve_with_incoming(grpc_incoming));
+        match self.grpc_tls_config {
+            Some(tls_config) => {
+                let tls_acceptor = TlsAcceptor::from(Arc::new(tls_config));
+                let grpc_incoming =
+                    futures::stream::unfold((grpc_listener, tls_acceptor), |(listener, acceptor)| async move {
+                        loop {
+                            match listener.accept().await {
+                                Ok((stream, _)) => match acceptor.accept(stream).await {
+                                    Ok(stream) => return Some((Ok::<_, io::Error>(stream), (listener, acceptor))),
```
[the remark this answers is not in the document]
lucastemb: Fixed in [e4ef0f4](https://github.com/DataDog/saluki/pull/2220/commits/e4ef0f4907553de0f744b6c8f442340fa36cabf4)

lib/saluki-components/src/common/otlp/mod.rs
```diff
@@ -176,8 +197,30 @@ impl OtlpServerBuilder {
         let grpc_listener = tokio::net::TcpListener::bind(grpc_socket_addr)
             .await
             .map_err(|e| generic_error!("Failed to bind OTLP gRPC listener on '{}': {}", grpc_socket_addr, e))?;
-        let grpc_incoming = tonic::transport::server::TcpIncoming::from(grpc_listener);
-        thread_pool_handle.spawn_traced_named("otlp-grpc-server", grpc_server.serve_with_incoming(grpc_incoming));
+        match self.grpc_tls_config {
+            Some(tls_config) => {
+                let tls_acceptor = TlsAcceptor::from(Arc::new(tls_config));
```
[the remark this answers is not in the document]
lucastemb: Fixed in [e4ef0f4](https://github.com/DataDog/saluki/pull/2220/commits/e4ef0f4907553de0f744b6c8f442340fa36cabf4)

lib/agent-data-plane-config/src/domains/otlp.rs
```diff
@@ -81,6 +81,54 @@ impl FromStr for HistogramMode {
     }
 }
 
+/// TLS settings for an inbound OTLP receiver.
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Eq, Serialize)]
+pub struct TlsConfig {
```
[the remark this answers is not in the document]
lucastemb: Fixed in [e4ef0f4](https://github.com/DataDog/saluki/pull/2220/commits/e4ef0f4907553de0f744b6c8f442340fa36cabf4)

lib/saluki-components/src/common/otlp/mod.rs
```diff
@@ -128,10 +138,21 @@ impl OtlpServerBuilder {
         Self {
             http_endpoint,
             grpc_endpoint,
+            http_tls_config: None,
+            grpc_tls_config: None,
```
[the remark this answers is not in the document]
lucastemb: Deferred since future of proxy mode uncertain at the time of writing.

docs/agent-data-plane/configuration/configuration.md
```diff
@@ -430,46 +430,54 @@ ways that are not yet fully characterized.
...
+| `dogstatsd_tcp_port`                                            | TCP listen port for DSD                             |                |
+| `flush_timeout_secs`                                            | Encoder flush timeout (secs)                        |                |
+| `memory_limit`                                                  | Process memory limit                                |                |
+| `memory_slop_factor`                                            | Memory accounting slop fraction                     | 0.25           |
+| `otlp_allow_context_heap_allocs`                                | Allow heap allocations for OTLP contexts            |                |
+| `otlp_cached_contexts_limit`                                    | Max cached OTLP metric contexts                     |                |
+| `otlp_cached_tagsets_limit`                                     | Max cached OTLP tagsets                             |                |
+| `otlp_config.receiver.protocols.grpc.tls.ca_file`               | OTLP gRPC TLS compatibility certificate authorities |                |
```
thieman, line 467: Is the goal to fully implement the options from the OTEL side? Seems like we're missing quite a bit if so? https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/README.md

lib/saluki-components/src/common/otlp/mod.rs
```diff
@@ -176,8 +201,51 @@ impl OtlpServerBuilder {
...
+                let grpc_incoming = futures::stream::unfold(incoming_rx, |mut rx| async {
+                    rx.recv().await.map(|stream| (stream, rx))
+                });
+                thread_pool_handle
+                    .spawn_traced_named("otlp-grpc-server", grpc_server.serve_with_incoming(grpc_incoming));
+            }
+            None => {
+                let grpc_incoming = tonic::transport::server::TcpIncoming::from(grpc_listener);
+                thread_pool_handle
+                    .spawn_traced_named("otlp-grpc-server", grpc_server.serve_with_incoming(grpc_incoming));
```
thieman, line 246: Nit: can this `thread_pool_handle` statement be out of the `match`? Seems identical in both cases?

lib/saluki-components/src/common/otlp/mod.rs
```diff
@@ -192,14 +260,139 @@ impl OtlpServerBuilder {
...
+            generic_error!(
+                "Failed to parse OTLP TLS private key file '{}': {}",
+                tls.key_file,
+                error
+            )
+        })?
+        .clone_key();
+
+    // The Collector loads `ca_file` into `tls.Config.RootCAs`. A pure inbound receiver does not consult those roots,
+    // but loading and validating the file preserves the Collector's configuration behavior.
```
thieman, line 319: We don't actually do anything meaningful with the ca_file? What's the point, then? Per the OTEL docs, sounds like we should be using this to validate client certificates?

lucastemb: Yes. As far as I can tell, the Core Agent behaves the same as well. It's effectively a no-op in both since the receiver is a server accepting inbound connections and that config option is only consulted when its a client validating a server.

docs/agent-data-plane/configuration/configuration.md
lucastemb: I believe so. I scoped this PR to the options that are present [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/config.md#configtls-tlsserversetting), but I am confused about what the authoritative and comprehensive list is.

thieman: My guess is that your link is outdated and that the `.schema.yaml` files strewn about are authoritative, but maybe we should check with someone on the OTEL side. Here's the schema for TLS: https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/config.schema.yaml

Is our actual goal to 100% support all OTEL config options?

lib/saluki-components/src/common/otlp/mod.rs
thieman: OTEL seems to think it should work both ways:

> Path to the CA cert. For a client this verifies the server certificate. For a server this verifies client certificates. If empty uses system root CA. (optional)

From https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/config.schema.yaml#L22C22-L22C182

lib/saluki-components/src/common/otlp/mod.rs
lucastemb: Each `thread_pool_handle` takes in different async streams of accepted connections to serve gRPC requests: one with TLS and another just with standard TCP. The alternative is using `dyn Future` and `Box::` I think this might be the easier to understand and maintain. 😅

lib/saluki-components/src/common/otlp/mod.rs
lucastemb: Since the changes in this PR are scoped to just the receiver, we're handling only inbound connections. Upstream, `ca_file` is loaded and validated but the config key that asks for clients to present certification to connect to the server is `client_ca_file` in our case. The OTEL exporter would be the one in charge of using `ca_file`, but that doesn't appear to be in the Core Agent's OTLP ingest path for metrics.

lib/agent-data-plane-config-system/src/saluki_only.rs
```diff
@@ -311,16 +314,28 @@ pub struct OtlpConfigReceiver {
...
     /// OTLP HTTP receiver knobs (`otlp_config.receiver.protocols.http.*`).
     pub http: OtlpConfigReceiverProtocolsHttp,
 }
 
+/// `otlp_config.receiver.protocols.grpc.*` values absent from the Datadog schema.
+#[derive(Clone, Debug, Default, Deserialize)]
+#[serde(default)]
+pub struct OtlpConfigReceiverProtocolsGrpc {
+    /// OTLP gRPC receiver TLS configuration.
+    pub tls: TlsConfig,
```
webern, line 328: The fact that these are being added as saluki_only keys is suspicious. If these are not part of the Agent's configurable surface, i.e. if they are part of the OTEL collector but not configurable when running OTEL Ingest via the Agent, then they would be out of scope for the current project.

webern (commented):
Are these settings actually configurable when using the OTLP Ingest in the Datadog Agent? If so, and if they are not in the schema, then it must be a novel configuration codepath that we haven't anticipated. Is that what's happening? Is the Agent adopting and using OTEL configs somehow without declaring them in the schema?

webern (changes requested):
This one needs to hold or close because of #2283 

As it stands this is modeling the keys as "saluki-only" which is not correct.

PR #2185, opened 2026-07-22
feat(dogstatsd): add `dogstatsd top`

thieman:
## Summary

This PR adds the retained-context diagnostic exposed by the Datadog Agent as `dogstatsd top` to the `agent-data-plane` binary.

It adds:

- `agent-data-plane dogstatsd top` for an online privileged-API snapshot followed by local analysis.
- `agent-data-plane dogstatsd dump-contexts` for dump-only workflows.
- `agent-data-plane dogstatsd top --path <artifact>` for offline and off-host analysis.
- A privileged API route that coordinates a snapshot through the DogStatsD aggregate owner and returns only the completed artifact path.
- Bidirectional Datadog Agent artifact interoperability, deterministic Agent-compatible reporting, secure atomic artifact publication, focused lifecycle/concurrency coverage, and recorded high-cardinality measurements.

The diagnostic reports contexts currently retained by DogStatsD aggregation. It does **not** report packet rate, sample frequency, or contexts observed during a collection window.

## Tracking

- Jira: [DADP-153 — Add `dogstatsd top` command to show top DogStatsD contexts](https://datadoghq.atlassian.net/browse/DADP-153)

## Reviewer TL;DR

The intended review is primarily about the approach and contracts rather than every test helper or injected failure seam.

The central design is:

1. `dsd_agg` remains the sole owner of `AggregationState`.
2. The API requests a snapshot over a bounded channel serviced in the aggregate task's existing `select!` loop.
3. The owner clones only shared context handles and small type/unit metadata, then resumes ingestion.
4. JSON serialization, zstd compression, synchronization, and atomic publication happen off the aggregate owner.
5. The API returns a JSON path, never the potentially large context set.
6. The CLI reads the artifact with a streaming decoder and renders the Agent-compatible report.

The most useful review order is:

1. Operator contract: [`docs/agent-data-plane/dogstatsd-top.md`](https://github.com/DataDog/saluki/blob/dcd2a8efa2b1689b7c3b46ef60b15fba1ce21eff/docs/agent-data-plane/dogstatsd-top.md)
2. CLI surface: [`bin/agent-data-plane/src/cli/dogstatsd/top.rs`](https://github.com/DataDog/saluki/blob/dcd2a8efa2b1689b7c3b46ef60b15fba1ce21eff/bin/agent-data-plane/src/cli/dogstatsd/top.rs)
[...]

PR #2237, opened 2026-07-29
fix(tls): require mTLS for privileged API

thieman:
## Summary

Require mandatory mutual TLS on the Agent Data Plane privileged API.

The privileged listener now requests a client certificate during the TLS handshake and accepts it only when the presented leaf certificate's DER bytes exactly match the configured Agent IPC certificate. Rustls still verifies the handshake signature, so a matching certificate without the corresponding private key is insufficient.

This applies before protocol routing to both privileged HTTP routes and remote-agent gRPC services.

## Security model

- The Agent IPC certificate/private key is one shared authorization identity.
- Clients without a certificate are rejected during the TLS handshake.
- Clients presenting any different certificate are rejected during the TLS handshake.
- The server and clients pin the same exact certificate rather than trusting a broader CA hierarchy.
- This does not introduce per-route or per-client authorization distinctions.
- Certificate rotation has no overlap window. The server and clients must restart or reload against the same certificate version.

## Key changes

- Add an exact-DER Rustls `ClientCertVerifier` to the shared Agent IPC TLS server helper.
- Preserve provider-backed TLS 1.2/TLS 1.3 handshake-signature verification and FIPS configuration validation.
- Allow `saluki-io` HTTP clients to consume a complete Rustls client configuration without losing a custom verifier or client identity.
- Migrate every existing ADP privileged CLI operation away from `danger_accept_invalid_certs` to the shared IPC mTLS configuration.
- Add machine-readable `agent-data-plane config --json` output and a `--runtime` selector for the translated runtime configuration view.
- Add target-aware Panoramic ADP/Core Agent CLI actions for Linux, macOS, and Windows test runtimes.
- Migrate privileged route and dynamic-config assertions to authenticated CLI calls.
- Verify the Core Agent performs a real inbound remote-agent status RPC over mTLS.
- Migrate the Antithesis overlapping DogStatsD stats workload to the shared IPC identity without exposing the Agent bearer token.
- Integrate the merged DogStatsD top feature: offline analysis avoids client construction, while online top, dump-contexts, and direct API validation use the IPC mTLS identity.
- Document exact-certificate semantics, shared-principal behavior, and restart-coordinated rotation.

## Scope review

A dedicated simplicity and test-quality pass removed roughly half of the original additions. It removed the bespoke anonymous-TLS Panoramic assertion, consolidated target CLI actions, replaced duplicated TLS, config, process, and Antithesis test harnesses with focused ownership-level coverage, and moved the JSON primitive fix into the shared scrubber. After merging the latest `main` and adapting the now-merged DogStatsD top feature, the current diff is approximately 1,980 additions across 45 files.

## Integration with DogStatsD top

#2185 is now part of `main`. Its offline `top --path` workflow still avoids API-client and certificate loading. Every online DogStatsD command—including `top` and `dump-contexts`—now constructs the shared mTLS client asynchronously. The Linux direct-API integration loads `ipc_cert.pem`, and the obsolete anonymous route probes were removed from both Linux and Windows cases.

## Test plan

Passed locally after rebasing onto the latest `main`:

- `make fmt`
- `make check-docs`
- `cargo check --workspace`
- `cargo nextest run -p datadog-agent-commons` — 22 passed
- `cargo nextest run -p saluki-common` — 116 passed
- `cargo nextest run -p saluki-io` — 138 passed
- `cargo nextest run -p agent-data-plane` — 226 passed
- `cargo nextest run -p panoramic` — 51 passed
- `cargo check -p antithesis-scenario-general --all-targets`
- `cargo build -p antithesis-scenario-general --bins`
- `make test-all`
  - 2,494 normal tests passed
  - 21 property tests passed
  - 8 Miri tests passed
  - 8 Loom tests passed
- Windows cross-compilation with tests for:
  - `datadog-agent-commons`
  - `saluki-io`
  - `agent-data-plane`
  - `panoramic`
- Fresh integration image build
- Complete Linux integration suite — 35 passed, 0 failed
- Focused macOS host-process integration — 5 passed, 0 failed:
  - `privileged-api-endpoints`
  - `adp-rar-registration`
  - `adp-config-dynamic-basic`
  - `adp-config-dynamic-runtime`
  - `adp-config-otlp`

Focused TDD coverage exercises matching, missing-client, and mismatched-client certificate handshakes through the production TLS helpers.

## CI validation

The previous branch head passed Linux, macOS, Windows, and FIPS CI. CI is rerunning after the latest `main` merge. Local merge validation passed workspace checks, Windows cross-checks, `make test-all`, the complete 35-case Linux integration suite, and the five affected macOS integration cases. No hosted Antithesis run was launched for this PR; the scenario binaries and configuration were built and validated locally.


bin/agent-data-plane/src/cli/config.rs
```diff
@@ -1,54 +1,172 @@
...
+    /// Controls whether the command emits compact machine-readable JSON.
+    #[argh(
+        switch,
+        description = "emit one compact JSON value to standard output for machine parsing"
+    )]
+    pub json: bool,
+
+    /// Controls whether the command selects the translated runtime configuration.
+    #[argh(switch, description = "select the translated runtime configuration view")]
+    pub runtime: bool,
```
thieman, line 22: Describe in response to this comment why we needed to add these two configs

bin/agent-data-plane/src/cli/config.rs
```diff
@@ -1,54 +1,172 @@
...
+}
+
+fn scrub_json_string(
```
thieman: Did you get scrubbing as a requirement from somewhere? If this is only called by authenticated callers, and only during our own test setup (right?) then scrubbing doesn't seem necessary. Talk through with me first before removing.

bin/correctness/panoramic/src/actions/mod.rs
```diff
@@ -45,3 +64,130 @@ pub fn create_action(config: &ActionConfig) -> Result<Box<dyn Action>, GenericEr
 pub fn default_core_agent_config_endpoint_template() -> String {
     DEFAULT_CORE_AGENT_CONFIG_ENDPOINT_TEMPLATE.to_string()
 }
+
+#[cfg(all(test, unix))]
+mod tests {
```
thieman, line 69: Do a targeted self-review against all tests added in this changeset verifying that they comply with repo-wide testing standards (see AGENTS.md). Remember also that anything added for `correctness` counts as test code, we want to minimize this unless it's valuable.

docs/agent-data-plane/configuration/configuration.md
```diff
@@ -581,6 +581,31 @@ while diagnosing payload content.
 ADP does not dump the exact encoded JSON or protobuf HTTP request body, and it does not log
 compressed wire payload bytes.
 
+### Privileged API authentication
+
+`data_plane.secure_api_listen_address` configures ADP's privileged HTTP and gRPC API. The address
+must include a URL scheme, such as `tcp://`. Every client must present the Agent IPC certificate
```
thieman: It doesn't make sense to me that either an HTTP or gRPC API would be pointed to via `tcp://`

docs/agent-data-plane/configuration/configuration.md
```diff
@@ -581,6 +581,31 @@ while diagnosing payload content.
 ADP does not dump the exact encoded JSON or protobuf HTTP request body, and it does not log
 compressed wire payload bytes.
 
+### Privileged API authentication
```
thieman, line 584: Is this generated from the corresponding .tmpl file? How did you do that / are you sure it hasn't drifted?

lib/datadog-agent/commons/src/ipc/tls.rs
```diff
@@ -65,14 +63,68 @@ impl ServerCertVerifier for DatadogAgentServerCertVerifier {
...
+    provider: Arc<CryptoProvider>,
+}
+
+impl DatadogAgentClientCertVerifier {
+    fn from_certificate_and_provider(cert: CertificateDer<'static>, provider: Arc<CryptoProvider>) -> Self {
+        Self { cert, provider }
+    }
+}
+
+impl ClientCertVerifier for DatadogAgentClientCertVerifier {
```
thieman, line 78: I'm surprised we need to add all this boilerplate, we can't achieve this with the existing APIs?

lib/saluki-io/src/net/client/http/client.rs
```diff
@@ -332,9 +353,17 @@ impl HttpClientBuilder {
...
     /// # Errors
     ///
-    /// If there was an error building the TLS configuration for the client, an error will be returned.
+    /// If there was an error building the TLS configuration for the client, or if a supplied complete TLS
+    /// configuration fails FIPS validation in a FIPS build, an error will be returned.
     pub fn build(self) -> Result<HttpClient, GenericError> {
-        let tls_config = self.tls_builder.build()?;
+        let tls_config = match self.client_tls_config {
+            Some(mut config) => {
+                ensure_client_config_fips_compliant(&config)?;
```
thieman, line 361: Review all changes that touch code surface related to FIPS and verify that FIPS invariants are maintained

bin/agent-data-plane/src/cli/config.rs
thieman: [GPT-5.6 Sol (OpenAI)] `--json` gives Panoramic a deterministic machine-readable result without parsing tracing output. `--runtime` is needed because the existing config integration tests inspect two different views: `/config` is the source/effective `GenericConfiguration`, while `/config/internal` is the translated `SalukiConfiguration` actually handed to runtime components. After mTLS, Panoramic cannot call those routes anonymously, so it invokes the target ADP CLI, which already knows the platform-specific config and IPC certificate paths on Linux, macOS, and Windows. The default human `config` behavior is unchanged.

bin/agent-data-plane/src/cli/config.rs
thieman: [GPT-5.6 Sol (OpenAI)] Scrubbing was existing `agent-data-plane config` behavior before this PR; I did not derive it from the new authentication requirement or remove it. Authentication limits who can call the route, but the command can still write configuration into terminals and CI artifacts, so preserving the existing redaction behavior seems appropriate. The structural implementation was needed for the runtime view because byte-level scrubbing changed a sensitive `null` field into invalid JSON. It now parses first, scrubs string leaves with key context, preserves null/bool/number types, and fails closed. I recommend keeping it, but I have not made a removal decision without this discussion.

bin/correctness/panoramic/src/actions/mod.rs
thieman: [GPT-5.6 Sol (OpenAI)] Completed a targeted review against the repository testing standards, including Panoramic/correctness code. I removed the general-purpose anonymous-TLS assertion and its local TLS servers, collapsed duplicate ADP/Core Agent action tests, reduced the config assertion harness to ownership-level endpoint/retry/redaction coverage, reduced the IPC and saluki-io TLS harnesses to the changed security invariants, removed the duplicated ADP client round trip, and removed mechanical Antithesis tests. The remaining tests reach production behavior: production Rustls builders/acceptor, a real HTTP mTLS round trip, the spawned ADP binary, or actual target child commands. This simplicity pass reduced the PR from roughly 4,000 additions across 41 files to roughly 1,950 additions across 34 files.

docs/agent-data-plane/configuration/configuration.md
thieman: [GPT-5.6 Sol (OpenAI)] `data_plane.secure_api_listen_address` is a `ListenAddress`, so `tcp://` selects the underlying socket transport; HTTP and gRPC are the application protocols above TLS on that listener. The wording incorrectly called it a URL scheme. Fixed in 545bd62f86 to say “transport scheme” and explain the layering.

docs/agent-data-plane/configuration/configuration.md
thieman: [GPT-5.6 Sol (OpenAI)] Yes. I edited `configuration.md.tmpl` as the source and regenerated `configuration.md` with `make build-schema-overlay`; the config inventory text comes from `schema_overlay.yaml` in the same generation pass. I then ran `make fmt` and `make check-docs` and checked the generated diff. The follow-up wording fix in 545bd62f86 was made in both source files and regenerated into the tracked Markdown.

lib/datadog-agent/commons/src/ipc/tls.rs
thieman: [GPT-5.6 Sol (OpenAI)] Rustls does not provide a built-in verifier for “accept exactly these leaf DER bytes.” `WebPkiClientVerifier` expresses CA/root-chain trust and applies WebPKI certificate constraints, which both broadens the identity model and can reject the Agent IPC certificate shape for reasons outside this contract. Implementing `ClientCertVerifier` is therefore required. The apparent boilerplate is the trait contract: advertise/require client auth, provide root hints, compare the leaf, expose supported schemes, and verify TLS 1.2/1.3 handshake signatures. The signature methods delegate to the installed provider and are what prove possession of the matching private key.

lib/saluki-io/src/net/client/http/client.rs
thieman: [GPT-5.6 Sol (OpenAI)] Reviewed the FIPS path end to end. Both IPC builders use the process-wide provider selected by `saluki-tls`; the custom verifier delegates supported schemes and TLS 1.2/1.3 signature verification to that provider. The client and server builders still call `ensure_client_config_fips_compliant` / `ensure_server_config_fips_compliant`, including server key-log and secret-extraction hardening. The new complete `ClientConfig` path in `saluki-io` also validates `config.fips()` before only normalizing ALPN, which does not alter cryptographic parameters. CI passed `check-features-linux-amd64`, `check-features-linux-arm64`, and the amd64/arm64 FIPS image builds.

bin/agent-data-plane/src/cli/config.rs
thieman: Do we actually need both versions (`--runtime` and not) or is that just because of how the integration tests are written today? Could we rewrite them to simplify?

bin/agent-data-plane/src/cli/config.rs
thieman: > changed a sensitive null field into invalid JSON.

Is that just a targeted bug or config option you could add in the scrubber rather than adding all this additional scrubbing code? Look for ways to simplify.

bin/correctness/panoramic/src/actions/mod.rs
thieman: That was the pass you already did, before I left this comment. I wanted you to look at what's in there NOW (the 2000 line changeset) and evaluate each of the tests. Per the testing guidelines, each test should map to specific production behaviors being evaluated.

bin/agent-data-plane/src/cli/config.rs
thieman: [GPT-5.6 Sol (OpenAI)] They are orthogonal rather than two versions of the same setting: `--json` selects machine-readable output, while `--runtime` selects the translated configuration view. The current integration cases intentionally verify both stages: `adp-config-dynamic-basic` checks that a Core Agent update reaches the source/effective map, while `adp-config-dynamic-internal` and `adp-config-otlp` check translation, normalization, defaults, and precedence in `SalukiConfiguration`. We could rewrite the source case to use only the runtime view, but that would drop direct source/config-stream coverage and would not remove a flag unless we made output format implicitly change which configuration is returned. Keeping format and view selection separate is the simpler API and preserves both contracts.

bin/agent-data-plane/src/cli/config.rs
thieman: [GPT-5.6 Sol (OpenAI)] Implemented the targeted scrubber fix in 1f9e7a7ae9. The shared password/token rules are now split between quoted JSON strings and plain/YAML values, so JSON `null`, booleans, and numbers under sensitive-looking keys remain typed and valid while string secrets are still masked. With that fixed centrally, I removed the config CLI recursive leaf-wrapping implementation and restored a single scrub-then-parse pass. Scrubbing remains enabled; focused Linux config integration, `saluki-common`, ADP, Windows cross-checks, and `make test-all` pass.

bin/correctness/panoramic/src/actions/mod.rs
thieman: [GPT-5.6 Sol (OpenAI)] Completed a fresh test-by-test audit of the current reduced diff against the production-reachability standard. For every added test I identified the production branch and a concrete production mutation that would make it fail. The audit found and I applied these changes in 1cbc1861c6: deleted a duplicated stdout executor test; deleted two shallow ActionConfig serde mirrors; folded dotted missing-key behavior into the real retry test and removed the private-helper test; consolidated missing/mismatched TLS rejection into one production-acceptor test; removed the redundant Docker port probe; and strengthened cancellation coverage so removing `kill_on_drop(true)` fails. The remaining groups map to distinct production behavior: production IPC accept/reject, complete Rustls HTTP client identity/ALPN, spawned ADP config mTLS/output, target CLI selection/redaction/cancellation, config endpoint mapping/retry, live privileged routes/Core Agent gRPC, and Antithesis transport. Current `make test-all` and focused Linux integration pass.

bin/agent-data-plane/tests/config_cli.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    timeout(PROCESS_TIMEOUT, command.output())
+        .await
+        .expect("agent-data-plane config command should exit before timeout")
+        .expect("agent-data-plane config command should launch")
+}
+
+#[tokio::test]
+async fn json_config_views_use_the_expected_mtls_routes_and_emit_only_compact_scrubbed_json() {
```
thieman: Do a targeted review of this very complex test. What are you trying to verify here, and can you do so maybe with a series of smaller, simpler tests instead of this big one?

bin/agent-data-plane/src/cli/config.rs
thieman: [GPT-5.6 Sol (OpenAI)] Follow-up: aligned the naming in 39970b0947 so the translated view is now consistently `runtime` everywhere: `/config/runtime`, `config_runtime.rs`, `ConfigRuntime*`, `config-runtime-api`, Panoramic selectors, integration case names, and documentation. There is no `/config/internal` alias.

lib/saluki-common/src/scrubber.rs
```diff
@@ -149,25 +149,43 @@ impl Default for Scrubber {
...
 
-        // Capture the optional closing `"` as $4 so the replacement preserves it for JSON values without breaking
-        // unquoted values (plain text / YAML). Without $4, `"password":"secret"` → `"password":"********` (invalid JSON).
-        // `:[ ]?` matches both compact JSON (`"password":"secret"`) and spaced YAML (`password: secret`).
-        let password_replacer = Replacer {
-            regex: Some(Regex::new(r#"(?i)(\"?(?:pass(?:word)?|pswd|pwd)\"?)((?:=| = |:[ ]?)\"?)([0-9A-Za-z#!$%&'()*+,\-./:;<=>?@\[\\\]^_{|}~]+)(\"?)"#).unwrap()),
-            repl: Some(b"$1$2********$4".to_vec()),
+        // JSON string values need both quotes to be part of the match. Keeping the key and separator captures while
+        // replacing the complete value string prevents non-string JSON values from being mistaken for secrets.
+        let json_password_replacer = Replacer {
```
thieman, line 154: Confirm that we added (or already have) sufficient regression tests for all expected JSON types with and without scrubbing

bin/agent-data-plane/tests/config_cli.rs
thieman: [GPT-5.6 Sol (OpenAI)] The original test combined four concerns: launching the real ADP binary, proving it presents the IPC identity to the production mTLS server, checking source/runtime route selection, and checking machine-readable scrubbed output. I split it in 5867500992. A shared one-request fixture now owns only the unavoidable real-process/certificate/server setup. `source_json_config_emits_compact_scrubbed_stdout_over_production_mtls` verifies authentication plus the compact/scrubbed stdout contract; `runtime_json_config_selects_the_runtime_route_and_response` verifies `--runtime` selects `/config/runtime`. The server no longer contains multi-route dispatch, and the file is 22 lines shorter.

lib/saluki-common/src/scrubber.rs
thieman: [GPT-5.6 Sol (OpenAI)] Confirmed and expanded in 5867500992. The structural matrix now covers every JSON value category: null, boolean, number, array, and object values remain valid and structurally unchanged; an ordinary string remains unchanged; and nested sensitive string fields inside arrays/objects are still scrubbed. A separate direct-string test covers password/token/JWT values. Existing dedicated JSON tests continue to cover API keys, URI passwords, Bearer tokens, compact/spaced JSON, and large single-line payloads. `saluki-common` (116 tests), ADP (226 tests), xwin, and `make test-all` pass.

lib/saluki-common/src/scrubber.rs
```diff
@@ -149,25 +149,43 @@ impl Default for Scrubber {
...
             hints: None,
             repl_func: None,
         };
 
-        // Redacts the value of any key ending in `token` or `jwt` (for example, `auth_token`,
-        // `cluster_agent.auth_token`, `refresh_token`). Mirrors `password_replacer`: the trailing `"` in the
-        // key group plus the `$4` closing-quote capture keep compact JSON (`"auth_token":"x"`) valid, and the
-        // optional leading `"` lets the match start mid-key (so dotted keys like `cluster_agent.auth_token`
-        // match after the `.`). `hints` is intentionally `None`: the hint check is case-sensitive, so a
-        // `"token"` hint would skip an uppercase `AUTH_TOKEN` key that `(?i)` would otherwise match.
```
webern, line 167: I don't like it when the clankers delete these sorts of helpers instead of updating them.

webern (approved):
This looks clean to me. Nice.

One note, the `/config` and `/config/internal` (which I think you renamed to `/config/runtime` duplication is due to the dual typed/untyped state of configuration. Eventually we will only support one of these (and it will be that `/config/internal|runtime` will be promoted to `/config`. I hope this will be soon before too many things can depend on it.

lib/saluki-common/src/scrubber.rs
thieman: [GPT-5.6 Sol (OpenAI)] The regex-specific part of the old comment was inaccurate after splitting one replacer into JSON and plain/YAML variants: there is no longer an optional `$4` quote capture, and JSON matching now deliberately requires a quoted string value. However, the explanation of supported token/JWT key shapes, dotted keys, and why hints remain disabled was still useful. I should have rewritten that portion rather than dropping it. Restored updated comments for both replacers in 25219f86f3.

tobz (approved):
Seems reasonable enough to me if tests are passing. 👍🏻

PR #2244, opened 2026-07-31
chore(config): align configuration more closely with Datadog Agent to remove previous compatibility hacks

tobz:
## Summary

This PR aligns our configuration loading/layout more closely with the Datadog Agent configuration (and, honestly, idiomatic Rust) by removing the previous hacks in place to allow properly handling the different sources that configuration data could be loaded from.

Prior to this PR, our configuration had built up a number of patches, hacks, and contortions that were put in place to allow ADP/Saluki to read a Datadog Agent configuration -- whether from YAML, environment variables, and so on -- and handle them in the same way that the Datadog Agent itself would. This mean things like how a configuration setting key would be mapped to an environment variable, and then how we would ensure that the setting made it to the right field on the right struct after deserialization. This _worked_, but it was clunky and fragile and ultimately a consequence of building this system prior to the existence of the configuration schema that the Datadog Agent now exports.

This PR replaces our previous hacks with a more straightforward solution. In no particular order:

- remove the concept of "key aliases": we don't need them any more, because we're aligned to the natural nested structure of the incoming configuration
- remove `DatadogRemapper` as we no longer need to map nested keys to flattened keys
- add a new `DatadogEnvProvider` config provider that maps directly from known env vars (source from the Agent's config schema) to their nested paths)
- remove the "env overlay" machinery as it's now obsolete with `DatadogEnvProvider` is place
- updated various config structs that previously renamed their fields using flattened keys, or used `#[serde(flatten)]`, to use a normal nested layout instead where the keys are nested on the Datadog Agent side

We do still have some duplication around environment providers, where there's a generic Datadog Agent one in `datadog-agent-config` and a more-specific ADP one in `agent-data-plane-config-system`, which is a superset of the generic one. This is due to needing a way to properly inject environment variables into configs for tests. We can always clean this up in the future but it might require some more reworking to generalize the underlying code and I wanted to leave it out for now.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

## References

DADP-2


tobz:
* **#2244** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2244?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2244?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/saluki-components/src/common/datadog/data_plane.rs
```diff
@@ -0,0 +1,78 @@
...
+#[derive(Clone, Debug, Default, Deserialize, Facet)]
+#[cfg_attr(test, derive(PartialEq, serde::Serialize))]
+pub(crate) struct EncoderDataPlaneConfiguration {
+    /// ADP-specific zstd compression level, taking precedence over the Core Agent's
+    /// `serializer_zstd_compressor_level`.
+    ///
+    /// Defaults to unset, in which case
+    /// [`resolve_zstd_compressor_level`](super::resolve_zstd_compressor_level) picks the level.
+    #[serde(default)]
+    pub(crate) serializer_zstd_compressor_level: Option<i32>,
```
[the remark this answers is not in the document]
tobz: Fixed in 4dff1295efab38ca9eef38d0e82f8f9a4ad41f23.

lib/agent-data-plane-config/src/shared.rs
```diff
@@ -126,8 +126,23 @@ pub struct Compression {
     /// Which compression algorithm the encoder uses.
     pub compressor_kind: String,
 
-    /// Compression level used when the algorithm is zstd.
+    /// Compression level used when the algorithm is zstd, as the Core Agent configures it.
```
lucastemb, line 129: nit: Non blocking, but I know that in the past, it's been said we should avoid mentioning the Core Agent in our comments. Obviously, since this feature relates to reconciling configuration options that we are siphoning from upstream, it makes sense, but I guess for future reference, where do we draw the line?

lib/datadog-agent/config/src/env_provider.rs
```diff
@@ -0,0 +1,125 @@
+//! A Figment provider that reads the Datadog schema's environment variables into their canonical
```
lucastemb, line 1: Non-blocking, but perhaps worth a follow-up PR? Maybe I'm missing something, but this looks a lot similar to the earlier implementation in `agent-data-plane-config-system/src/env_provider.rs`. Is there any reason that this is duplicated?

lucastemb (approved):
Just left a few non-blocking comments, but overall, everything looks sound to me.

lib/agent-data-plane-config/src/shared.rs
tobz: In this, ADP is the public artifact tied to the Datadog Agent, so we're fine to mention it here.

lib/datadog-agent/config/src/env_provider.rs
tobz: Yep, this is covered in the PR description:

> We do still have some duplication around environment providers, where there's a generic Datadog Agent one in `datadog-agent-config` and a more-specific ADP one in `agent-data-plane-config-system`, which is a superset of the generic one. This is due to needing a way to properly inject environment variables into configs for tests. We can always clean this up in the future but it might require some more reworking to generalize the underlying code and I wanted to leave it out for now.

webern:
Progresses #2169

PR #2252, opened 2026-08-03
feat(datadog): add stateless Foldspace retry fallback

aqian01:
## Summary

- add endpoint-scoped reliable Foldspace senders for logs after endpoint routing, with one outstanding stateful payload per logical sender
- convert retained stateful payloads into complete stateless log transactions before transport retries, high-priority overflow retries, shutdown flushing, or disk persistence
- reconstruct each failed payload from the checked snapshot as of `payload_id.prev()`, applying the seed and retained wire datums as separate decoder batches
- preserve log fields, UUIDs, endpoint path, API-key routing, request headers, compression, and transaction metadata
- keep endpoint/API-key state and ACK handling independent, and never create or rotate Foldspace streams for ordinary stateless retry attempts
- add explicit conversion, error, and oversized-drop telemetry together with focused stream, race, persistence, and recovery tests
- add the opt-in `logs_config.use_grpc` configuration and generated configuration documentation

## Why

ADP prioritizes fresh payloads over retries. Newer stateful log payloads can therefore redefine Foldspace patterns before an older failed payload is retried. Replaying the old stateful bytes against the latest stream state can reconstruct the wrong logs.

Before any stateful payload enters the existing retry or persisted queues, this change uses Foldspace's retained-payload recovery API to rebuild the original logical logs and creates an ordinary self-contained transaction accepted by the same intake endpoint. The payload leaves stateful replay only through that conversion path, and its changelog retention is released only after retry-queue acceptance or an explicit terminal drop.

Permanent HTTP failures keep the existing drop behavior. If stateless expansion exceeds retry-queue limits, ADP records an oversized conversion and drops it; it never falls back to enqueueing stateful bytes.

## Foldspace dependency

- recovery API draft: DataDog/foldspace#30
- pinned revision: `c585fd1fc09758eb61e133b0e4e023a01c82cb26`
- #30 is stacked on Ryan's POC in DataDog/foldspace#23

## Initial POC scope and durability

- logs only
- `DispatchPolicy::Reliable` only
- at most one outstanding stateful payload per logical sender
- separate Foldspace state per endpoint/API-key destination

Persisted entries are complete stateless transactions and disk recovery has no Foldspace stream-state dependency. This does not provide full crash durability: a process crash before conversion can still lose an unacknowledged stateful payload, and the persisted queue retains its existing dequeue crash window.

## Checks

- `cargo nextest run -p saluki-components stateful_logs` (9 passed)
- `cargo nextest run -p agent-data-plane-config-system saluki_only` (5 passed)
- `cargo nextest run -p datadog-agent-config-testing` (4 passed)
- `cargo check -p saluki-components --tests`
- `cargo check --workspace`
- `make build-schema-overlay`
- `make fmt`
- `make check-all`


lib/agent-data-plane-config-system/src/saluki_only.rs
```diff
@@ -213,6 +215,14 @@ pub struct DataPlaneChecks {
     pub enabled: Option<bool>,
 }
 
+/// `logs_config.*` transport settings absent from the vendored Datadog schema.
+#[derive(Clone, Debug, Default, Deserialize)]
+#[serde(default)]
+pub struct LogsConfig {
+    /// Whether stateful Foldspace transport is enabled (`logs_config.use_grpc`).
+    pub use_grpc: Option<bool>,
```
webern, line 223: Could this be named more clearly? `use_grpc` does not imply, to me, `enable_foldspace`.

Issue #2254, opened 2026-08-03
Add Cluster Agent metadata support for ADP static tags


lucastemb:
ADP currently has narrow Cluster Agent support for forwarding autoscaling-failover metric series to the Cluster Agent `/series` endpoint. It does not provide a general authenticated read client for Cluster Agent metadata.

As a result, ADP’s OTLP metrics static-tag path does not add Cluster-Agent-derived tags such as `orch_cluster_id`.

This is non-blocking for OTLP metrics ingest: ADP can receive, translate, tag, and forward OTLP metrics without Cluster Agent communication. We should revisit this when ADP has broader Cluster Agent client support, including endpoint discovery, authentication, retries, and a defined metadata API surface.

**Out of scope for this issue:**

- OTLP ingestion, translation, and forwarding behavior.
- Existing autoscaling-failover Cluster Agent forwarding.
- Adding ad hoc Cluster Agent calls solely for one static tag.

> **Note:** This issue draft was AI-assisted. A human reviewer should validate the stated architecture, scope, and proposed follow-up before implementation.

PR #2255, opened 2026-08-03
feat(config): add provider_kind and EKS Fargate static tags

### LINKED ISSUE #2066: Add provider_kind and EKS Fargate static tags to OTLP metrics

## Background

The Agent adds deployment-specific static tags to every OTLP metric. When `provider_kind` is set,
it adds a `provider_kind:<value>` tag. When it runs as an EKS Fargate sidecar, it adds configured
`DD_TAGS` and `DD_EXTRA_TAGS` plus EKS-specific tags. ADP does not currently apply either behavior
to OTLP metrics.

### Agent behavior

`GetStaticTagsSlice` in `pkg/util/tags/static_tags.go:82-95` appends
`provider_kind:<value>` when the `provider_kind` config key is non-empty. The key is bound to
`DD_PROVIDER_KIND` in `pkg/config/setup/common_settings.go:793`. This branch is independent of
Fargate detection.

The same function calls `getFargateStaticTags` when `fargate.IsSidecar()` is true. That function
starts with `configUtils.GetConfiguredTags`, which concatenates the `tags` and `extra_tags` config
lists (`pkg/config/utils/tags.go:13-29`; `DD_TAGS` and `DD_EXTRA_TAGS`). For EKS Fargate it then
adds `eks_fargate_node:<node>`, preserves a manually supplied `kube_cluster_name:<name>` or adds
the discovered cluster name, optionally adds `orch_cluster_id:<id>`, and adds
`kube_distribution:eks` (`pkg/util/tags/static_tags.go:25-75`). EKS Fargate is enabled through the
`eks_fargate` config key / `DD_EKS_FARGATE` feature detection
(`pkg/config/env/environment_containers.go:222-226`).

`comp/otelcol/otlp/config.go:91-107` joins the static tag slice with commas and writes it to
`otlp_config.metrics.tags`, replacing the configured value when static tags are present. The
serializer exporter splits the value and prepends the resulting tags to every series and sketch in
`comp/otelcol/otlp/components/exporter/serializerexporter/consumer.go:132-182`.

### agent-data-plane behavior

ADP's `MetricsConfig` in `lib/saluki-components/src/common/otlp/config.rs:162-168` currently has
only an `enabled` field. It has no `tags` field, and `otlp_config.metrics.tags` is marked ignored
in `lib/datadog-agent/config/schema/schema_overlay.yaml:3738`. ADP therefore has no OTLP static-tag
configuration or attachment path.

ADP already handles `provider_kind` for DogStatsD. `DogStatsDConfiguration` declares the field in
`lib/saluki-components/src/sources/dogstatsd/mod.rs:632-639`, and `additional_tags` appends
`provider_kind:<value>` in `:702-711`. The schema inventory limits that support to the DogStatsD
pipeline (`lib/datadog-agent/config/schema/schema_overlay.yaml:1897-1903`); no equivalent OTLP
path exists.

ADP also has no EKS Fargate sidecar detection, `DD_TAGS`/`DD_EXTRA_TAGS` global-tag mechanism,
or derivation of `eks_fargate_node`, `kube_cluster_name`, or `kube_distribution`. The only related
`AwsEcsFargateKind` code is based on incoming OTLP resource attributes and is not the Agent's
EKS-sidecar detection path.

After the tags plumbing from #2064 is available, the OTLP source can carry the configured and
derived tags from `OtlpConfiguration::build` in
`lib/saluki-components/src/sources/otlp/mod.rs:178-227` into the translator. The translator
constructs the per-resource/scope tag set in
`lib/saluki-components/src/sources/otlp/metrics/translator.rs:412-455`, before passing it to
`map_to_dd_format`. That is the existing attachment point for static tags on emitted metrics.

### Requirements tracking

- OM_CLDP_559
- OM_CLDP_193

## Definition of done

- [x] Add the OTLP static-tag plumbing required to consume `otlp_config.metrics.tags` after #2064.
- [x] Add `provider_kind:<value>` to every ADP OTLP metric when `DD_PROVIDER_KIND` is set.
- [x] Add EKS Fargate detection and reproduce the Agent's configured and derived static tags,
      including `DD_TAGS`, `DD_EXTRA_TAGS`, `eks_fargate_node`, manual cluster-name precedence, and
      `kube_distribution:eks`.
- [ ] Add red correctness coverage for provider kind and EKS Fargate static tags, then make it
      green against the Agent baseline.
- [ ] Preserve existing OTLP resource, scope, and datapoint tags and update the compatibility
      schema entries for the implemented configuration.

## References

- Depends on #2064
- Extends #2006
- Revisits #1640

lucastemb:
## Summary

Added `provider_kind` support and EKS Fargate related static tags as part of the ongoing OTLP Ingest metrics pipeline related work. 

Support will be pipeline agnostic _(meaning it will also apply to DogStatsD)_ since investigation reveals that this is not scoped simply to just the OTLP pipeline.

> We have explicitly chosen to omit some Cluster Agent related tags (now tracked by https://github.com/DataDog/saluki/issues/2254) due to ADP's current narrow implementation and non-blocking nature to our current work

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests. 

## References

- Closes: #2066


lib/saluki-components/src/sources/otlp/mod.rs
```diff
@@ -64,6 +64,50 @@ fn parse_configured_metric_tags(raw: &str) -> SharedTagSet {
     tags.into_shared()
 }
 
+/// Resolves static tags that the Core Agent adds to OTLP metrics.
```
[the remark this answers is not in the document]
lucastemb: Fixed in [0b443fa](https://github.com/DataDog/saluki/pull/2255/commits/0b443fae413163e4cb31d75c9f6fc09417162b4b)

lib/agent-data-plane-config/src/domains/otlp.rs
```diff
@@ -48,6 +48,18 @@ pub struct Metrics {
     /// Comma-separated list of tags to add to every emitted metric.
     pub tags: String,
 
+    /// Static provider classification added to every OTLP metric when non-empty.
+    pub provider_kind: String,
+
+    /// Whether the Agent runs as an EKS Fargate sidecar.
+    pub eks_fargate: bool,
```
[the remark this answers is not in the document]
lucastemb: Fixed in [0b443fa](https://github.com/DataDog/saluki/pull/2255/commits/0b443fae413163e4cb31d75c9f6fc09417162b4b)

lib/saluki-components/src/sources/otlp/mod.rs
```diff
@@ -77,11 +121,17 @@ pub struct OtlpConfiguration {
...
 
 impl OtlpConfiguration {
-    /// Creates a new `OtlpConfiguration` from the resolved OTLP configuration.
-    pub fn from_configuration(otlp: &domains::otlp::Domain) -> Self {
+    /// Creates a new `OtlpConfiguration` from the resolved OTLP and shared configuration.
+    pub fn from_configuration(otlp: &domains::otlp::Domain, shared: &shared::SharedConfiguration) -> Self {
+        let mut otlp = otlp.clone();
+        let static_tags = resolve_static_metric_tags(&otlp, &shared.tags, |key| std::env::var(key).ok());
+        if !static_tags.is_empty() {
+            otlp.metrics.tags = static_tags.join(",");
```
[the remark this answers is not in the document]
lucastemb: That is intentional parity as seen [here](https://github.com/DataDog/datadog-agent/blob/main/comp/otelcol/otlp/config.go#L102-L105). We're _correctly_ overwriting `otlp_config.metrics.tags` when resolved static tags are non-empty.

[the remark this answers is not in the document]
lucastemb: This is addressed in the comment below. The Codex bot is effectively pointing out the same thing.

lib/saluki-components/src/sources/otlp/mod.rs
```diff
@@ -64,6 +64,57 @@ fn parse_configured_metric_tags(raw: &str) -> SharedTagSet {
...
+    tags.extend(global_tags.extra_tags.iter().cloned());
+
+    if !otlp.metrics.eks_fargate {
+        return tags;
+    }
+
+    if !otlp.metrics.kubernetes_kubelet_nodename.is_empty() {
+        tags.push(format!("eks_fargate_node:{}", otlp.metrics.kubernetes_kubelet_nodename));
+    } else {
+        warn!("Couldn't build the 'eks_fargate_node' tag: kubernetes_kubelet_nodename is not configured.");
```
tobz: Is this log message (and the equivalent one below) meant to explicitly match a log message from the Core Agent?

It might be good to try and reword it to make it more friendly to users, in a "problem-solution" pattern:

```
Tag 'eks_fargate_node' will be missing from telemetry ingested via OTLP due to missing configuration data. Ensure 'kubernetes_kubelet_nodename' is set in the configuration.
```

If this is just something they can set (but more on that in a second), then we should explicitly be telling them not only what to set but _where_ to set it. As is, we're not being that explicit that this is a configuration setting.

Separately: is this actually something users are expected to set manually, or something that would be configured by the Helm chart or Operator? If it's the latter, then we actually might not want to have the "solution" part if there's no way for them to act on it.

lib/saluki-components/src/sources/otlp/mod.rs
```diff
@@ -64,6 +64,57 @@ fn parse_configured_metric_tags(raw: &str) -> SharedTagSet {
...
+    E: Fn(&str) -> Option<String>,
+{
+    let mut tags = Vec::new();
+
+    if !otlp.metrics.provider_kind.is_empty() {
+        tags.push(format!("provider_kind:{}", otlp.metrics.provider_kind));
+    }
+
+    let ecs_fargate =
+        environment("AWS_EXECUTION_ENV").as_deref() == Some("AWS_ECS_FARGATE") || environment("ECS_FARGATE").is_some();
```
webern: This is bothering me. It's an environmental side-effect buried in an otherwise relatively pure function. I feel like this check belongs somewhere that is dedicated to setup (right?) and that its environment determination should be passed in to this function. I'm not sure if there's a "place" for that, but it jumped out at me as a kind of hidden form of config/environment sensitivity.

lucastemb: You're right. Fixed in c299b79. I made it such that the check for `ECS_Fargate` occurs much earlier and is propagated through instead of belonging to that hacky conditional statement you called out.

lib/saluki-components/src/sources/otlp/mod.rs
lucastemb: Seems like it's something that users set manually. I do think the problem-solution pattern for the error message is better than what I had and was added in [c299b79](https://github.com/DataDog/saluki/pull/2255/commits/c299b7903116d73c0611d9c4270e12e40f6e4b15) 😄

lib/saluki-env/src/features/mod.rs
```diff
@@ -53,6 +53,11 @@ fn is_env_var_present(name: &str) -> bool {
     !get_env_var_or_empty(name).is_empty()
 }
 
+/// Returns whether the process runs in an Amazon ECS Fargate environment.
+pub fn is_ecs_fargate() -> bool {
```
webern, line 57: Based on what else is in this file, such as "running in docker", this does seem like a reasonable place to put this function.

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -982,7 +998,12 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
     }
 
     fn consume_provider_kind(&mut self, value: String) {
-        self.config.domains.dogstatsd.listeners.provider_kind = value;
+        self.config.domains.dogstatsd.listeners.provider_kind = value.clone();
+        self.config.domains.otlp.metrics.provider_kind = value;
```
webern: This is a bit suspicious. I wonder if there is a better home for it in `config.shared`. I recall that early in the typed config process it was considered desirable that we avoid duplication of a value into two different places in the typed config structure.

webern (commented):
As I was looking at this, the major question that came up in my mind, which I'm having trouble answering is: why did it come up now that we need the `tags` and `extra_tags` configuration fields when we did not need these for DogStatsD.

My limited understanding of this is, ADP's `HostTagsConfiguration` is temporary because normally it's handled by the backend, but a window is needed before the Agent can sent the host and host_tags to the backend. After that, the backend associates those host tags with data from the host. 

These `tags` and `extra_tags` fields feel suspiciously related to that mechanism, and seemingly, to me, would behave the same way for OTLP-sourced data in the backend as they would for DSD-sourced data.

To state the halting point I have on this PR differently, or again...

It feels like either this behavior is needed for DogStatsD as well as OTLP, or that it is needed by neither of them. It is entirely possible that I am missing something though because this is a very confusing area of Agent/ADP contract.

Edit: After reading the above, AI is helpfully pointing out that:

> OTLP could legitimately differ if its metrics bypass the mechanism by which backend host association works for DogStatsD. That is probably the exact question you want need to answer.

tobz (approved):

PR #2271, opened 2026-08-05
feat(dogstatsd): allowlist tag values by metric prefix

lukesteensen:
## Summary

Add `metric_tag_value_allowlist`, which limits DogStatsD metric cardinality by retaining selected values for a tag and removing or replacing values outside the allow-list.

Rules select metrics using case-sensitive metric-name prefixes:

```yaml
metric_tag_value_allowlist:
  - metric_prefix: requests.
    tag_name: customer_id
    values: [customer-1, customer-2]
    on_miss: replace
    replacement: other
```

This change also moves the existing `metric_tag_filterlist` component from component-local deserialization and `GenericConfiguration` access to the typed DogStatsD configuration domain. Dynamic whole-tag updates continue through a narrow `Live<T>` view, while value allow-list rules remain static.

### Design decisions

- **Match final metric names.** Prefixes match after DogStatsD mapper rewrites and `statsd_metric_namespace` prefixing, so rules describe names as aggregation sees them.
- **Reject overlapping distinct prefixes.** Distinct prefixes cannot overlap, even when rules target different tags, so each metric resolves to at most one prefix. Multiple tags may share the exact same prefix; duplicate prefix-and-tag pairs are invalid.
- **Run whole-tag filtering first.** An `include` rule must retain the value-filtered tag name or the whole-tag rule removes it before value filtering.
- **Distinguish bare and empty-valued tags.** A bare tag such as `customer_id` has no value and remains unchanged. `customer_id:` has the value `""` and is subject to the allow-list.
- **Keep value rules static initially.** Changing `metric_tag_value_allowlist` requires an ADP restart. Remote Config updates to `metric_tag_filterlist` do not replace value rules.
- **Preserve the existing metric and tag scope.** Rules apply to counters and sketch-backed metrics, including instrumented and origin tags. Other metric types pass through unchanged.
- **Validate configuration strictly.** ADP rejects empty prefixes and tag names, `:` in tag names, overlapping distinct prefixes, and duplicate prefix-and-tag pairs. Configured strings preserve surrounding whitespace. Empty replacements and replacement collisions remain operator policy choices.
- **Index value rules by metric prefix.** Prefixes are sorted and binary-searched once per metric, then rules are looked up by tag name within the matching prefix. Matching outcomes are cached, while rule misses are not cached so unrelated traffic cannot evict policy-relevant entries.
- **Count filtered originals in telemetry.** A replaced tag contributes to the filtered-tag counter; replacement insertions are not counted separately.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

- `make fmt`
- `cargo check --workspace`
- `make test-all`
  - 2,520 unit tests passed
  - 26 property tests passed
  - Workspace doctests passed
  - 49 Miri tests passed
  - 8 Loom tests passed
- Full-workspace Clippy passed.
- Configuration-system and configuration smoke tests passed.
- Dependency advisory, license, unused-dependency, and feature-matrix checks passed.
- Vale passed on the documentation changed by this PR.
- The assembled DogStatsD topology test verifies that a real allow-list retains a listed value and removes an unlisted value after mapper and namespace processing.
- Existing `dsd-tag-filterlist` correctness case
  - Baseline and comparison emitted the same 178 unique non-internal metrics.
- `git diff --check`

`make check-all` passes formatting and full-workspace Clippy, then stops on an existing Vale spelling error in unchanged `docs/development/contributing.md` (`PR's`). The remaining check targets were run separately and passed.

## References

N/A


lib/agent-data-plane-config-system/src/saluki_only.rs
```diff
@@ -129,6 +130,8 @@ pub struct SalukiOnly {
     pub dogstatsd_minimum_sample_rate: Option<f64>,
     /// Mapper string interner entry count (`dogstatsd_mapper_string_interner_size`).
     pub dogstatsd_mapper_string_interner_size: Option<u64>,
+    /// Per-metric tag value allow-list rules (`metric_tag_value_allowlist`).
+    metric_tag_value_allowlist: Option<Vec<RawMetricTagValueAllowlistEntry>>,
```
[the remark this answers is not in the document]
lukesteensen: Confirmed and fixed in f92621d5c3. The Saluki-only environment tracer treated every sequence as a whitespace-separated string list, so this structured sequence could not load from DD_METRIC_TAG_VALUE_ALLOWLIST. I added an explicit serde newtype marker for JSON-encoded structured sequences; the tracer maps only marked sequences to EnvDecode::JsonValue, while existing Vec<String> leaves retain StringList behavior. An end-to-end regression test now loads a JSON object array from DD_METRIC_TAG_VALUE_ALLOWLIST and asserts the typed DogStatsD rule, including values, replace action, and replacement. Config-system tests, config smoke tests, workspace test compilation, Clippy, formatting, and diff checks pass.

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -793,7 +789,8 @@ async fn add_dsd_pipeline_to_blueprint(
     let dsd_mapper_config = DogStatsDMapperConfiguration::from_configuration(config)?;
     let dsd_enrich_config =
         ChainedConfiguration::default().with_transform_builder("dogstatsd_mapper", dsd_mapper_config);
-    let dsd_tag_filterlist_config = TagFilterlistConfiguration::from_configuration(config)
+    let dogstatsd_config = config_system.live(|config| &config.domains.dogstatsd);
+    let dsd_tag_filterlist_config = TagFilterlistConfiguration::from_configuration(dogstatsd_config)
```
webern, line 763: Good change. Switching from `GenericConfiguration` to `Live<dogstatsd::Domain>`.

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -1019,11 +1023,6 @@ mod tests {
                 "statsd_metric_namespace_blocklist": [],
                 "metric_filterlist": ["tenant.raw.blocked"],
                 "metric_filterlist_match_prefix": false,
-                "metric_tag_filterlist": [{
-                    "metric_name": "tenant.mapped.requests",
-                    "action": "exclude",
-                    "tags": ["remove"]
-                }],
```
webern, line 998: Makes sense because Metric Tag Filter list switched to typed config. 👍

lib/agent-data-plane-config-system/src/saluki_env_overlay.rs
```diff
@@ -182,11 +182,17 @@ impl<'de> Deserializer<'de> for PathRecorder<'_, '_> {
         visitor.visit_some(self)
     }
 
-    fn deserialize_newtype_struct<V>(self, _name: &'static str, visitor: V) -> Result<V::Value, Self::Error>
+    fn deserialize_newtype_struct<V>(mut self, name: &'static str, visitor: V) -> Result<V::Value, Self::Error>
```
webern, line 188: I don't understand this change but trust that the coding agent does! (For context this is how we derive environment variable keys from nested struct keys).

lib/agent-data-plane-config-system/src/saluki_only.rs
```diff
@@ -66,16 +66,57 @@
...
+use agent_data_plane_config::domains::dogstatsd::{MetricTagValueAllowlistEntry, TagValueMismatchAction};
 use agent_data_plane_config::domains::traces::{OttlErrorMode, OttlFilter, OttlTransform};
 use agent_data_plane_config::SalukiConfiguration;
 use bytesize::ByteSize;
 use saluki_config::DurationString;
+use serde::de::Visitor;
 use serde::{Deserialize, Deserializer};
 
+// The environment path recorder recognizes this serde newtype marker and decodes the leaf as JSON.
+pub(crate) const JSON_SEQUENCE_MARKER: &str = "SalukiJsonSequence";
```
webern, line 81: Oh, I see we are encoding JSON as a value in the environment variable and need a way to recognize that during deserialization.

bin/agent-data-plane/src/components/tag_filterlist/mod.rs
```diff
@@ -293,12 +407,14 @@ impl Transform for TagFilterlist {
                     }
                     None => break,
                 },
-                (_, new_entries) = watcher.changed::<Vec<MetricTagFilterEntry>>() => {
-                    self.filters = compile_filters(new_entries.as_deref().unwrap_or(&[]));
+                new_entries = self.entries.changed() => {
+                    self.filters = compile_all_filters(&new_entries, &self.value_allowlists)
+                        .expect("the static value allow-list was validated when the transform was built");
```
webern: This makes me a little bit nervous. If the validation at the config level is sufficient, then ideally the `compile_all_filters` function would not need to return a `Result`. But if it legitimately can fail because the config layer cannot be expected to sufficiently validate incoming filters, then something better than panicking probably needs to be done.

I think the problem to resolve is this:
- If the config layer can't completely validate the incoming filter list, and
- later, the component determines the list to be invalid...

There isn't really a graceful way out. You have accepted configuration that the component cannot express. If that's really the case then perhaps a panic *is* the right thing to do. But a better solution would be to try and make sure a bad config cannot be accepted by the system.

Note that, the way the config system is set up right now, on system startup, bad config fails the system. Later once we are serving, a bad partial config update (i.e. a remote config change) will report an error but we will continue running on the last good config.

webern (changes requested):
This looks great over all, except for one sticking point which is what should happen if bad config reaches the component (currently a panic).

Also I have a question. Is this feature development that is mirroring an Agent behavior? Or is this "green field" feature development on the ADP-side only?

This is an important question for testing because currently we would use a correctness test for integration to see that we match Agent behavior. But we don't have a great mechanism for testing greenfield behavior at the moment.

bin/agent-data-plane/src/components/tag_filterlist/mod.rs
```diff
@@ -29,71 +36,56 @@ use saluki_core::{
...
+
+#[derive(Eq, PartialEq)]
+enum CompiledMismatchAction {
+    Remove,
+    Replace(Tag),
 }
 
-use self::telemetry::Telemetry;
+struct CompiledTagValueAllowlist {
+    allowed_values: HashSet<String, FoldHashState>,
```
tobz: Can we just import `FastHashSet` and use the alias directly here? I realize we're explicitly importing `hashbrown` versions here, vs the recommended aliases pointing at stdlib versions, but it doesn't seem like we actually use anything from `hashbrown` that we can't get from stdlib, so...

(Same with `HashMap` further down.)

bin/agent-data-plane/src/components/tag_filterlist/mod.rs
```diff
@@ -112,7 +104,10 @@ pub enum FilterMetricTagsOutcome {
 /// - Same metric name + same action → union of tag key sets.
 /// - Same metric name + conflicting actions → `exclude` wins.
 pub fn compile_filters(entries: &[MetricTagFilterEntry]) -> CompiledFilters {
-    let mut filters: CompiledFilters = HashMap::with_hasher(FoldHashState::default());
+    let mut filters = CompiledFilters {
```
tobz: We should just be able to derive `Default` on `CompiledFilters` and then use `CompiledFilters::default()` here.

bin/agent-data-plane/src/components/tag_filterlist/mod.rs
```diff
@@ -142,14 +140,132 @@ pub fn compile_filters(entries: &[MetricTagFilterEntry]) -> CompiledFilters {
...
 
+/// Adds value allow-list entries to a compiled filter table.
+///
+/// Metric prefixes for the same tag must not overlap, ensuring that a metric/tag pair matches at most one rule.
+/// Rules for different tags may use the same or overlapping prefixes.
+///
+/// # Errors
+///
+/// Returns an error when two entries for the same tag have overlapping metric prefixes.
+pub fn add_value_allowlists(
```
tobz, line 154: A general comment on the logic here: can we just do all of the string trimming automatically on behalf of the user?

bin/agent-data-plane/src/components/tag_filterlist/mod.rs
tobz: This makes _me_ nervous because I don't actually think it's correct?

If we assume that we validated the value allowlist at first by virtue of making it to this point, there's still the logic in `add_value_allowlists` that checks for metric prefix overlap, which might occur at runtime if the new incoming metric tag filter configuration changes significantly?

Maybe I'm misreading it, but that looks sus to me... and it's not immediately clear how to reconcile that.

tobz (changes requested):

bin/agent-data-plane/src/components/tag_filterlist/mod.rs
lukesteensen: Yeah, tried to rework this to be clearer. I won't claim my understanding of the config system is 100% yet, but we _should_ now be fully parsing and validating in the config system itself. Runtime updates should also end up replacing the full key after validation, so partial updates can't introduce an overlap.

bin/agent-data-plane/src/components/tag_filterlist/mod.rs
lukesteensen: I revisited this and got rid of the trimming to be consistent with what we do for the full tag feature. Seemed a little overly aggressive on second thought.

webern (approved):
Nice changes. LGTM.

bin/agent-data-plane/src/components/tag_filterlist/mod.rs
```diff
@@ -309,29 +381,108 @@ impl Transform for TagFilterlist {
...
 pub fn filter_metric_tags(
     metric: &mut Metric, state: &mut TagSetMutViewState, filters: &CompiledFilters,
 ) -> FilterMetricTagsOutcome {
-    let Some((is_exclude, tag_names)) = filters.get(metric.context().name().as_ref()) else {
+    let metric_name = metric.context().name().clone();
+    let key_filter = filters.key_filters.get(metric_name.as_ref());
+    let has_value_rule = filters
+        .value_prefix_filters
+        .iter()
+        .any(|filter| metric_name.starts_with(&filter.metric_prefix));
```
tobz: The value prefix filters are supposed to be deduplicated at compilation time, right?

If so, it seems like we should actually be getting the reference to the `CompiledValuePrefixFilter` once here, and then having `filter_tag` dive through `value_allowlists` to match whatever tag is being processed.

Saves a _little_ bit of work, at least.

bin/agent-data-plane/src/components/tag_filterlist/mod.rs
```diff
@@ -309,29 +381,108 @@ impl Transform for TagFilterlist {
...
-    tag_set_view.retain_tags(|tag| should_keep_tag(tag, *is_exclude, tag_names));
-    tag_set_view.retain_origin_tags(|tag| should_keep_tag(tag, *is_exclude, tag_names));
+    tag_set_view.retain_tags(|tag| match filter_tag(tag, key_filter, metric_name.as_ref(), filters) {
+        TagFilterAction::Keep => true,
+        TagFilterAction::Remove => false,
+        TagFilterAction::Replace(replacement) => {
+            tag_replacements.push(replacement.clone());
+            false
+        }
+    });
```
tobz: Could we maybe just make `filter_tag` take `&mut Vec<Tag>`, populate the replacement value if necessary, and return the boolean result directly?

Then we could go back to two lines instead of this duplicated logic here.

tobz (changes requested):

bin/agent-data-plane/src/components/tag_filterlist/mod.rs
lukesteensen: I had actually reworked this a bit, but I realized that commit went into the next PR in the stack. And then when I looked closer, it was more complex than it needed to be because it allowed overlapping prefixes if they applied to different tags. So I brought that commit into this PR and simplified it a bit more, no overlapping prefixes at all, and should no longer have the n^2 compile issue that was present here.

tobz (approved):

PR #2275, opened 2026-08-05
enhancement(dogstatsd): batch datagram decoder dequeues

thieman:
## Summary

- Dequeue up to 32 queued DogStatsD datagrams per shared receiver lock acquisition.
- Amortize Tokio MPSC receive, async mutex, wakeup, and scheduling overhead across a small batch.
- Preserve the existing global queue, decoder worker pool, per-packet socket context, and shutdown draining behavior.

This is stacked on #2079.

`Receiver::recv_many` wakes as soon as one item is available; it does not wait for the batch to fill. Low-volume traffic therefore follows the same wait-for-one-message behavior as before.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Performance

### Automated validation

- `make fmt`
- `cargo nextest run -p saluki-components dogstatsd`
  - 154 tests passed
- `cargo xwin check --target x86_64-pc-windows-msvc -p saluki-components --tests`
- Repository pre-commit checks, including workspace Clippy and API documentation build

### Closed-loop UDS throughput benchmark

I used a local experimental Docker/Lading harness rather than committing additional benchmark infrastructure in this PR. The benchmark used:

- the converged Agent image, toggling only Core versus ADP DogStatsD ownership;
- UDS datagrams;
- the existing deterministic 3,000-context Lading DogStatsD profile;
- a two-CPU target (`0-1`) and separate two-CPU Lading generator (`2-3`);
- two decoder workers;
- a 2 GiB target memory limit;
- five seconds of warm-up;
- a 30-second measurement at a saturated 1,600 MiB/s requested rate;
- blocking Lading sends, so overload appears as reduced achieved throughput rather than client drops.

| Run | One-at-a-time dequeue | Batched dequeue |
|---|---:|---:|
| 1 | 956.2 MiB/s | 1,026.6 MiB/s |
| 2 | 979.3 MiB/s | 1,065.7 MiB/s |
| Mean | **967.7 MiB/s** | **1,046.1 MiB/s** |

Mean throughput improved **8.1%**. Mean packet rate increased from approximately **263k datagrams/s** to **284k datagrams/s**.

I also prototyped Linux `recvmmsg` with per-message UDS credentials. It improved the realistic profile by approximately 0.9% and a small-datagram profile by approximately 1.9%, within local variance, so that additional unsafe and platform-specific code is not included.

### Trade-offs

- A worker may hold up to 32 messages outside the global channel while decoding them. Highly heterogeneous per-packet decode costs could therefore create short-lived worker load imbalance.
- The 100 ms event-buffer flush timer is not polled while a batch is decoded. At the measured packet rate, a 32-packet batch represents substantially less than one millisecond of work.
- Datagram completion order is already not guaranteed across parallel decoder workers; batching does not introduce a new ordering guarantee.
- The batch vector holds at most 32 existing queued messages per worker and does not allocate additional packet buffers.

## References

- #2079
- [DADP-150](https://datadoghq.atlassian.net/browse/DADP-150)

[DADP-150]: https://datadoghq.atlassian.net/browse/DADP-150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aqian01 (approved):

bin/correctness/millstone/src/main.rs
webern, line 76: I found this description slightly lacking. Here is a more suggested, fuller description.

> When a completion file is requested, Millstone writes a marker after to signal that it has finished sending, then remains alive instead of exiting. This lets callers detect that sending is complete while preserving the sender’s process identity for origin detection.

bin/correctness/panoramic/src/correctness/runner.rs
webern, line 48: Writing this comment after the suggestion in main.rs. I am slightly confused as to whether millstone will wait indefinitely when writing a completion file or whether it is interacting with `FLUSH_WAIT` somehow.

lib/saluki-components/src/sources/dogstatsd/origin.rs
webern, line 22: Nit: just because Pinned and Unpinned mean something specific in Rust code.

```suggestion
    Resolved(Option<EntityId>),
    Unresolved(u32),
```

webern, line 23: nit: Pinned and Unpinned are overloaded because of the Rust meanings. Also I thought a bit of documentation could help the reader here.

```suggestion
/// The outcome of origin resolution at the time of message decoding.
#[derive(Clone, Debug, Eq, PartialEq)]
pub(super) enum ProcessOrigin {
    /// The resolver ran when the message was decoded.
    Resolved(Option<EntityId>),
    /// Only the PID is available, the resolver did not run.
    Unresolved(u32),
    /// The traffic is generated by replay, use the replay PID.
    Replay(u32),
```

lib/saluki-components/src/sources/dogstatsd/mod.rs
webern, line 1662: If I understand correctly, this provides backpressure before starting another receive cycle. Is it the allocations downstream from here that are causing the RSS regressions in the regression detector? Or is it the larger number of `BytesBuffers` themselves?

webern (commented):
Mostly nits other than the memory regression detector.

PR #2285, opened 2026-08-06
feat(otlp): infra-attribute host and entity tag enrichment

### LINKED ISSUE #2070: Implement OTLP infra-attribute host and entity tag enrichment.

## Background

The Agent enriches OTLP data with tags from the entities that produced it. It resolves entity IDs
from resource attributes, looks up tags, and applies global tags. ADP currently resolves only three
entity IDs and uses the first entity that returns tags. ADP does not yet support the Agent's full
entity resolution, tag merge behavior, global tags, or configuration for this feature.

### Agent behavior

The shared `infraTagsProcessor.ProcessTags` function is used by the metrics, logs, traces, and
profiles processors:

- `comp/otelcol/otlp/components/processor/infraattributesprocessor/common.go:79`
- Metrics call it from `metrics.go:45`; logs from `logs.go:41`; traces from `traces.go:35`.

Before tag lookup, the processor can synthesize a container ID when `container.id` is absent. The
fallback order is direct container ID, process ID, cgroup inode, and pod UID plus container name:
`comp/core/tagger/impl/tagger.go:276-333`.

It then checks these entity IDs in order: container ID, OCI manifest digest, ECS task ARN, Kubernetes
deployment plus namespace, Kubernetes namespace, Kubernetes node, Kubernetes pod UID, and process
PID (`comp/otelcol/otlp/components/processor/infraattributesprocessor/common.go:220-252`). It
merges tags from every matching entity with first-writer-wins precedence. Global tags are merged
last, so they apply even when no entity matches (`common.go:97-114`).

The processor config includes `cardinality`, `allow_hostname_override`, and the trace-only
`trace_container_tag_promotion` field (`config.go:47-77`). Metrics and logs pass container-tag
promotion as off.

### agent-data-plane behavior

ADP has shared origin-tag code for metrics and logs in
`lib/saluki-components/src/common/otlp/origin.rs:11-61`. Metrics reach it through the context
resolver (`sources/otlp/metrics/translator.rs:637-641`); logs call it directly
(`sources/otlp/logs/translator.rs:44-49`). Traces do not use the resolver.

`raw_origin_from_attributes` reads only literal `container.id`, `k8s.pod.uid`, and `process.pid`
values (`common/otlp/attributes/mod.rs:225-250`). The resolver checks those three values in that
order and returns after the first non-empty tag set (`common/otlp/origin.rs:23-46`). It does not
synthesize container IDs, merge all matching entities, or query `EntityId::Global` for tags.
Cardinality therefore defaults to `OriginTagCardinality::Low`.

`MetricsConfig` contains only `enabled` (`common/otlp/config.rs:162-167`). There is no
`allow_hostname_override` field. The compatibility schema marks `otlp_config.metrics.tags` and
`otlp_config.metrics.tag_cardinality` as ignored, and does not contain an
`allow_hostname_override` key (`lib/datadog-agent/config/schema/schema_overlay.yaml:3737-3738`).
The Agent's direct OTLP metrics pipeline also hardcodes low cardinality; the
`otlp_config.metrics.tag_cardinality` key is dead configuration and should not be wired merely for
parity.

The implementation should extend the shared OTLP code in `saluki-components`, use the existing
`WorkloadProvider` and `EntityId` types from `saluki-env`, and thread configuration through the
metrics, logs, and traces translators as appropriate. The exact design should be confirmed during
implementation.

### Requirements Tracking

- OM_TAGI_417
- OM_TAGI_053
- OM_TAGI_299
- OM_TAGI_088
- OM_TAGI_341
- OM_TAGI_182
- OM_TAGI_463

## Definition of Done

- [ ] Demonstrate the current gaps with red correctness coverage against the Agent baseline.
- [ ] Add shared OTLP entity extraction and tag resolution for metrics and logs, and make the
      resolver available to traces where required by the supported behavior.
- [ ] Add container-ID fallback resolution using process ID, cgroup inode, and pod UID plus
      container name where the ADP runtime can provide those inputs.
- [ ] Check the Agent's entity-ID list in the same order and merge all matching entity tags with
      the Agent's precedence rules.
- [ ] Apply global tags independently of whether an entity ID matches.
- [ ] Preserve the Agent's direct OTLP metrics behavior of low cardinality, and document whether
      ADP needs a separate user-facing cardinality override rather than wiring the dead
      `otlp_config.metrics.tag_cardinality` key.
- [ ] Add and wire `allow_hostname_override` if the supported ADP design requires it, including
      compatibility schema entries.
- [ ] Add unit coverage for entity extraction, fallback order, merge precedence, global tags, and
      configuration behavior.
- [ ] Add or extend a `kubernetes_in_docker` correctness case that exercises known workload entities
      and makes the Agent and ADP outputs green and equivalent.

## References

- Augments #2064
- Extends #2006
- Relates to #2016

lucastemb:
## Summary

As part of the ongoing work to reach behavioral parity for the OTLP Ingest metrics pipeline, we need to enrich data with tags from the entities that produced it. 

This PR aims to:

- Expand recognized entity ID types
- Establishes precedence rules 
- Merges tags following a "first-writer-wins" protocol (e.g. `container -> service:x` `pod -> service:y`, `service:x` will be the tag 
- Tag cardinality now recognized

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests
## References

- Closes: #2070

### DIFF
--- a/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
+++ b/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
@@ -936,6 +936,24 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
         }
     }
 
+    fn consume_otlp_config_metrics_tag_cardinality(&mut self, value: String) {
+        let cardinality = match value.to_ascii_lowercase().as_str() {
+            "low" => OriginTagCardinality::Low,
+            "orchestrator" => OriginTagCardinality::Orchestrator,
+            "high" => OriginTagCardinality::High,
+            "none" => OriginTagCardinality::None,
+            _ => {
+                self.record_error(TranslateError::new(
+                    "otlp_config.metrics.tag_cardinality",
+                    format!("unknown OTLP tag cardinality `{value}`; expected low, orchestrator, high, or none"),
+                ));
+                return;
+            }
+        };
+
+        self.config.domains.otlp.metrics.tag_cardinality = cardinality;
+    }
+
     fn consume_otlp_config_metrics_tags(&mut self, value: String) {
         self.config.domains.otlp.metrics.tags = value;
     }
@@ -1190,6 +1208,7 @@ mod tests {
             "dd_url": "https://custom.example.com",
             "dogstatsd_port": 9125,
             "dogstatsd_tag_cardinality": "high",
+            "otlp_config": { "metrics": { "tag_cardinality": "orchestrator" } },
             "expected_tags_duration": "15s",
             "telemetry": { "dogstatsd_origin": true },
         }))
@@ -1211,6 +1230,11 @@ mod tests {
         assert_eq!(
             config.domains.dogstatsd.origin.tag_cardinality,
             OriginTagCardinality::High
+        );
+        // The OTLP metrics cardinality uses the same source values but remains scoped to the OTLP metrics domain.
+        assert_eq!(
+            config.domains.otlp.metrics.tag_cardinality,
+            OriginTagCardinality::Orchestrator
         );
         // Driven `format: duration` parse: a Go duration string becomes a `Duration`.
         assert_eq!(config.shared.tags.expected_tags_duration, Duration::from_secs(15));
--- a/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
+++ b/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
@@ -605,21 +605,10 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
     }
 
     fn consume_dogstatsd_tag_cardinality(&mut self, value: String) {
-        // TODO: consider moving the enum to agent-data-plane-config
-        let cardinality = match value.to_ascii_lowercase().as_str() {
-            "low" => OriginTagCardinality::Low,
-            "orchestrator" => OriginTagCardinality::Orchestrator,
-            "high" => OriginTagCardinality::High,
-            "none" => OriginTagCardinality::None,
-            other => {
-                self.record_error(TranslateError::new_with_message(
-                    "dogstatsd_tag_cardinality",
-                    format!("unknown tag cardinality `{other}`"),
-                ));
-                return;
-            }
-        };
-        self.config.domains.dogstatsd.origin.tag_cardinality = cardinality;
+        match value.parse::<OriginTagCardinality>() {
+            Ok(cardinality) => self.config.domains.dogstatsd.origin.tag_cardinality = cardinality,
+            Err(error) => self.record_error(TranslateError::new("dogstatsd_tag_cardinality", error)),
+        }
     }
 
     fn consume_dogstatsd_tags(&mut self, value: Vec<String>) {
@@ -937,21 +926,10 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
     }
 
     fn consume_otlp_config_metrics_tag_cardinality(&mut self, value: String) {
-        let cardinality = match value.to_ascii_lowercase().as_str() {
-            "low" => OriginTagCardinality::Low,
-            "orchestrator" => OriginTagCardinality::Orchestrator,
-            "high" => OriginTagCardinality::High,
-            "none" => OriginTagCardinality::None,
-            _ => {
-                self.record_error(TranslateError::new(
-                    "otlp_config.metrics.tag_cardinality",
-                    format!("unknown OTLP tag cardinality `{value}`; expected low, orchestrator, high, or none"),
-                ));
-                return;
-            }
-        };
-
-        self.config.domains.otlp.metrics.tag_cardinality = cardinality;
+        match value.parse::<OriginTagCardinality>() {
+            Ok(cardinality) => self.config.domains.otlp.metrics.tag_cardinality = cardinality,
+            Err(error) => self.record_error(TranslateError::new("otlp_config.metrics.tag_cardinality", error)),
+        }
     }
 
     fn consume_otlp_config_metrics_tags(&mut self, value: String) {
--- a/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
+++ b/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
@@ -1231,7 +1231,6 @@ mod tests {
             config.domains.dogstatsd.origin.tag_cardinality,
             OriginTagCardinality::High
         );
-        // The OTLP metrics cardinality uses the same source values but remains scoped to the OTLP metrics domain.
         assert_eq!(
             config.domains.otlp.metrics.tag_cardinality,
             OriginTagCardinality::Orchestrator
--- a/lib/saluki-components/src/common/otlp/origin.rs
+++ b/lib/saluki-components/src/common/otlp/origin.rs
@@ -1,5 +1,6 @@
 use std::sync::Arc;
 
+use agent_data_plane_config::domains::dogstatsd::OriginTagCardinality as ConfigOriginTagCardinality;
 use otlp_protos::opentelemetry::proto::common::v1::{self as otlp_common, any_value::Value};
 use saluki_common::collections::FastHashSet;
 use saluki_context::{
@@ -22,6 +23,7 @@ const PROCESS_PID: &str = "process.pid";
 const CGROUP_INODE: &str = "datadog.container.cgroup_inode";
 const INIT_CONTAINER: &str = "datadog.container.is_init";
 
+/// Recognized OTLP resource attributes used for entity tag resolution.
 #[derive(Default)]
 struct OtlpResourceAttributes<'a> {
     container_id: Option<&'a str>,
@@ -78,11 +80,14 @@ impl OtlpOriginTagResolver {
 
     /// Resolves entity and global tags for one OTLP metrics resource by entity precedence.
     pub fn resolve_resource_tags(
-        &self, attributes: &[otlp_common::KeyValue], tag_cardinality: OriginTagCardinality,
+        &self, attributes: &[otlp_common::KeyValue], tag_cardinality: ConfigOriginTagCardinality,
     ) -> SharedTagSet {
-        if tag_cardinality == OriginTagCardinality::None {
-            return SharedTagSet::default();
-        }
+        let tag_cardinality = match tag_cardinality {
+            ConfigOriginTagCardinality::Low => OriginTagCardinality::Low,
+            ConfigOriginTagCardinality::Orchestrator => OriginTagCardinality::Orchestrator,
+            ConfigOriginTagCardinality::High => OriginTagCardinality::High,
+            ConfigOriginTagCardinality::None => return SharedTagSet::default(),
+        };
 
         let attributes = OtlpResourceAttributes::from_attributes(attributes);
         let mut tags = TagSet::default();
@@ -334,7 +339,7 @@ mod tests {
         ];
 
         assert_eq!(
-            tags(resolver.resolve_resource_tags(&resource, OriginTagCardinality::Low)),
+            tags(resolver.resolve_resource_tags(&resource, ConfigOriginTagCardinality::Low)),
             vec![
                 "container:present",
                 "deployment:present",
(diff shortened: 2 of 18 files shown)


lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -936,6 +936,24 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
         }
     }
 
+    fn consume_otlp_config_metrics_tag_cardinality(&mut self, value: String) {
+        let cardinality = match value.to_ascii_lowercase().as_str() {
```
webern: Check other enums to see if there's a different pattern. Normally we would implement `FromStr` on the enum then use it here.

Note to other reviewers: the Datadog schema doesn't give us enough information to generate enums right now but it may in the future.

lib/saluki-components/src/common/otlp/origin.rs
```diff
@@ -1,12 +1,71 @@
...
     origin::{OriginTagCardinality, OriginTagsResolver, RawOrigin},
-    tags::SharedTagSet,
+    tags::{SharedTagSet, TagSet},
 };
-use saluki_env::WorkloadProvider;
+use saluki_env::{workload::EntityId, WorkloadProvider};
+use stringtheory::MetaString;
 use tracing::trace;
 
+const CONTAINER_ID: &str = "container.id";
```
webern, line 14: Did we copy these from the Agent? I think maybe this is a request for documentation. Generally "We copied these from the Agent" isn't necessarily the documentation we want, but if we did, how is it documented there. Or maybe it just needs to say, "these are the resources(?) that we recognize in OTLP data convert to Datadog-canonical tags" or something like that.

lib/saluki-components/src/common/otlp/origin.rs
```diff
@@ -1,12 +1,71 @@
...
+const K8S_DEPLOYMENT_NAME: &str = "k8s.deployment.name";
+const K8S_NAMESPACE_NAME: &str = "k8s.namespace.name";
+const K8S_NODE_NAME: &str = "k8s.node.name";
+const K8S_POD_UID: &str = "k8s.pod.uid";
+const PROCESS_PID: &str = "process.pid";
+const CGROUP_INODE: &str = "datadog.container.cgroup_inode";
+const INIT_CONTAINER: &str = "datadog.container.is_init";
+
+#[derive(Default)]
+struct OtlpResourceAttributes<'a> {
```
webern, line 28: Same idea here. Document. I guess this is an object that we are using to aggregate the recognized OTLP resources so that we can translate them to Datadog-tags.

lib/saluki-components/src/common/otlp/origin.rs
```diff
@@ -17,12 +76,108 @@ impl OtlpOriginTagResolver {
         Self { workload_provider }
     }
 
+    /// Resolves entity and global tags for one OTLP metrics resource by entity precedence.
+    pub fn resolve_resource_tags(
```
webern, line 82: There could be faster ways to do this with fewer allocations, but my AI agent estimated that this only happens 10's of times every 10-60 seconds, so I think we're good.

lib/saluki-components/src/common/otlp/origin.rs
```diff
@@ -17,12 +76,108 @@ impl OtlpOriginTagResolver {
...
+
+        fallback_entities.into_iter().find(|entity_id| {
+            self.workload_provider
+                .get_tags_for_entity(entity_id, tag_cardinality)
+                .is_some_and(|tags| !tags.is_empty())
+        })
+    }
+
+    fn container_entity_from_external_data(&self, attributes: &OtlpResourceAttributes<'_>) -> Option<EntityId> {
+        let pod_uid = attributes.k8s_pod_uid?;
```
webern, line 142: Woah, how have I never used this syntax. Cool. (Returning a None like a Result)

lib/saluki-components/src/common/otlp/origin.rs
```diff
@@ -58,3 +213,213 @@ impl OriginTagsResolver for OtlpOriginTagResolver {
         }
     }
 }
+
+fn entity_ids_from_attributes(attributes: &OtlpResourceAttributes<'_>) -> Vec<EntityId> {
+    let mut entity_ids = Vec::with_capacity(7);
+
+    if let Some(oci_manifest_digest) = attributes.oci_manifest_digest {
+        if let Some((_, digest)) = oci_manifest_digest.split_once("@sha256:") {
```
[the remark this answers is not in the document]
webern: Seems like a legit bug:

> The OTel semantic convention for oci.manifest.digest defines its value as the bare digest, e.g. sha256:e4ca62c0d62f... — not an image-reference form like redis@sha256:.... The code only extracts a digest when the string contains @sha256: (line 221), which matches the Docker-style "image@sha256:digest" reference, not the standard semconv value.
>
> So any standards-compliant OTel SDK sending the plain sha256:<hex> form hits no match, and entity_ids_from_attributes silently omits the ContainerImageMetadata entity — meaning image-derived tags never get looked up or merged for that resource. It's silent: no error, just missing tags.

lib/saluki-components/src/sources/otlp/metrics/config.rs
```diff
@@ -11,6 +13,8 @@ const DEFAULT_SWEEP_INTERVAL: Duration = Duration::from_secs(1800);
 #[derive(Debug, Clone, Copy)]
 #[allow(dead_code)]
```
webern, line 13: Oops. Did I do that? Is this really dead code? If so, delete it. If not, delete the `allow`.

```suggestion
```

lib/saluki-components/src/sources/otlp/metrics/config.rs
```diff
@@ -44,6 +48,16 @@ impl OtlpMetricsTranslatorConfig {
         Ok(())
     }
 
+    pub fn with_tag_cardinality(mut self, tag_cardinality: ConfigOriginTagCardinality) -> Self {
+        self.tag_cardinality = match tag_cardinality {
+            ConfigOriginTagCardinality::Low => OriginTagCardinality::Low,
```
webern: I don't know if we've fully established a pattern for this or not, but it seems like a redundant type. If this really is not dead code then maybe we should just use the enum from the config instead of translating to another version of it.

lib/saluki-components/src/sources/otlp/metrics/translator.rs
```diff
@@ -79,6 +80,8 @@ pub struct OtlpMetricsTranslator {
     config: OtlpMetricsTranslatorConfig,
     default_hostname: MetaString,
     context_resolver: ContextResolver,
+    origin_tag_resolver: Option<OtlpOriginTagResolver>,
```
webern: I was hoping maybe this doesn't have to be optional... AI helped chase it down:

> The only production call site (run.rs:911-913) always calls `.with_workload_provider(env_provider.workload().clone())`, and `EnvironmentProvider::workload()` isn't
 optional either — every real provider has a concrete Workload. The only place we build an `OtlpConfiguration` without one is tests (config_with_metrics, etc.).
>
> `Option<Arc<dyn WorkloadProvider...>>` on `OtlpConfiguration` could become a plain `Arc<dyn WorkloadProvider...>`, with `with_workload_provider` staying as the setter. Test call sites that skip it would need a no-op/stub WorkloadProvider impl instead of relying on None. That removes the Option from `origin_tag_resolver` in `translator.rs` too, and the silent `unwrap_or_default()` enrichment-skip path goes away.
> 
> Separately: `build()` already returns `Result<_, GenericError>` and already validates other preconditions (receiver enabled, endpoint parsing) before constructing the Otlp source. So if we don't make the field non-optional, at minimum build() should reject a missing workload_provider the same way — self.workload_provider.as_ref().ok_or_else(|| generic_error!("workload provider must be set for OTLP source"))? — rather than silently building a source with enrichment disabled. Right now a missing with_workload_provider() call fails silently at runtime instead of at startup.

lib/saluki-env/src/workload/entity.rs
webern, line 1: I asked it this: Do the changes in lib/saluki-env/src/workload/entity.rs affect data paths other than OTLP and is that OK?

And it said this (seems fine...):

Yes — EntityId (in saluki-env/src/workload/entity.rs) is shared infrastructure, not OTLP-specific, so these changes ripple into DogStatsD too. Here's what's affected and whether it's OK:

1. New variants recognized by tagger.rs — remote_entity_id_to_entity_id (bin/agent-data-plane) is the single conversion point from the remote tagger's string-prefixed IDs into EntityId, shared by every signal,
not just OTLP. So DogStatsD's tagger lookups can now also surface EcsTask, KubernetesNode, Process, etc. That's fine — it's the same underlying tagger stream, just previously under-recognized.

2. dogstatsd/replay/writer.rs round-trip encode/decode — updated purely so the traffic-capture replay format stays exhaustive over the now-larger EntityId enum (compiler forces this). DogStatsD doesn't emit
these new entity types itself today, but if the tagger ever hands one to a captured DogStatsD session, replay now round-trips it correctly instead of failing to compile or panicking.

3. Ord/precedence reordering (precedence_value, HighestPrecedenceEntityIdRef) — this is the one I'd actually check. It's used in exactly one place outside entity.rs: bin/.../workload/api.rs:125, sorting
entities for the /workload/remote_agent/tags/dump diagnostic JSON endpoint. Cosmetic only — display ordering, not tag-merge logic.

DogStatsD's real merge logic (dogstatsd/origin.rs::collect_origin_tags) hardcodes its own precedence order independent of entity.rs's Ord impl, and OTLP's new resolve_resource_tags does the same. So the
reordering (Container now outranking PodUid, etc.) doesn't silently change DogStatsD tag-merge behavior — it only reshuffles a debug dump.

lib/saluki-components/src/common/otlp/origin.rs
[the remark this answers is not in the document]
lucastemb: I forgot to respond to this Codex comment yesterday, but I believe that the Core Agent mimics the same behavior (`oci.manifest.digest` has to contain `@sha256:`). Addressing this comment would be intentional divergence.

lib/saluki-components/src/common/otlp/origin.rs
```diff
@@ -58,3 +213,213 @@ impl OriginTagsResolver for OtlpOriginTagResolver {
...
+            entity_ids.push(EntityId::ContainerImageMetadata(format!("sha256:{digest}").into()));
```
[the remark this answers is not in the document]
lucastemb: Rationale listed above: 
> I forgot to respond to this Codex comment yesterday, but I believe that the Core Agent mimics the same behavior (`oci.manifest.digest` has to contain `@sha256:`). Addressing this comment would be intentional divergence.

lib/saluki-components/src/common/otlp/origin.rs
```diff
@@ -17,12 +76,108 @@ impl OtlpOriginTagResolver {
...
+                ?entity_id,
+                cardinality = tag_cardinality.as_str(),
+                "No tags found for entity."
+            );
+            return;
+        };
+
+        for tag in &entity_tags {
+            let Some((key, value)) = tag.as_str().split_once(':') else {
+                continue;
```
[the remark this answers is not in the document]
lucastemb: Again, this is how the Core Agent handles tags as well. Tags must have a non-empty `key:value` form. [Proof](https://github.com/DataDog/datadog-agent/blob/main/comp/otelcol/otlp/components/processor/infraattributesprocessor/common.go#L282-L288)

lib/saluki-components/src/sources/otlp/metrics/config.rs
lucastemb: I think that this might have to stay temporarily while we resolve some functionality because I believe some of these unused keys we'll probably utilize down the line. For example, I think that `with_remapping` will apply to DDOT related stuff sometime in the near future if I'm not mistaken. I had to remove a call site as part of a separate PR since I think it was an experimentally enabled to `true` and we have no use at the moment.

lib/saluki-components/src/common/otlp/origin.rs
lucastemb: Ok, I investigated this briefly and we only make two heap allocations for vectors and they're bounded per resource at `3` entries for `fallback_entities` and `7` for `entity_ids`. So yea, I agree I think we can leave this as is.

webern (approved):

PR #2308, opened 2026-08-11
feat(config): adds support for full suite of otlp_config.receiver keys

### LINKED ISSUE #2283: Support non-schema-defined OTEL Receiver configuration
### Problem Statement

The way that the Agent handles OTEL Collector receiver configuration is incompatible with how ADP is consuming the remote Agent config stream. For OTLP Ingest, the Agent configures its OTEL Collector using the `otlp_config.receiver` section of Datadog config.

Some keys in that section are defined in the Datadog schema, and are first-class Agent configurations. But the entire section is mapped into the OTEL configuration as a passthrough landing at `receivers.otel`.

We have verified that non-schema values *do* arrive via the Remote Agent config stream (all of them, e.g. `foo=bar` is sent), so the good news is that we receive all of this configuration. The bad news is that we don't have an easy way to model and deserialize it. With typed config (see #2169) we have rejected the notion of grab-bag maps so we need to provide a proper deserialization home for these configurations.

Furthermore, assuming that there are some configurations in that namespace that we either cannot or currently do not support, we need to give them the same treatment that we give to Datadog schema values we do not support. Namely, we should warn or error/exit based on severity for OTEL receiver configurations we do not support when the OTEL pipeline(s) are active.

### Solution Outline

Fortunately, OTEL has *very recently* added a [configuration schema](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/config.schema.yaml). We should make use of this in the same way that we have with the Datadog schema.

The design idea I have is to vendor the OTEL schema as a sibling to the Datadog schema and to "patch" the relevant OTEL section into the correct Datadog section prior to additional processing. This means that the OTEL keys will be added to our total keyspace and will need to be represented in `schema_overlay.yaml` just like all Datadog keys.

Justification: these OTEL keys are explicitly part of Agent configuration based on the guidance provided here https://github.com/DataDog/datadog-agent/blob/33156494bdb5132100dc45b7bb6ff19af6edd76a/pkg/config/config_template.yaml#L4998-L5000 As such, we need to treat them as part of the interface that ADP is supporting for OTLP Ingest.

By patching our JSON-schema to include these keys prior to build-time processing, we get a lot of out-of-the-box benefits that should "just-work"
- A generated deserialization struct in the right place
- Inventory tracking without inventing a new mechanism
- Generated documentation
- Warnings/errors for unsupported keys

### Design

Treat the effective OpenTelemetry OTLP receiver configuration accepted by the Datadog Agent as part of the Agent configuration interface. Vendor the upstream `otlpreceiver` schema and the fixed schemas it references. Keep those files pristine and record their source revision. Include only settings that can affect the Agent’s embedded Collector: gRPC, HTTP, networking, TLS, keepalive, CORS, compression, timeouts, headers, and other self-contained server behavior. Omit `auth` and `middlewares`, because they reference Collector extensions that Agent users cannot define through `otlp_config.receiver` and therefore cannot affect embedded-Collector behavior.

Add a build-time normalization step that resolves OpenTelemetry’s `$defs`, package-qualified `$ref` values, and `allOf` composition, then converts the result into the Datadog schema dialect. Merge manually transcribed defaults from a maintained sidecar, with brief provenance pointing to the corresponding upstream Go default constructors. Patch the normalized receiver subtree into `otlp_config.receiver` before any schema validation, overlay validation, inventory generation, classifier generation, documentation generation, or typed deserializer generation. All consumers must operate on the same composed schema.

The composed leaves become part of the existing Datadog keyspace and must each be represented in `schema_overlay.yaml`. Existing overlay semantics determine which leaves generate typed fields and witness methods, which are documented as behavioral differences, and which trigger pipeline-gated compatibility diagnostics. Supported settings must flow through `DatadogConfiguration`, the witness translator, and the OTLP domain model into the receiver implementation. Unsupported non-default values continue through the existing flattened-key classifier, which already maps severity to debug, warning, or startup failure when OTLP is active. `data_plane.otlp.proxy.receiver` is out of scope.

#### Vendor the applicable OpenTelemetry schemas

- Add a separate tree under `lib/datadog-agent/config/schema/otel/` containing pristine copies of `receiver/otlpreceiver/config.schema.yaml` and every transitively referenced fixed schema needed by the included receiver settings.
- Pin the files to the OpenTelemetry Collector revision ~~used by the Datadog Agent rather than tracking `main`~~. (Human edit: not possible, initially, to match the Agent version because the OTEL schema was added upstream more recently than the Agent's OTEL collector version). Add a short README recording the repository, revision, copied files, and update procedure.
- Include receiver settings that work within the Agent’s fixed embedded-Collector assembly. This includes the gRPC and HTTP protocol objects, signal-specific HTTP URL paths, networking, TLS, keepalive, CORS, compression, request limits, timeouts, buffers, headers, and similar self-contained server behavior.
- Remove the `auth` and `middlewares` properties while deriving the composed schema. They require Collector extension instances that users cannot define through `otlp_config.receiver`; they must not be patched into the Datadog schema and therefore must not appear in `schema_overlay.yaml`.
- Do not modify the pristine vendored OTel files during normalization.

#### Normalize the OpenTelemetry schema dialect

The current resolver in `config-overlay-model/src/lib.rs` only replaces a mapping containing `$ref: <filename>` with the referenced file. The OTel schemas instead use local `$defs`, references such as `$ref: protocols`, package-qualified references such as `$ref: /config/configgrpc.server_config`, sibling keywords beside `$ref`, and `allOf` composition. Do not pass those files directly to the existing resolver.

Add a dedicated OTel normalization step that:

1. Resolves local references against the current file’s `$defs`.
2. Resolves package-qualified references against the corresponding vendored package schema and selected `$defs` member.
3. Recursively resolves transitive references and fails with the source file and reference when a reference is missing or cyclic.
4. Merges `allOf` object fragments without dropping sibling `properties`, descriptions, or constraints. Fail on incompatible duplicate definitions rather than choosing one silently.
5. Preserves JSON Schema information used by current generation, including `type`, `format`, `items`, `additionalProperties`, `enum`, and `description`.
6. Converts object containers and leaves into the Datadog schema dialect expected by `schema_gen`: containers use `node_type: section`; leaves use `node_type: setting`.
7. Preserves OTel optionality. In particular, absent `grpc` or `http` protocol sections must remain absent, while an explicitly present empty protocol section receives that protocol’s field defaults.
8. Produces a deterministic derived receiver schema under an `otel/generated/` directory. Treat that artifact as generated output and regenerate it through `make build-schema-overlay`.

Add `cargo:rerun-if-changed` directives for the complete OTel vendor tree, the defaults sidecar, and normalization code.

#### Maintain receiver defaults and Agent-specific metadata

- Add a sidecar outside the pristine vendor directory that maps receiver leaf paths to their OpenTelemetry defaults.
- Hand-transcribe defaults from the Go structs and default constructors at the pinned OTel revision. Briefly document how each group of defaults was discovered, including the relevant package, source file or constructor, and revision.
- Merge the sidecar into the derived schema during normalization. Validate that every sidecar path exists in the normalized schema and fail on stale, duplicate, or type-incompatible entries.
- Keep protocol-object presence separate from field defaults. A default endpoint inside `grpc` must not cause an absent `grpc` section to be created.
- Preserve the existing Datadog schema metadata for receiver leaves already defined there, including Datadog defaults and environment-variable behavior.
- OTel-only passthrough keys are YAML/config-stream settings, not automatically Agent-bound environment variables. Mark them `no-env` in the derived Datadog dialect unless the Agent explicitly binds that key. Do not accidentally generate new `DD_OTLP_CONFIG_...` behavior that the Datadog Agent does not provide.

#### Compose before every schema consumer

Introduce one canonical composed-schema loading path. It must load the Datadog schema, load or generate the normalized OTel receiver schema, replace the existing `otlp_config.receiver` subtree, and return the composed document. Refactor schema processing so no consumer accidentally reads the unpatched Datadog schema.

The composed schema must be used by:

- schema key collection and `schema_overlay.yaml` exhaustiveness validation;
- schema-map and environment metadata generation;
- classifier generation and default comparison;
- generated documentation and configuration registry tooling;
- `DatadogConfiguration`/typify generation; and
- witness generation.

Do not patch only `datadog_config_gen.rs`; doing so would generate a receiver struct while leaving overlay validation, inventory, documentation, and runtime classification unaware of the same keys.

#### Update `schema_overlay.yaml` and typed configuration

- Add every leaf introduced by the composed receiver schema to `lib/datadog-agent/config/schema/schema_overlay.yaml`. The build must continue to enforce that every composed schema leaf appears exactly once in the overlay.
- Determine an initial support classification for every patched key using the existing overlay model. Use the existing metadata requirements for each classification, including pipeline affinity, severity where applicable, descriptions, documentation, and test support.
- Reassess the existing hand-authored `otlp_config.receiver` entries against the normalized upstream definitions and remove duplicate or obsolete schema definitions.
- Run `make build-schema-overlay` and commit all changed generated artifacts.
- For every receiver leaf classified as supported or partially supported, ensure typify generates it under `DatadogConfiguration.otlp_config.receiver`. Implement the generated witness method in `DatadogTranslator`, add an appropriately typed destination in `agent-data-plane-config`’s OTLP domain, and wire the value into `saluki-components`’ OTLP server construction.
- Keep defaults and parsing at the typed configuration boundary. Runtime components must not read `GenericConfiguration` or reintroduce a receiver grab-bag map.
- Do not add unsupported receiver leaves to the runtime domain merely to deserialize them. The existing classifier operates on `ConfigurationSystem::raw_map().flattened_keys()` and can diagnose unsupported leaves independently of the typed supported-key projection.
- Do not change `data_plane.otlp.proxy.receiver`; it is outside this issue.

#### Tests and acceptance criteria

Add focused tests for:

- local `$defs`, package-qualified `$ref`, transitive references, sibling keywords, and `allOf` merging;
- unresolved references, cycles, and conflicting merged properties producing actionable build errors;
- deterministic generated schema output;
- removal of `auth` and `middlewares` from the composed schema and overlay keyspace;
- inclusion of representative gRPC, HTTP, TLS, keepalive, CORS, timeout, header, compression, request-size, and signal URL-path leaves;
- sidecar default injection, provenance coverage, path/type validation, and protocol-section optionality;
- preservation of existing Datadog receiver metadata and the absence of generated environment bindings for OTel-only keys;
- exact composed-schema coverage by `schema_overlay.yaml`;
- generated receiver types and witness methods for supported/partially supported keys;
- deserialization of representative Remote Agent config snapshots, including non-Datadog-schema receiver values;
- supported settings reaching OTLP server construction and affecting behavior;
- partial and unsupported non-default settings producing the existing pipeline-gated compatibility behavior; and
- OTLP-inactive configurations not producing OTLP-specific compatibility diagnostics.

The implementation is complete when `make build-schema-overlay` succeeds from a clean checkout, the generated artifacts are clean on a second run, every patched receiver leaf is covered by `schema_overlay.yaml`, supported receiver settings have a typed source-to-runtime path, and unsupported settings are diagnosed through the existing classifier when an OTLP pipeline is active. Run `make fmt`, `cargo check --workspace --tests`, and `make check-all` after generation.

lucastemb:
## Summary

As part of the ongoing work related to OTLP Ingest, we noticed that some of the embedded OTEL configuration options were not being broadcasted. This PR seeks to explicitly list all `otlp_config.receiver` options that are recognized. This will lay the ground work for a series of future PRs addressed at accepting or rejecting support for the full suite of config keys. 

The full list can be seen [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/config.schema.yaml) or [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/README.md)

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

N/A
## References

- Closes: #2283


lib/datadog-agent/config/schema/core/core_schema.yaml
webern, line 1: We cannot "patch" the core schema like this because we need to be able to overwrite with a new version.

What we need to do instead is vendor (i.e. check-in to this repo) the otel configuration schema (or the relevant parts of it) here in, e.g. `schema/otel` next to `schema/core`. Then during `build.rs` we deserialize both and perform the "patch" in memory before handing it off to code gen or other processes.

webern (changes requested):
Headed in the right direction, but instead of doing a one-time patch and checking-in the resultant, patched, schema, we need to deserialize and do the patching in memory at `build.rs` execution.

Your version of things here in the PR currently works and gives us what we want, but it cannot be maintained because the core schema will be overwritten when we update it.

In general, both schema's should be checked into our repository in a pristine state, no edits or patches. Anything we need to do with them happens in-memory during `build.rs`.

lib/datadog-agent/config-overlay-model/src/lib.rs
```diff
@@ -558,6 +564,291 @@ fn resolve_refs(value: &mut serde_yaml::Value, schema_dir: &Path) -> Result<(),
...
+
+/// Load the Datadog schema with the OTel receiver schema patched in.
+///
+/// This loads the pristine Datadog schema (`core_schema.yaml`), loads and resolves the pristine
+/// OTel receiver schema (`otel/config.schema.yaml`), and replaces the `otlp_config.receiver`
+/// subtree in the Datadog schema with the resolved OTel subtree.
+pub fn load_composed_schema(datadog_schema: &Path, otel_schema_dir: &Path) -> Result<serde_yaml::Value, Error> {
+    let mut datadog_schema = load_resolved_schema(datadog_schema)?;
+    let otel_receiver = load_otel_receiver(otel_schema_dir)?;
+    patch_receiver(&mut datadog_schema, otel_receiver)?;
```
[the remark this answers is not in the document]
lucastemb: This is incorrect. The Core Agent does "own" these keys, but the processing is delegated and offloaded to the embedded OTEL Collector. It would be misleading to classify them as Saluki-only. Also, it was decided that composition should happen in-memory at build time.

lib/datadog-agent/config/build.rs
```diff
@@ -15,18 +15,18 @@ fn main() {
     let manifest_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
     let files = Files::default();
 
-    let schema_dir = files.schema.parent().expect("schema file must have a parent directory");
-    println!("cargo:rerun-if-changed={}", schema_dir.display());
+    println!("cargo:rerun-if-changed={}", files.datadog_schema.display());
```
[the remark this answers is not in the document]
lucastemb: Fair. Fixed in [95e873b](https://github.com/DataDog/saluki/pull/2308/commits/95e873be3102de46827d108db986cf901c0341e8)

lib/datadog-agent/config-overlay-model/src/lib.rs
```diff
@@ -558,6 +564,291 @@ fn resolve_refs(value: &mut serde_yaml::Value, schema_dir: &Path) -> Result<(),
...
+        .and_then(|v| v.get_mut("otlp_config"))
+    else {
+        return Ok(());
+    };
+
+    let Some(receiver) = otlp_config.get_mut("properties").and_then(|v| v.get_mut("receiver")) else {
+        return Ok(());
+    };
+
+    *receiver = otel_receiver;
```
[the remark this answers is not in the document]
lucastemb: This _is_ a good catch. Fixed in (https://github.com/DataDog/saluki/pull/2308/commits/95e873be3102de46827d108db986cf901c0341e8). We should honor the DataDog defaults for `otlp_config.receiver` since they were configured for our specific use cases.

lib/datadog-agent/config-overlay-model/src/lib.rs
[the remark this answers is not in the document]
webern: I agree, this is by design as we have though about this and decided that, since these appear in Datadog configuration, and affect the observable behavior of the Agent (by being copied into the embedded OTeL collector), these are effectively Datadog configurations. Furthermore, doing this makes all of our downstream processes work correctly for said keys.

lib/datadog-agent/config/build.rs
```diff
@@ -15,18 +15,22 @@ fn main() {
...
     println!("cargo:rerun-if-changed=build/witness_gen.rs");
 
-    let schema_path = files.schema.clone();
+    let schema_path = files.datadog_schema.clone();
+    let schema_map = schema_gen::load_schema(&files.datadog_schema, &files.otel_schema_dir);
     let overlay = SchemaOverlay::load(files).unwrap_or_else(|e| panic!("{e}"));
-    let schema_map = schema_gen::load_schema(&schema_path);
 
     classifier_gen::generate(&overlay, &schema_map, &manifest_dir);
     datadog_config_gen::generate(&overlay, &schema_path, &schema_map, &manifest_dir);
```
webern: Could we pass the composed schema to datadog_config_gen instead of this raw path? schema_map uses the composed schema, but datadog_config_gen::generate reloads this path with load_resolved_schema. Typed configuration generation therefore still sees the unpatched core schema. The issue requires every build consumer to use the same composed schema.

Put differently... any caller that is using the schema should take the composed schema instead of a file path.

lib/datadog-agent/config-overlay-model/src/lib.rs
```diff
@@ -558,6 +564,299 @@ fn resolve_refs(value: &mut serde_yaml::Value, schema_dir: &Path) -> Result<(),
...
+    receiver_section.insert(
+        serde_yaml::Value::String("properties".to_string()),
+        serde_yaml::Value::Mapping(properties),
+    );
+
+    Ok(serde_yaml::Value::Mapping(receiver_section))
+}
+
+/// Merge the resolved OTel receiver subtree into the Datadog schema's `otlp_config.receiver` while
+/// honoring default values not native to OTel.
```
webern, line 604: I think OTeL defaults are an unsolved problem. They don't exist in the spec, apparently. So as soon as we enable one of these keys, we are going to have to figure out its default and how that value arrives into the code.

I think we can move forward for now, but we will hit this next.

lib/datadog-agent/config-overlay-model/src/lib.rs
```diff
@@ -558,6 +564,299 @@ fn resolve_refs(value: &mut serde_yaml::Value, schema_dir: &Path) -> Result<(),
...
+        for (_, v) in props.iter_mut() {
+            resolve_otel_refs(v, otel_dir, current_defs)?;
+        }
+    }
+
+    Ok(())
+}
+
+/// Resolve a single `$ref` target, returning the definition value and its source `$defs`.
+fn resolve_otel_ref_target(
```
webern, line 755: Somewhere in here we should either mark these as `no_env` (I think?) with a TODO, that says see if the Agent accepts these by `DD_WHATEVER`. Or, if feeling ambitious, maybe do an experiment to see if these can be set by DD env var and if so, use the pattern to populate the right vars here.

webern (commented):
The only thing is that I think the gen_datadog_config is not using the merged schema.

The other comments are less immediate.

lib/datadog-agent/config-overlay-model/src/lib.rs
lucastemb: After poking around (and running a correctness test) it seems like only the 13 that are advertised in `core_schema.yaml` can be set via DD env var. According to the LLM, it seems that `no-env` won't suppress generation of `DD_*` env entries, but I'm unclear what the intended effect is.

lucastemb: Fixed in [b9a7bdd](https://github.com/DataDog/saluki/pull/2308/commits/b9a7bdd666457e08e1a79715dabc1d1165612030), but admittedly, I'm not sure if this is what you were looking for

lib/datadog-agent/config/build.rs
lucastemb: Fixed in [843c6e4](https://github.com/DataDog/saluki/pull/2308/commits/843c6e456c65a7ab7a122b8b8a5aaa4e00ee5f65)

webern (approved):
I want to unblock this merge because I will be out for three days next week and my review basically amounts to, this looks good and we probably still need to deal with defaults, which we can do in a future PR, or as we enable the fields.

PR #2317, opened 2026-08-12
fix(config): migrate endpoints to typed config for provenance awareness

### LINKED ISSUE #1965: Investigate whether `dd_url`'s schema default makes `site` unreachable in ADP.

### LINKED ISSUE #2310: Migrate Datadog metrics configuration to typed config for #1965
## Context

PR #2279 carries configuration provenance into typed configuration, so a default-sourced `dd_url` can be distinguished from an explicitly configured URL. This is one of the migrations needed to complete #1965.

## Work

Migrate `DatadogMetricsConfiguration` / `dd_metrics_encode` from the raw configuration map to typed configuration. Use the typed `site` and `Endpoints.dd_url.is_explicit()` when resolving the primary endpoint and when evaluating `any_series_endpoint_matches`. Include the MRF and autoscaling-failover metrics encoder instances.

Parent issue: #1965

### LINKED ISSUE #2312: Migrate Datadog forwarder configuration to typed config for #1965
## Context

PR #2279 carries configuration provenance into typed configuration, so a default-sourced `dd_url` can be distinguished from an explicitly configured URL. This is one of the migrations needed to complete #1965.

## Work

Migrate `DatadogForwarderConfiguration` / `dd_out` and `mrf_dd_out` from the raw configuration map to typed configuration. Use the shared typed endpoint configuration rather than independently deserializing `site` and `dd_url`, so the forwarder and metrics encoder agree on the primary endpoint.

Parent issue: #1965

### LINKED ISSUE #2313: Migrate shared forwarder configuration to typed config for #1965
## Context

PR #2279 carries configuration provenance into typed configuration, including the retry-queue settings needed by forwarders.

## Work

Migrate `ForwarderConfiguration`, the common adapter used by both forwarders, so it overlays the typed endpoint and retry values onto the remaining raw/deserialized forwarder settings.

Parent issue: #1965

### LINKED ISSUE #2314: Make endpoint configuration provenance-aware for #1965
## Context

PR #2279 makes it possible to distinguish a default-sourced `dd_url` from an explicitly configured URL.

## Work

Update `EndpointConfiguration`, where primary endpoint resolution happens, so a defaulted `dd_url` is treated as no override while an explicit `dd_url`—including `https://app.datadoghq.com`—remains an override.

Parent issue: #1965

### LINKED ISSUE #2315: Make retry configuration provenance-aware for #1965
## Context

PR #2279 carries provenance for the forwarder retry-queue settings.

## Work

Update `RetryConfiguration::queue_max_size_bytes()` to prefer `retry_queue_payloads_max_size` only when explicit, then fall back to `retry_queue_max_size` only when explicit.

Parent issue: #1965

### LINKED ISSUE #2316: Migrate cluster-agent forwarder configuration to typed config for #1965
## Context

PR #2279 carries provenance through the shared forwarder configuration used by the cluster-agent path.

## Work

Migrate `ClusterAgentForwarderConfiguration` / `cluster_agent_out` to receive the provenance-aware retry settings through the shared `ForwarderConfiguration`. Its destination URL is supplied programmatically, so it does not need global `site`/`dd_url`.

Parent issue: #1965

webern:
## Human Summary

This migrates all components affected by #1965 (we were unable to detect explicitly set to default vs default config values) to typed config to pick up the fix provided in #2279.

After implementation I ran a clean-room Opus audit with this prompt:

> please do a clean-room audit of this commit `f4ae503e807b398ca55ee3f6275f554889cff645`
>
> Our goal is to correct a previous incorrect behavior in which we were not sensitive to whether a configuration value was set by the Agent as a default or whether the customer explicitly set the value (and happened to set it to the default value).
>
> Other than that specific intentional bug fix, no other behavioral changes are intended.
>
> Validate that only the desired behavioral change has occurred and that configuration defaults are in-tact.
>
> Read AGENTS.md and /config-system

It only found two actionable items, which I fixed. One was that we were previously rejecting empty API keys, so I retained that behavior. The other was that an MRF endpoint consisting only of whitespace could be treated as a valid setting. Also fixed. All defaults stood up to scrutiny.

Edit: @aqian01 found a flaw that went undetected which became #2325.

### Review Guide

I tried breaking this up, which is why there are a bunch of subissues being closed all at once, but it did not divide very well. As such, the diff a bit large:

Focus on these files:
- `lib/saluki-components/*`
- `bin/agent-data-plane/src/cli/run.rs`
- `lib/agent-data-plane-config*`

Deprioritize generated code and inventory churn in:
- `lib/datadog-agent/*`

## AI Summary

Migrate Datadog endpoint, retry, proxy, forwarder, metrics encoder, MRF, and Cluster Agent configuration consumers from raw `GenericConfiguration` deserialization to resolved typed configuration.

Primary endpoint resolution now happens once in the typed configuration layer:

- A default-sourced `dd_url` no longer shadows `site`.
- An explicitly configured `dd_url`, including the schema-default URL, remains an override.
- MRF and Cluster Agent destination overrides cannot be overwritten by global endpoint settings.
- Retry queue size precedence uses configuration provenance, so explicit zero values are preserved.
- Raw configuration access remains only for live API-key refresh, secrets retry behavior, and the `run_path` compatibility fallback whose schema default is still an unresolved placeholder.

The schema inventories, generated registries, smoke-test metadata, and component tests are updated for the typed consumers.

## Change Type

## How did you test this PR?

- unit tests updated and created throughout
- integration test modified to test this behavior. this was proven to have caught the original bug (`red -> green`)

## References

- Closes: #1965
- Closes: #2310
- Closes: #2312
- Closes: #2313
- Closes: #2314
- Closes: #2315
- Closes: #2316
- Related to: #2279


webern:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/DataDog/saluki/2317?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#2317** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2317?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2317?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2325** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2325?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

.claude/skills/config-system/SKILL.md
```diff
@@ -223,7 +227,9 @@ the witnessed model.
     affected configurations.
 
 A cutover should be behaviorally transparent. If the old behavior conflicts with the source schema
-or typed-system invariants, surface the conflict rather than silently choosing one.
+or typed-system invariants, surface the conflict rather than silently choosing one. After cutover
+artifacts of deserialization logic should not be left behind. For example, no `derive(Deserialize)`
+and if possible a held `Option<SomeType>` should collapse to a held `SomeType` if possible.
```
webern, line 248:
```suggestion
and if possible a held `Option<SomeType>` should collapse to a held `SomeType`.
```

lib/saluki-components/src/common/datadog/protocol.rs
```diff
@@ -5,79 +5,8 @@ use std::collections::HashMap;
...
-
-fn default_v3_series_shadow_sites() -> Vec<String> {
-    vec!["datadoghq.com".to_string()]
-}
-
-fn default_use_v3_api_series_enabled() -> String {
-    "datadog_only".to_string()
-}
-
-#[derive(Deserialize)]
```
aqian01, line 26: Do we want to remove this? I think this is a behavior change since this causes us to not accept true as a boolean now?

Ran this regression test to confirm
```
#[test]
fn boolean_use_v3_api_series_enabled_is_normalized() {
    let sources = SourceTree::all_explicit(json!({
        "use_v3_api": {
            "series": {
                "enabled": true
            }
        }
    }));

    let config = translate_strict(&sources)
        .expect("a boolean V3 series mode should translate");

    assert_eq!(
        config.shared.metrics_encoding.v3_series_mode.mode,
        "true"
    );
}
```

aqian01 (changes requested):

lib/saluki-components/src/common/datadog/protocol.rs
webern: This was a very important finding (perhaps, see below). I fixed it in a separate PR that is now a precursor to this one (this one has now been stacked on the fix).

To prove the where it was detected, I added your test in *this* PR at `lib/agent-data-plane-config-system/src/system.rs` around line 419.

The flaw may not have been as bad as it seemed at first. I believe the Agent actually sends us remote config post-type-coercion, and the Agent-ported coercion algorithms were already being used for ADP-local environment variables. So I think this would have only surfaced when ADP was parsing a local YAML file itself.

In any event, the fix is in #2325 and your helpful test now passes!

.claude/skills/config-system/SKILL.md
```diff
@@ -236,7 +240,9 @@ the witnessed model.
...
```
webern, line 248:
```suggestion
and if possible a held `Option<SomeType>` should collapse to a held `SomeType`.
```

aqian01 (approved):

PR #2318, opened 2026-08-12
feat(metrics): add stateful Foldspace transport

aqian01:
## Summary

- route authoritative V3 metric series as versioned, compressed logical batches
- add independent Foldspace gRPC workers per endpoint/API key with bounded ordered in-flight state
- retain and retry complete logical metrics in memory or on disk, encoding only when gRPC has usable capacity
- keep sketches and unit-bearing series on their existing HTTP paths

## Motivation

Stateful metric encoding must happen after endpoint routing so each destination owns independent dictionary and stream state. The complete logical batch remains retained until ACK, which allows an unacknowledged batch to be encoded again against confirmed state after a transport failure or reconnect.

A failure after transmission has unknown delivery, so a resent logical batch may be duplicated, but its metric fields and identity are preserved.

## Retry behavior

- no HTTP fallback for stateful metric retries
- no retry dequeue or encoding while the stream is disconnected or lacks send capacity
- reconnect backoff with jitter
- fresh-data priority with bounded progress for memory and disk retries
- explicit memory/disk limits, drop telemetry, and graceful-shutdown persistence

## Configuration

- `serializer_experimental_use_v3_api.stateful_metrics_enabled`
- `stateful_metrics_max_inflight_payloads` (default: `32`)

## Dependency

Depends on [DataDog/foldspace#39](https://github.com/DataDog/foldspace/pull/39) and pins Foldspace commit `7a5feff72311a66e8b6ce004f78b578f733a2b8f`.

## Validation

- `cargo test -p saluki-components common::datadog::stateful_metrics` — 6 passed
- `cargo test -p saluki-components encoders::datadog::metrics` — 71 passed
- `cargo test -p saluki-components common::datadog::config` — 34 passed
- `cargo test -p saluki-io net::util::retry` — 34 passed
- `cargo clippy -p saluki-components -p saluki-core -p saluki-io --all-targets -- -D warnings`
- `cargo +nightly fmt --all -- --check`
- `git diff --check`


.cargo/config.toml
```diff
@@ -6,3 +6,6 @@ rustflags = ["--cfg", "tokio_unstable", "-C", "target-feature=+crt-static"]
 
 [env]
 JEMALLOC_SYS_WITH_MALLOC_CONF = "abort_conf:true,max_background_threads:1,narenas:1,tcache:false,thp:never,oversize_threshold:32768,dirty_decay_ms:1000,muzzy_decay_ms:0"
+
+[registries]
+datadog-rust-testing = { index = "sparse+https://depot-read-api-rust.us1.ddbuild.io/datadog/rust-testing/" }
```
tobz, line 11: What is this?

lib/saluki-components/src/encoders/datadog/metrics/mod.rs
```diff
@@ -932,11 +976,18 @@ async fn run_request_builder(
...
+                    // intake rejects as an empty value set. The stateful destination accounts for its own drops.
                     if !v1::has_emittable_point(&metric) {
                         debug!(metric = %metric.context().name(), "Dropping series metric with no finite points.");
-                        telemetry.events_dropped_encoder().increment(1);
+                        if !uses_stateful_series {
+                            telemetry.events_dropped_encoder().increment(1);
+                        }
+                        continue;
+                    }
+                    if uses_stateful_series && v2_series_builder.is_none() {
```
tobz, line 990: Why are we skipping metrics when stateful encoding is enabled but the V2 encoding mode is disabled?

lib/saluki-components/src/encoders/datadog/metrics/mod.rs
```diff
@@ -968,7 +1019,7 @@ async fn run_request_builder(
                         *batch_id = Some(Uuid::now_v7());
                     }
                     let active_batch_id = needs_batch_id.then_some(batch_id.as_ref()).flatten();
-                    let should_buffer_v3 = endpoint_mode.needs_v3() || series_shadow_active;
+                    let should_buffer_v3 = (endpoint_mode.needs_v3() || series_shadow_active) && !uses_stateful_series;
```
tobz, line 1022: It's unclear to me why we'd mess with metric buffering here, since it's specific to V3 shadowing, just because stateful encoding is enabled?

lib/saluki-components/src/common/datadog/stateful_metrics.rs
```diff
@@ -0,0 +1,1668 @@
...
+const STATE_REQUEST_BYTES: u64 = 5 * 1024 * 1024;
+const LOGICAL_METRIC_BATCH_VERSION: u16 = 1;
+const LOGICAL_METRIC_COMPRESSION_LEVEL: i32 = 3;
+const FRESH_BURST_LIMIT: usize = 8;
+const RETRIES_BEFORE_DISK: usize = 4;
+const RECONNECT_JITTER_FRACTION: f64 = 0.2;
+const MAX_RECONNECT_BACKOFF: Duration = Duration::from_secs(30);
+
+/// Configuration for the Foldspace metrics destination.
+pub struct StatefulMetricsDestinationConfiguration {
```
tobz, line 62: High-level thing: we shouldn't have a component defined in a "common" code module.

Put this under `src/destinations/datadog/metrics_stateful`.

lib/saluki-components/src/common/datadog/stateful_metrics.rs
```diff
@@ -0,0 +1,1668 @@
...
+                                continue;
+                            }
+                            metrics.push(metric);
+                        }
+                        if metrics.is_empty() {
+                            continue;
+                        }
+
+                        let metadata = metadata_from_metrics(&metrics);
+                        let metrics: Arc<[Metric]> = metrics.into();
```
tobz, line 215: This is going to reallocate a brand new underlying slice... which doesn't make a ton of sense.

Just use `Arc<Vec<Metric>>`.

lib/saluki-components/src/common/datadog/stateful_metrics.rs
```diff
@@ -0,0 +1,1668 @@
...
+                        endpoint_senders.retain(|sender| !sender.is_closed());
```
tobz, line 216: Why _would_ a particular sender end up closed?

If it was closed, is there any good reason for the component to not fail since, I'm assuming, a sender corresponding to a configured endpoint? We'd basically just stop sending to a configured endpoint if the sender had an issue, so quasi-silent data loss?

lib/saluki-components/src/common/datadog/stateful_metrics.rs
```diff
@@ -0,0 +1,1668 @@
...
+        Self {
+            additional_tags,
+            max_metrics_per_batch,
+            max_points_per_batch,
+            pending_series: Vec::with_capacity(max_metrics_per_batch),
+            pending_points: 0,
+        }
+    }
+
+    fn push(&mut self, metric: &Metric) -> (Vec<LogicalMetricBatch>, bool) {
```
tobz, line 365: `Vec::new` _is_ allocation-free, but.. why not just make this `Option<Vec<...>>` to better signal that something was intentionally returned or not?

lib/saluki-components/src/common/datadog/stateful_metrics.rs
```diff
@@ -0,0 +1,1668 @@
...
+
+        let started_batch = self.pending_series.is_empty();
+        self.pending_points = self.pending_points.saturating_add(point_count);
+        self.pending_series.push(series);
+        if self.pending_series.len() >= self.max_metrics_per_batch || self.pending_points >= self.max_points_per_batch {
+            batches.extend(self.take());
+            return (batches, false);
+        }
+
+        (batches, started_batch)
```
tobz, line 392: Why is it relevant for the caller to know if there's an outstanding/pending batch after this call?

Why wouldn't they just call `has_pending`?

PR #2323, opened 2026-08-12
chore(agent-data-plane): share shutdown signal wait logic

jszwedko:
## Human Summary

Refactors the shutdown wait handler into a shared module that can be used where ever we need to wait to ensure we consistently rely on the same signals, cross-platform.

## AI Summary

#2322 added SIGTERM/SIGINT handling for graceful shutdown, but the same signal-waiting logic was duplicated between the `run` command and `dogstatsd replay`'s cancellation setup. An earlier version of this PR tried to close a related gap -- a signal delivered before the supervisor starts falling through to the OS's default disposition -- by registering signal handlers eagerly at process start. Per [discussion on #2322](https://github.com/DataDog/saluki/pull/2322#discussion_r3770948218), that gap doesn't need graceful handling: an abrupt exit before the supervisor is running is acceptable. This PR instead does a pure refactor, extracting the existing lazy signal-wait logic into a shared `saluki_app::util::wait_for_shutdown_signal` helper used by both `run` and `dogstatsd replay`, removing the duplication and making it reusable by any binary built on the saluki crates, not just agent-data-plane.

## Test plan

- [x] `cargo check --workspace`, `cargo clippy -p agent-data-plane -p saluki-app --no-deps` pass.
- No new automated tests added; this is a non-behavioral refactor of existing signal-handling code.


bin/agent-data-plane/src/cli/run.rs
```diff
@@ -254,43 +260,84 @@ pub async fn handle_run_command(
...
+    sigterm: tokio::signal::unix::Signal,
 
-        let mut sigterm = signal(SignalKind::terminate()).expect("failed to install SIGTERM handler");
+    #[cfg(windows)]
+    ctrl_c: tokio::signal::windows::CtrlC,
+    #[cfg(windows)]
+    ctrl_break: tokio::signal::windows::CtrlBreak,
+}
+
+impl ShutdownSignal {
```
tobz: Can we actually create a new module under `saluki_app`, `saluki_app::util`, and put this in there?

Will be useful to other projects like ACR.

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -254,43 +260,84 @@ pub async fn handle_run_command(
...
-/// `dd-procmgr` (which manages ADP as a subprocess on Windows) sends via `GenerateConsoleCtrlEvent` to request a
-/// graceful stop.
-async fn wait_for_shutdown_signal() {
+/// `ShutdownSignal::install` must be called as early as possible in startup -- before any `.await` point -- because
+/// installing a signal handler happens synchronously when the registration function is called, not when the
+/// resulting listener is later awaited. `tokio::signal::ctrl_c()` is an `async fn`, so calling it inline at the
+/// point where we're ready to wait (as this used to) defers that registration until first polled: any `SIGTERM` (or
+/// `SIGINT`/`CTRL_BREAK`) delivered during bootstrap -- for example, while waiting on the initial configuration from
+/// the Datadog Agent -- falls through to the OS's default disposition and kills the process immediately, skipping
+/// cleanup like PID file removal.
```
tobz: Nit: rambling agent comment about the _past_, instead of simply documenting the _present_.

tobz (changes requested):

bin/agent-data-plane/src/cli/run.rs
jszwedko: 😂 true, I hadn't actually reviewed this PR yet.

bin/agent-data-plane/src/cli/run.rs
jszwedko: Updated this PR to remove the early registration and just refactor and share the handler.

tobz (approved):

PR #2325, opened 2026-08-13
fix(config): permissively coerce scalar config leaves

### LINKED ISSUE #1965: Investigate whether `dd_url`'s schema default makes `site` unreachable in ADP.

### LINKED ISSUE #2310: Migrate Datadog metrics configuration to typed config for #1965

### LINKED ISSUE #2312: Migrate Datadog forwarder configuration to typed config for #1965

### LINKED ISSUE #2313: Migrate shared forwarder configuration to typed config for #1965

### LINKED ISSUE #2314: Make endpoint configuration provenance-aware for #1965

### LINKED ISSUE #2315: Make retry configuration provenance-aware for #1965

### LINKED ISSUE #2316: Migrate cluster-agent forwarder configuration to typed config for #1965

webern:
## Human Summary

In #2317 @aqian01 detected a flaw in the typed config deserialization mechanism. The ported, permissive, Viper type coercions that the Agent uses were only being applied to environment variables. This means that when we parsed a config file ourselves, we would not have applied those permissive coercions. This is now fixed so that we more completely mirror Agent config parsing behavior when migrating to typed config.

## AI Summary

The Datadog Agent accepts configuration values through `cast`, so a leaf's declared type determines which alternate spellings it accepts. `dogstatsd_port: "8125"` and a schema-`string` leaf written as a YAML boolean (`use_v3_api.series.enabled: true`, read with `GetString`) are both valid Agent configuration. ADP's generated model instead deserialized each leaf strictly according to its schema type, causing startup translation to fail or rejecting the entire runtime configuration snapshot.

This change:
- Adds serde deserializers for boolean, integer, float, and string schema scalar types.
- Adds a build-time `permissivize` pass that applies the appropriate deserializer to every generated scalar leaf based on its generated Rust type.
- Makes `env_decode` use the same parsers, aligning file, environment, and configuration-stream inputs.
- Fails code generation for unrecognized leaf types, so new schema types require an explicit coercion decision.
- Retains hard errors for values that cannot be converted, rather than accepting `cast`'s zero values.
- Removes the now-unnecessary `input_shape: string_or_integer` metadata and related types.

Field types remain unchanged; only deserialization behavior is permissive. Numeric strings are accepted in decimal form only. The generated model diff adds deserializer attributes, and the change covers the full generated scalar model rather than individual keys.

## Change Type
- [x] Bug fix

## How did you test this PR?

- Unit tests cover accepted and rejected spellings for each scalar type.
- A schema-driven test exercises every scalar leaf with a castable value and verifies the coerced value through a serialization round-trip.
- Translation-gate tests cover a boolean on the V3 series mode string leaf and a quoted `dogstatsd_port`.
- `make build-schema-overlay`, `cargo check --workspace --tests`, and `cargo clippy --workspace --tests` pass; config crate tests pass.

## References

- Related: #2317

---

## Note: #2317 merged into the wrong base

I merged #2317 into `m/confra-cast` (this PR) instead of `main`. This PR now carries both changes. The #2317 description follows verbatim.

---

# Merged from #2317: fix(config): migrate endpoints to typed config for provenance awareness

## Human Summary

This migrates all components affected by #1965 (we were unable to detect explicitly set to default vs default config values) to typed config to pick up the fix provided in #2279.

After implementation I ran a clean-room Opus audit with this prompt:

> please do a clean-room audit of this commit `f4ae503e807b398ca55ee3f6275f554889cff645`
>
> Our goal is to correct a previous incorrect behavior in which we were not sensitive to whether a configuration value was set by the Agent as a default or whether the customer explicitly set the value (and happened to set it to the default value).
>
> Other than that specific intentional bug fix, no other behavioral changes are intended.
>
> Validate that only the desired behavioral change has occurred and that configuration defaults are in-tact.
>
> Read AGENTS.md and /config-system

It only found two actionable items, which I fixed. One was that we were previously rejecting empty API keys, so I retained that behavior. The other was that an MRF endpoint consisting only of whitespace could be treated as a valid setting. Also fixed. All defaults stood up to scrutiny.

Edit: @aqian01 found a flaw that went undetected which became #2325.

### Review Guide

I tried breaking this up, which is why there are a bunch of subissues being closed all at once, but it did not divide very well. As such, the diff a bit large:

Focus on these files:
- `lib/saluki-components/*`
- `bin/agent-data-plane/src/cli/run.rs`
- `lib/agent-data-plane-config*`

Deprioritize generated code and inventory churn in:
- `lib/datadog-agent/*`

## AI Summary

Migrate Datadog endpoint, retry, proxy, forwarder, metrics encoder, MRF, and Cluster Agent configuration consumers from raw `GenericConfiguration` deserialization to resolved typed configuration.

Primary endpoint resolution now happens once in the typed configuration layer:

- A default-sourced `dd_url` no longer shadows `site`.
- An explicitly configured `dd_url`, including the schema-default URL, remains an override.
- MRF and Cluster Agent destination overrides cannot be overwritten by global endpoint settings.
- Retry queue size precedence uses configuration provenance, so explicit zero values are preserved.
- Raw configuration access remains only for live API-key refresh, secrets retry behavior, and the `run_path` compatibility fallback whose schema default is still an unresolved placeholder.

The schema inventories, generated registries, smoke-test metadata, and component tests are updated for the typed consumers.

## Change Type

## How did you test this PR?

- unit tests updated and created throughout
- integration test modified to test this behavior. this was proven to have caught the original bug (`red -> green`)

## References

- Closes: #1965
- Closes: #2310
- Closes: #2312
- Closes: #2313
- Closes: #2314
- Closes: #2315
- Closes: #2316
- Related to: #2279


webern:
* **#2317** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2317?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2325** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2325?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2325?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.vale/styles/config/vocabularies/technical/accept.txt
```diff
@@ -152,6 +152,7 @@ cooldown
 crypto
 deserializable
 deserializer
+deserializers
```
webern:
```suggestion
deserializer(s?)
```

.claude/skills/config-system/SKILL.md
```diff
@@ -123,6 +123,19 @@ Reserve `#[serde(flatten)]` for a struct that genuinely groups several *top-leve
...
+`GetInt`, `GetString`, ...), so a leaf's permissiveness follows from its declared type, not from the
+key: `dogstatsd_port: "8125"`, or a `string`-typed leaf written as a YAML boolean, are configurations
+the Agent accepts. `datadog-agent/config/src/cast_de.rs` ports those casts, codegen attaches one to
+every scalar leaf by type (`permissivize`), and `env_decode` shares the same parsers, so every source
+accepts the same spellings. A value the cast cannot convert is a hard error rather than the Agent's
+silent zero value.
+
+Do not hand-write a per-key tolerant deserializer for a schema-typed scalar; give the coercion to the
+type instead. A generated field whose type has no classification fails the build.
+
```
webern, line 135: This is much too wordy for a skill text and may not be needed at all. What is the actual concern that this is addressing. I need you to tell me what mistake an agent might make if this were omitted entirely, then we need to find the fewest number of words that would be likely to prevent that mistake.

lib/agent-data-plane-config-system/src/system.rs
```diff
@@ -546,6 +546,26 @@ mod tests {
...
+            Some(json!({
+                "use_v3_api": { "series": { "enabled": true } },
+                "dogstatsd_port": "8126",
+            })),
+            None,
+        )
+        .await
+        .expect("scalars in Agent-castable forms boot");
+
+        assert_eq!(system.config().shared.metrics_encoding.v3_series_mode.mode, "true");
```
webern: What's going on here? Are we storing `system.config().shared.metrics_encoding.v3_series_mode.mode` as a `String` instead of a `bool`? Why? In typed config we should be storing values in system-native types (i.e. as a primitive boolean) not as a config-friendly-string that the system later has to coerce.

```suggestion
        assert_eq!(system.config().shared.metrics_encoding.v3_series_mode.mode, true);
```

lib/datadog-agent/config/src/cast_de.rs
```diff
@@ -0,0 +1,446 @@
...
+
+    fn expecting(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("a boolean, a boolean string, or an integer")
+    }
+
+    fn visit_bool<E: de::Error>(self, value: bool) -> Result<bool, E> {
+        Ok(value)
+    }
+
+    fn visit_i64<E: de::Error>(self, value: i64) -> Result<bool, E> {
```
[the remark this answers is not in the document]
webern: We hates it, precious. But we fixed it in https://github.com/DataDog/saluki/pull/2325/commits/f21f6592cb36d486f396b363bd3c7ae529fb350a

lib/datadog-agent/config/src/cast_de.rs
```diff
@@ -0,0 +1,446 @@
...
+}
+
+/// `cast.ToInt64E` for a string.
+///
+/// # Errors
+///
+/// Returns a message naming the value when it is not a decimal integer.
+pub(crate) fn parse_i64(raw: &str) -> Result<i64, String> {
+    raw.trim()
+        .parse::<i64>()
```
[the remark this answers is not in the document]
webern: We hates it, precious. But we fixed it in https://github.com/DataDog/saluki/pull/2325/commits/f21f6592cb36d486f396b363bd3c7ae529fb350a

lib/agent-data-plane-config-system/src/system.rs
webern: Turns out this field is not a boolean. It is an enum (defined in schema documentation rather than typed), with the values `true`, `false` and `datadog_only`. So https://github.com/DataDog/saluki/pull/2325/commits/ea9db6756666f6ee41b9be6308b9dbc07cc8bd3c models it as an enum following an established pattern we have for these and documents the pattern in the config-system skill.

aqian01 (changes requested):
I think the start up path still uses the raw config

Code trace:
1. [passes raw map to dogstatsd](https://github.com/DataDog/saluki/blob/m/confra-cast/bin/agent-data-plane/src/cli/run.rs#L436-L439)
2. [Uses the raw config](https://github.com/DataDog/saluki/blob/m/confra-cast/bin/agent-data-plane/src/cli/run.rs#L725-L731)
3. [Constructed from raw config](https://github.com/DataDog/saluki/blob/m/confra-cast/bin/agent-data-plane/src/cli/run.rs#L768-L775)
4. [And the field requires u16 ](https://github.com/DataDog/saluki/blob/m/confra-cast/lib/saluki-components/src/sources/dogstatsd/mod.rs#L332-L338)

Ran a regression test to confirm

```
#[tokio::test]
async fn dogstatsd_accepts_quoted_port() {
    let values = json!({ "dogstatsd_port": "8126" });
    let (config, _) = ConfigurationLoader::for_tests(Some(values), None, false).await;
    let config = DogStatsDConfiguration::from_configuration(&config)
        .expect("quoted port should work");

    assert_eq!(config.port, 8126);
}
```

webern:
> I think the start up path still uses the raw config
> 
> Code trace:
> 
> 1. [passes raw map to dogstatsd](https://github.com/DataDog/saluki/blob/m/confra-cast/bin/agent-data-plane/src/cli/run.rs#L436-L439)
> 2. [Uses the raw config](https://github.com/DataDog/saluki/blob/m/confra-cast/bin/agent-data-plane/src/cli/run.rs#L725-L731)
> 3. [Constructed from raw config](https://github.com/DataDog/saluki/blob/m/confra-cast/bin/agent-data-plane/src/cli/run.rs#L768-L775)
> 4. [And the field requires u16 ](https://github.com/DataDog/saluki/blob/m/confra-cast/lib/saluki-components/src/sources/dogstatsd/mod.rs#L332-L338)
> 
> Ran a regression test to confirm
> 
> ```
> #[tokio::test]
> async fn dogstatsd_accepts_quoted_port() {
>     let values = json!({ "dogstatsd_port": "8126" });
>     let (config, _) = ConfigurationLoader::for_tests(Some(values), None, false).await;
>     let config = DogStatsDConfiguration::from_configuration(&config)
>         .expect("quoted port should work");
> 
>     assert_eq!(config.port, 8126);
> }
> ```

This is the behavior on main as well, confirmed by running your test there. This PR does not change any raw config path behavior. It makes the typed configuration model coerce scalar leaves the way the Agent does, which will cover `dogstatsd_port` when migrated. `DogStatsDConfiguration` still reads raw_map(), so the fix for `dogstatsd_port` arrives when that component is migrated to typed config.

aqian01 (approved):

PR #2329, opened 2026-08-13
chore(dev): simplify PR title requirements

tobz:
## Summary

As stated in the PR title.

More or less:

- while I love conventional/semantic PR titles, it's been somewhat cumbersome when the scope to pick is non-obvious
- requiring a scope, even when there's only a few, can sometimes be duplicative of a well-written description portion (`fix(datadog): properly do foo in Datadog forwarder`, etc)
- bot PRs throw a wrench in these requirements and have necessitated the writing of _other_ workflows to fix them up (😓)

In this PR, we've updated the PR title check to relax our constraints: only a type is required (`fix`, `enhancement`, etc) and we skip the check entirely when the PR is labeled such that we determine it's a bot PR. We can still specify a scope if we want to.... and, to be clear, we should! Scopes are part of good commit message/PR description hygiene. However, we no longer _require_ doing so. As part of that, we're now able to get rid of the supporting Make target and scripts for generating the list of accepted scopes.

Likewise, we've removed the workflow for renaming bot PR titles since we just ignore the PR title check for them.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Merging is required before the workflow changes can be run.

## References

DADP-2


aqian01 (approved):

.github/workflows/check-pr-title.yml
```diff
@@ -25,76 +25,13 @@ jobs:
...
-          requireScope: true
           types: |
             chore
             enhancement
             feat
             fix
             docs
             revert
-
-          scopes: |
```
jszwedko, line 37: Is there any way to keep the list of scopes such that, if the optional scope is provided, it has to match one of them? My concern is we may end up with a wide variety of scopes picked by LLMs. Though, maybe this is fine 🤔

tobz: If we specify a list at all, then yeah, that becomes the allowlist.

My reason for removing the list at all, instead of just removing the `requireScope` bit... is that it means we have to generate that list of allowed scopes, and that's where it gets annoying, having to either craft a suitable fixed list _or_ have a script to generate them from a known set of rules but then making sure that list stays in sync, etc etc.

It just felt easier to remove the need for a list at all. I'm not against adding it back if LLMs go off the rails with scope choice, though.

PR #2345, opened 2026-08-14
fix(agent-data-plane): fail the build when release metadata is missing

jszwedko:
## Human Summary

Ensure that the ADP metadata is populated for release builds. This actually caught and fixes an issue with the Windows and AIX builds using the wrong env var name for the build time.

Separately, also caught an issue with the docker image build where it wasn't propagating `APP_DEV_BUILD`.

## AI Summary
`saluki-metadata` silently substitutes placeholders (`unknown`, `0.0.0`, `0000-00-00 00:00:00`) for any `APP_*` variable a build doesn't supply. That's what makes a bare `cargo build` work, but it also means a *release* build whose tooling forgot or misspelled one of those variables ships without anyone noticing — the binary just reports `unknown` to users and telemetry. Two instances were already live in `main`, which is what prompted this. Rather than only fixing them, this makes the whole class of bug loud: once a build supplies application identity **and** declares itself a release build, the build script fails instead of quietly falling back.

This is the first step of a broader cleanup of how ADP's build metadata is plumbed; it's self-contained and doesn't depend on where that lands.

### The two live bugs
- **Release images were marked as dev builds.** `.gitlab/build.yml` passes `--build-arg APP_DEV_BUILD`, but `docker/Dockerfile.agent-data-plane` never declared a matching `ARG`, so Docker dropped it and the build script fell back to `true`. Tagged release images therefore emitted `running{version="1.6.0-dev-<sha>"}` instead of `1.6.0`.
- **Windows binaries had no build time.** `windows-build-adp.ps1` set `APP_BUILD_DATE`, but `saluki-metadata` reads `APP_BUILD_**TIME**`, and the Windows job sets nothing else — so those binaries reported `0000-00-00 00:00:00`.

### Why the check isn't keyed on `APP_DEV_BUILD` alone
`APP_DEV_BUILD` is set at the **workflow** level in `.gitlab-ci.yml`, so on a tag pipeline it's in scope for every job — including the `.gitlab/test.yml` jobs (unit tests, clippy, miri), which have no `rules:` and no reason to supply application metadata. Enforcing purely on `APP_DEV_BUILD=false` would have failed the entire release pipeline. The guard therefore only engages when something actually tried to identify the application, which no test or lint job does. Forgetting a single variable still trips it, since the others remain set.

### Also included
Small fixes that keep the guard honest rather than papering over it: the Makefile's dead `APP_BUILD_DATE` assignments are corrected (the right value already reached cargo via the global `export` at the top of the file, so these were misleading rather than broken), `build-adp-aix.sh` now stamps a real timestamp since it defaults to a release build and would otherwise trip the new check, and the build-metadata docs are brought in line with reality.

## Test plan
Verified the guard against the full matrix by invoking `cargo build -p saluki-metadata` with each environment, confirming no false positives on the paths that must keep working:

| Scenario | Expected | Result |
| --- | --- | --- |
| Plain dev build, no vars | builds | builds |
| Test job on a **tag** pipeline (`APP_DEV_BUILD=false` only) | builds | builds |
| Test job on a tag + Makefile-exported git hash/build time | builds | builds |
| Release build, complete metadata | builds | builds |
| Release build, `APP_BUILD_TIME` forgotten | **fails** | fails |
| Release build, `APP_IDENTIFIER` misspelled | **fails** | fails |
| Dev build with metadata (non-tag CI) | builds | builds |

- [x] Failure messages name the specific offending variable (verified for the misspelling case)
- [x] `make -n build-adp` confirms the corrected `APP_BUILD_TIME` reaches cargo
- [ ] Worth confirming on a tag pipeline before this leaves draft, since the release path is the thing being protected


ci/tooling/windows-build-adp.ps1
```diff
@@ -46,15 +46,15 @@ Install-CachedZipTool `
...
+# ADP_APP_VERSION / ADP_APP_BUILD_TIME in Makefile) so the Windows binary identifies itself
 # the same way as the linux/darwin binaries do.
 $env:APP_FULL_NAME = "Agent Data Plane"
 $env:APP_SHORT_NAME = "data-plane"
 $env:APP_IDENTIFIER = "agent-data-plane"
 $env:APP_VERSION = $env:ADP_VERSION
 # Windows PowerShell 5.1 (the default `powershell.exe` in the LTSC2022 build image) doesn't
 # have Get-Date's -AsUTC switch (added in PS 7.1). ToUniversalTime() works on both.
-$env:APP_BUILD_DATE = (Get-Date).ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ssZ")
+$env:APP_BUILD_TIME = (Get-Date).ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ssZ")
```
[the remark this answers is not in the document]
jszwedko: Fixed in [03247a3](https://github.com/DataDog/saluki/pull/2345/commits/03247a3e9cf52bdd9c38040e5d99aa51138b5d2c)

lib/saluki-metadata/build.rs
```diff
@@ -33,6 +33,45 @@ fn main() {
...
         .unwrap_or(0);
 
+    // Release builds shouldn't silently ship the placeholder values above, so treat any that survive as a build
+    // failure. We can't key off APP_DEV_BUILD alone: CI sets it at the workflow level, so on a tag pipeline it's also
+    // set for test/lint jobs that supply no metadata, hence only enforcing when something identified the application.
+    let identifying_build = ["APP_FULL_NAME", "APP_SHORT_NAME", "APP_IDENTIFIER", "APP_VERSION"]
+        .iter()
+        .any(|var_name| env_var_present(var_name));
+
+    if !app_dev_build && identifying_build {
```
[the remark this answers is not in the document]
jszwedko: Fixed in [03247a3](https://github.com/DataDog/saluki/pull/2345/commits/03247a3e9cf52bdd9c38040e5d99aa51138b5d2c)

ci/tooling/windows-build-adp.ps1
jszwedko, line 57: Was using the wrong name.

ci/tooling/windows-integration-tests.ps1
```diff
@@ -59,9 +59,21 @@ Invoke-Native docker version
 Write-Host "[*] Building Panoramic and Agent Data Plane for Windows..."
 # saluki-metadata reads these at build time. They must match the values that
 # the Linux Makefile passes through, otherwise ADP's log subagent prefix
-# renders as "UNKNOWN" instead of "DATAPLANE".
+# renders as "UNKNOWN" instead of "DATAPLANE". Keep the set complete: saluki-metadata
+# rejects partial metadata on release builds, so a gap here would break this job if it
+# were ever handed APP_DEV_BUILD=false.
 $env:APP_FULL_NAME = "Agent Data Plane"
 $env:APP_SHORT_NAME = "data-plane"
+$env:APP_IDENTIFIER = "adp"
```
jszwedko, line 67: Sets missing variables.

lib/saluki-metadata/build.rs
```diff
@@ -33,6 +33,45 @@ fn main() {
         .and_then(|v| v.parse::<u32>().ok())
         .unwrap_or(0);
 
+    // Release builds shouldn't silently ship the placeholder values above, so treat any that survive as a build
+    // failure. We can't key off APP_DEV_BUILD alone: CI sets it at the workflow level, so on a tag pipeline it's also
+    // set for test/lint jobs that supply no metadata, hence only enforcing when something identified the application.
```
tobz, line 38: I guess my thought is: why even have this differentiating logic in the first place? If we're going through the trouble here of making this more misuse-resistant, let's just make them all required 100% of the time... remove the fallback logic, etc etc.

tobz (changes requested):

lib/saluki-metadata/build.rs
jszwedko: Yeah, that's fair. I was trying to make bare `cargo build`s work locally without setting a bunch of variables (also integration tests would need updates I think). The alternative is that we push people through the Makefile all of the time. I'll poke at this some more.

jszwedko: After some more poking I landed on keeping it as-is so that a bare `cargo build` continues to work for local development.

I briefly investigated a build.rs-based approach to support `cargo build` while populating the variables but ran into the following problems for packaging in the git hash:

> 1. .dockerignore excludes .git/, and every docker build uses context .. Un-ignoring it is a non-starter: .git is 700 MB and the Dockerfile does COPY . /adp.
> 2. The antithesis build images don't install git at all — only build-essential ca-certificates make cmake gcc g++ perl protobuf-compiler curl unzip rustup. (The main ADP build image does have it.)
> 3. Worktree layout. Here .git is a file, and HEAD resolves to /Users/…/saluki/.git/worktrees/default-app-variables/HEAD — outside the crate and outside the workspace. So cargo:rerun-if-changed=.git/HEAD silently does nothing; you'd need git rev-parse --git-path HEAD plus the resolved ref. Since this repo is actively developed in worktrees, getting it wrong means a silently stale hash, which is worse than unknown — a wrong-but-plausible SHA is harder to spot than an obvious placeholder.

In particular, 3 swayed me: the idea of having a stale hash seems troublesome.

Build time could be derivable if we have `build.rs` always run.

Curious what you think though.

tobz: Ehhhhh, fair enough.

tobz (approved):

PR #2346, opened 2026-08-14
enhancement(agent-data-plane): let the binary declare its own identity

jszwedko:
> **Stacked on #2345.** Base is `jszwedko/fail-on-placeholder-build-metadata`; review that one first. The diff shown here is only this step.

## Human Summary

Move static metadata or inferable metadata (version) into ADP to remove the need to set some of the `APP_` variables. This, I think, will remove one source of bugs resulting from inconsistent setting as was seen in https://github.com/DataDog/saluki/pull/2343 and enable easier local development (don't need to remember to set the variables). Instead, we only set the metadata that is only known at build time (like 🥁 the build time).

## AI Summary
`saluki-metadata` bakes the application's name and version into a shared library at compile time — something it fundamentally cannot do per-binary, since one compiled copy is shared by every binary in the workspace. The only lever scoped to a single build command is the process environment, which is why every build entry point has had to independently remember the same four values. That's thirteen places, and the drift it produces is what the last two PRs were cleaning up. This removes the cause rather than the symptoms.

The fix is to split the eight fields by what they actually describe:

| Per-**binary** — moves to the binary | Per-**build** — stays in the build script |
| --- | --- |
| `full_name`, `short_name`, `identifier`, `version` | `git_hash`, `build_time`, `dev_build`, `target_arch` |

ADP declares the left column once, in `main.rs`:

```rust
pub const APP_DETAILS: AppDetails = saluki_metadata::declare_app_details!(
    full_name  = "Agent Data Plane",
    short_name = "data-plane",
    identifier = "adp",
);
```

The version comes from Cargo — `env!` inside the macro expands in the *calling* crate, so it picks up ADP's version, and `CARGO_PKG_VERSION_{MAJOR,MINOR,PATCH}` replaces the hand-rolled version splitting in the build script. The right column is genuinely shared (one build has one git hash regardless of how many binaries it produces), so it stays put — as does the release-metadata guard from #2345, narrowed to the fields it still covers. `target_arch` has to stay there regardless: `TARGET` is only set for build scripts.

**Registration happens as the first statement of `main`.** The details are a compile-time constant with no relationship to configuration, so they're registered before anything else rather than being routed through the config-driven bootstrap phase — which also means the window where config loading and logging translation run is already covered. The `get_app_details()` signature is unchanged, so every existing call site is untouched, including `saluki-app` and the `version` command.

Net effect: `APP_FULL_NAME`, `APP_SHORT_NAME`, `APP_IDENTIFIER`, and `APP_VERSION` stop being build inputs entirely and disappear from the Makefile, `Dockerfile.agent-data-plane`, `.gitlab/build.yml`, both Windows scripts, the AIX script, and both antithesis Dockerfiles. This also makes the value fixes in #2343 and the metadata completion in #2345 moot by deletion — those lines no longer exist to drift.

## Also included
`run_inner`'s `Action::Version` arm was unreachable — `version` returns early, before configuration is loaded, so that reporting the version never depends on usable config. The dead arm quietly implied the opposite, so it now asserts the invariant instead: removing the early return fails loudly rather than silently turning `version` into a config-requiring command.

## Behavior changes
- **A bare `cargo build` now produces a correctly identified binary.** This was the original motivation, and it's per-binary rather than the workspace-wide `.cargo/config.toml` default we ruled out earlier.
- **Windows version strings are fixed.** Non-tag Windows builds embedded `v12345-abc1234` (a pipeline slug) as the version; they now report the real version, matching every other platform. No other platform's version string changes — the container and tarball paths already used the Cargo version.
- `ADP_APP_VERSION` survives in the Makefile, but only because release artifacts are named after it.

## Test plan
- [x] `cargo build` with no environment at all reports `Agent Data Plane` / `data-plane` / `adp` / `1.6.0` — previously `unknown`
- [x] `make build-adp` reports the same, plus a real git hash
- [x] `agent-data-plane version` still prints `v1.6.0-<sha>`; the leading `v` is display-only and unaffected
- [x] `cargo clippy --workspace --all-targets` clean; 1235 tests pass across `saluki-metadata`, `saluki-app`, `saluki-components`, `datadog-agent-commons`, and `agent-data-plane`
- [x] Confirmed no dangling references to the four removed variables anywhere in the repo
- [x] New `tests/version_cli.rs` asserts the built binary reports its real identity — verified it actually catches a regression by moving the registration below the version early-return, which compiles fine and fails the test
- [ ] Container, Windows, and AIX builds are unexercised locally — worth a pipeline run before this leaves draft

## Follow-up, not in this PR
Library unit tests observe the unregistered fallback, because there's no `main()` to register. That's true today as well (the build script yields `unknown` under `cargo test`), so nothing regresses — `endpoints.rs` asserts `0-0-0-unknown.agent` against itself either way. What changes is that it becomes *fixable*: identity can now be injected at runtime. Doing it properly needs an idempotent helper, since a `OnceLock` is shared across parallel tests in one binary.


ci/tooling/build-adp-aix.sh
```diff
@@ -28,14 +28,10 @@ export RANLIB="${ADP_AIX_RANLIB}"
 export CARGO_HOME
 export CARGO_TARGET_DIR
 
-ADP_APP_VERSION_AUTO="$(grep -E '^version = "' "${repo_root}/bin/agent-data-plane/Cargo.toml" | head -n 1 | cut -d '"' -f 2)"
 APP_GIT_HASH_AUTO="$(git -C "${repo_root}" rev-parse --short HEAD 2>/dev/null || echo not-in-git)"
 
-export APP_FULL_NAME="${APP_FULL_NAME:-${ADP_APP_FULL_NAME:-Agent Data Plane}}"
-export APP_SHORT_NAME="${APP_SHORT_NAME:-${ADP_APP_SHORT_NAME:-data-plane}}"
-export APP_IDENTIFIER="${APP_IDENTIFIER:-${ADP_APP_IDENTIFIER:-adp}}"
+# Only build metadata is passed in; ADP declares its own name and Cargo supplies its version.
```
jszwedko:
```suggestion
```

ci/tooling/windows-build-adp.ps1
```diff
@@ -44,14 +44,10 @@ Install-CachedZipTool `
...
-# ADP_APP_VERSION / ADP_APP_BUILD_TIME in Makefile) so the Windows binary identifies itself
-# the same way as the linux/darwin binaries do.
-$env:APP_FULL_NAME = "Agent Data Plane"
-$env:APP_SHORT_NAME = "data-plane"
-$env:APP_IDENTIFIER = "agent-data-plane"
-$env:APP_VERSION = $env:ADP_VERSION
+# saluki-metadata reads these at build time. Must match what the Makefile passes through
+# (ADP_APP_GIT_HASH / ADP_APP_BUILD_TIME) so the Windows binary reports the same build
+# metadata as the linux/darwin binaries do. The application's name and version aren't here:
+# ADP declares those itself and Cargo supplies the version.
```
jszwedko:
```suggestion
# saluki-metadata reads these at build time. Must match what the Makefile passes through
# (ADP_APP_GIT_HASH / ADP_APP_BUILD_TIME) so the Windows binary reports the same build
# metadata as the linux/darwin binaries do.
```

ci/tooling/windows-integration-tests.ps1
```diff
@@ -57,21 +57,8 @@ if (-not (Get-Command docker -ErrorAction SilentlyContinue)) {
...
-$env:APP_IDENTIFIER = "adp"
-# Mirrors ADP_APP_VERSION_AUTO in the Makefile: the version comes from ADP's manifest.
-$AdpManifest = Join-Path $RepoRoot "bin\agent-data-plane\Cargo.toml"
-$AdpVersionMatch = Select-String -Path $AdpManifest -Pattern '^version = "(.+)"' | Select-Object -First 1
-if (-not $AdpVersionMatch) {
-    throw "Could not read the ADP version from ${AdpManifest}"
-}
-$env:APP_VERSION = $AdpVersionMatch.Matches[0].Groups[1].Value
+# saluki-metadata reads these at build time. ADP declares its own name and version, so only
+# build metadata is passed in here.
```
jszwedko:
```suggestion
# saluki-metadata reads these at build time.
```

docs/agent-data-plane/releasing.md
```diff
@@ -81,13 +81,13 @@ during the build process and is used to drive a number of behaviors:
...
 - special constant identifiers that are used to populate things like HTTP request headers (user agent, etc)
 
-This build metadata is calculated with a Make target—`emit-adp-build-metadata`—which populates it during local builds or
-regular CI builds. The relevant build arguments are all prefixed with `APP_` and are as follows:
+ADP's identity—its full name, short name, identifier, and version—isn't passed in by the build. The binary declares it
+in `bin/agent-data-plane/src/main.rs` using `saluki_metadata::declare_app_details!`, and Cargo supplies the version from
+`bin/agent-data-plane/Cargo.toml`. Changing any of those means editing that declaration, not the build tooling.
+
+What remains describes the build rather than the application, and is calculated with a Make
+target—`emit-adp-build-metadata`—which populates it during local builds or regular CI builds:
```
jszwedko: Leave this description as it was. We don't need to note that some of the metadata has been moved.

Makefile
```diff
@@ -11,9 +11,10 @@ export APP_GIT_HASH := $(or $(CI_COMMIT_SHA),$(shell git rev-parse --short HEAD
 export APP_BUILD_TIME := $(or $(CI_PIPELINE_CREATED_AT),0000-00-00T00:00:00-00:00)
 
 # ADP-specific settings used during builds.
-export ADP_APP_FULL_NAME := Agent Data Plane
-export ADP_APP_SHORT_NAME := data-plane
-export ADP_APP_IDENTIFIER := adp
+#
+# ADP's name and version aren't here: the binary declares them itself (see `declare_app_details!` in
+# bin/agent-data-plane/src/main.rs) and Cargo supplies the version, so builds no longer pass them in. ADP_APP_VERSION
+# survives only because release artifacts are named after it.
```
jszwedko:
```suggestion
```

bin/agent-data-plane/src/main.rs
```diff
@@ -217,7 +231,8 @@ async fn run_inner(
         Action::Debug(cmd) => handle_debug_command(local_config, cmd).await,
         Action::Config(cmd) => handle_config_command(local_config, cmd).await,
         Action::Dogstatsd(cmd) => handle_dogstatsd_command(local_config, cmd).await,
-        Action::Version(v) => handle_version_command(v.json).await,
+        // Handled before bootstrap, so that reporting the version never depends on there being usable configuration.
+        Action::Version(_) => unreachable!("version is handled before bootstrap"),
```
jszwedko, line 235: This is an unrelated change, but 🤖 noticed it while I was in the area.

lib/saluki-metadata/build.rs
```diff
@@ -1,63 +1,32 @@
...
     // failure. We can't key off APP_DEV_BUILD alone: CI sets it at the workflow level, so on a tag pipeline it's also
-    // set for test/lint jobs that supply no metadata, hence only enforcing when something identified the application.
-    let identifying_build = ["APP_FULL_NAME", "APP_SHORT_NAME", "APP_IDENTIFIER", "APP_VERSION"]
-        .iter()
-        .any(|var_name| env_var_present(var_name));
+    // set for test/lint jobs that supply no build metadata, hence only enforcing when some of it was supplied.
+    let metadata_supplied = env_var_present("APP_GIT_HASH") || env_var_present("APP_BUILD_TIME");
 
-    if !app_dev_build && identifying_build {
+    if !app_dev_build && metadata_supplied {
```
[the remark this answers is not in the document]
jszwedko: Fixed in [3cabab8](https://github.com/DataDog/saluki/pull/2346/commits/3cabab83820a088977f43fa5475acbf4599ea675)

ci/tooling/windows-build-adp.ps1
```diff
@@ -44,14 +44,9 @@ Install-CachedZipTool `
...
-# (ADP_APP_FULL_NAME / ADP_APP_SHORT_NAME / ADP_APP_IDENTIFIER / ADP_APP_GIT_HASH /
-# ADP_APP_VERSION / ADP_APP_BUILD_TIME in Makefile) so the Windows binary identifies itself
-# the same way as the linux/darwin binaries do.
-$env:APP_FULL_NAME = "Agent Data Plane"
-$env:APP_SHORT_NAME = "data-plane"
-$env:APP_IDENTIFIER = "agent-data-plane"
-$env:APP_VERSION = $env:ADP_VERSION
+# saluki-metadata reads these at build time. Must match what the Makefile passes through
+# (ADP_APP_GIT_HASH / ADP_APP_BUILD_TIME) so the Windows binary reports the same build
+# metadata as the linux/darwin binaries do.
```
tobz: This comment is now orphaned and should be removed.

ci/tooling/windows-integration-tests.ps1
```diff
@@ -57,21 +57,7 @@ if (-not (Get-Command docker -ErrorAction SilentlyContinue)) {
...
-$env:APP_SHORT_NAME = "data-plane"
-$env:APP_IDENTIFIER = "adp"
-# Mirrors ADP_APP_VERSION_AUTO in the Makefile: the version comes from ADP's manifest.
-$AdpManifest = Join-Path $RepoRoot "bin\agent-data-plane\Cargo.toml"
-$AdpVersionMatch = Select-String -Path $AdpManifest -Pattern '^version = "(.+)"' | Select-Object -First 1
-if (-not $AdpVersionMatch) {
-    throw "Could not read the ADP version from ${AdpManifest}"
-}
-$env:APP_VERSION = $AdpVersionMatch.Matches[0].Groups[1].Value
+# saluki-metadata reads these at build time.
```
tobz: This comment is now orphaned and should be removed.

lib/saluki-metadata/src/lib.rs
```diff
@@ -1,56 +1,116 @@
...
+    (full_name = $full_name:expr, short_name = $short_name:expr, identifier = $identifier:expr $(,)?) => {
+        $crate::AppDetails::new(
+            $full_name,
+            $short_name,
+            $identifier,
+            $crate::Version::new(
+                env!("CARGO_PKG_VERSION"),
+                $crate::const_parse_u32(env!("CARGO_PKG_VERSION_MAJOR")),
+                $crate::const_parse_u32(env!("CARGO_PKG_VERSION_MINOR")),
+                $crate::const_parse_u32(env!("CARGO_PKG_VERSION_PATCH")),
```
tobz, line 69: You should be able to replace this with [`u32::from_str_radix`](https://doc.rust-lang.org/stable/std/primitive.u32.html#method.from_str_radix), which is already `const` capable.

tobz (changes requested):

lib/saluki-metadata/src/lib.rs
jszwedko: Ooo, TIL. Yes, that looks better.

jszwedko: Changed in [4abca41](https://github.com/DataDog/saluki/pull/2346/commits/4abca41a974a8ed92ac4fec750e9bddc9272b3b8)

tobz (approved):

PR #2355, opened 2026-08-17
feat(agent-data-plane): filter the tagger stream and detect new entity prefixes

jszwedko:
# Human Summary

This PR does two things:

1. Introduces a mechanism to alert us of new entity prefix additions that ADP doesn't recognize.  It works by introducing a test that validates against a fixture file that contains the Agent's list of prefixes. It then also introduces a GitHub workflow that runs nightly to update it from datadog-agent main. That PR will alert us of the new prefix where we will decide to handle or ignore it.
2. When querying the Agent entity stream, send a filter to the Agent to only send us entities ADP actually cares about. I kept the ignore list in ADP, as well, as a safe guard, but I'm open to dropping this.

## AI Summary

#2353 stops ADP warning about entity ID prefixes it has no use for, but ADP still asks the Agent for all of them and discards what it doesn't want. This asks only for the nine prefixes it converts, so the Agent stops serializing and streaming entities that are dropped on arrival. The catch is that server-side filtering makes the next new entity type *silent* — where today an unclassified prefix at least produces a log line, a filtered subscription simply never delivers it. So this also adds a tripwire: a fixture recording the Agent's declared prefixes, a test asserting each one is either handled or explicitly ignored, and a nightly workflow that refreshes the fixture from the Agent's `main` branch and opens a PR when it changes.

The point of the workflow is that the PR *is* the notification, and the fix lands on the same branch:

```mermaid
flowchart LR
    A["Agent adds a prefix<br/>on main"] --> B["nightly workflow<br/>refreshes fixture"]
    B --> C{"fixture<br/>changed?"}
    C -->|no| D["done, nothing opened"]
    C -->|yes| E["opens PR"]
    E --> F["CI fails: prefix is<br/>neither handled nor ignored"]
    F --> G["we classify it on<br/>that branch, CI goes green"]
    G --> H["merge"]
```

This catches additions ~24h after they merge to Agent master, rather than when we next bump the Agent version. For the case that motivated all of this: `kueue_workload` merged upstream on 2026-07-09 and first appeared in our logs on 2026-08-02, so a nightly would have raised it about three and a half weeks earlier — and before it reached a release.

## Notable design decisions

**The test asserts a subset, not equality.** Every prefix the Agent declares must be classified, but our lists may contain prefixes the Agent no longer declares. ADP runs alongside a range of Agent versions, so a removal upstream must not force us to drop an arm we still need for older Agents. In practice: additions fail CI, removals show up as a deleted fixture line with CI green, and a human decides whether to prune.

**The ignored arm is driven by the list.** `remote_entity_id_to_entity_id` matches on `IGNORED_ENTITY_ID_PREFIXES` rather than repeating the literals, so the list and the match can't disagree. `HANDLED_ENTITY_ID_PREFIXES` does still exist alongside its match arms — that pairing is what `prefix_lists_agree_with_conversion` checks, since a prefix we subscribe to but don't convert would silently drop entities we asked for.

**Extraction fails loudly.** The dangerous failure isn't a crash, it's a silent shrink: a 404, an HTML error page, or an upstream refactor of that `const` block all yield zero prefixes, and an empty fixture makes the test pass while checking nothing. Both the workflow and `agent_entity_id_prefix_fixture_is_intact` require a plausible minimum plus long-standing anchor prefixes (`container_id`, `kubernetes_pod_uid`, `internal`).

**Branch names embed a content hash**, so each distinct change gets its own branch and a change we've already reported is recognizable. The workflow checks for that branch before pushing anything and skips the run if it exists: the branch carries a commit that isn't on `main`, so pushing to it would be rejected as a non-fast-forward and fail every night until the PR was dealt with, making a stale run indistinguishable from a broken one. (`update-agent-version.yml` has the same ordering issue, but its weekly cadence rarely exposes it.) Older PRs from the workflow are closed with a comment pointing at the newer one.

One consequence worth knowing: if a drift PR is closed *without* deleting its branch, that exact prefix set won't be reported again. Branch deletion on close is the norm, in which case the next nightly re-opens it — which is the nagging behavior we want.

## One thing to confirm before merging

**Does the octo-sts token permit reading `DataDog/datadog-agent`?** The new policy (`.github/chainguard/self.update-agent-tagger-prefixes.create-pr.sts.yaml`) is scoped to `DataDog/saluki`, matching the existing version-bump policy. `update-agent-version.yml` already calls the API against `DataDog/datadog-agent` with an equivalent token and works, and the repo is public, so I expect `repos.getContent` to be fine — but it's untested until this runs on `main`.

## Related

Pairs with DataDog/datadog-agent#55012, which adds `@DataDog/agent-data-plane` as a code owner of `comp/core/tagger/types/entity_id.go` so we also hear about a new prefix at Agent review time, before it merges.

## Test plan

- [x] `agent_entity_id_prefixes_are_all_classified` fails with the intended actionable message when a prefix is present in the fixture but classified nowhere — verified by appending a fake prefix locally
- [x] `agent_entity_id_prefix_fixture_is_intact` guards against a truncated or empty fixture
- [x] `prefix_lists_agree_with_conversion` covers handled→`Some`, ignored→`None`, and no overlap between the lists
- [x] Ran the workflow's extraction regex against the real `entity_id.go` at `origin/main`: yields exactly the 17 prefixes in the fixture, and re-running the fixture rewrite produces a byte-identical file, so a nightly run on an unchanged Agent opens no PR
- [ ] Workflow itself needs a `workflow_dispatch` run on `main` to validate the octo-sts token path


bin/agent-data-plane/src/internal/env/workload/collectors/tagger.rs
```diff
@@ -216,6 +220,46 @@ impl MemoryBounds for RemoteAgentTaggerMetadataCollector {
...
+const IGNORED_ENTITY_ID_PREFIXES: &[&str] = &[
+    "gpu",
+    "kueue_workload",
+    "kubernetes_kueue_queue",
+    "kueue_resource_flavor",
+    "crd",
+    "kubernetes_capabilities",
+    "host",
+    "kubelet",
+];
```
tobz: Constants should be declared at the top of a file, after imports.

bin/agent-data-plane/src/internal/env/workload/collectors/tagger.rs
```diff
@@ -304,27 +334,87 @@ mod tests {
         }
     }
 
+    /// The Agent's set of tagger entity ID prefixes, regenerated nightly by
+    /// `.github/workflows/update-agent-tagger-prefixes.yml`.
+    const AGENT_ENTITY_ID_PREFIXES: &str =
+        include_str!("../../../../../tests/fixtures/agent_tagger_entity_id_prefixes.txt");
```
tobz: I think you could replace the `../../..` bits with `env!("CARGO_MANIFEST_DIR")` which should, if I'm reading the docs right, give us the path to where the package's manifest (aka `Cargo.toml`) lives... which is exactly where the path backtracking gets you.

Basically, you'd end up with `include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/tests/fixtures/agent_tagger_entity_id_prefixes.txt"))`.

lib/datadog-agent/commons/src/ipc/client/mod.rs
```diff
@@ -134,16 +134,23 @@ impl RemoteAgentClient {
...
+    ///
+    /// The server only sends events for entities whose ID carries one of `prefixes`, which avoids paying the
+    /// serialization and bandwidth cost of entities that the caller has no use for. If `prefixes` is empty, the server
+    /// sends events for every prefix it knows about.
     ///
     /// If there is an error with the initial request, or an error occurs while streaming, the next message in the
     /// stream will be `Some(Err(status))`, where the status indicates the underlying error.
-    pub fn get_tagger_stream(&mut self, cardinality: TagCardinality) -> StreamingResponse<StreamTagsResponse> {
+    pub fn get_tagger_stream(
+        &mut self, cardinality: TagCardinality, prefixes: Vec<String>,
```
tobz: Can we make `prefixes` optional and then just do `prefixes: prefixes.unwrap_or_default()` when constructing the request payload?

Makes the intent clearer to pass `None` vs `Vec::new()` if someone doesn't want any filtering, IMO.

tobz (changes requested):

bin/agent-data-plane/src/internal/env/workload/collectors/tagger.rs
jszwedko: Ah, doh, yes, that is better.

lib/datadog-agent/commons/src/ipc/client/mod.rs
jszwedko: Agreed 👍

jszwedko: Updated in [6e89bf0](https://github.com/DataDog/saluki/pull/2355/commits/6e89bf0b7a360e1d954bf04332dac2c27462107c)

bin/agent-data-plane/src/internal/env/workload/collectors/tagger.rs
jszwedko: Updated in [6e89bf0](https://github.com/DataDog/saluki/pull/2355/commits/6e89bf0b7a360e1d954bf04332dac2c27462107c)

bin/agent-data-plane/src/internal/env/workload/collectors/tagger.rs
jszwedko: Updated in [6e89bf0](https://github.com/DataDog/saluki/pull/2355/commits/6e89bf0b7a360e1d954bf04332dac2c27462107c)

.github/workflows/update-agent-tagger-prefixes.yml
```diff
@@ -0,0 +1,245 @@
...
+          git diff -- "$FIXTURE"
+
+          # Name the branch after the content, so that an unrelated later change gets its own branch and so we can
+          # recognize a change we've already reported.
+          BRANCH_NAME="ci/update-agent-tagger-prefixes-$(sha256sum "$FIXTURE" | cut -c1-12)"
+
+          # We report each distinct change once. The branch for a change we've already reported carries a commit that
+          # isn't on main, so pushing to it would be rejected as a non-fast-forward and fail every run until the pull
+          # request is dealt with -- which would leave us unable to tell a stale run from a broken one.
+          if git ls-remote --exit-code --heads origin "$BRANCH_NAME" > /dev/null; then
```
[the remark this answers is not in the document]
jszwedko: Updated in [6e89bf0](https://github.com/DataDog/saluki/pull/2355/commits/6e89bf0b7a360e1d954bf04332dac2c27462107c)

tobz (approved):

.github/workflows/update-agent-tagger-prefixes.yml
```diff
@@ -0,0 +1,245 @@
...
+            echo "This change has already been reported on ${BRANCH_NAME}; leaving the existing pull request alone."
+            echo "needs_update=false" >> $GITHUB_OUTPUT
+            exit 0
```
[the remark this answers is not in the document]
jszwedko: Updated in [6e89bf0](https://github.com/DataDog/saluki/pull/2355/commits/6e89bf0b7a360e1d954bf04332dac2c27462107c)

PR #2358, opened 2026-08-18
fix(config): ignore core-Agent-only dogstatsd_experimental_http keys

vickenty:
## Summary

Downgrade experimental dogstatsd http settings from high severity to excluded. 

HTTP dogstatsd traffic does not reach ADP and these settings do not impact anything about regular dogstatsd processing. Both can coexist at the same time with no issue, so there is no reason to block ADP from running if dogstatsd http is enabled.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

### DIFF
--- a/docs/agent-data-plane/configuration/configuration.md
+++ b/docs/agent-data-plane/configuration/configuration.md
@@ -19,17 +19,6 @@ If you find an error on this page, please [open an issue].
 
 ## Unsupported Settings
 
-<!-- section:unsupported-in-progress -->
-### Being Worked On
-
-The following settings are not yet supported in ADP but are planned with GitHub issue links for
-tracking.
-
-| Config Key                                   | Description                                     | Issue   |
-| -------------------------------------------- | ----------------------------------------------- | ------- |
-| `dogstatsd_experimental_http.enabled`        | Enable experimental HTTP/H2C DSD listener       | [#1682] |
-| `dogstatsd_experimental_http.listen_address` | Bind address for experimental HTTP DSD listener | [#1682] |
-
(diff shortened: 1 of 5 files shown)


docs/agent-data-plane/configuration/configuration.md
```diff
@@ -19,17 +19,6 @@ If you find an error on this page, please [open an issue].
...
-<!-- section:unsupported-in-progress -->
-### Being Worked On
-
-The following settings are not yet supported in ADP but are planned with GitHub issue links for
-tracking.
-
-| Config Key                                   | Description                                     | Issue   |
-| -------------------------------------------- | ----------------------------------------------- | ------- |
-| `dogstatsd_experimental_http.enabled`        | Enable experimental HTTP/H2C DSD listener       | [#1682] |
-| `dogstatsd_experimental_http.listen_address` | Bind address for experimental HTTP DSD listener | [#1682] |
```
thieman, line 31: Is the plan then that these config keys will never work through ADP? What does the timeline for "moving out of experimental" and also "enabling in ADP" look like?

vickenty: I think it is safe to say that the experimental flags would never work in ADP. There is no timeline for moving out of experimental or adding support in ADP yet.

jszwedko (approved):
Spoke with @vickenty about this. He pointed out that until ADP supports DogStatsD-HTTP it will be served by the Core Agent even if ADP is running and serving traditional DogStatsD traffic. Given that, it seems safe to ignore these options.

PR #2361, opened 2026-08-18
fix(config): remove defunct V3 metrics validation references and fix dsd-forwarding tests

lucastemb:
## Summary

[54581](https://github.com/DataDog/datadog-agent/pull/54581) removed validation of V3 metrics, which sent identical V2 and V3 metrics simultaneously. Saluki continued to honor that obsolete config causing nightly correctness failures caused by duplicate metrics.  

This PR aims to remove defunct unit tests, correctness tests, config keys, etc. It also disables `preflight_mode` for `dsd-forwarding-*` correctness tests since the probe metric being sent to fake intake caused a mismatch between `baseline` and `comparison`. To my knowledge, `preflight_mode` should have no intended effect on the forwarding path. 

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

Correctness tests

## References

N/A


lib/saluki-components/src/forwarders/cluster_agent/mod.rs
```diff
@@ -309,7 +304,7 @@ mod tests {
         );
         assert!(config.forwarder_config.use_v3_api_series().endpoints.is_empty());
         assert!(config.forwarder_config.v3_api().series.endpoints.is_empty());
-        assert!(config.forwarder_config.v3_api().series.shadow_sites.is_empty());
+        assert!(config.forwarder_config.v3_api().series.endpoints.is_empty());
```
tobz: Duplicate line, looks like.

bin/correctness/datadog-intake/src/main.rs
```diff
@@ -50,6 +51,15 @@ async fn run() -> Result<(), GenericError> {
...
         .error_context("Failed to bind DogStatsD forwarding capture socket.")?;
+
+    let raw_socket = dogstatsd_forwarding_socket
+        .into_std()
+        .error_context("Failed to convert forwarding capture socket to std.")?;
+    if let Err(e) = SockRef::from(&raw_socket).set_recv_buffer_size(838_860) {
+        warn!(error = %e, "Failed to set UDP receive buffer size for forwarding capture socket.");
+    }
+    let dogstatsd_forwarding_socket = UdpSocket::from_std(raw_socket)
+        .error_context("Failed to re-create forwarding capture socket after buffer sizing.")?;
```
tobz: You shouldn't need to do this back and forth: `SockRef::from` will take `&S where S: std::os::fd::AsFd`, which `UdpSocket` implements... so `SockRef::from(&dogstatsd_forwarding_socket)` _should_ be enough.

tobz (changes requested):
Two small things, but otherwise looking good. 👍🏻

tobz (approved):

PR #2384, opened 2026-08-20
enhancement(core): Add efficient trace payload (aka ETP aka v1) decoder

ajgajg1134:
## Summary
Adds a decoder for the Datadog v1.0 APM trace wire format (the `idx`/`etp` msgpack layout) at `lib/saluki-components/src/decoders/datadog`, producing one unified `Trace` per trace chunk. Decode-only; the streaming string table is resolved to owned strings at decode time.

Unknown field numbers are skipped for forward compatibility, and any inline strings they carry are harvested into the string table. String references in this format are purely positional, so a skipped field carrying a new string would otherwise shift every subsequent index in the known fields that follow, silently resolving to the wrong string or failing as out of range. This matches the reference Go decoder's `harvestUnknownFieldStrings` (DataDog/datadog-agent#53918).

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?
Tested against a golden payload produced by the reference Go encoder, covering all seven `AnyValue` types, span links, span events, 128-bit trace IDs, and string deduplication, plus unit tests for the unknown-field harvesting at every struct level, the depth and size caps, and a proptest asserting arbitrary bytes never panic the decoder.


lib/saluki-components/src/decoders/datadog/read.rs
```diff
@@ -0,0 +1,323 @@
...
+pub fn read_array_len(r: &mut &[u8], context: &'static str) -> Result<u32, DecodeError> {
+    let len = rmp::decode::read_array_len(r).map_err(mp(context))?;
+    if len > MAX_SIZE {
+        return Err(DecodeError::OversizeHeader {
+            context,
+            len,
+            max: MAX_SIZE,
+        });
+    }
+    check_slab_count(len, MIN_BYTES_PER_ARRAY_ELEMENT, r, context)?;
```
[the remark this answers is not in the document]
ajgajg1134: This is acceptable, it is unlikely but technically allowed and the reference decoder in the trace-agent has the same limits.

lib/saluki-components/src/decoders/datadog/value.rs
```diff
@@ -0,0 +1,170 @@
...
+
+use super::error::DecodeError;
+use super::read;
+use super::string_table::StringTable;
+
+/// Maximum `AnyValue` nesting depth.
+///
+/// Without a bound, a deeply nested payload could drive the decoder into unbounded recursion and
+/// overflow the stack. Matches the reference decoder's limit.
+const MAX_ANY_VALUE_DEPTH: usize = 200;
```
tobz: 200 is an awfully deep call stack by itself, not taking into account that we'd likely already be 30-40 frames deep by the time we get to this point.

Is there _any_ information about why it's 200 in the reference implementation?

lib/saluki-components/src/decoders/datadog/value.rs
```diff
@@ -0,0 +1,170 @@
...
+
+/// Reads a streaming string and resolves it to an owned [`MetaString`].
+///
+/// A streaming string is encoded either as a literal string (appended to `strings`, its new index
+/// returned implicitly) or as a `uint32` index into `strings`.
```
tobz, line 21: This sounds invalid since we're always returning `MetaString`?

lib/saluki-components/src/decoders/datadog/read.rs
```diff
@@ -0,0 +1,323 @@
...
+
+/// Returns `true` if the marker byte is a MessagePack fixstr.
+fn is_fixstr(b: u8) -> bool {
+    b & 0xe0 == 0xa0
+}
+
+/// Maximum nesting depth when skipping an unknown value.
+///
+/// Bounds recursion so a deeply nested unknown field cannot overflow the stack.
+const MAX_SKIP_DEPTH: usize = 200;
```
tobz: Constants should be declared at the top of the file, below imports.

(Sorry, no good formatter lint for this to make it automatic. 😓)

lib/saluki-components/src/decoders/datadog/tests.rs
```diff
@@ -0,0 +1,676 @@
...
+fn trailing_bytes_after_payload_rejected_valid_value() {
+    let mut buf = Vec::new();
+    encode::write_map_len(&mut buf, 0).unwrap(); // empty, otherwise-valid tracer payload
+    encode::write_uint(&mut buf, 42).unwrap(); // an otherwise-valid MessagePack value
+
+    let err = decode_v1_payload(&buf).unwrap_err();
+    assert!(matches!(err, DecodeError::TrailingBytes { len: 1 }));
+}
+
+proptest! {
```
tobz, line 727: A small note here: we'll potentially want to consider moving this into `saluki-io` with borrowed data representations similar to how we do it for the DogStatsD codec.

This makes it a _little_ more amenable to running through our fuzz testing harness which already exists for `saluki-io`.

(To be clear: I'm just saying this out loud so I don't forget, and I don't expect you to do this here because it's an easy enough migration to do later.)

tobz (changes requested):

lib/saluki-components/src/decoders/datadog/value.rs
ajgajg1134: This limit in the reference decoder is fairly new and is just a generous limit to avoid extremely deep structures from instrumentation mistakes. I think we picked 200 as it seemed unlikely to be hit by any real use-case and would prevent the decoding goroutines from consuming far too much data.

This is a good point though, so I decided to poke around more and happened across this (relatively) recent otel change https://github.com/open-telemetry/opentelemetry-specification/pull/5186 which didn't exist when we originally wrote our spec. I think the `64` value they settled on is reasonable for us to use too. I'll update this and the reference decoder

lib/saluki-components/src/decoders/datadog/value.rs
tobz: Chatted with Andrew separately about this, and realized I was reading the code incorrectly.

lib/saluki-components/src/decoders/datadog/value.rs
ajgajg1134: Updated to 64 and will open a separate change in the datadog agent

lib/saluki-components/src/decoders/datadog/read.rs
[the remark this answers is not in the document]
tobz: I agree that it's satisfactory to simply match the reference decoder for the purposes of this PR, but I do think this is an interesting weakness that could be exploited / triggered unintentionally leading to a denial of service attack.

I wonder if there's anything that could be done in the future to either parse more incrementally... like we detect that the proclaimed span N is a no-op, so we never generate it, and it just leads to us never actually allocating 25 million `Span` slots because they all failed the vibe check as we went through them.

Kind of a fine line between trying to pre-allocate exactly vs simply choosing a good enough default (and decoding incrementally) such that we don't do too many resizes and don't constantly pre-allocate _too_ much.

tobz (approved):

tobz (approved):

PR #2446, opened 2026-08-21
enhancement(core): add registry for external resources

tobz:
## Summary

This PR introduces `ResourceRegistry`, a new centralized mechanism for coordinating the acquisition of external resources, such as network sockets, between components.

When a data plane starts up, it goes through the process of configuring itself, which generally involves things like binding network sockets to expose API endpoints for administrative purposes, or accepting data, and so on. These network sockets, or "external" resources, are very cut-and-dry: if we can't bind a port we need, we exit. This currently works well enough because we construct our topology once, and if it succeeds, then process runs and we don't have to worry about issues with binding sockets again at runtime because they're already bound.

However, in order to support a future where components can recover from failure, or even be intentionally added/removed/replaced dynamically at runtime, we have to contend with a new reality: we might need to destroy the object that holds an acquired external resource before creating our new component, and in doing so, introduce a risk that something else comes along and takes that resource.

This PR introduces a new mechanism for managing these resources at a higher level, `ResourceRegistry`, that is designed to solve these issues. `ResourceRegistry` provides a way to define "specifications" for types of resources -- a way to uniquely identify the resource, and how to create it -- and then is exposed to callers that need to create such resources. The registry owns resources for the lifetime of the process (or until the resource is manually discarded) and provides "leases" to callers, such that one instance of a component can hold temporary ownership of the resource, and then the resource is automatically returned when the component goes away. When the component comes back, it can reacquire the resource without the resource having ever been destroyed. This ensures that another external process cannot _steal_ a resource the data plane depends on, and it also keeps it alive for the purposes of zero-downtime changes to the topology.

Initially, the goal is to use this for network sockets, but the design can be scaled to anything, really, that we deem as a unique, mutually exclusive resource that we want to create and hold long-term, beyond the lifecycle of a single component task.

In terms of this PR, we've done the following:

- introduced `ResourceRegistry` and the various directly related types (`ResourceSpecification`, `ResourceLease`, `AcquireError`, etc etc)
- created a new `BuildContext` type that holds both `ComponentContext` and `ResourceRegistry`, and replaces the `ComponentContext` we provide to `build` methods in component builder traits to expose the ability to acquire resources from the registry when building a component
- exposed a new API endpoint, `/resources/status`, to query the state of the resource registry in terms of all registered resources, which process owns them, and so on

The vastly majority of changed files are simply related to the `build` method signature change to take `BuildContext`, while the rest are related to adding `ResourceRegistry` itself. This PR does not yet migrate anything to actually use it, and that will happen in a follow-up PR.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

New and existing tests.

## References

DADP-2

### DIFF
--- a/lib/saluki-core/src/runtime/state/resources/mod.rs
+++ b/lib/saluki-core/src/runtime/state/resources/mod.rs
@@ -1,0 +1,656 @@
+//! External resource management.
+
+use std::{
+    any::{Any, TypeId},
+    fmt, mem,
+    ops::{Deref, DerefMut},
+    sync::{Arc, Mutex},
+};
+
+use async_trait::async_trait;
+use saluki_common::collections::FastHashMap;
+use saluki_error::GenericError;
+use serde::Serialize;
+use snafu::Snafu;
+use stringtheory::MetaString;
+use tracing::{debug, warn};
+
+use crate::{runtime::process::Id as ProcessId, support::SubsystemIdentifier};
+
+mod api;
+pub use self::api::{ResourceRegistryAPIHandler, ResourceRegistryState};
+
+mod worker;
+pub use self::worker::ResourceRegistryWorker;
+
+#[cfg(test)]
+mod tests;
+
+/// The kind of an external resource.
+///
+/// Deliberately a closed set: the registry coordinates a known, bounded collection of scarce things, so introducing a
+/// kind is a design decision that belongs here rather than in a downstream crate. This names the kind only; it implies
+/// no dependency on the types that implement it.
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ResourceKind {
+    /// A bound network socket.
+    Socket,
+
+    /// A synthetic kind for exercising cross-kind behavior in tests.
+    #[cfg(test)]
+    Test,
+}
+
+impl ResourceKind {
+    /// Returns the string representation of this kind.
+    pub const fn as_str(&self) -> &'static str {
+        match self {
+            Self::Socket => "socket",
+            #[cfg(test)]
+            Self::Test => "test",
+        }
+    }
+}
+
+impl fmt::Display for ResourceKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.as_str())
+    }
+}
+
+/// A specification naming one resource, and the blueprint for creating it.
+///
+/// A specification describes *what* to create without creating it, so it can be built and compared long before the
+/// underlying resource exists. It also names exactly one resource type, which is what keeps two callers from acquiring
+/// the same resource as two different types: there is no way to express the mismatch.
+#[async_trait]
+pub trait ResourceSpecification: Clone + fmt::Debug + Send + Sync + 'static {
+    /// Type of the resource this specification creates.
+    type Resource: Send + 'static;
+
+    /// Kind of the resource this specification names.
+    const KIND: ResourceKind;
+
+    /// Returns the key for the resource this specification names.
+    ///
+    /// Keys must be unique within their kind: two specifications of the same kind that yield the same key are
+    /// understood to name the same underlying scarce thing and will conflict with each other. Keys of different kinds
+    /// never collide, so a key only has to distinguish a resource from its siblings.
+    ///
+    /// Only the parts of a specification that identify the underlying thing belong in the key. Settings that do not
+    /// change *which* resource is named should be left out, so that two specifications differing only in their
+    /// settings are correctly recognized as naming the same resource.
+    fn key(&self) -> MetaString;
+
+    /// Creates the resource.
+    ///
+    /// A resource that is only meaningful as a set of underlying handles -- several sockets bound to one address with
+    /// `SO_REUSEPORT`, say -- holds all of them itself. Creating them together in a single call is what makes them
+    /// atomic: if any one fails, this returns an error and nothing is registered.
+    ///
+    /// # Errors
+    ///
+    /// If the resource can't be created, an error is returned and nothing is registered.
+    async fn create(&self) -> Result<Self::Resource, GenericError>;
+
+    /// Prepares a returning resource for its next holder.
+    ///
+    /// Called when a lease is dropped, before the resource becomes available again. Implement this only for a resource
+    /// that accumulates state over the course of a single lease and must start clean for the next one -- a listener
+    /// tracking how many of its pre-bound sockets it has handed out, for example. Everything the resource is *for*,
+    /// such as the sockets themselves, must survive: the point of the registry is that it outlives its holders.
+    ///
+    /// Defaults to doing nothing, which is right for a resource that carries no per-lease state.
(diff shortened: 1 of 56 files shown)


tobz:
* **#2446** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2446?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2446?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/saluki-core/src/runtime/state/resources/worker.rs
```diff
@@ -0,0 +1,59 @@
...
+
+            // Register our API routes before we actually start running.
+            dataspace.assert(resource_routes, "resource-registry-api");
+
+            // Expose our diagnostic artifact via the diagnostics control surface.
+            let diagnostics = DiagnosticsEmitter::from_dataspace(
+                SubsystemIdentifier::from_segments(["resource-registry"]),
+                dataspace,
+            );
+            diagnostics.register_collector("resources.json", move || resource_registry.snapshot_json());
```
thieman: Do these diagnostics and the API route automatically get hooked into anything (flares?) just by being registered? Or is that a separate step somewhere

tobz: Yep, that's the purpose of this worker: it registers `/resources/status` on the unprivileged API, and it emits this diagnostic bundle which will automatically land a file named `resources.json` in any flare we generate.

lib/saluki-core/src/runtime/state/resources/mod.rs
```diff
@@ -0,0 +1,555 @@
...
+    },
+
+    /// The key is registered, but to a different type of resource.
+    #[snafu(display(
+        "resource '{}' is registered as kind '{}', but was requested as kind '{}'",
+        key,
+        existing_kind,
+        requested_kind
+    ))]
+    KindMismatch {
```
thieman: If Kind needs to potentially be aligned between multiple callers, should it be an enum?

lib/saluki-core/src/runtime/state/resources/mod.rs
```diff
@@ -0,0 +1,555 @@
...
+/// In many cases, a data plane process will have to interact with the outside world by way of exposing network
+/// endpoints, or exposing files, and so on... referred to here as "resources." These resources are unique, or are
+/// conceptually meant to be unique: there should be no other processes trying to take ownership of them, and only one
+/// part of the code in the data plane process should own them.
+///
+/// A [`ResourceRegistry`] owns those resources on behalf of the whole process and lends them out. A task never owns a
+/// resource, but instead holds a [`ResourceLease`]. When the lease drops -- including when the process holding it dies
+/// -- the resource returns to the registry intact and still live, ready for the next acquirer. Since the registry
+/// outlives the components that use its resources, a component can be torn down and rebuilt without the underlying
+/// resource ever being released back to the operating system.
```
thieman: It seems like we'd also want a way to permanently release resources depending on the types of changes we expect to make at runtime. If I start out binding a UDP port and then disable it in configuration which is applied at runtime, I wouldn't expect the process to keep holding that port forever

lib/saluki-core/src/runtime/state/resources/mod.rs
tobz: So this is an admittedly sharper edge than I'd like.

What's happening here is that we need a unique key by which to store a given resource. That key is provided by the "resource specification", so like, for a socket, it's gonna be something like `udp://localhost:8125`.

However, the design of `Resource` and `ResourceSpecification` is such that multiple different types (aka "kind") can use the same resource specification if they want to... so if you try to acquire a resource of type V where the resource spec gives us key A, and we go and see that there's already a registered resource for key A, but it's attached to resource of type T... that's a kind mismatch.

I'm still mulling over if we can solve this cleanly, since it mostly originates from how we expose the resource specification type as an associated type constant on the `Resource` trait... which we explicitly did to support the ability to use a single resource specification type for multiple resources (specifically, `Listener` vs `ConnectionOrientedListener`).

lib/saluki-core/src/runtime/state/resources/mod.rs
tobz: Ah, yeah, this documentation section is stronger than it should be: we do have the ability to explicitly release a resource and remove it from the registry entirely, and we need it precisely for the use case you've identified.

This section is just trying to drive home the point that resources can't be unwittingly released back to the OS except when done so intentionally.

I'll reword.

tobz: Reworded in bb91ce1b.

lib/saluki-core/src/runtime/state/resources/mod.rs
```diff
@@ -0,0 +1,494 @@
...
+                            spec_desc,
+                            state: EntryState::Creating(lease_info.clone()),
+                            acquisitions: 0,
+                        },
+                    );
+                }
+            }
+        }
+
+        let created = R::create(&spec).await;
```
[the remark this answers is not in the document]
thieman: Is this just missing the claim_guard entirely or is it saying something about its implementation isn't working?

lib/saluki-core/src/runtime/state/resources/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+            .entries
+            .iter()
+            .map(|(key, entry)| ResourceStatus {
+                key: key.to_string(),
+                kind: entry.kind,
+                spec: entry.spec_desc.clone(),
+                state: entry.state.as_str(),
+                owner: entry.state.holder().map(|info| info.owner.to_string()),
+                acquisition_process_id: entry.state.holder().map(|info| info.acquisition_process_id.as_usize()),
+                acquisitions: entry.acquisitions,
```
thieman: Would this mapping make more sense as a method on the RegistryState?

lib/saluki-core/src/runtime/state/resources/mod.rs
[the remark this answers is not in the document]
tobz: This comment is to prior to adding `CreationGuard`, so it's now resolved.

thieman (approved):

lib/saluki-core/src/runtime/state/resources/mod.rs
tobz: Yeah, that's reasonable.

lib/saluki-core/src/runtime/state/resources/mod.rs
tobz: So I actually went back and tweaked the design so now it's one-to-one, which is to say that a resource specification explicitly maps to a single resource type (as in `ListenAddress` only ever being used to create a `Listener`, etc) _and_ introduced a high-level enum for resource "kind" (as in, socket vs file vs ...) which are used together to provide a less wishy-washy determination of uniqueness, and make it harder to unwittingly overlap two unrelated resource types.

We do still have a runtime "type mismatch" check, but it's more constrained now overall and, so, hopefully, a little simpler to understand.

PR #2460, opened 2026-08-21
fix(cgroups): tolerate paths that vanish mid-traversal

jszwedko:
## Summary

Fixes [DADP-98](https://datadoghq.atlassian.net/browse/DADP-98) / #499. Walking the cgroups hierarchy is inherently racy — cgroups are removed as the workloads attached to them exit, and we can't hold the hierarchy still while we traverse it — but the traversal treated any I/O error as fatal. The resulting `ERROR | Failed to visit cgroups hierarchy` log had no path and no indication of where in the walk it failed, and worse, the truncated result was handed back to the metadata collector as if it were complete. This makes the traversal tolerate paths that disappear underneath it, and makes an incomplete result distinguishable from an exhaustive one so callers stop drawing conclusions from what they didn't see.

## Problem

`get_child_cgroups()` returns a `Vec<Cgroup>` and propagates the first error out of `visit_subdirectories` with `?`, so a single `ENOENT` truncates the list. `SynchronousCgroupsManager::poll` then diffs that list against the cgroups it already knows about and deletes anything missing:

```mermaid
sequenceDiagram
    participant P as poll (every 2s)
    participant T as visit_subdirectories
    participant S as active_cgroups
    P->>T: walk hierarchy
    T-->>T: read a/ ✓, read b/ ✓
    Note over T: container c exits,<br/>c/ is removed
    T-->>P: Err(ENOENT) — d/, e/ never read
    P->>P: ERROR "Failed to visit cgroups hierarchy."
    P->>S: saw {a, b} → remove aliases for {c, d, e}
    Note over S: d and e are still running, so origin<br/>enrichment breaks for them until the next poll
```

Only `c` actually went away. `d` and `e` lose their inode → container ID alias, which is what client-based origin detection in DogStatsD resolves against, until the next successful poll re-adds them.

For comparison, the core Agent's `pkg/util/cgroups` aborts the walk on `ENOENT` the same way (it only special-cases `fs.ErrPermission`), but `Reader.RefreshCgroups` assigns `r.cgroups` *after* the error check, so a failed refresh keeps serving the last good snapshot rather than invalidating it. ADP had no equivalent protection.

## What changed

- `visit_subdirectories` skips individual paths it can't read, logging the specific path, and counts them. Only a failure to read the hierarchy root itself is still fatal, since that points at a misconfigured `cgroupfs` path rather than a transient condition.
- Each skip is classified by whether it could have hidden a cgroup that **still exists**, which is the only thing that makes "absent" unsafe to read as "removed":

  | Skip reason | Can it hide a live, already-tracked cgroup? |
  |---|---|
  | `NotFound` | No — the directory is gone, so its descendants are gone. Deleting their aliases is correct, just a poll early. |
  | `PermissionDenied` | No — a subtree we could never read never contributed entries, so there's nothing to spuriously delete. This matches the Agent, which skips unreadable directories and still uses the walk. |
  | Anything else (`EIO`, `ENOTDIR`, …) | Yes — the subtree exists and we simply failed to read it this time. |

  Only the last category marks the traversal incomplete. Treating permission errors as incomplete would be worse than the bug being fixed: one permanently unreadable directory would suppress the removal pass forever, and tracked cgroups would accumulate for the life of the process.

  The residual trade-off: a directory that was readable and *becomes* unreadable mid-run is misclassified as safe, and its aliases get dropped. That needs a runtime permissions change on cgroupfs, and the alternative is the unbounded growth above.

  One exception to that classification: the **root**. Treating a permission error as non-obscuring is sound for a subtree, where we never reported anything from it in the first place, and unsound for the root, where it means we saw nothing at all. A root we can't list is therefore an error, not a skip, so `get_child_cgroups` reports it as incomplete. Note that this is reachable even after a successful `stat`: listing a directory needs read permission, while stat'ing it only needs to traverse its parent.

  Worth flagging that this makes us deliberately stricter than the Agent here. `filepath.WalkDir` hands a root `ReadDir` failure to the same `fs.ErrPermission` branch, which returns `filepath.SkipDir`; `walkDir` converts that to `nil`, so `parseCgroups` returns an empty map with no error and [`RefreshCgroups`](https://github.com/DataDog/datadog-agent/blob/8b4e9e14f054dbadfce67705875a1fd3b581d059/pkg/util/cgroups/reader.go#L237-L250) overwrites `r.cgroups` with it. The Agent gets away with it because its reader is a query cache — consumers just get "not found" and skip — whereas ADP's collector turns absence into `remove_alias` operations against shared state.
- `get_child_cgroups` returns a `TraversalResult` carrying the cgroups plus whether the walk was exhaustive. `visit_subdirectories` creates and fills it, and its callback returns the cgroup it identified rather than capturing a `Vec`, so the result owns the cgroups alongside the skip counts.
- The collector runs its removal pass only when the traversal would have seen a cgroup if it still existed. Holding a stale alias for one more poll interval is the cheaper mistake than dropping a live one.
- Three counters, all prefixed `cgroups_metadata_collector_`: `traversals_complete_total` and `traversals_incomplete_total` account for every traversal (so they can be summed for a total or ratioed to see how often traversals are cut short), and `traversal_paths_skipped_total` counts the individual paths lost. This makes the rate of these races visible without relying on log volume.

Also picked up the previously-unused `HierarchyReader::root_path()` in place of the inline match in `get_child_cgroups`.

## Handled separately

Three other divergences from the Agent turned up while comparing. All are unrelated to this race, so they got their own PRs rather than being folded in here:

- **#2461** (merged) — `extract_container_id` applied its `.mount` / `crio-conmon-` exclusions to the regex *match* (a bare hex string) rather than the folder name, so neither filter could ever fire. Also adds the `libpod-conmon-` exclusion the Agent carries.
- **#2462** (merged) — ADP accepted any inode; the Agent treats inodes ≤ 2 as unknown.
- **#2464** (open) — ADP matched only the leaf directory name; the Agent's `ContainerFilter` falls back to walking path segments leaf → root.

This branch has been merged up with `main` to pick up #2461 and #2462.

## Test plan

- [x] New unit tests in `lib/saluki-env/src/workload/helpers/cgroups.rs` cover: a full walk, a missing root, a non-directory root, a directory removed mid-traversal, an unreadable directory, and `get_child_cgroups` reporting complete / incomplete / missing-root traversals.
- [x] The mid-traversal removal test reproduces the race deterministically rather than by timing: every subdirectory is visited and pushed onto the stack before any is read back, so removing an already-visited one guarantees the later read fails with `ENOENT`.
- [x] `cargo test -p saluki-env` and `cargo clippy --all-targets -p saluki-env -- -D warnings` run in a Linux container, since this module is `#[cfg(target_os = "linux")]`. Also run as a non-root user so the permission-denied tests actually assert instead of taking their root early-out.
- [x] `cargo check --workspace --tests` on Linux.
- [ ] Worth a look in staging to confirm the `ERROR` log is gone and the new counters show the expected low rate.

[DADP-98]: https://datadoghq.atlassian.net/browse/DADP-98?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/saluki-env/src/workload/helpers/cgroups.rs
```diff
@@ -233,21 +238,55 @@ impl CgroupsReader {
...
+            complete: outcome.obscured == 0,
+            skipped: outcome.skipped,
         }
-
-        cgroups
     }
 }
 
+/// The result of traversing the cgroups hierarchy.
+pub struct CgroupsTraversal {
```
tobz: It seems like we should collapse this into `TraversalResult`, and have it hold/create the `Vec<Cgroup>` for us, and pass around `&mut TraversalResult`... rather than having two structs.

tobz (changes requested):

lib/saluki-env/src/workload/helpers/cgroups.rs
jszwedko: 👍 This does seem cleaner. I pushed up a refactoring to use one type that also holds the cgroups.

tobz (approved):
LGTM although I think this just needs a rebase to incorporate the other cgroups-related changes you made that got merged in.

Issue #1033, opened 2025-12-01
Improve installation/configuration root detection.


tobz:
## Context

Currently, we depend exclusively on the configuration path passed to the ADP binary, along with environment variables, for configuring ADP. This includes the initial discovery of where the Core Agent (control plane) is running and how to connect to it.

While this is _fine_ from a configuration standpoint, it depends on statically knowing where the binary is installed, in terms of potentially locating other relevant binaries/files -- IPC certificate, SGC (Datadog Secrets Backend), etc -- which breaks under the model of ADP running via Datadog Installer/Fleet Automation. DI/FA install the Agent (and thus ADP) in a version-specific directory alongside other versions, which means it's no longer enough to simply know that we look in path `/var/foo` when on Linux, but `/opt/foo` when on macOS, and so on.

We should add better support for detecting where ADP is running from, locating the installation root if possible, and using that as a primary factor in determining where the Agent/ADP are installed for extrapolating to other files, then falling back on platform-specific logic.

jszwedko:
I tested this out in a VM with the datadog-installer. agent-data-plane seems to start fine since its boot-time dependencies (e.g. auth_token and ipc_cert.pem) are in `/etc/datadog-agent`, which continues to be the location of configuration even when installed via the installer (as opposed to the versioned directories like `/opt/datadog-packages/datadog-agent/7.82.2-1/run`). Other configuration options like `run_path` are resolved by the Core Agent and sent to ADP (e.g. `/opt/datadog-packages/datadog-agent/7.82.2-1/run`).

PR #2463, opened 2026-08-21
feat(commons): support the DD_COMMON_ROOT filesystem layout on Linux

jszwedko:
## Summary

The Core Agent supports being installed under a single writable root — `DD_COMMON_ROOT` collapses `/etc/datadog-agent` and `/var/log/datadog` into `{root}/etc` and `{root}/logs`, so the Agent can run where the root filesystem is read-only. ADP hardcoded the FHS locations, which means that under that layout it would fail to find its configuration file, IPC auth token, and IPC certificate, and would try to log to a directory that does not exist. The layout is not shipped yet, but ADP should not be the thing that blocks it, so this teaches ADP to resolve the same paths the same way.

The important constraint is *where* the resolution happens. Paths that ADP receives over the configuration stream (`run_path` and friends) have already been resolved by the Core Agent against its own root, and must not be transformed again. Only ADP's own local defaults — the ones it needs during bootstrap, before it has registered as a remote agent and can receive anything — are affected.

```mermaid
flowchart LR
    subgraph bootstrap["Bootstrap (local, before registration)"]
        E["DD_COMMON_ROOT"] --> P["PlatformSettings"]
        P --> C["datadog.yaml"]
        P --> A["auth_token"]
        P --> I["ipc_cert.pem"]
        P --> L["agent-data-plane.log"]
    end
    subgraph connected["Connected (configstream)"]
        CA["Core Agent"] -->|already resolved| R["run_path, ..."]
    end
    C --> CA
    A --> CA
    I --> CA
```

Semantics match the Core Agent's `pkg/util/defaultpaths` exactly: unset leaves the FHS layout in place, set-but-empty selects `/opt/datadog-agent`, and any other value is used as the root. Linux and AIX only — macOS already uses a common-root-shaped layout, and Windows resolves its root from the registry.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Added unit tests in `linux_impl.rs` covering the three environment variable states (unset, set-but-empty, set) and the config/log directory derivation with and without a common root. The path resolution is factored into pure helpers so the tests do not mutate process-global environment state.

Not yet exercised end to end — the layout has no shipping consumer to test against.

## References

- Part of DataDog/saluki#1033
- Core Agent implementation: `pkg/util/defaultpaths/common_nix.go`, `pkg/util/defaultpaths/path_nix.go` (`commonRootOrPath`)

### DIFF
--- a/lib/datadog-agent/commons/src/platform/linux_impl.rs
+++ b/lib/datadog-agent/commons/src/platform/linux_impl.rs
@@ -1,20 +1,144 @@
-use std::path::Path;
+//! Linux (and AIX) platform settings.
+//!
+//! Two filesystem layouts are supported, matching the Core Agent:
+//!
+//! - the FHS layout (the default), where configuration lives in `/etc/datadog-agent` and logs live in
+//!   `/var/log/datadog`, and
+//! - the "common root" layout, selected by the `DD_COMMON_ROOT` environment variable, where every directory the Agent
+//!   owns is collapsed under a single root: `{root}/etc`, `{root}/logs`, `{root}/run`, and so on.
+//!
+//! The common root layout exists to support installs where the root filesystem is read-only and the Agent cannot
+//! scatter writes across `/etc` and `/var`. The Core Agent implements it in `pkg/util/defaultpaths`
+//! (`commonRootOrPath`), and reads the environment variable during package initialization rather than from its own
+//! configuration, because the layout determines where the configuration file itself is found.
+//!
+//! We resolve it the same way, and for the same reason: the paths derived here (the configuration file, the IPC auth
+//! token, and the IPC certificate) are all needed during bootstrap, before ADP has registered with the Core Agent and
+//! can receive its resolved configuration. Paths that arrive over the configuration stream have already been resolved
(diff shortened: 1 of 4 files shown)


tobz (changes requested):
Just to tighten things up here, can we hide all of the existing consts (like `DATADOG_AGENT_CONF_DIR`) in `linux_impl.rs` and the other impl source files?

We don't actually use them outside of the files they're declared in, and in the case of Linux, we especially don't want anyone to use them since we don't want the common root logic to be circumvented.

jszwedko:
> Just to tighten things up here, can we hide all of the existing consts (like `DATADOG_AGENT_CONF_DIR`) in `linux_impl.rs` and the other impl source files?
> 
> We don't actually use them outside of the files they're declared in, and in the case of Linux, we especially don't want anyone to use them since we don't want the common root logic to be circumvented.

👍  I like that idea.

jszwedko:
Made the constants private in [440d962](https://github.com/DataDog/saluki/pull/2463/commits/440d96232e1d389c51046460413ab84f04be6887)

tobz (approved):

Issue #2013, opened 2026-07-06
Support gRPC keepalive ServerParameters config for the OTLP source


lucastemb:
Sub-issue of #2005.

## Background

The OpenTelemetry Collector's OTLP receiver exposes a `keepalive.server_parameters` config
group under `configgrpc` ([docs](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/config.md#configgrpc-keepaliveserverconfig)),
mapping to Go's [`grpc/keepalive.ServerParameters`](https://pkg.go.dev/google.golang.org/grpc/keepalive#ServerParameters).
ADP's OTLP gRPC server (`OtlpServerBuilder::build()` in
`lib/saluki-components/src/common/otlp/mod.rs`) does not currently configure any connection-age
or ping/pong keepalive behavior, so long-lived or idle connections are never proactively
recycled — a gap relative to Agent/Collector parity, and a potential issue for load-balanced
deployments that rely on periodic connection cycling.

## Fields to support

| Field | Type | Default | Description |
|---|---|---|---|
| `max_connection_idle` | duration | infinity | Duration after which an idle connection (no outstanding RPCs) is closed via `GOAWAY`. |
| `max_connection_age` | duration | infinity | Maximum lifetime of a connection before the server sends `GOAWAY` (±10% jitter applied by gRPC to avoid connection storms). |
| `max_connection_age_grace` | duration | infinity | Additional grace period after `max_connection_age` before the connection is forcibly closed. |
| `time` | duration | `2h` | Idle-connection ping interval — if no activity is seen after this duration, the server pings the client to check liveness. Values below `1s` are clamped up to `1s` by gRPC. |
| `timeout` | duration | `20s` | How long the server waits for ping activity before closing an unresponsive connection. |

## Work

- [ ] Add a `KeepaliveServerParametersConfig` struct to `lib/saluki-components/src/common/otlp/config.rs`, following the `GrpcConfig` pattern (`Deserialize` + `Facet`, `#[serde(default = "default_...")]` per field, explicit `impl Default` matching gRPC's own defaults above).
- [ ] Nest it under `GrpcConfig` (e.g. `keepalive.server_parameters`), matching the Collector's config shape (pairs with the `enforcement_policy` group from the sibling issue).
[...]

PR #2465, opened 2026-08-21
feat(otlp): added support for select keepalive config keys for gRPC

### LINKED ISSUE #2012: Support gRPC keepalive EnforcementPolicy config for the OTLP source
Sub-issue of #2005.

## Background

The OpenTelemetry Collector's OTLP receiver exposes a `keepalive.enforcement_policy` config
group under `configgrpc` ([docs](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/config.md#configgrpc-keepaliveserverconfig)),
which maps directly to Go's [`grpc/keepalive.EnforcementPolicy`](https://pkg.go.dev/google.golang.org/grpc/keepalive#EnforcementPolicy).
ADP's OTLP gRPC server currently exposes no equivalent — `OtlpServerBuilder::build()`
(`lib/saluki-components/src/common/otlp/mod.rs`) calls `Server::builder()` without any
enforcement policy configured, so ADP silently falls back to whatever `tonic`/`h2` defaults
apply, which does not match Agent/Collector parity.

## Fields to support

| Field | Type | Default | Description |
|---|---|---|---|
| `min_time` | duration | `5m` | Minimum time a client should wait between keepalive pings before the server considers it a policy violation. |
| `permit_without_stream` | bool | `false` | If `true`, allows the server to accept keepalive pings even when there are no active RPC streams. If `false`, a ping without an active stream causes the server to send `GOAWAY` and close the connection. |

## Work

- [ ] Add a `KeepaliveEnforcementPolicyConfig` struct to `lib/saluki-components/src/common/otlp/config.rs`, following the existing pattern used by `GrpcConfig` (`Deserialize` + `Facet`, per-field `#[serde(default = "default_...")]`, explicit `impl Default`).
- [ ] Nest it under `GrpcConfig` (e.g. `keepalive.enforcement_policy`), matching the Collector's config shape.
- [ ] Wire the parsed values into the tonic `Server` builder in `OtlpServerBuilder::build()` via [`tonic::transport::Server::tcp_keepalive`](https://docs.rs/tonic/latest/tonic/transport/server/struct.Server.html) / the underlying `h2`/`tonic` equivalent of `grpc.KeepaliveEnforcementPolicy` (confirm exact tonic API — enforcement policy may require using `tonic`'s `http2_keepalive_*` combined with the `h2` server builder's connection-abuse settings; verify tonic exposes a direct enforcement-policy knob or if this needs a custom interceptor).
- [ ] Document the new config fields per `AGENTS.md` documentation guidelines (defaults, behavior, who should tune them).
- [ ] Add unit/config tests validating defaults and deserialization.

## Notes

Verify during implementation whether `tonic` exposes `EnforcementPolicy` directly (as of recent
tonic/h2 versions this may not be a first-class option) — if not, this issue may need to document
that limitation or track it as a follow-up rather than a full implementation.

----

Oops: I shouldn't have replicated the research on this one, will be more careful with the rest:

## Background

The OTLP gRPC receiver accepts HTTP/2 PING frames from connected clients. The Agent applies a
keepalive enforcement policy that limits how frequently clients may send these frames. ADP does not
accept this configuration or enforce an equivalent policy, so clients can send PING frames without
the Agent's connection-level protection.

This may require a change upstream to Tonic see below.

### Agent behavior

The Agent sets the default value of
`otlp_config.receiver.protocols.grpc.keepalive.enforcement_policy.min_time` to `5m` in
`pkg/config/setup/otlp_settings.go`. It then passes the complete `otlp_config.receiver` subtree to
the OpenTelemetry Collector OTLP receiver through `comp/otelcol/otlp/config.go` and
`comp/otelcol/otlp/map_provider.go`.

The Collector's `configgrpc.KeepaliveEnforcementPolicy` contains two fields:

```go
type KeepaliveEnforcementPolicy struct {
    MinTime             time.Duration `mapstructure:"min_time,omitempty"`
    PermitWithoutStream bool          `mapstructure:"permit_without_stream,omitempty"`
}
```

`min_time` defaults to five minutes. `permit_without_stream` is not explicitly defaulted by the
Agent and therefore uses the gRPC default of `false`. Users may still set either field in the nested
receiver configuration because the Agent passes that configuration to the Collector.

When the Collector constructs its gRPC server, it converts these values to
`keepalive.EnforcementPolicy` and passes them to `grpc.KeepaliveEnforcementPolicy`. grpc-go counts
PING frames that violate the policy. After more than two violations, it sends an HTTP/2 GOAWAY with
`ENHANCE_YOUR_CALM` and debug data `too_many_pings`, then closes the connection.

Relevant reference code:

- `pkg/config/setup/otlp_settings.go:36-43` sets the Agent's gRPC receiver defaults.
- `comp/otelcol/otlp/config.go:41-48` loads the receiver configuration.
- `comp/otelcol/otlp/map_provider.go:97-101` adds it to the Collector receiver map.
- `go.opentelemetry.io/collector/config/configgrpc.KeepaliveEnforcementPolicy` defines the fields.
- `configgrpc.ServerConfig.getGrpcServerOptions` installs the gRPC enforcement policy.
- `google.golang.org/grpc/internal/transport.(*http2Server).handlePing` counts violations and sends
  GOAWAY.

### agent-data-plane behavior

ADP's gRPC receiver configuration is `GrpcConfig` in
`lib/saluki-components/src/common/otlp/config.rs`. It currently contains only `endpoint`,
`transport`, and `max_recv_msg_size_mib`. The OTLP source reads this configuration in
`lib/saluki-components/src/sources/otlp/mod.rs` and passes the endpoint and message-size limit to
`OtlpServerBuilder`.

`OtlpServerBuilder` constructs the gRPC server with an unconfigured `tonic::transport::Server` in
`lib/saluki-components/src/common/otlp/mod.rs:166-180`:

```rust
let grpc_server = Server::builder()
    .add_service(grpc_metrics_server)
    .add_service(grpc_logs_server)
    .add_service(grpc_traces_server);
```

No keepalive enforcement settings are applied. Tonic 0.14 exposes server-initiated HTTP/2 keepalive
interval and timeout settings, but those settings do not restrict client-initiated PING frames.
Tonic and its h2 dependency do not expose grpc-go's minimum client PING interval,
permitted-idle-PING, ping-strike, or `too_many_pings` GOAWAY behavior through their server builders.

Add nested configuration for `keepalive.enforcement_policy.min_time` and
`keepalive.enforcement_policy.permit_without_stream`, preserving the Agent's defaults. Before
selecting an implementation, determine whether the enforcement can be added to Tonic or h2 upstream
or requires a lower-level connection implementation in ADP. Do not map these fields to Tonic's
`http2_keepalive_interval` or `http2_keepalive_timeout`; those control server-generated PING frames
and have different behavior.

### Requirements tracking

- OM_CONF_318

lucastemb:
## Summary

Added relevant `keepalive` config options for gRPC. 

`keepalive.enforcement_policy` is not currently supported because the library we utilize, `tonic/h2`, has no way to configure these keys externally. 

`permit_without_stream` and `min_time` make it such that `PINGs` and `PONGs` can be sent without an active stream and will close the connection if a client sends `PINGs` more than once in a 5 minute interval. This is to protect against malicious actors or spam in the general case, but since the clients (in our case), are trusted, we can safely defer this. 

Additionally, `max-connection-idle` is not currently supported because `tonic/h2` has no equivalent; however, as it stands, the default is set to `INF` anyways. 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests.

## References

- Closes: #2012 #2013


docs/agent-data-plane/configuration/configuration.md
```diff
@@ -219,6 +222,74 @@ ADP's architecture does not have the same batching mechanism and is unaffected b
...
+defaults propagated unchanged through the OpenTelemetry Collector and into the Datadog Agent
+config schema; they are not a considered decision for ADP's deployment model.
+
+ADP's clients are trusted OpenTelemetry SDKs on the same host, not adversarial actors, so the
+threat model this policy defends against does not apply. Additionally, ADP uses tonic/h2 for its
+HTTP/2 stack, which has no enforcement policy hook: h2 auto-acks every inbound PING
+unconditionally, with no observation point to rate-limit or send GOAWAY.
+
+Setting `min_time` has no effect in ADP. The default (`5m`) is accepted silently; a non-default
+value emits a warning.
```
webern, line 273: Ok, this seems fair. I agree that the customer's own client should not be expected to overping keepalives and it seems extremely unlikely that a customer would depend on the behavior described here.

docs/agent-data-plane/configuration/configuration.md
```diff
@@ -219,6 +222,74 @@ ADP's architecture does not have the same batching mechanism and is unaffected b
...
+These defaults propagated unchanged through the OpenTelemetry Collector and into the Datadog
+Agent config schema; they are not a considered decision for ADP's deployment model.
+
+ADP's clients are trusted OpenTelemetry SDKs on the same host that benefit from warm idle
+connections, so the threat model this policy defends against does not apply. Additionally, ADP
+uses tonic/h2 for its HTTP/2 stack, which has no enforcement policy hook: h2 auto-acks every
+inbound PING regardless of active streams, with no observation point to reject no-stream pings.
+
+Setting `permit_without_stream` has no effect in ADP. ADP always permits PINGs without an active
+stream.
```
webern, line 293: Ok, seems fine and seems like it would be hard with our libraries.

docs/agent-data-plane/configuration/configuration.md
```diff
@@ -219,6 +222,74 @@ ADP's architecture does not have the same batching mechanism and is unaffected b
...
+per-connection hooks (active stream count, GOAWAY trigger) through tonic's server loop. The raw
+`h2::server::Connection` has the primitives (`has_streams`, `graceful_shutdown`), but tonic
+owns the connection lifecycle and does not hand the `h2` connection back to caller code.
+
+The grpc-go default is infinity, so the Core Agent at its default does not reclaim warm-idle
+connections either: it relies on `time`/`timeout` for dead-connection detection, which ADP
+supports. At the default configuration, ADP's behavior matches the Core Agent on this axis.
+
+Setting `max_connection_idle` has no effect in ADP. The default (infinity) is accepted silently;
+a non-default value emits a warning.
```
webern, line 321: I guess it feels a *little bit* more possible that someone could rely on this behavior... maybe? Or, that there could be *some* risk of resource leakage if a client is constantly crashing and leaving streams open or something (is that possible?)

However, the warning does seem sufficient for this one as well.

lib/datadog-agent/config/src/generated/datadog_configuration.rs
```diff
@@ -2073,6 +2133,18 @@ mod duration_defaults {
...
+    }
+    pub(super) fn otlp_config_receiver_protocols_grpc_keepalive_server_parameters_max_connection_age_grace() -> std::time::Duration {
+        std::time::Duration::from_nanos(0)
+    }
+    pub(super) fn otlp_config_receiver_protocols_grpc_keepalive_server_parameters_time() -> std::time::Duration {
+        std::time::Duration::from_nanos(0)
+    }
+    pub(super) fn otlp_config_receiver_protocols_grpc_keepalive_server_parameters_timeout() -> std::time::Duration {
+        std::time::Duration::from_nanos(0)
+    }
```
webern, line 2215: How are these defaults being determined?

If we are relying on the schema, then it makes sense why they would all be zero, because the schema is not carrying these for us. I think there is something that needs to be looked into here.

What are the real defaults? If they are not zero, how are we going to populate the real defaults?

lib/saluki-components/src/sources/otlp/mod.rs
```diff
@@ -72,6 +72,25 @@ fn cors_configuration(cors: &domains::otlp::Cors) -> CorsConfiguration {
...
+///
+/// Unset `time` and `timeout` are replaced with their defaults. The age and grace fields default
+/// to no limit, which matches tonic's `None` default.
+fn resolve_grpc_keepalive(keepalive: &Option<domains::otlp::KeepaliveServerParameters>) -> GrpcKeepalive {
+    let params = keepalive.as_ref();
+    GrpcKeepalive {
+        http2_keepalive_interval: Some(
+            params
+                .and_then(|p| p.time)
+                .unwrap_or(domains::otlp::DEFAULT_GRPC_KEEPALIVE_TIME),
```
webern: Unfortunate to see this happening here instead of in the config layer.

lib/saluki-components/src/sources/otlp/mod.rs
```diff
@@ -72,6 +72,25 @@ fn cors_configuration(cors: &domains::otlp::Cors) -> CorsConfiguration {
...
+        ),
+        http2_keepalive_timeout: params.and_then(|p| p.timeout),
```
webern: What's happening here? Are we just converting from one option type to another?

lib/datadog-agent/config/src/generated/datadog_configuration.rs
lucastemb: The defaults are determined by the upstream Go gRPC library that the embedded OTEL Collector uses. If I'm not mistaken, everything is `0` by default to represent infinite or unset. 

I do believe for `timeout` and `time` they should be non-zero (`20s` and `2h` respectively). I'm thinking this could be done at the translation layer similar to how it's done for `max_recv_msg_size_mib`

docs/agent-data-plane/configuration/configuration.md
lucastemb: _Hmm_. Yeah, worst case scenario, a bit of CPU usage is wasted keeping the stream open in the most common case. I think that the `time` and `timeout` keys provide protection against the case you're describing since if it doesn't receive a `PONG` from the client, it will terminate the connection.

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -78,6 +78,43 @@ impl<'a> DatadogTranslator<'a> {
     fn record_error(&mut self, error: TranslateError) {
         self.errors.push(error);
     }
+
+    /// Returns a mutable reference to the keepalive server parameters, initializing the struct if
+    /// keepalive has not yet been configured.
+    fn keepalive_server_parameters_mut(&mut self) -> &mut KeepaliveServerParameters {
```
webern: Seems a little weird. Unnecessary abstraction?

lib/agent-data-plane-config/src/domains/otlp.rs
webern, line 2:
```suggestion
//! OTLP domain: the OTLP receiver (gRPC/HTTP transports, logs/metrics activation), the OTLP proxy
//! gating, and OTLP context sizing. OTLP trace handling lives in the `traces` domain.
//!
//! Unlike normal Datadog schema settings, some of these have been imported from the OTLP schema,
//! which does not define defaults. This file, therefore, must provide certain OTLP defaults.
//! Other than this exceptional case, schema-derived fields should obtain their defaults from the
//! Datadog schema.
```

lib/saluki-components/src/sources/otlp/mod.rs
```diff
@@ -74,6 +74,22 @@ fn cors_configuration(cors: &domains::otlp::Cors) -> CorsConfiguration {
     }
 }
 
+/// Converts keepalive server parameters into tonic-compatible settings.
+///
+/// Defaults are already resolved by the configuration layer; this function only maps the typed
+/// values into the tonic-facing struct.
+fn resolve_grpc_keepalive(keepalive: &Option<domains::otlp::KeepaliveServerParameters>) -> GrpcKeepalive {
```
webern: Could be done, perhaps with `the_optional.map(|params| /* construct the GrpcKeepalive */).unwrap_or_default()`.

Depending on how often it comes up, I guess.

webern (approved):

PR #2466, opened 2026-08-22
fix(agent-data-plane): align DogStatsD /compat metrics with go_expvar semantics

jszwedko:
## Summary

We're directing operators away from the Core Agent's `go_expvar` check and toward scraping ADP's `/compat/metrics` with the OpenMetrics check (DADP-160). For that swap to be lossless, the numbers on `/compat` have to mean what the DogStatsD expvars mean. Auditing ADP against every variable the Core Agent actually publishes turned up three places where they diverge — one metric that over-counts, one that doesn't exist, and one that attributes UDP failures to UDS. This is the DogStatsD slice of that audit; aggregator and serializer follow separately.

```mermaid
flowchart LR
    subgraph src["DogStatsD source telemetry"]
        P["component_packets_received_total<br/>{listener_type, state}"]
        E["component_errors_total<br/>{listener_type, error_type}"]
    end

    P -->|"state:ok only<br/>(was: ok + error)"| PK["dogstatsd_udp_packets<br/>dogstatsd_uds_packets"]
    P -->|state:error| PE["dogstatsd_*_packet_reading_errors"]
    E -->|"error_type:unterminated<br/>(new)"| U["dogstatsd_unterminated_metric_errors"]
    E -->|"error_type:origin_detection<br/>+ listener_type:unix/unixgram"| O["dogstatsd_uds_origin_detection_errors"]
    E -->|error_type:framing| X["no expvar equivalent"]

    style U fill:#d4edda
    style X fill:#f8d7da
```

**Packet counts included failed reads.** The rules matched on `listener_type` alone, and since the renderer drops the `state` tag it merged `state:ok` and `state:error` into a single value. The Core Agent only increments `udpPackets` / `udsPackets` on a successful read.

**`dogstatsd/UnterminatedMetricErrors` had no equivalent.** ADP implements the same behavior — `NewlineFramer` rejects payloads without a trailing newline when `dogstatsd_eol_required` is set — but reported it as a generic `component_errors_total{error_type="framing"}`, mixed in with length-delimited and oversized-frame failures. Missing delimiters now get a distinct `FramingError::MissingDelimiter` variant and a dedicated `error_type:unterminated` counter, which *replaces* the framing increment rather than adding to it so `error_type` rollups stay correct.

**Origin detection errors conflated transports.** This was the only DogStatsD error counter registered without a `listener_type` tag, so UDP origin-detection failures were being remapped into `dogstatsd_uds_origin_detection_errors`, which the Core Agent only ever increments for UDS.

`dogstatsd/PacketsLastSecond` is intentionally excluded — ADP already documents `dogstatsd_stats_*` as unsupported.

> [!NOTE]
> Adding `listener_type` to the origin-detection counter changes the tag shape of that series on `/metrics`, not just `/compat/metrics`.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

New unit tests in `bin/agent-data-plane/src/state/metrics/mod.rs` cover each fix: that packet counts reflect only `state:ok` while reading errors reflect only `state:error`, that unterminated payloads are counted separately from other framing errors, and that UDP origin-detection failures no longer reach the UDS metric. `render_compat_telemetry` was updated to feed realistic `state`-tagged series.

On the source side, `decoder_continues_connectionless_but_stops_connected_on_framing_error` now asserts the split between `error_type:unterminated` (missing newline) and `error_type:framing` (oversized frame), and the `NewlineFramer` / DogStatsD framer tests assert the new error variant.

Still to do before this leaves draft: end-to-end verification against a running ADP, diffing `/compat/metrics` against a Core Agent `/debug/vars` dump under the same DogStatsD load.

## References

- DADP-177 — go_expvar telemetry audit
- DADP-179 — extend `/compat/metrics` to remaining go_expvars subsystems
- DADP-160 — decision to use OpenMetrics over `go_expvar`


tobz:
Going to close this as we're moving to deprecate use of the expvar endpoint for the Agent as a whole.

PR #2472, opened 2026-08-24
chore(io): add `Stream` impl for conn-oriented listener to use for `GrpcServer`

tobz:
## Summary

This PR introduces the ability to convert `ConnectionOrientedListener` into a `Stream`-compatible type for the purpose of easily being used in `GrpcServer`.

Currently, `GrpcServer` contains a decent chunk of duplicate logic to support running on either a TCP or UDS endpoint, which is suboptimal because we have a bunch of perfectly sufficient abstracted network I/O utilities at hand, like `ConnectionOrientedListener`. However, these types were initially designed with an `async fn`-first mindset, so they aren't amenable to the lower-level approaches like having a `Stream` of accepted connections, which is what `tonic` and `hyper` prefer operating over.

This PR introduces the ability to convert `ConnectionOrientedListener` into a new type, `ConnectionOrientedListenerStream`, which does just that. Callers convert their `ConnectionOrientedListener` to `ConnectionOrientedListenerStream` by calling a new method, `into_stream`, and everything just works.

We've changed a number of bits here to make it work well, namely:

- refactored `GrpcServer` to use it (duh)
- implemented `Connected` for `Connection` so gRPC requests can check connection info if necessary
- reworked how we acquire peer credentials for UDS sockets in stream mode

Prior to this PR, we used the "out-of-band" mode for both datagram and stream mode which meant we got the "connection address" with every single datagram/message we received. However, for UDS sockets in stream mode, we can actually just query the peer credentials once, after we accept the connection, and then store it alongside the socket and avoid the overhead of grabbing it on every receive operation. This helps us out a little in existing cases like the DogStatsD source, but was necessary to do for this work because there's no way to even change the underlying code to read it out-of-band anyways.

A nice side effect of this is that it actually makes the code able to get peer credentials on more operating systems, at least for UDS sockets in stream mode. Not that we'd be doing anything with the credentials on those platforms since we don't build for them, but... nice to get to remove code _and_ get more benefit at the same time.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

New and existing unit tests, integration tests, etc.

## References

DADP-2


tobz:
* **#2472** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2472?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2472?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

aqian01 (approved):

webern (approved):
You are good at async code.

tobz:
Closing this since it's superseded by all of the `GrpcServer` deduplication work.

Issue #2293, opened 2026-08-11
Migrate DogStatsD pipeline to typed config.


webern:
Migrate this to typed config:

```rust
    if dp.dogstatsd_enabled() {
        let dsd_control_surface =
            add_dsd_pipeline_to_blueprint(&mut blueprint, &config_system.raw_map(), config_system, env_provider)
                .await?;
        control_surfaces.attach_dogstatsd(dsd_control_surface);
    }
```

Edit: **Migration inventory and order**

| Order | Item | Location | `Live<T>` | Notes / gotchas |
|-------|------|----------|-----------|-----------------|
| 1 | `AggregateConfiguration` | `lib/saluki-components/src/transforms/aggregate` | no | Smallest static surface (3 keys, no watches). `domains.dogstatsd.aggregation` is already populated. Keep the `dogstatsd_expiry_seconds` alias behavior. |
| 2 | `DogStatsDMapperConfiguration` | `lib/saluki-components/src/transforms/dogstatsd_mapper` | no | Static, but `dogstatsd_mapper_profiles` accepts a JSON string or a sequence (`PickFirst`). Settles the profiles/mappings shape in the model. |
| 3 | `DogStatsDDebugLogConfiguration` | `lib/saluki-components/src/destinations/dsd_debug_log` | **yes** | First `Live<T>`, smallest reactive surface: one watched key (`dogstatsd_metrics_stats_enable`). Default log path still comes from `PlatformSettings`. |
| 4 | `TagFilterlistConfiguration` | `bin/agent-data-plane/src/components/tag_filterlist` | **yes** | Watches `metric_tag_filterlist`. Local `MetricTagFilterEntry` duplicates the model type; collapse to one. `data_plane.dogstatsd.aggregator_tag_filter_cache_capacity` is modeled under `aggregation`, not `tag_filterlist`. |
| 5 | `DogStatsDPostAggregateFilterConfiguration` | `bin/agent-data-plane/src/components/dogstatsd_post_aggregate_filter` | **yes** | First four-watch conversion, and the simpler of the two. Defines the shared filterlist/blocklist precedence projection. `HistogramSuffixes` validation stays in the component. |
| 6 | `DogStatsDPrefixFilterConfiguration` | `bin/agent-data-plane/src/components/dogstatsd_prefix_filter` | **yes** | Reuses the projection from 5, then adds namespace prefixing, the default blocklist, and the `statsd_metric_namespace_blacklist` alias. |
[...]

PR #2473, opened 2026-08-24
chore(config): declare object-array element types

webern:
## Human Summary

TODO

## AI Summary

Declare source types for `dogstatsd_mapper_profiles` and
`metric_tag_filterlist` in `datadog-agent-config::array_objects`.

The core schema declares both arrays without `items` schemas, so generated
config exposes `Vec<serde_json::Value>`. The translator now deserializes each
value into a named source type instead of declaring local shims.

Field defaults and action handling are unchanged. Tests cover required and
defaulted source fields.

## Change Type

## How did you test this PR?

- `make fmt`
- `make check-docs`
- `make check-clippy`
- `cargo nextest run -p datadog-agent-config -p agent-data-plane-config-system -p agent-data-plane-config`
- `cargo nextest run -p agent-data-plane -E 'test(mapper) + test(tag_filter) + test(filterlist)'`

## References

- Progresses: #2293

### DIFF
--- a/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
+++ b/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
@@ -7,7 +7,9 @@
 //! to its `SalukiConfiguration` destination.
 //!
 //! Most keys assign a single field directly. The endpoint keys (`api_key`, `dd_url`, `site`,
-//! `additional_endpoints`) are copied into the model without selecting a primary endpoint here.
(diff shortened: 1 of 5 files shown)


webern:
Closing: not needed after further thought. The `serde_json::Value` this addressed is at the translator boundary, which is the layer whose job is converting Datadog source shapes into model types. Relocating the element type there does not remove the raw JSON, so the change bought a rename rather than an architectural fix.

PR #2475, opened 2026-08-24
feat(dogstatsd): add replay support on AIX, macOS, and Windows

thieman:
## Summary

Implements DADP-180 Phase 1: `adp dogstatsd replay` is enabled on Linux, macOS, Windows, and AIX.

## Jira

[DADP-180](https://datadoghq.atlassian.net/browse/DADP-180)

## Key changes

- Selects replay transport by platform:
  - **Linux:** existing Unix datagram transport with `SCM_CREDENTIALS` captured-origin replay.
  - **macOS and AIX:** plain Unix datagrams through `dogstatsd_socket`.
  - **Windows:** newline-framed named-pipe writes through `dogstatsd_pipe_name`.
- Removes redundant replay-specific platform gates: the project supports only Unix and Windows targets, so remaining gates select only actual implementation differences.
- Makes replay-session API client support available on all supported transports.
- Warns non-Linux replay users that captured PID-based origin tags cannot be preserved; client-supplied/live origin enrichment can still apply.
- Captures completed Windows named-pipe frames so the native capture-to-replay workflow is available on both platforms.
- Adds a shared Linux/macOS Unix-datagram case and a Windows named-pipe capture/replay integration case.
- Verifies replayed metric names by exact match against the DogStatsD stats Metric column.

## Test plan

- [x] `make check-all`
- [x] `make test-all`
- [x] Shared Unix case on Linux: `make build-datadog-agent-image` and `cargo run --release --bin panoramic -- run -d test/integration/cases --runtime linux -t dogstatsd-replay-unix --no-tui -p 1`
- [x] Shared Unix case on macOS: `make test-integration-macos-ci CASE=dogstatsd-replay-unix`
- [x] `make build-windows-cross WINDOWS_CROSS_CARGO_ARGS='--package agent-data-plane --tests'`
- [x] Native Linux container: `cargo check -p agent-data-plane --tests`
- [x] AIX: native `ci/tooling/build-adp-aix.sh`, focused replay-sender unit test, and Tokio `UnixDatagram` smoke test on `soaix499`.
- [ ] Windows integration case reruns in `test-integration-windows-amd64` CI.

## Review

Independent review approved both the initial cross-platform implementation and the AIX/cfg-simplification follow-ups.

[DADP-180]: https://datadoghq.atlassian.net/browse/DADP-180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -1,40 +1,42 @@
 use std::cmp::Reverse;
 use std::collections::{HashMap, HashSet};
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "linux", target_os = "macos", target_os = "windows"))]
```
thieman: Most of these cfg gates amount to excluding AIX. Look into how easy it would be to support AIX here (if it already works in Unix so maybe it's fine?) and answer in this thread. Do not code anything yet.

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -282,14 +284,20 @@ async fn handle_dogstatsd_capture(
...
 async fn handle_dogstatsd_replay(
     api_client: &mut DataPlaneAPIClient, config: &GenericConfiguration, cmd: ReplayCommand,
 ) -> Result<(), GenericError> {
-    let socket_path = dogstatsd_socket_path(config)?;
+    let target = dogstatsd_replay_target(config)?;
 
     info!("Preparing DogStatsD replay from '{}'.", cmd.replay_file_path.display());
 
+    #[cfg(any(target_os = "macos", target_os = "windows"))]
```
thieman: This cfg gate should instead be "not Linux"

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] I checked this. It looks like a small implementation change, but not zero-risk until we exercise it on an AIX host.

AIX is already a first-class build target here (`build-adp-aix`, the AIX Rust SDK profile, and AIX process-memory support). The replay path can use the same Unix-datagram transport as macOS: Tokio exposes `UnixDatagram` for `cfg(unix)`, and AIX has AF_UNIX support. Linux would remain the only credential-stamping branch; AIX would send plain datagrams and receive the existing non-Linux origin-enrichment warning.

The relevant upstream caveat is that Mio uses its poll(2) fallback on AIX rather than epoll/kqueue. That should still support the socket, but it makes a native AIX smoke test mandatory rather than something we should infer from macOS.

The code simplification would be worthwhile: replace the repeated Linux/macOS/Windows target lists with `cfg(unix)` for the Unix transport/session path, keep Windows as the named-pipe branch, and reserve the unsupported branch for `not(any(unix, windows))`. I estimate this as a small follow-up (roughly the replay CLI gates, replay API-client gates, and one AIX/plain-Unix transport test), plus native AIX build and runtime validation. I have not changed code yet.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] Changed in f37470bb1d. The warning and plain Unix-datagram send path now use `not(target_os = "linux")`; AIX is enabled through the Unix transport alongside macOS.

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -1,40 +1,42 @@
 use std::cmp::Reverse;
 use std::collections::{HashMap, HashSet};
-#[cfg(target_os = "linux")]
+#[cfg(any(unix, windows))]
```
thieman: This seems like it would encompass all of our supported platforms, if so then it's unnecessary and should be deleted. Same for any gates like this. Try to minimize how many config gates are needed generally.

thieman: [GPT-5.6 Terra (OpenAI)] Simplified in e217302619. Since every supported target is Unix or Windows, I removed the redundant `any(unix, windows)` gates from this module and the replay API client. The remaining gates only select real implementation differences: Unix versus Windows transport, Linux credential stamping, and the non-Linux plain-datagram warning.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] Native AIX validation is now complete on `soaix499`: `ci/tooling/build-adp-aix.sh` built the current source successfully, and `cargo test -p agent-data-plane --bin agent-data-plane non_linux_unix_replay_sender_delivers_the_capture_payload -- --nocapture` passed. I also ran a standalone native Tokio `UnixDatagram` sender/receiver smoke test successfully. The full package test command is blocked by an unrelated IBM Cargo 1.92 integration-test environment-variable issue (`CARGO_BIN_EXE_agent-data-plane`), so I used the binary test target containing the replay sender regression instead.

test/integration/cases/dogstatsd-replay-unix/config.yaml
```diff
@@ -0,0 +1,49 @@
+type: integration
```
thieman, line 1: Do we need the linux and mac tests to be separate or can they be combined? They look very similar?

thieman: [GPT-5.6 Terra (OpenAI)] Combined in 97d8242d21. The replacement `dogstatsd-replay-unix` case has `runtimes: [linux, mac]`, uses transport-neutral names/paths, and removes the duplicate macOS case. I ran the shared case successfully on both the Linux container runtime and the native macOS host-process runtime.

webern (approved):
This may be a naive question, but is the whole point of this that we want to run it through the actual listening socket/pipe/whatever? I guess I'm just wondering why go through all the trouble of doing the platform-specific things instead of just, I don't know, telling the process where the payload file is and letting it inject the data itself "behind" the socket/pipe/whatever?

bin/correctness/panoramic/src/actions/dogstatsd_replay.rs
```diff
@@ -0,0 +1,217 @@
...
+            let (stats_result, replay_result) = tokio::join!(stats, replay);
+
+            match replay_result {
+                Ok(_) => match stats_result {
+                    Ok(stats_output) => {
+                        let missing_metrics: Vec<&str> = self
+                            .expected_metrics
+                            .iter()
+                            .map(String::as_str)
+                            .filter(|metric| !stats_output.contains(metric))
```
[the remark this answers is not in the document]
thieman: [GPT-5.6 Terra (OpenAI)] Valid finding. Fixed in 57ca80a006: the replay action now parses the Metric column from the formatted stats table and compares complete trimmed names, rather than searching the whole output for substrings. Added a regression test that rejects `replay.request` when only `replay.requests` is present. The full Panoramic unit suite and the Linux replay integration case pass with the exact-match parser.

thieman:
@webern What I think happened here is that the original Linux implementation mandated UDS for two reasons:

1. We wanted to prefer a less-lossy transport, which implies disallowing UDP.
2. The Linux replay can both capture and replay origin detection information, which it does by setting options on the Unix socket. I think that could also have been accomplished using the StatsD protocol extensions, but maybe those weren't around yet?

Reason 2 doesn't apply on these other platforms since we don't support origin detection, but Reason 1 does still apply. So I kept it opinionated to match the Linux implementation and chose named pipes on Windows (which are closer to TCP than UDP).

PR #2477, opened 2026-08-25
chore(config): migrate DogStatsDPrefixFilterConfiguration to typed config

webern:
## Human Summary

Migrate `DogStatsDPrefixFilterConfiguration` to typed configuration.

## AI Summary

Move the DogStatsD prefix filter to typed configuration.

- Build namespace prefixing from the typed DogStatsD domain.
- Use the shared live metric-filter view for filter updates.
- Preserve schema defaults and the `statsd_metric_namespace_blocklist` alias.
- Remove raw-map deserialization and key watches.

## Change Type
- [x] Non-functional (chore, refactoring, docs)

## How did you test this PR?

- `make build-schema-overlay`
- `make fmt`
- `make check-docs`
- `make check-clippy`
- Targeted `cargo nextest` runs for the prefix filter, typed translation, and live precedence
- `make test-correctness-case CASE=dsd-mapper-blocklist`

## References

- Progresses: #2293

PR #2474, opened 2026-08-24
chore(config): migrate DogStatsDPostAggregateFilterConfiguration to typed configuration

webern:
## Human Summary

Migrates `DogStatsDPostAggregateFilterConfiguration` to typed configuration. A follow-up audit suggested documenting `metric_filterlist` and `statsd_metric_blocklist` precedence behavior, so that was added as well as a test.

## AI Summary

Move the DogStatsD post-aggregate metric filter to typed configuration.

- Resolve current filterlist and legacy blocklist precedence at the translation boundary.
- Pass the transform a narrow live `MetricFilter` view.
- Rebuild histogram filtering after accepted typed configuration updates.
- Keep histogram suffix validation in the post-aggregate component.
- Retain the raw prefix-filter consumer in the configuration inventory until its migration.

## Change Type
- [x] Non-functional (chore, refactoring, docs)

## How did you test this PR?

- `make fmt`
- `make check-docs`
- `make check-clippy`
- `cargo check -p agent-data-plane-config -p agent-data-plane-config-system -p agent-data-plane --tests`
- Targeted `cargo nextest` runs for the post-aggregate filter and translator precedence tests
- Configuration smoke test
- `make test-correctness-case CASE=dsd-mapper-blocklist`

## References

- Progresses: #2293

PR #2471, opened 2026-08-24
chore(config): migrate TagFilterlistConfiguration to typed config

webern:
## Summary

- Replace `TagFilterlistConfiguration::from_configuration` with `new(...)`.
- Pass a narrow `Live<Vec<MetricTagFilterEntry>>` to the component instead of the full DogStatsD domain.
- Pass value allowlists and context cache capacity as static component inputs.
- Warn on an unrecognized `action` and fall back to `exclude` without recording a translation error.
- This matches the core Agent recovery behavior and avoids rejecting a configuration the Agent runs with.
- Keep an entry without `tags` with an empty tag list, matching the Agent's `mapstructure` decode.
- Reject and drop an entry without `metric_name` while retaining valid neighboring entries.
- Reject negative `data_plane.dogstatsd.aggregator_tag_filter_cache_capacity` values during translation instead of clamping them.

## Change Type

- [x] Non-functional (chore, refactoring, docs)

## How did you test this PR?

- `make fmt`
- `make check-docs`
- `make check-clippy`
- Targeted `cargo nextest` runs for the configuration system, `datadog-agent-config`, and the tag-filterlist / DogStatsD pipeline tests

## References

- Progresses: #2293

PR #2470, opened 2026-08-24
chore(config): migrate dogstatsd debug log to typed config

webern:
## Human Summary

Migrates `DogStatsDDebugLogConfiguration` to typed config. This includes a dynamic field served by `Live<T>`.  `run_starts_and_stops_logging_with_metrics_stats_setting` surprisingly exercises this.

## Summary

Migrate the DogStatsD debug log destination from `GenericConfiguration` to typed configuration.

- Build static file settings from the typed DogStatsD configuration.
- Keep `dogstatsd_metrics_stats_enable` dynamic through a narrow `Live<bool>` view.
- Move log-size parsing and roll-count validation to the configuration translator.
- Update the configuration inventory and generated registry for typed ownership.
- Preserve debug-log writing, rotation, and runtime enable/disable behavior.

## Change Type
- [x] Non-functional (chore, refactoring, docs)

## How did you test this PR?

- `make build-schema-overlay`
- `make fmt`
- `make check-all`
- `make test-all`
- Targeted `cargo nextest` runs for the debug-log and configuration-system tests

## References

- Progresses: #2293

PR #2478, opened 2026-08-25
chore(config): model run_path in typed configuration

webern:
## Human Summary

The state directory, `run_path`, was being read directly by key in a few places. Now we read this from typed configuration preserving existing `unset` behavior by using an `Option<PathBuf>` as the type. The coding agent noticed that the schema has a placeholder, `${run_path}` as a literal default string, so we check for that literal, warn, and discard it in the highly unlikely case that an Agent bug passes it to us.

## AI Summary

- Model `run_path` in typed configuration.
- Use it for retry storage, DogStatsD captures, and context dumps.

## Change Type

## How did you test this PR?

- `make fmt`
- `make check-clippy`
- `make check-docs`
- Targeted `cargo nextest` tests for retry, DogStatsD capture, and context dumps

## References

- Progresses: #2293

### DIFF
--- a/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
+++ b/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
@@ -1104,6 +1104,23 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
         self.config.shared.endpoints.proxy.no_proxy = value;
     }
 
+    fn consume_run_path(&mut self, value: String) {
+        // In the vendored schema, run_path is defaulted to the placeholder ${run_path}. Though it
+        // seems highly unlikely this could slip through to ADP, we check for it, warn and treat
+        // the value as unset.
+        //
+        // Note that for this config, we do not care whether provenance is explicit or default. We
+        // just want to know whether we have a value or not.
+        if value == "${run_path}" {
+            warn!("`run_path` contains the unresolved schema placeholder '${{run_path}}'. Treating it as unset.");
+            self.config.shared.run_path = None;
+        } else if value.is_empty() {
+            self.config.shared.run_path = None;
+        } else {
+            self.config.shared.run_path = Some(PathBuf::from(value))
+        }
+    }
+
     fn consume_serializer_compressor_kind(&mut self, value: String) {
         self.config.shared.endpoints.compression.compressor_kind = value;
     }
--- a/lib/agent-data-plane-config/src/shared.rs
+++ b/lib/agent-data-plane-config/src/shared.rs
@@ -37,6 +37,13 @@ pub struct SharedConfiguration {
     /// Verbosity of the internal telemetry emitted about the runtime itself. (not in Datadog Agent
     /// config schema)
     pub metrics_level: String,
+
+    /// Base directory for runtime-state files.
+    ///
+    /// ADP uses this to derive default locations for forwarder retry data, DogStatsD captures, and
+    /// DogStatsD context dumps. Defaults to unset when configuration does not provide a concrete
+    /// `run_path`.
+    pub run_path: Option<PathBuf>,
 }
 
 /// Inputs used to derive deployment-wide static tags.
--- a/bin/agent-data-plane/src/cli/run.rs
+++ b/bin/agent-data-plane/src/cli/run.rs
@@ -30,11 +30,10 @@ use saluki_components::{
     relays::otlp::OtlpRelayConfiguration,
     sources::{ChecksIPCConfiguration, DogStatsDConfiguration, OtlpConfiguration},
     transforms::{
-        aggregate_context_snapshot_channel, AggregateConfiguration, AggregateContextSnapshotHandle,
-        ApmStatsTransformConfiguration, AutoscalingFailoverGatewayConfiguration, ChainedConfiguration,
-        DogStatsDMapperConfiguration, DogStatsDMapperProfile, DogStatsDMetricMapping, HistogramConfiguration,
-        HostEnrichmentConfiguration, MrfMetricsGatewayConfiguration, TraceObfuscationConfiguration,
-        TraceSamplerConfiguration,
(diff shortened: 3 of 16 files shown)


webern:
* **#2499** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2499?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2498** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2498?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2497** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2497?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2496** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2496?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2488** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2488?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2487** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2487?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2479** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2479?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2478** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2478?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2478?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2477** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2477?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2474** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2474?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2471** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2471?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2470** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2470?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2469** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2469?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -1080,6 +1080,23 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
...
     }
 
+    fn consume_run_path(&mut self, value: String) {
+        // In the vendored schema, run_path is defaulted to the placeholder ${run_path}. Though it
+        // seems highly unlikely this could slip through to ADP, we check for it, warn and treat
+        // the value as unset.
+        //
+        // Note that for this config, we do not care whether provenance is explicit or default. We
+        // just want to know whether we have a value or not.
+        if value == "${run_path}" {
```
thieman, line 1114: Did we let the config team know about this? Also, do we not need some actual default? Probably `/opt/datadog-agent/run` on Linux/Mac/AIX?

Which leads to another question, how do we handle platform-specific defaults and what should that be on Windows?

webern: We have now! https://github.com/DataDog/datadog-agent/issues/55426

Platform/runtime/environment-specific defaults are values that both JSON schema and our own config parsing layer should probably not attempt to resolve. I don't think I would request the JSON schema to specify a Linux default that is sensitive to deployment method, distro, etc. In the issue I just opened, I suggested adding the various runtime defaults to the documentation instead.

How we are handling this now: if a default can only be derived at runtime or based on platform, then the config system considers the default to be `None` and the runtime has to deal with it. In practice, these defaults are unreachable unless we are in standalone mode since the Agent will always resolve and provide them.

I think the guard I added here is unreachable in production unless a bug were introduced into the Agent.

AI answer on what happens when run_path is unset:

> In normal connected operation, the current Agent resolves and supplies run_path. Standalone configurations may legitimately leave it unset, in which case ADP continues running but default-location retry persistence, captures, and context dumps are unavailable.

Me again:

1. We could, potentially, provide some sort of platform-dependent lookup table to provide defaults such as...
   - Linux/macOS/AIX: `/opt/datadog-agent/run`
   - Windows: `<effective ProgramData/ConfigRoot>\run`, normally `C:\ProgramData\Datadog\run`
 
2. Another option would be to use `Tempfile` which works on Windows and Unix/Linux (but not Android, apparently). And maybe this would not be great since a restarted ADP would lose the previous retry queue.

3. The last option would be to fail startup and error out.

However, each of those is a behavioral change and, in general, my config migration PRs attempt to be behaviorally transparent (although in this one I decided to change the behavior that could lead to a path like `/\$\{run_path\}/transactions_to_retry`).

So the final `TL;DR`: other than guarding against the silly path, behavior is unchanged and ADP will run **without** retry-on-disk or dump/capture support _if it is in standalone and `run_path` is not provided_.

lib/agent-data-plane-config/src/shared.rs
```diff
@@ -37,6 +37,13 @@ pub struct SharedConfiguration {
     /// Verbosity of the internal telemetry emitted about the runtime itself. (not in Datadog Agent
     /// config schema)
     pub metrics_level: String,
+
+    /// Base directory for runtime-state files.
+    ///
+    /// ADP uses this to derive default locations for forwarder retry data, DogStatsD captures, and
+    /// DogStatsD context dumps. Defaults to unset when configuration does not provide a concrete
+    /// `run_path`.
```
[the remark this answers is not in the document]
webern: I think this is clear enough without introducing the documentation into a second place.

thieman (approved):

PR #2482, opened 2026-08-25
feat(otlp): added support for self-telemetry usage beacon metrics

### LINKED ISSUE #2092: Emit OTLP self-telemetry usage beacon metrics

## Background

The Agent emits self-telemetry metrics for OTLP ingest. ADP currently emits no equivalent metrics.
Add the usage beacon series and attach the resolved host to each series so OTLP ingest reports the
same usage information in ADP.

### Agent behavior

The Agent emits these metrics from the OTLP serializer exporter on each `ConsumeMetrics` call, after
translating the incoming batch and before sending the accumulated series:
`comp/otelcol/otlp/components/exporter/serializerexporter/exporter.go:201-217`.

`addTelemetryMetric` appends `datadog.agent.otlp.metrics` as a gauge with value `1`, no tags, and
`SourceTypeName: "System"` (`consumer.go:188-196`). `addRuntimeTelemetryMetric` appends one
`datadog.agent.otlp.runtime_metrics` gauge with value `1` for every language detected in the
translated batch. Each series has one tag, `language:<language>` (`consumer.go:223-233`).

`addGatewayUsage` appends `datadog.otel.gateway` when `gatewayUsage.Gauge()` reports that the
feature is enabled. Its value is the returned gauge value. It also always appends
`datadog.otel.gateway.configured`, whose value comes from `gatewayUsage.EnvVarValue()`. Both are
gauges with no tags (`consumer.go:236-261`).

All four series use the same `hostname` value as their `Host` field. The exporter calls
`e.hostGetter(ctx)` directly once per `ConsumeMetrics` call (`exporter.go:207-210`), returns the
error if host resolution fails, and passes the successful value to all three telemetry helpers
(`exporter.go:212-214`). This is separate from registering the same function as the translator's
fallback source provider for individual datapoints (`exporter.go:97-113`); beacon host resolution
does not use a translated datapoint's dimensions.

The Agent's serializer consumer is created for each exporter call and sends the accumulated series
directly through the injected serializer (`consumer.go:280-306`). The Agent-injected factory path
uses this consumer (`factory.go:72-75`), while the OSS collector path uses a different consumer that
does not emit the two `datadog.agent.otlp.*` series (`consumer_collector.go:33-65`).

### agent-data-plane behavior

ADP has no equivalent of these four series. A scoped search of
`lib/saluki-components/src/sources/otlp/`, `lib/saluki-components/src/common/otlp/`, and
`lib/saluki-components/src/forwarders/otlp/` found no `datadog.agent.otlp`,
`datadog.otel.gateway`, usage-beacon, or self-telemetry series.

OTLP requests enter `SourceHandler::handle_metrics`, which splits an
`ExportMetricsServiceRequest` into individual `ResourceMetrics` messages and sends them through an
`mpsc` channel (`lib/saluki-components/src/sources/otlp/mod.rs:351-362`). The translator then
processes each `ResourceMetrics` in `OtlpMetricsTranslator::translate_metrics`
(`lib/saluki-components/src/sources/otlp/metrics/translator.rs:412-447`), and the converter flushes
the `BufferedDispatcher` on its timer (`lib/saluki-components/src/sources/otlp/mod.rs:390-499`).
There is currently no hook that emits synthetic Datadog metric events for an incoming request or
converter flush.

Translated metrics are constructed with `Metric::from_parts` and wrapped in `Event::Metric` by
`record_metric_event` (`lib/saluki-components/src/sources/otlp/metrics/translator.rs:630-655`).
New beacon metrics can use that same pattern with `MetricValues::gauge`, a resolved context, and
`Event::Metric`. ADP's existing OTLP counters are Prometheus-style internal counters registered by
`saluki_metrics::MetricsBuilder` (`lib/saluki-components/src/common/otlp/mod.rs:56-107`); they are
not converted into Datadog metric events.

The translator has a `default_hostname` field (`translator.rs:74-81`). The value is passed into the
translator when the OTLP source is built (`lib/saluki-components/src/sources/otlp/mod.rs:76,
141-143, 221, 244, 259-276`) and is stored at construction (`translator.rs:391-406`). It is used as
the fallback host for datapoints lacking a resource-derived host (`translator.rs:419-427`). It is
not re-resolved for each request. Beacon emission should use this same host value, while keeping
beacon creation independent of per-datapoint host selection.

The runtime metric mapping table contains language prefixes, including Go, .NET, and JVM, but it is
currently unused (`lib/saluki-components/src/sources/otlp/metrics/runtime_metrics.rs:23-31`). ADP
has no existing per-batch language accumulation. If the runtime beacon is implemented, collect the
languages represented by runtime metrics while translating the request or flush and emit one
`language:<language>` series for each distinct language.

To preserve the Agent's request-level behavior, the implementation may need to carry a beacon
message or request-scoped state through the channel into `run_converter`; `handle_metrics` currently
has no dispatcher handle. A simpler implementation at `translate_metrics` would operate per
`ResourceMetrics`, but must avoid emitting duplicate beacons for every individual datapoint and
should document the chosen boundary.

### Requirements tracking

- OM_RESZ_063
- OM_RESZ_275

## Definition of done

- [ ] Demonstrate the missing usage beacon series and host tags with a red correctness test.
- [ ] Emit `datadog.agent.otlp.metrics` as a gauge with value `1` and the resolved host.
- [ ] Emit one `datadog.agent.otlp.runtime_metrics` gauge with the resolved host and
      `language:<language>` for each distinct runtime-metric language in the batch.
- [ ] Emit `datadog.otel.gateway` when gateway usage is enabled and always emit
      `datadog.otel.gateway.configured`, matching the Agent's values and host behavior where the
      gateway concept applies to ADP.
- [ ] Ensure beacon host resolution uses ADP's configured default hostname and is independent of
      per-datapoint resource-host translation.
- [ ] Add focused unit tests for metric names, types, values, tags, host, language deduplication,
      and the selected request/flush boundary.
- [ ] Demonstrate a green correctness test against the Agent baseline.

## References

- Relates to #2006
- Sub-issue of #2005

lucastemb:
## Summary

This PR aims to add functionality to emit the following self-telemetry beacon metrics `datadog.agent.otlp.runtime_metrics` and `datadog.agent.otlp.metrics`. They report OTLP metrics ingest activity, as well as recognized language runtimes (Go, .NET, Java). 

Inclusion of `datadog.otel.gateway` and `datadog.otel.gateway.configured` are omitted for this PR because we don't _actually **run**_ the Open Telemetry collector; therefore, is no notion of collector deployment mode (gateway vs. agent) for these metrics to report on.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests

## References

- Closes: #2092


lib/saluki-components/src/sources/otlp/mod.rs
```diff
@@ -365,19 +365,29 @@ impl Source for Otlp {
...
     Metrics(OtlpResourceMetrics),
     Logs(OtlpResourceLogs),
     Traces(OtlpResourceSpans),
 }
 
+/// Messages flowing through the converter channel.
+///
+/// Most messages carry an OTLP signal (`Signal`), but control signals like
+/// `MetricsRequestEnd` mark request boundaries for usage beacon emission.
+enum OtlpMessage {
```
webern: I think the design is broken because simultaneous requests can land on the channel interleaved. So the `MetricsRequestEnd` signal could conflate the runtime counts of more than one request.


Options, briefly:

1. Send the whole protobuf request through the channel.
2. Send a self-contained `Vec<ResourceMetrics>` request batch.
3. Translate the complete request in the network handler.
4. Add request IDs and track request state in the converter.

1 seems best to me, and you could do the runtime-language accounting while translating the request's `ResourceMetrics` for efficiency.

2 seems ok, but is nearly equivalent to 1; it only avoids passing the protobuf wrapper through the channel.

3 puts more work in the network handler, and I think the idea with these processing pipelines is to decouple receiving from processing, so 1 and 2 seem better.

4 would also work, I suppose, and would be closest to your current design, but might be a little heavy on the bookkeeping.

> Below is some AI output looking for other options besides `#1` which is what I seeded it with...

---

## Several Design Options

### 1. Send the whole request through the channel — **recommended**

Matt's idea:

```text
handler → Metrics(ExportMetricsServiceRequest) → converter
```

The converter then:

1. creates request-local `HashSet<Language>`;
2. walks and translates every `ResourceMetrics`;
3. records languages while translating;
4. emits customer metrics;
5. emits the generic and language beacons through the same metrics dispatcher.

This puts request-level behavior in a request-level stage. It requires no request IDs or persistent translator state. It probably does not require a new topology component—the existing converter task can be that stage.

Trade-off: channel capacity counts variable-sized whole requests rather than individual resources.

### 2. Send a self-contained request batch

Instead of the protobuf request type:

```rust
OtlpMessage::MetricsRequest(Vec<ResourceMetrics>)
```

Architecturally equivalent to `#1`, but the internal channel avoids depending on the outer OTLP service type. Slightly more explicit about what the converter consumes.

### 3. Translate the complete request in the handler

The handler performs request-local translation and sends resulting `Event::Metric` batches through the channel:

```text
request → Vec<Event> including beacons → channel
```

Clean request semantics, but moves compute-heavy work back onto network-server tasks—the converter exists specifically to avoid this.

### 4. Add request IDs

Send:

```rust
Metrics { request_id, resource }
MetricsRequestEnd { request_id }
```

The converter maintains:

```rust
HashMap<RequestId, RequestState>
```

Supports arbitrary interleaving, but introduces ID generation, cleanup, memory bounds, and abandoned-request handling.

## Clanker's opinion

Matt's design—**`#1` or `#2`**—is the strongest. Request-level knowledge should remain locally scoped to request processing. After that stage discovers the distinct languages, both translated customer metrics and synthetic beacons can enter the same dispatcher exactly as they do today. This keeps the channel contract simple and removes the need for mutable cross-request state.

webern (changes requested):
I think this isn't correct as written. See comment arbitrarily anchored to the new enum.

webern (approved):
Had to rely on a 🤖 to speed things up. Looks good. There are some findings left, but I think we can move forward. 1 seems real. 2 and 3 are nits.

1. Language detection is broader than the Agent's (the only real semantic divergence). Go calls extractLanguageTag inside the runtimeMetricsMappings[md.Name()] hit branch (metrics_translator.go:559-560); this PR checks the prefix map unconditionally on every metric name(translator.rs:542-547). Because the map contains the bare prefix "jvm", any unmapped customer metric named jvm.* will emit a language:jvm beacon where the Agent would not. Over-reports usage.
2. Two vestigial tests from the old design that now assert nothing: emit_usage_beacons_clears_languages_after_emission (there's no state left to clear — it just passes an empty set) and emit_usage_beacons_does_not_increment_metrics_received_counter, whose own comment admits "we can't
    directly read the counter." Delete both.
3. Trivia: now_ns is computed in nanos then divided by 1_000_000_000 — just use .as_secs(). And channel capacity 1024 now counts whole requests rather than individual ResourceMetrics, so worst-case in-flight bytes went up; probably fine given the memory limiter, but worth a thought.

PR #2494, opened 2026-08-26
feat(otlp): support remaining OTLP receiver protocol settings

### LINKED ISSUE #2068: Support remaining OTLP receiver protocol settings in ADP

Sub-issue of #2005.

## Background

The Agent accepts configuration for gRPC stream limits, connection buffers, request metadata, and
HTTP request-body size on its OTLP receiver. ADP currently accepts only the receiver endpoints,
transport, and gRPC message-size limit. Add the supported settings to ADP, and document any setting
that cannot be implemented with the current server libraries.

This is the residual receiver-configuration issue. It does not include Unix transport (#2010),
gRPC keepalive (#2011, #2012, and #2013), TLS (#2014), or HTTP CORS (#2015); those issues own those
configuration groups.

### Agent behavior

The Agent passes the `otlp_config.receiver` subtree to the OpenTelemetry Collector OTLP receiver.
The receiver consumes these six fields:

- `grpc.max_concurrent_streams` defaults to `0` and sets the HTTP/2 maximum concurrent streams per
  connection when nonzero.
- `grpc.read_buffer_size` defaults to `524288` bytes and sets the gRPC server read buffer.
- `grpc.write_buffer_size` defaults to `0` and sets the gRPC server write buffer; zero uses the
  gRPC default.
- `grpc.include_metadata` defaults to `false` and controls propagation of incoming gRPC metadata
  to downstream consumers.
- `http.max_request_body_size` defaults to `0`; the HTTP receiver then applies its 20 MiB default
  limit. A positive value sets the limit in bytes.
- `http.include_metadata` defaults to `false` and controls propagation of incoming HTTP headers as
  request metadata.

The Agent's defaults are defined in `pkg/config/setup/otlp_settings.go:36-48`. The raw receiver
configuration is retained in `comp/otelcol/otlp/collector.go:133-134`, inserted into the collector
configuration in `comp/otelcol/otlp/map_provider.go:96-101`, and unmarshaled by the Collector's
`configgrpc.ServerConfig` and `confighttp.ServerConfig`. The Collector applies the gRPC settings in
`configgrpc/configgrpc.go:502-507,566-575,611` and the HTTP settings in
`confighttp/server.go:191-192,226-227,295`.

### agent-data-plane behavior

ADP's `saluki-components` crate defines `GrpcConfig` and `HttpConfig` in
`lib/saluki-components/src/common/otlp/config.rs:56-91`, but neither struct has these fields. The
receiver is built in `lib/saluki-components/src/common/otlp/mod.rs:157-213`.

Tonic exposes `Server::max_concurrent_streams`, so that setting can be added to `GrpcConfig` and
applied to the `Server::builder()` chain. Tonic does not expose the Agent's read/write socket-buffer
settings; decide whether an equivalent is required, whether an HTTP/2 window setting is acceptable,
or whether the difference should be documented. gRPC metadata is available from each tonic request,
but ADP's `OtlpHandler` currently receives only the encoded body, so metadata propagation requires a
handler/API change.

The HTTP server uses an Axum `Router` in `mod.rs:207-213` and `saluki-io`'s `HttpServer`. Add the
body-size setting to `HttpConfig` and apply it with Axum's `DefaultBodyLimit` layer, preserving the
Agent's 20 MiB behavior when the configured value is zero. HTTP handlers currently extract only
`Bytes`; metadata propagation requires extracting headers and extending the handler API. Note that
Axum's current default body limit is not the Agent's default and must not be left implicit.

### Requirements tracking

- OM_CONF_112
- OM_CONF_238

## Definition of done

- [ ] Add and validate the six receiver fields with the Agent defaults and documented units.
- [ ] Wire `max_concurrent_streams` and `max_request_body_size`; resolve or document the gRPC
  buffer limitation.
- [ ] Define how request metadata is represented and propagate it for both protocols, or document
  the intentional divergence.
- [ ] Add component tests for config defaults, stream/body limits, and metadata propagation or the
  documented unsupported cases.
- [ ] Add a red/green correctness or integration test covering the supported receiver behavior.

lucastemb:
## Summary

Added support for the remaining configuration options `grpc.max_concurrent_streams`  and `http.max_request_body_size` 

`grpc.read_buffer_size`, `grpc.write_buffer_size`, `grpc.include_metadata`, `http.include_metadata` are, for the time being, designated as unsupported due to differences in how metadata can be propagated through through the topology, lack of customer adoption, and differences in buffer behavior. 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests

## References

- Closes: https://github.com/DataDog/saluki/issues/2068


lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -1038,6 +1038,10 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
         }
     }
 
+    fn consume_otlp_config_receiver_protocols_grpc_max_concurrent_streams(&mut self, value: i64) {
+        self.config.domains.otlp.receiver.grpc.max_concurrent_streams = u32::try_from(value.max(0)).unwrap_or(0);
```
webern: > Matt's bot:

This clamps bad input instead of rejecting it. Every other conversion in this file does
`match u32::try_from(value) { Ok(v) => ..., Err(_) => self.record_error(...) }`, so this is
the odd one out.

It also fails in a bad direction. `0` means "no limit" here. So a typo like `4294967296`
turns into `0`, which quietly removes the stream cap instead of erroring.

Please use `try_from` and `record_error` like the rest of the file, so out-of-range values
are reported and the previous value is left alone.

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -1066,6 +1070,10 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
         }
     }
 
+    fn consume_otlp_config_receiver_protocols_http_max_request_body_size(&mut self, value: i64) {
+        self.config.domains.otlp.receiver.http.max_request_body_size = value.max(0) as u64;
```
webern: > Matt's bot:

Same thing here: `value.max(0) as u64` silently turns a negative number into `0`, and `0`
means "use the 20 MiB default". A bad value should be an error, not a silent fallback.
Please use `record_error` like the other conversions in this file.

lib/saluki-components/src/common/otlp/mod.rs
```diff
@@ -706,6 +743,33 @@ mod tests {
         panic!("Unix socket at {} did not appear within 500ms", path.display());
     }
 
+    #[tokio::test]
+    async fn http_body_limit_rejects_oversized_request() {
```
webern, line 924: > Matt's bot:

This test builds its own `Router` and its own `DefaultBodyLimit::max(64)`, so it only proves
that axum enforces a limit. It never calls `OtlpServerBuilder`, so it would still pass if the
new `.layer(DefaultBodyLimit::max(...))` line were deleted.

What I'd like covered instead:

1. `max_request_body_size: 0` really results in a 20 MiB limit.
2. A configured value really results in that limit.

Also, nothing covers `max_concurrent_streams` at all right now.

lib/saluki-io/src/net/server/grpc.rs
```diff
@@ -65,6 +67,22 @@ impl GrpcServer {
...
 
+    /// Sets the HTTP/2 maximum concurrent streams per connection.
+    ///
+    /// A value of `0` means no limit (the tonic default). A positive value sets the
+    /// `SETTINGS_MAX_CONCURRENT_STREAMS` HTTP/2 setting, bounding how many concurrent streams a
+    /// single connection may have.
+    ///
+    /// Defaults to no limit (`None`).
+    pub fn with_max_concurrent_streams(mut self, max_concurrent_streams: u32) -> Self {
+        self.max_concurrent_streams = if max_concurrent_streams == 0 {
```
webern: > Matt's bot:

The "0 means unlimited" rule is now decoded in two places: `OtlpServerBuilder` stores a raw
`0`, and then this method converts `0` into `None`. If someone changes one side later, the
two disagree.

I'd rather the magic `0` be translated once, as early as possible (ideally at the config
boundary), and everything below it take an `Option<u32>` that already means what it says.

> Edit, `Option<NonZeroU32>` would be better if we actually want to do with this is saying.

lib/saluki-io/src/net/server/grpc.rs
```diff
@@ -95,6 +113,13 @@ impl Supervisable for GrpcServer {
     async fn initialize(&self, process_shutdown: ShutdownHandle) -> Result<SupervisorFuture, InitializationError> {
         let routes = self.routes.clone().unwrap_or_default();
         let shutdown_timeout = self.graceful_shutdown_timeout.unwrap_or(Duration::MAX);
+        let max_concurrent_streams = self.max_concurrent_streams;
```
webern: > Matt's bot:

Small one: this local doesn't add anything, since `self.max_concurrent_streams` is used only
in the `match` right below it. You can match on `self.max_concurrent_streams` directly.

webern (commented):
Please excuse the bot comments in my name. Had to hurry. I'll take a closer look with eyes tomorrow. Consider the auto comments I posted here to be suggestions. Sorry for the overly strong wording on some of them... see you tomorrow.

lib/saluki-components/src/common/otlp/mod.rs
```diff
@@ -854,6 +896,95 @@ mod tests {
...
+        });
+
+        tokio::task::yield_now().await;
+
+        // Build with a small body limit (64 bytes) through the builder.
+        let max_body_size: usize = 64;
+        supervisor_handle
+            .scope(
+                OtlpServerConfiguration::new(http_endpoint, grpc_endpoint, 4 * 1024 * 1024)
+                    .with_http_max_request_body_size(max_body_size as u64)
```
webern, line 959: Can we also have a test case where `max_request_body_size=0` and verifying that the resultant limit is `20 MiB`?

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
webern: Resolved

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
webern: Resolved

lib/saluki-io/src/net/server/grpc.rs
webern: Resolved

lib/saluki-io/src/net/server/grpc.rs
webern: Closing as you seem to have rejected the idea of `Option<NonZeroU32>` in favor of a `0` sentinel. Which is sensible and fine and my idea was perhaps weird anyway.

lib/saluki-components/src/common/otlp/mod.rs
```diff
@@ -198,6 +198,13 @@ impl OtlpTlsConfiguration {
     }
 }
 
+/// OTLP server configuration.
```
webern: nit: This appears copied and probably just the default description below is needed.

lib/saluki-components/src/common/otlp/mod.rs
```diff
@@ -242,6 +253,24 @@ impl OtlpServerConfiguration {
         self
     }
 
+    /// Sets the HTTP/2 maximum concurrent streams per connection for the gRPC receiver.
+    ///
+    /// A value of `None` (the default) means no limit. A `Some` value sets the
```
webern: nit: This still describes the reverted `Option<NonZeroU32>` API. It should say that `0` means no limit and positive values set the limit.

webern (approved):
There are a few small items left, but looks good with those fixes.

PR #2501, opened 2026-08-27
chore(config): model secrets settings for the retry gate

webern:
## Human Summary

We use the secrets configs to determine whether or not 403s are retriable. Now we do so from typed config and document the fields correctly in the config inventory. (Previously they had been used to actually resolve secrets, but now ADP defers that responsibility to the Agent and just uses these as a signal as to retriability of intake 403s).

## AI Summary

Move the forwarder retry policy's secrets gate onto typed configuration.

The gate decides whether an intake's `403 Forbidden` is worth retrying: it is,
when secret resolution might replace the rejected API key. It read
`secret_backend_command` and `secret_refresh_on_api_key_failure_interval` from
the raw map on every 403, and both keys were excluded from the inventory.

- Model both keys in a new `shared.secrets` slice, with `Secrets::in_use` naming
  the predicate the gate asks for.
- Give the retry policy a `Live<Secrets>` view, read per response, so turning
  secrets management on or off still changes the gate without rebuilding the
  service.
- Group the forwarder's runtime configuration inputs into one
  `LiveForwarderConfiguration`, which also keeps the argument count in bounds.

Behavior notes:

- A negative refresh interval is clamped to zero rather than recorded as a
  translation error. The Core Agent accepts it, it means the same thing as no
  interval, and startup translation is strict, so rejecting it would stop ADP
  from starting on a configuration the Agent is happy with.
- The gate now follows the typed model, so an update that fails translation
  leaves the last known good value in place instead of taking a raw value the
  typed system rejected.

The raw map stays on the Datadog forwarder for API key validation, which is
migrated separately.

## Change Type

## How did you test this PR?

- `make build-schema-overlay`
- `make fmt`
- `make check-docs`
- `make check-clippy`
- `cargo nextest run -p saluki-components -p agent-data-plane-config -p agent-data-plane-config-system -p datadog-agent-config-testing`

## References

- Progresses: #2293

### DIFF
--- a/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
+++ b/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
@@ -1257,7 +1257,11 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
 
     fn consume_secret_refresh_on_api_key_failure_interval(&mut self, value: i64) {
         // A negative interval is not an error the Core Agent reports, and it means the same thing as
-        // no interval at all, so clamp rather than refuse to start.
+        // no interval at all, so clamp rather than reporting an error.
+        if value < 0 {
+            warn!("`secret_refresh_on_api_key_failure_interval` is negative ({value}). Treating it as 0 (disabled).");
+        }
+
         self.config.shared.secrets.refresh_on_api_key_failure_interval = value.max(0) as u64;
     }
 
--- a/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
+++ b/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
@@ -1251,6 +1251,16 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
         }
     }
 
+    fn consume_secret_backend_command(&mut self, value: String) {
+        self.config.shared.secrets.backend_command = non_empty_trimmed(value);
+    }
+
+    fn consume_secret_refresh_on_api_key_failure_interval(&mut self, value: i64) {
+        // A negative interval is not an error the Core Agent reports, and it means the same thing as
+        // no interval at all, so clamp rather than refuse to start.
+        self.config.shared.secrets.refresh_on_api_key_failure_interval = value.max(0) as u64;
+    }
+
     fn consume_serializer_compressor_kind(&mut self, value: String) {
         self.config.shared.endpoints.compression.compressor_kind = value;
     }
@@ -2060,6 +2070,49 @@ mod tests {
         assert_eq!(None, mrf.site);
         assert_eq!(None, mrf.dd_url);
         assert_eq!(None, mrf.metrics_endpoint_url());
+    }
+
+    #[test]
+    fn secrets_settings_default_to_schema_values_when_unset() {
+        let (config, errors) = translate_explicit(json!({}));
+
+        assert!(errors.is_none());
+        let secrets = &config.shared.secrets;
+        assert_eq!(None, secrets.backend_command);
+        assert_eq!(0, secrets.refresh_on_api_key_failure_interval);
+        assert!(!secrets.in_use());
+    }
+
+    #[test]
+    fn secrets_settings_translate_and_normalize() {
+        let (config, errors) = translate_explicit(json!({
+            "secret_backend_command": "  /usr/bin/fetch-secrets  ",
+            "secret_refresh_on_api_key_failure_interval": 5
+        }));
+
+        assert!(errors.is_none());
+        let secrets = &config.shared.secrets;
+        assert_eq!(Some("/usr/bin/fetch-secrets"), secrets.backend_command.as_deref());
+        assert_eq!(5, secrets.refresh_on_api_key_failure_interval);
+        assert!(secrets.in_use());
+
+        // A blank command says no more than an absent one.
+        let (config, errors) = translate_explicit(json!({ "secret_backend_command": "   " }));
+
+        assert!(errors.is_none());
+        assert_eq!(None, config.shared.secrets.backend_command);
+        assert!(!config.shared.secrets.in_use());
+    }
+
+    #[test]
+    fn a_negative_secret_refresh_interval_is_clamped_rather_than_rejected() {
+        // The Core Agent accepts a negative interval without complaint, and it means the same thing as
+        // no interval, so translation must not fail and leave ADP unable to start.
+        let (config, errors) = translate_explicit(json!({ "secret_refresh_on_api_key_failure_interval": -1 }));
+
+        assert!(errors.is_none());
+        assert_eq!(0, config.shared.secrets.refresh_on_api_key_failure_interval);
+        assert!(!config.shared.secrets.in_use());
     }
 
     #[test]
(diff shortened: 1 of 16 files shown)


webern:
* **#2523** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2523?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2501** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2501?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2501?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2512** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2512?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -1251,6 +1251,16 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
         }
     }
 
+    fn consume_secret_backend_command(&mut self, value: String) {
+        self.config.shared.secrets.backend_command = non_empty_trimmed(value);
+    }
+
+    fn consume_secret_refresh_on_api_key_failure_interval(&mut self, value: i64) {
+        // A negative interval is not an error the Core Agent reports, and it means the same thing as
+        // no interval at all, so clamp rather than refuse to start.
```
webern:
```suggestion
        // no interval at all, so clamp rather than reporting an error.
```

Also add a warning if it is negative as that as an unexpected value.

lib/datadog-agent/config/schema/schema_overlay.yaml
```diff
@@ -2915,6 +2915,24 @@ inventory:
       additional_attributes:
         config_registry_filename: shared.rs
 
+  secret_backend_command:
+    support: full
```
webern: I think both of these should be partial, severity::low.

Documentation should explain that ADP defers to the Agent for secrets resolution, and how ADP uses these.

lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -408,14 +435,19 @@ where
             api_key_refresher.spawn();
         }
 
+        // The retry classifiers cannot await, so a task carries secrets changes into the gate they read.
+        let secrets_refresher = SecretsGateRefresher::new(live.secrets);
+        let secrets = secrets_refresher.gate.clone();
+        secrets_refresher.spawn();
+
```
webern, line 442: Hmmm. Is this the right way to handle this in this codebase? Is this a supervised task?

lib/saluki-components/src/common/datadog/retry.rs
```diff
@@ -1,20 +1,112 @@
...
+    fn store(&self, secrets: &Secrets) {
+        self.in_use.store(secrets.in_use(), Ordering::Relaxed);
+    }
+
+    fn is_open(&self) -> bool {
+        self.in_use.load(Ordering::Relaxed)
+    }
+}
+
+/// Updates a [`SecretsGate`] as typed configuration changes.
```
webern: Awkward

```suggestion
/// Updates a [`SecretsGate`] when configuration changes.
```

lib/saluki-io/src/net/util/retry/classifier/mod.rs
```diff
@@ -7,5 +7,5 @@ pub use self::http::{HttpRetryPredicate, StandardHttpClassifier};
 /// whether or not a request should be retried from the logic of determining how long to wait before retrying.
 pub trait RetryClassifier<Res, Error> {
     /// Returns `true` if the original request should be retried.
-    fn should_retry(&self, response: &Result<Res, Error>) -> bool;
+    fn should_retry(&mut self, response: &Result<Res, Error>) -> bool;
```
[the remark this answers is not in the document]
webern: We don't have downstream crate users, so the justification for this comment is wrong. However, I also wonder about these `mut` changes from the standpoint of safety and want this audited:
- is it necessary?
- is it safe?

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
webern: > GPT Sol: Applied the requested wording and added a warning before clamping a negative value to zero. Fixed in [85ef85968d](https://github.com/DataDog/saluki/commit/85ef85968d).

lib/datadog-agent/config/schema/schema_overlay.yaml
webern: > GPT Sol: Switched both settings to `partial` and documented that the Agent owns secret resolution while ADP reads these settings only to decide whether a `403 Forbidden` response may be retriable. The generated registry and configuration docs were regenerated. Fixed in [85ef85968d](https://github.com/DataDog/saluki/commit/85ef85968d).
>
> I did not add `severity: low`: partial inventory entries have no severity field. Severity currently exists only for `none` and `unknown`; adding it here would be silently ignored. Extending the partial-support model to carry severity would need to be a separate schema-model change.

lib/saluki-io/src/net/util/retry/classifier/mod.rs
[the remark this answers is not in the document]
webern: > GPT Sol: The mutability was not necessary. The audit found two production implementations: `StandardHttpClassifier`, which only reads its predicates, and `SecretsHttpClassifier`, which reads an `AtomicBool` through `SecretsGate`. The only production caller already has mutable policy access but does not need mutable classifier access. The change was behaviorally safe inside this tree, but it needlessly changed the public trait, so I restored `&self` throughout. Fixed in [85ef85968d](https://github.com/DataDog/saluki/commit/85ef85968d).

lib/saluki-components/src/common/datadog/io.rs
webern: > GPT Sol: It is not a supervised task in the runtime sense; it is a detached Tokio task created with `spawn_traced_named`. That is consistent with the existing tasks owned by `TransactionForwarder`, including the endpoint I/O loops and `ApiKeyRefresher`. `TransactionForwarder` has no supervisor handle; the supervised unit is the enclosing component.
>
> A task is needed because `Live<T>` only advances its local snapshot through `changed().await`, while `should_retry` is synchronous. The refresher pumps those changes into a gate the classifier can read, and the gate is seeded before spawning to avoid a stale initial value. I therefore left this unchanged.
>
> For process clarity: GPT Sol drafted this under Matt's supervision and review. This was not unsupervised development.

lib/saluki-components/src/common/datadog/retry.rs
webern: > GPT Sol: Applied. Fixed in [85ef85968d](https://github.com/DataDog/saluki/commit/85ef85968d).

lib/saluki-components/src/common/datadog/io.rs
webern: @tobz is this right?

tobz: This is right insofar as I have not yet gotten to the point of supervisor-ifying everything yet, `TransactionForwarder` being among the remaining bits and bobs.

tobz (approved):

PR #2503, opened 2026-08-27
feat(otlp): added origin metadata to metrics

### LINKED ISSUE #2078: Add origin metadata to OTLP metrics for billing attribution

## Background

ADP emits OTLP metrics without origin metadata. The missing metadata prevents the intake payload
from identifying the product, subproduct, and receiver that produced each metric. This affects
billing and UI attribution of OTLP metric volume. Add the Agent's origin metadata behavior to ADP
for OTLP metrics, including ordinary and sketch metrics.

### Agent behavior

The Agent represents origin metadata as `originProduct`, `originSubProduct`, and
`originProductDetail`. `origin.go:22-31` defines product values for the Agent (`10`), the Datadog
Exporter (`19`), and Unknown (`0`). OTLP translation uses subproduct OTLP (`17`) and derives
product detail from the OpenTelemetry instrumentation scope name. The scope-to-detail table
contains about 40 known Collector receiver names, including `hostmetricsreceiver`,
`kubeletstatsreceiver`, and `prometheusreceiver`; unknown receivers and scopes use detail `0`.

The reference mapping is in
`pkg/opentelemetry-mapping-go/otlp/metrics/origin.go:22-215`. The scope name is read from the
`InstrumentationScope` while translating `ScopeMetrics` in
`pkg/opentelemetry-mapping-go/otlp/metrics/metrics_translator.go:504-540`. The resulting origin
fields are placed in the metric dimensions at `metrics_translator.go:614-628`.

The Agent's serializer then maps each known `OriginProductDetail` to an internal
`MetricSource` in `comp/otelcol/otlp/components/exporter/serializerexporter/consumer.go:35-80` and
`:139-183`. That source is converted to the origin fields written to the intake payload by
`pkg/serializer/internal/metrics/origin_mapping.go` and the v1/v3 metric serializers. The
`MetricSource` step is part of the Agent's local collector/exporter and serializer pipeline; ADP
should reproduce the resulting wire metadata, not copy that internal enum unless it needs an ADP
internal equivalent.

### agent-data-plane behavior

ADP already has the wire-level representation. `MetricOrigin::OriginMetadata` in
`lib/saluki-core/src/data_model/event/metric/metadata.rs:118-160` contains `product`, `subproduct`,
and `product_detail`, and the v2 encoder writes these as origin product, category, and service in
`lib/saluki-components/src/encoders/datadog/metrics/v2/mod.rs:428-459`.

The OTLP metrics translator does not populate this field. Both `record_metric_event` and
`record_sketch_event` construct metrics with `MetricMetadata::default()` at
`lib/saluki-components/src/sources/otlp/metrics/translator.rs:632-660` and `942-966`, leaving
origin as `None`. The translator already sees each `ScopeMetrics` object and uses its scope for
optional instrumentation-scope tags at `translator.rs:430-456`, but it discards the scope before
building the metric metadata.

Add an OTLP origin constructor alongside `MetricOrigin::dogstatsd()` and `jmx_check()`, and port
the Agent's receiver-name-to-product-detail table into an OTLP metrics origin-mapping module. Keep
product/subproduct/detail constants and the receiver values identical to the Agent. Thread the
scope name through the translation context and attach the resulting `MetricOrigin` at both metric
and sketch emission sites. Confirm whether the Agent baseline emits product `10` or the OTLP
serializer's product `19` for this path; the correctness test should establish the expected wire
value.

### Requirements tracking

- OM_ORIG_417
- OM_ORIG_052
- OM_ORIG_803
- OM_ORIG_671

## Definition of Done

- [ ] Demonstrate the missing origin metadata with a red OTLP correctness test.
- [ ] Populate product, OTLP subproduct `17`, and receiver-derived product detail on metric and
      sketch events.
- [ ] Match the Agent's known receiver mappings and unknown-scope fallback.
- [ ] Verify the v2 intake payload matches the Agent baseline for origin metadata and billing
      attribution.
- [ ] Make the correctness test green.

## References

- Sub-issue of #2005
- Relates to #2006
- Relates to #2016
- Relates to #2065
- Relates to #2070

lucastemb:
## Summary

Adds origin metadata (`originProduct`, `originSubProduct`, and `originProductDetail`) for OTLP Ingest metrics. 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests

## References

- Closes: #2078

### DIFF
--- a/bin/correctness/panoramic/src/correctness/analysis/metrics/types.rs
+++ b/bin/correctness/panoramic/src/correctness/analysis/metrics/types.rs
@@ -202,13 +202,28 @@ impl NormalizedMetrics {
         for metric in metrics {
             let context = NormalizedMetricContext::from_stele_context(metric.context().clone());
             let metric_type = metric_value_type(metric.values());
+            let metric_origin = metric.origin().cloned();
             let key = (context, metric_type);
             let (context_values, origin) = aggregated_context_values
                 .entry(key)
                 .or_insert_with(|| (Vec::new(), None));
             context_values.extend_from_slice(metric.values());
-            if origin.is_none() {
-                *origin = metric.origin().cloned();
+
+            // Track the origin for this group. The same metric may arrive via multiple wire formats
+            // (V1 JSON, V2 series/sketch, V3), and we take the first non-`None` origin. If a later
+            // metric in the same group has a different non-`None` origin, that indicates a wire-format
+            // discrepancy — we reject it rather than silently dropping the conflicting value.
+            match (&*origin, &metric_origin) {
+                (None, Some(new_origin)) => *origin = Some(new_origin.clone()),
+                (Some(existing), Some(new_origin)) if existing != new_origin => {
+                    return Err(generic_error!(
+                        "Conflicting origin metadata for metric '{}': {:?} vs {:?}",
+                        metric.context().name(),
+                        existing,
+                        new_origin
+                    ));
+                }
+                _ => {}
             }
         }
 
--- a/bin/correctness/panoramic/src/correctness/analysis/metrics/types.rs
+++ b/bin/correctness/panoramic/src/correctness/analysis/metrics/types.rs
@@ -4,7 +4,7 @@ use std::{
 };
 
 use saluki_error::{generic_error, ErrorContext as _, GenericError};
-use stele::{Metric, MetricContext, MetricValue};
+use stele::{Metric, MetricContext, MetricOrigin, MetricValue};
 
 /// A metric context and type pair, used for comparing metrics between different sets.
 type ContextTypePair<'a> = (&'a NormalizedMetricContext, MetricType);
@@ -101,16 +101,17 @@ pub struct NormalizedMetric {
     context: NormalizedMetricContext,
     raw_values: Vec<(u64, MetricValue)>,
     value: MetricValue,
+    origin: Option<MetricOrigin>,
 }
 
 impl NormalizedMetric {
-    /// Attempts to create `NormalizedMetric` from the given context and raw metric values.
+    /// Attempts to create `NormalizedMetric` from the given context, raw metric values, and origin.
     ///
     /// # Errors
     ///
     /// If the raw values are empty, or if the values can't be normalized, an error is returned.
     pub fn try_from_values(
-        context: NormalizedMetricContext, mut raw_values: Vec<(u64, MetricValue)>,
+        context: NormalizedMetricContext, mut raw_values: Vec<(u64, MetricValue)>, origin: Option<MetricOrigin>,
     ) -> Result<Self, GenericError> {
         // We need to first sort the raw values by timestamp, to ensure we have proper ordering semantics.
         raw_values.sort_by_key(|a| a.0);
@@ -121,6 +122,7 @@ impl NormalizedMetric {
             context,
             raw_values,
             value,
+            origin,
         })
     }
 
@@ -149,7 +151,15 @@ impl NormalizedMetric {
     pub fn normalized_value(&self) -> &MetricValue {
         &self.value
     }
+
+    /// Returns the origin metadata of the metric, if present.
+    pub fn origin(&self) -> Option<&MetricOrigin> {
+        self.origin.as_ref()
+    }
 }
+
+/// Aggregated values and origin for a single (context, type) group during normalization.
+type AggregatedMetric = (Vec<(u64, MetricValue)>, Option<MetricOrigin>);
 
 /// A set of normalized metrics.
 ///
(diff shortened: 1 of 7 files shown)


lib/saluki-core/src/data_model/event/metric/metadata.rs
```diff
@@ -159,6 +161,19 @@ impl MetricOrigin {
         }
     }
 
+    /// Creates a `MetricOrigin` for any metric ingested via OTLP.
+    ///
+    /// OTLP metrics use product `19` and subproduct `0`. The product detail is provided by the
+    /// caller— typically derived from the instrumentation scope name— and falls back to `0` (unknown) when the
+    /// scope is absent or does not map to a known receiver.
```
tobz: Can we drop these comment lines here?

tobz (approved):
One small comment nit but other than that, LGTM.

lib/saluki-core/src/data_model/event/metric/metadata.rs
```diff
@@ -3,8 +3,10 @@ use std::{fmt, sync::Arc};
 use stringtheory::MetaString;
 
 const ORIGIN_PRODUCT_AGENT: u32 = 10;
```
webern, line 8: Adjunct to tobz's request below, can we just say what these are at a quick glance for an uninitiated reader.

Basically just one or two human sentences for the clueless newb.

```suggestion
// These come from... where?
const ORIGIN_PRODUCT_AGENT: u32 = 10;
```

lib/saluki-components/src/sources/otlp/metrics/translator.rs
```diff
@@ -88,6 +89,12 @@ enum MetricTypeOverrideWarningKind {
 struct TranslationContext<'a> {
     resource_attributes: &'a [OtlpKeyValue],
     metrics: &'a Metrics,
+    /// The `InstrumentationScope` name of the enclosing `ScopeMetrics`, or an empty string when no scope was
+    /// present.
+    ///
+    /// Used to derive the metric origin's product detail. An empty
```
webern, line 96: nit: wonky wrapping

lib/saluki-components/src/sources/otlp/metrics/origin.rs
```diff
@@ -0,0 +1,154 @@
...
+const COLLECTOR_RECEIVER_PREFIX: &str = "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/";
+
+/// Returns the origin product detail for the given instrumentation scope name.
+///
+/// Returns `0` (unknown) when the scope name is empty or does not begin with the Collector receiver prefix.
+/// Otherwise, extracts the receiver name—the first path segment after the prefix—and looks it up in the
+/// known-receiver table.
+pub fn product_detail_from_scope(scope_name: &str) -> u32 {
+    product_detail_from_scope_name(scope_name)
+}
```
webern: What is the purpose of this wrapper function?

webern (approved):
Some nits/requests and I opened #2519 for a someday follow-up. Otherwise LGTM

❌  only merge after all correctness tests pass!

lib/saluki-core/src/data_model/event/metric/metadata.rs
webern: > GPT Terra: appears resolved.

lib/saluki-components/src/sources/otlp/metrics/origin.rs
webern: > GPT Terra: appears resolved.

bin/correctness/panoramic/src/correctness/analysis/metrics/types.rs
```diff
@@ -183,19 +193,43 @@ impl NormalizedMetrics {
...
+            let (context_values, origin) = aggregated_context_values
+                .entry(key)
+                .or_insert_with(|| (Vec::new(), None));
             context_values.extend_from_slice(metric.values());
+
+            // Track the origin for this group. The same metric may arrive via multiple wire formats
+            // (V1 JSON, V2 series/sketch, V3), and we take the first non-`None` origin. If a later
+            // metric in the same group has a different non-`None` origin, that indicates a wire-format
+            // discrepancy — we reject it rather than silently dropping the conflicting value.
+            match (&*origin, &metric_origin) {
```
webern, line 216: My thought with the failing test is that we could make this origin-sensitive assertion a unique/different assertion that is only used for a test that wants it. And perhaps it should focus only on the origin behavior that we are specifically wanting to check.

In other words, your would have 
 - The existing assertion (currently broken): `telemetry matches`
 - A new opt-in assertion: `metric origins match`, Config: `compare_metric_origins`

If you split them then your test could use both assertions and others would not be opted-in.

> GPT: I traced this failure through the downloaded GitLab artifacts and the relevant Agent paths. It appears before baseline-versus-ADP comparison: the baseline normalizer encounters the same generated DogStatsD context with both product 10/category 0 and product 10/category 10. The timestamp-extension test sends timestamped samples through the Agent's no-aggregation path while untimestamped samples use normal aggregation, so one logical metric can legitimately arrive from those two paths with different origin metadata. Because this normalizer intentionally coalesces records by context and type across payloads and flushes, requiring every coalesced record to have one origin is stricter than existing baseline behavior and breaks an unrelated correctness case. I agree with Matt's idea that origin validation could work as a separate opt-in assertion: preserve existing value normalization, collect observed origins independently, and enable origin comparison only for focused OTLP correctness cases.

webern (changes requested):
I'm not sure if I understand the whole origin assertion thing correctly, but I left a suggestion of splitting it out into its own assertion. Marking as `Request changes` due to the test failure.

Issue #1634, opened 2026-05-12
dogstatsd: ADP silently drops metrics with invalid UTF-8 in tag block


jszwedko:
## Summary

When a DogStatsD datagram carries invalid UTF-8 bytes in the tag block (e.g. `\xff\xfe`), ADP rejects the entire datagram silently while the Datadog Agent's built-in DogStatsD implementation accepts the same payload. This is a silent migration regression for any client that emits binary-tainted tag values (locale bytes, hashed identifiers, upstream encoding bugs).

## Behavior

**ADP:** datagram is dropped. `adp__component_errors_total{component_id="dsd_in", error_type="decode", message_type="metrics"}` increments by 1. The metric never appears downstream. No log line names the dropped metric name.

**Core agent DogStatsD (ADP disabled):** No error. Metric is accepted and forwarded normally.

Note: `adp__component_events_dropped_total` does **not** increment when this occurs — the drop is invisible to that counter (separate issue).

## Root cause

`lib/saluki-io/src/deser/codec/dogstatsd/helpers.rs` — the `tags()` parser runs `simdutf8::basic::from_utf8` over the entire tag block before any parsing:

```rust
move |input| match simdutf8::basic::from_utf8(input) {
    Ok(tags) => Ok((&[], RawTags::new(tags, max_tag_count, max_tag_len))),
    Err(_) => Err(nom::Err::Error(Error::new(input, ErrorKind::Verify))),
}
```

Any invalid UTF-8 byte anywhere in the tag block causes the whole datagram to be dropped.

## Reproduction

```shell
# Send a gauge with an invalid UTF-8 byte in the tag block
printf 'mymetric:1|g|#env:prod,tag:\xff\xfe' | nc -u -w1 localhost 8125
```

ADP logs:
```
WARN | Failed to parse frame. ... error:"encountered error 'Verify' while processing message '...'"
```

Core agent: metric is accepted, no error.

## Expected behavior

ADP should match the core agent's behavior. The core agent's `parseTags` splits the raw tag bytes on `,` and converts each slice to a string via Go's `string([]byte)` — which performs no UTF-8 validation and passes invalid bytes through verbatim. There is no validity check on tag bytes anywhere in the core agent's parsing path.

ADP should accept datagrams with invalid UTF-8 tag bytes rather than dropping the entire metric. Invalid bytes should be preserved or replaced with `\u{FFFD}` — but the metric must not be silently discarded.

## Suggested mitigation

[...]

PR #2504, opened 2026-08-28
fix(dogstatsd): normalize invalid UTF-8 tag blocks

thieman:
## Summary

- Normalize invalid UTF-8 in every DogStatsD `#` tag block to one U+FFFD per contiguous invalid byte run, matching the backend behavior modeled for Agent-originated v2 series.
- Retain borrowed, zero-copy tag storage for valid inputs; use owned backing storage only after normalization is required.
- Refactor raw-tag iteration and DogStatsD well-known-tag extraction to borrow the backing representation, allowing metrics, events, and service checks to handle either storage form safely.
- Add regression coverage for the normalizer and for metric, event, and service-check tag blocks.

## Scope

This implements DADP-77 / #1634. It intentionally does not add recovery telemetry, an error subtype, or a user-facing configuration policy.

## Validation

- [x] Test-driven red phase: disabling recovery causes the invalid-tag event regression test to fail with the previous `Verify` parser error.
- [x] `cargo nextest run -p saluki-io` — 166 passed.
- [x] `make fmt`
- [x] `make check-all`
- [x] `make test-all`
- [x] Repository pre-commit hooks.

## Self-review

A fresh-context review found and the follow-up commit corrected the permissive-decoding documentation, which still claimed invalid UTF-8 tag bytes were rejected. No remaining correctness, lifetime-safety, normalization, special-tag/origin, allocation-path, telemetry-scope, or configuration-scope defects were identified.

### DIFF
--- a/lib/saluki-components/src/sources/dogstatsd/origin.rs
+++ b/lib/saluki-components/src/sources/dogstatsd/origin.rs
@@ -277,9 +277,12 @@ impl OriginTagsResolver for DogStatsDOriginTagResolver {
 }
 
 /// Builds an `RawOrigin` object from the given metric packet.
-pub fn origin_from_metric_packet<'packet>(
-    packet: &MetricPacket<'packet>, well_known_tags: &WellKnownTags<'packet>,
-) -> RawOrigin<'packet> {
+pub fn origin_from_metric_packet<'packet, 'tags>(
+    packet: &'tags MetricPacket<'packet>, well_known_tags: &'tags WellKnownTags<'tags>,
+) -> RawOrigin<'tags>
+where
+    'packet: 'tags,
+{
     let cardinality = packet.cardinality.or(well_known_tags.cardinality);
 
     let mut origin = RawOrigin::default();
@@ -291,9 +294,12 @@ pub fn origin_from_metric_packet<'packet>(
 }
 
 /// Builds an `RawOrigin` object from the given event packet.
-pub fn origin_from_event_packet<'packet>(
-    packet: &EventPacket<'packet>, well_known_tags: &WellKnownTags<'packet>,
-) -> RawOrigin<'packet> {
+pub fn origin_from_event_packet<'packet, 'tags>(
+    packet: &'tags EventPacket<'packet>, well_known_tags: &'tags WellKnownTags<'tags>,
+) -> RawOrigin<'tags>
+where
+    'packet: 'tags,
+{
     let cardinality = packet.cardinality.or(well_known_tags.cardinality);
 
     let mut origin = RawOrigin::default();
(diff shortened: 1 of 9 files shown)


lib/saluki-components/src/sources/dogstatsd/origin.rs
```diff
@@ -291,9 +294,12 @@ pub fn origin_from_metric_packet<'packet>(
...
 
 /// Builds an `RawOrigin` object from the given event packet.
-pub fn origin_from_event_packet<'packet>(
-    packet: &EventPacket<'packet>, well_known_tags: &WellKnownTags<'packet>,
-) -> RawOrigin<'packet> {
+pub fn origin_from_event_packet<'packet, 'tags>(
+    packet: &'tags EventPacket<'packet>, well_known_tags: &'tags WellKnownTags<'tags>,
+) -> RawOrigin<'tags>
+where
+    'packet: 'tags,
```
thieman, line 301: Does this mean these two lifetimes are identical? If so, why do we need both?

thieman: [GPT-5.6 Terra (OpenAI)] No. `'packet` is the lifetime of data borrowed *by* the packet: frame-backed metric name/local/external fields, plus the borrowed `RawTags` variant. `'tags` is the shorter lifetime of the borrow of the `MetricPacket` and `WellKnownTags` in the source handler.

They can happen to coincide at a call site, but they are not the same contract. After this change, `RawTags` may instead hold a `Cow::Owned` normalized string. `WellKnownTags` then borrows that packet-owned string, not the input frame. `RawOrigin` borrows both packet fields and well-known tag values, so it must be bounded by `'tags`; the `'packet: 'tags` bound says any frame-backed packet fields outlive that shorter borrow.

Returning `RawOrigin<'packet>` would incorrectly promise that well-known tag references outlive the packet object when their backing string may be owned by it. The resolver consumes `RawOrigin` synchronously in the handler, so the shorter lifetime is the intended boundary.

tobz (approved):

webern (approved):
Nice use of Cow::Borrowed to avoid allocations in the happy path.

Is `U+FFFD` the canonical behavior we are supposed to use (it is three bytes, apparently)?

tobz:
> Is `U+FFFD` the canonical behavior we are supposed to use (it is three bytes, apparently)?

@webern Yeah, this is the only sane decision we can make at the ADP/Saluki layer: downstream systems may choose to further sanitize (turn the replacement character into an underscore, for example) but we don't know that ahead of time so we're doing the minimum viable thing here to turn the tag into valid UTF-8 while preserving what we can.

PR #2507, opened 2026-08-31
feat(otlp): new error, drop, and latency telemetry for metrics

### LINKED ISSUE #2091: Add error, drop, and latency telemetry to OTLP metrics ingest

## Background

ADP records OTLP input bytes and the number of metric events it produces. It does not record
translation errors, dropped metric points, or the time spent processing OTLP metrics. Add self-
telemetry for these outcomes so operators can detect and measure ingestion problems. This is an
observability improvement; the Agent exposes request and event volume counters but does not expose
equivalent error, drop, or latency counters.

### Agent behavior

The Agent defines four OTLP metric volume counters in
`comp/otelcol/otlp/components/exporter/serializerexporter/telemetry.go:13-55`:
`datadog_agent_otlp_metrics_events`, `datadog_agent_otlp_metrics_requests`, and the corresponding
`ddot_otlp_metrics_events` and `ddot_otlp_metrics_requests` counters. The exporter increments the
selected request counter and adds the input metric count to the selected event counter at the start
of `ConsumeMetrics`, before translation, hostname resolution, or serialization
(`comp/otelcol/otlp/components/exporter/serializerexporter/exporter.go:186-196`).

The Agent returns translation errors from `MapMetrics`, hostname errors, and serializer errors
(`exporter.go:206-217`), but does not increment a failure counter for those paths. The consumer also
has uninstrumented drop paths: explicit and exponential histogram consumer callbacks are no-ops
(`consumer.go:113-119`), and client-stats payloads are silently discarded when encoding fails
(`consumer.go:124-129`). There is no OTLP-specific error counter, dropped-point counter, or latency
histogram in the serializer exporter package. Use this behavior as the reference for existing
volume telemetry, not as a request to copy its missing failure telemetry.

### agent-data-plane behavior

The shared OTLP server registers `component_events_received_total` for OTLP metrics and
`component_bytes_received_total` for OTLP input in
`lib/saluki-components/src/common/otlp/mod.rs:57-108`. The source translator increments the events
counter after producing output events (`lib/saluki-components/src/sources/otlp/metrics/translator.rs:520`),
while decode and channel failures are only returned or logged
(`sources/otlp/mod.rs:351-360`). The converter logs and drops a complete resource on translation
failure and logs individual dispatcher failures (`sources/otlp/mod.rs:412-429`), while the metrics
translator logs and drops unsupported temporalities or failed histogram conversions
(`sources/otlp/metrics/translator.rs:575-625, 1100-1110, 1190-1212`). None of these paths update a
counter. The path also has no latency measurement.

Add error and dropped-point counters, tagged with a bounded reason such as `decode`, `translate`,
`unsupported_temporality`, `histogram_conversion`, `dispatch`, or `flush`. Add a histogram for
translation or end-to-end ingest latency, with units and scope documented in the metric help. The
server-level failures can extend `common::otlp::Metrics`; translation-specific counters and the
latency histogram can be held by a metrics struct in `sources/otlp/metrics` and passed through the
translator. Register them with `MetricsBuilder::from_component_context` and
`register_counter_with_tags`/`register_histogram_with_tags`, following the existing metrics idiom.
Avoid unbounded metric labels and do not change the emitted OTLP metric data.

### Requirements tracking

- OM_RESZ_233

## Definition of done

- [ ] Demonstrate the missing error, dropped-point, and latency telemetry with focused red tests.
- [ ] Add bounded-reason error and dropped-point counters to the relevant ADP OTLP paths.
- [ ] Add and document a processing-latency histogram for OTLP metrics.
- [ ] Verify counter increments for decode, translation, conversion, dispatch, and flush failures,
      and for each intentional per-point drop path.
- [ ] Verify latency samples are recorded for successful and failed metric batches.
- [ ] Add or update tests using ADP's metrics debugging recorder/snapshotter pattern to assert
      counter and histogram values.
- [ ] Keep existing OTLP event/byte counters and emitted metric output unchanged.

## References

- Sub-issue of #2005
- Relates to #2006
- Relates to #2073

lucastemb:
## Summary

Adds new telemetry for metric drops, errors, and we now have latency reporting. New `Counters` `otlp_metrics_errors_total`  `otlp_metrics_dropped_points_total` and `Histogram` `otlp_metrics_processing_duration_seconds` - each with varying tag giving more granular report on the reason (e.g. `translate`, `unsupported_temporality`, etc.) 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests

## References

- Closes: #2091

### DIFF
--- a/lib/saluki-components/src/common/otlp/mod.rs
+++ b/lib/saluki-components/src/common/otlp/mod.rs
@@ -70,6 +70,9 @@ pub struct Metrics {
     logs_received: Counter,
     bytes_received: Counter,
     spans_received: Counter,
+    metrics_errors_decode: Counter,
+    metrics_errors_dispatch: Counter,
+    metrics_errors_flush: Counter,
 }
 
 impl Metrics {
@@ -89,6 +92,18 @@ impl Metrics {
         &self.bytes_received
     }
 
+    pub fn metrics_errors_decode(&self) -> &Counter {
+        &self.metrics_errors_decode
+    }
+
+    pub fn metrics_errors_dispatch(&self) -> &Counter {
+        &self.metrics_errors_dispatch
+    }
+
+    pub fn metrics_errors_flush(&self) -> &Counter {
+        &self.metrics_errors_flush
+    }
+
     /// Test-only helper to construct a `Metrics` instance.
     #[cfg(test)]
     pub fn for_tests() -> Self {
@@ -97,6 +112,9 @@ impl Metrics {
             logs_received: Counter::noop(),
             bytes_received: Counter::noop(),
             spans_received: Counter::noop(),
+            metrics_errors_decode: Counter::noop(),
+            metrics_errors_dispatch: Counter::noop(),
+            metrics_errors_flush: Counter::noop(),
         }
     }
 }
@@ -113,6 +131,10 @@ pub fn build_metrics(component_context: &ComponentContext) -> Metrics {
         bytes_received: builder.register_counter_with_tags("component_bytes_received_total", [("source", "otlp")]),
         spans_received: builder
             .register_counter_with_tags("component_events_received_total", [("message_type", "otlp_spans")]),
+        metrics_errors_decode: builder.register_counter_with_tags("otlp_metrics_errors_total", [("reason", "decode")]),
+        metrics_errors_dispatch: builder
+            .register_counter_with_tags("otlp_metrics_errors_total", [("reason", "dispatch")]),
+        metrics_errors_flush: builder.register_counter_with_tags("otlp_metrics_errors_total", [("reason", "flush")]),
     }
 }
 
@@ -457,6 +479,7 @@ async fn http_metrics_handler<H: OtlpHandler>(
         Ok(()) => (StatusCode::OK, "OK"),
         Err(e) => {
             error!(error = %e, "Failed to handle OTLP metrics.");
+            metrics.metrics_errors_decode().increment(1);
             (StatusCode::INTERNAL_SERVER_ERROR, "Internal processing error")
         }
     }
@@ -537,6 +560,7 @@ impl<H: OtlpHandler> MetricsService for GrpcServiceImpl<H> {
             Ok(()) => Ok(Response::new(ExportMetricsServiceResponse { partial_success: None })),
             Err(e) => {
                 error!(error = %e, "Failed to handle OTLP metrics.");
+                self.metrics.metrics_errors_decode().increment(1);
                 Err(Status::internal("Internal processing error"))
             }
         }
@@ -600,6 +624,7 @@ mod tests {
         components::{test_util::TestComponentSupervisor, ComponentContext},
         runtime::state::{DataspaceUpdate, IdentifierFilter},
     };
+    use saluki_error::generic_error;
     use saluki_io::net::server::BoundServerAddress;
     use saluki_metrics::test::TestRecorder;
     use saluki_tls::test_util::SelfSignedCert;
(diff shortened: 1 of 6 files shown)


lib/saluki-components/src/common/otlp/mod.rs
```diff
@@ -113,6 +137,12 @@ pub fn build_metrics(component_context: &ComponentContext) -> Metrics {
         bytes_received: builder.register_counter_with_tags("component_bytes_received_total", [("source", "otlp")]),
         spans_received: builder
             .register_counter_with_tags("component_events_received_total", [("message_type", "otlp_spans")]),
+        metrics_errors_decode: builder.register_counter_with_tags("otlp_metrics_errors_total", [("reason", "decode")]),
+        metrics_errors_channel: builder
+            .register_counter_with_tags("otlp_metrics_errors_total", [("reason", "channel")]),
+        metrics_errors_dispatch: builder
+            .register_counter_with_tags("otlp_metrics_errors_total", [("reason", "dispatch")]),
+        metrics_errors_flush: builder.register_counter_with_tags("otlp_metrics_errors_total", [("reason", "flush")]),
```
tobz: Can we actually make these `component_errors_total`?

We can always remap them in order to send them over RAR in the expected shape, if need be, but on the Saluki/ADP side... we want to standardize on _components_ emitting telemetry in the shape of `component_*_total` where possible.

lib/saluki-components/src/sources/otlp/metrics/telemetry.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+impl OtlpMetricsTranslatorMetrics {
+    /// Builds the translator telemetry from the given component context.
+    pub fn from_component_context(component_context: &ComponentContext) -> Self {
+        let builder = MetricsBuilder::from_component_context(component_context);
+
+        Self {
+            errors_translate: builder
+                .register_counter_with_tags("otlp_metrics_errors_total", [("reason", "translate")]),
```
tobz: Same thing here about `component_errors_total`.

lib/saluki-components/src/sources/otlp/metrics/telemetry.rs
```diff
@@ -0,0 +1,75 @@
...
+            dropped_unsupported_temporality: builder.register_counter_with_tags(
+                "otlp_metrics_dropped_points_total",
+                [("reason", "unsupported_temporality")],
+            ),
+            dropped_histogram_conversion: builder.register_counter_with_tags(
+                "otlp_metrics_dropped_points_total",
+                [("reason", "histogram_conversion")],
+            ),
+            dropped_invalid_value: builder
+                .register_counter_with_tags("otlp_metrics_dropped_points_total", [("reason", "invalid_value")]),
```
tobz: Are we legitimately tracking drops at the level of points here? Like... could there be a single metric where we don't drop the entire thing but only drop _some_ points from it?

I ask because if "point" here is _actually_ synonymous with "metric", then we should call this `component_events_dropped_total` instead.

lib/saluki-components/src/sources/otlp/metrics/translator.rs
```diff
@@ -471,8 +476,25 @@ impl OtlpMetricsTranslator {
...
+    /// translation succeeds or fails, so the histogram captures the full distribution of processing times.
     pub fn translate_metrics(
         &mut self, resource_metrics: OtlpResourceMetrics, metrics: &Metrics,
+    ) -> Result<(IntoIter<Event>, FastHashSet<&'static str>), GenericError> {
+        let start = Instant::now();
+        let result = self.translate_metrics_inner(resource_metrics, metrics);
+        self.translator_metrics
+            .processing_duration()
+            .record(start.elapsed().as_secs_f64());
+        if let Err(ref _e) = &result {
```
tobz: Just do `if result.is_err() {`.

tobz (changes requested):

lib/saluki-components/src/sources/otlp/metrics/telemetry.rs
lucastemb: Yeah, this is good call out. I think `point` is synonymous with `event` in this context so the update naming suggestion is more accurate

tobz (approved):

PR #2499, opened 2026-08-27
chore(config): refresh endpoint api keys from typed config

webern:
## Human Summary

HOLD: DO NOT REVIEW. I think I'm closing this and starting over.

Moves API-key refresh to typed config. Refreshing the API key was a pull-based mutation whenever the middleware accessed `api_key`, and this refactor retains that design. Supporting this pull-based live configuration view required a change to `Live<T>`'s interface. The `refresh()` function gets the latest value from configuration and updates the helpd snapshot of it without interfering with a different consumer that may be waiting for it through `changed()` (getting this right was the bulk of the back-and-forth work on this PR).

## AI Summary

Move endpoint API-key refresh from the raw configuration map to typed live views.

`ResolvedEndpoint` no longer stores a cloned `GenericConfiguration` or configuration key paths. Its `ApiKeySource` identifies one of four sources:

- a fixed key that configuration does not refresh;
- the required primary API key;
- the optional Multi-Region Failover API key;
- one position in `additional_endpoints`.

The request middleware still refreshes keys when it builds a request. `Live::refresh(&mut self) -> &T` supports this synchronous path. It loads the latest accepted projected value into that view's snapshot and returns a reference to it. It does not mark the watch cursor as observed. Separate `Live<T>` clones retain separate snapshots and watch cursors.

Endpoint lookup borrows from the refreshed snapshot. It no longer clones the full `additional_endpoints` map on every request. A changed key is copied into `ResolvedEndpoint`; a missing or blank key leaves the previously configured key in place.

Retry queue IDs remain based on an additional endpoint's configured URL and key position.

`api_key` and `additional_endpoints` drop `get_typed` from the overlay inventory. API-key validation still subscribes to raw configuration change notifications, but reads key values through the typed endpoint views.

### Behavior notes

- API-key normalization is now uniform. All keys are trimmed. Blank keys do not replace a configured key.
- The live path no longer accepts `additional_endpoints` as a JSON-encoded string. It uses the typed mapping form already required at startup.

`AdditionalEndpoints` was the last YAML-deserialized type in `saluki-components`, so the PR removes `serde_with` and `serde_yaml` from that crate and updates generated dependency files.

## Change Type

- [x] Non-functional (chore, refactoring, docs)

## How did you test this PR?

- `make fmt`
- `make check-docs`
- `make check-clippy`
- `cargo nextest run -p saluki-components -p agent-data-plane-config -p agent-data-plane-config-system`
[...]

PR #2512, opened 2026-09-01
chore(config): refresh endpoint api keys from typed config

webern:
## Human Summary

This PR replaces a previous attempt in #2499 with a different approach. In the previous attempt `Live<T>` was modified to support both push and pull apis for config updates because `api_key` pulled the latest value whenever it was called. It was a mess to try and give `Live<T>` two contracts at once in a sensible, least-surprising way.

So instead, this PR introduces a place, `ApiKeyCell`, where the event-style arrival of a new API key can be processed once when `changed()` fires. The middleware gets the latest processed key from this whenever it needs to build a request.

## AI Summary

Move endpoint API-key refresh from endpoint-side raw-map reads to typed live views.

Each endpoint holds its current key in an atomic `ApiKeyCell`. While the forwarder is built, the refresher reconciles every cell with its live view, closing the gap between the configuration snapshot used to build endpoints and the view's initial value. Once the forwarder starts, one task per bound view updates cells after `Live::changed()` reports a new configuration value. Requests read the cell without mutating `ResolvedEndpoint`.

API-key validation follows a signal emitted after a cell is replaced. It therefore validates the same key requests use and cannot publish readiness for a key that the endpoint has already replaced.

Keys are trimmed before comparison and installation. A missing, blank, or HTTP-header-invalid live update leaves the last usable key in place and does not trigger validation. An HTTP-header-invalid initial primary or endpoint-override key still prevents that endpoint from being built. An invalid initial additional-endpoint key is logged and skipped, like an empty or duplicate additional key; other configured endpoints still start.

Primary and metrics-primary endpoints follow the primary key view. Additional endpoints follow the configured URL and key position. Multi-Region Failover follows its optional region key. The Cluster Agent continues to use its independently supplied bearer token.

The endpoint set remains fixed at startup. Retry queue IDs continue to use the configured URL and key position. This removes endpoint-side raw-map parsing and pull-based key refresh; `Live<T>` remains push-only.

## Change Type

## How did you test this PR?

- `make fmt`
- `make check-docs`
- `make check-clippy`
- `cargo nextest run -p saluki-components common::datadog`
- `cargo nextest run -p saluki-components forwarders::`

## References

- Replaces: #2499
- Progresses: #2293


webern:
* **#2523** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2523?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2512** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2512?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2512?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/saluki-components/src/common/datadog/api_key.rs
```diff
@@ -0,0 +1,574 @@
...
+    fn store(&self, api_key: Option<&str>) -> StoreOutcome {
+        let Some(api_key) = api_key.map(str::trim).filter(|key| !key.is_empty()) else {
+            return StoreOutcome::NoUsableKey;
+        };
+
+        if **self.current.load() == *api_key {
+            return StoreOutcome::Unchanged;
+        }
+
+        self.current.store(Arc::new(shared(api_key)));
```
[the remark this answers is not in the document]
webern: Fix seems legit - Matt

> Opus: Fixed in [`8f5dbb8`](https://github.com/DataDog/saluki/commit/8f5dbb8e0264d2c861d5d7d478e37f62de2550ae).
> `ApiKeyCell` now rejects a key that is not a valid HTTP header value in both places it can arrive:
> `ApiKeyCell::new` returns an error, so endpoint construction fails for a configured key
> (`EndpointError::InvalidApiKey`), and `store` returns `StoreOutcome::InvalidHeaderValue` for a live
> update, which leaves the previous key in place and logs an error. The invalid key never reaches the
> request path, so the `HeaderValue::from_str` call in `middleware.rs` cannot see it.

lib/saluki-components/src/common/datadog/validation.rs
```diff
@@ -245,14 +245,15 @@ async fn validate_targets(client: &mut HttpClient, targets: &[ValidationTarget])
...
     let mut targets = Vec::new();
 
     for routable in endpoints {
-        // `api_key()` lazily refreshes the cached key from live configuration before validation.
-        let endpoint = routable.endpoint_mut();
-        let api_key = endpoint.api_key().trim().to_string();
+        // `api_key()` returns whatever the endpoint's refresher stored last, so validation follows a
+        // rotation without rebuilding the endpoint set.
+        let endpoint = routable.endpoint();
+        let api_key = endpoint.api_key().to_string();
```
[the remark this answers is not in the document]
webern: > Opus: Good catch, fixed in [`8f5dbb8`](https://github.com/DataDog/saluki/commit/8f5dbb8e0264d2c861d5d7d478e37f62de2550ae).
> Validation no longer subscribes to the raw configuration map. `ApiKeyRefresher` signals through a new
> `ApiKeyChanges` handle after it installs a key, and the validation loop waits on that signal, so it
> can only validate a key the endpoints already hold. A key the cell rejects installs nothing and
> signals nothing, so validation does not re-run for it, and the URL set of the endpoints is still
> fixed at startup. Covered by `a_rotated_key_is_revalidated_before_the_next_interval` in
> `validation.rs`, which uses an hour-long interval so only the key change can drive the second
> validation.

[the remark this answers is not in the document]
webern: I guess we were still responding to the `GenericConfiguration` event channel by accident. The fix looks legit. - Matt

> Opus: Good catch, fixed in [`8f5dbb8`](https://github.com/DataDog/saluki/commit/8f5dbb8e0264d2c861d5d7d478e37f62de2550ae).
> Validation no longer subscribes to the raw configuration map. `ApiKeyRefresher` signals through a new
> `ApiKeyChanges` handle after it installs a key, and the validation loop waits on that signal, so it
> can only validate a key the endpoints already hold. A key the cell rejects installs nothing and
> signals nothing, so validation does not re-run for it, and the URL set of the endpoints is still
> fixed at startup. Covered by `a_rotated_key_is_revalidated_before_the_next_interval` in
> `validation.rs`, which uses an hour-long interval so only the key change can drive the second
> validation.

lib/saluki-components/src/common/datadog/api_key.rs
```diff
@@ -0,0 +1,706 @@
...
+        StoreOutcome::Replaced
+    }
+}
+
+/// Returns `api_key` as a `MetaString` that clones without allocating, whatever its length.
```
tobz: This is not true in practice, because `Arc::<str>::from` will actually allocate to hold the `str` as a byte slice, although we won't allocate to go from `Arc<str>` to `MetaString`.

lib/saluki-components/src/common/datadog/api_key.rs
```diff
@@ -0,0 +1,706 @@
...
+        StoreOutcome::NoUsableKey => warn!(
+            endpoint,
+            "Configuration no longer supplies a usable API key for this endpoint. Continuing with the previously \
+             configured API key."
+        ),
+        StoreOutcome::InvalidHeaderValue => error!(
+            endpoint,
+            "Configuration supplied an API key that cannot be used as an HTTP header value. Continuing with the \
+             previously configured API key."
+        ),
```
tobz: Seems like we should just collapse these into `InvalidKey` or something, and maybe log the _reason_ why the key is invalid at the point where we actually try and parse/validate it pre-store?

lib/saluki-components/src/common/datadog/api_key.rs
```diff
@@ -0,0 +1,706 @@
...
+///
+/// Configuration always resolves the primary intake's key to a string, while a failover region's key
+/// can be left unset. The two shapes are kept apart so that neither view has to invent a value.
+#[derive(Clone, Debug)]
+pub(crate) enum ApiKeyView {
+    /// A key configuration always resolves, such as the primary intake's key.
+    Required(Live<String>),
+
+    /// A key configuration can leave unset, such as a failover region's key.
+    Optional(Live<Option<String>>),
```
tobz, line 143: I think I'm understanding the purpose of this variant -- handle cases like Cluster Autoscaling Failover where there's no normal API key, just a bearer token -- but I'm wondering if this can't just be `ApiKeyView(Option<Live<String>>)`?

We could still satisfy the methods we have on `ApiKeyView` here, but unless I'm missing something, everything that uses this -- normal DD forwarder, MRF, DCA autoscaling failover -- either has an API key _and_ requires it, or doesn't have an API key at all... which I think makes a case for the optionality being on the outside, rather than the inside: for cases where we need an API key, we never expect it to ever be legitimately unset.

lib/saluki-components/src/common/datadog/api_key.rs
webern: > GPT Sol: Fixed the documentation to state that creating the shared `MetaString` allocates once and subsequent clones do not allocate. https://github.com/DataDog/saluki/commit/2e6bc20c22

lib/saluki-components/src/common/datadog/api_key.rs
webern: Diff looks fine to me. - meatbag

> GPT Sol: Fixed by collapsing missing, blank, and header-invalid updates into `InvalidKey` and logging the specific reason where each update is validated. https://github.com/DataDog/saluki/commit/2e6bc20c22

lib/saluki-components/src/common/datadog/api_key.rs
webern: I think I understand this now. `ApiKeyView` exists only to erase the difference between:

- the main Datadog api_key: `Live<String>`
- the MRF key: `Live<Option<String>>`

`Option<Live<String>>` would express whether a forwarder has an API-key source at all (e.g. Cluster Autoscaling Failover), but it cannot help with this distinction of the underlying type.

Both source-schema fields are strings defaulting to `""`. But the translator leaves the primary `api_key` as a `String`:

```rust
fn consume_api_key(&mut self, value: String) {
    self.config.shared.endpoints.api_key = value;
}
```

And normalizes a blank MRF key to an optional value:

```rust
fn consume_multi_region_failover_api_key(&mut self, value: String) {
    self.config.domains.multi_region_failover.api_key = non_empty_trimmed(value);
}
```

**Conclusion**

We could change `SalukiConfiguration` so the MRF key is also a `String` and use `""` to mean unset. But it would make the model ever so slightly less expressive of reality: the main `api_key` is required, while an MRF key is optional. I'd vote to keep the small type-coalescing enum as-is, but I can go either way.

tobz (approved):

PR #2518, opened 2026-09-01
feat(otlp): added remaining e2e metrics tests 

### LINKED ISSUE #2006: Expand correctness test harness to exercise OTLP metrics features.
Expand the correctness test framework so it can exercise more of the OTLP metrics behavior supported by the Datadog Agent and reveal parity gaps in ADP.

The items below come from an initial AI source review and write-up and should be validated as part of the work.

AI report:

## Candidate coverage gaps

Millstone currently uses the pinned `lading_payload` OTLP metrics generator. That generator appears to emit only Gauges and Sums.

Extend the generated corpus to cover:

- [x] Histograms
- [x] Exponential Histograms
- [x] Summaries

Exercise the temporalities supported by the Agent for each metric type:

- [x] Delta Histograms
- [x] Cumulative Histograms
- [x] Delta Exponential Histograms
- [x] Confirm that cumulative Exponential Histograms should remain unsupported, matching the Agent
- [x] Summary behavior, which does not have an aggregation temporality

 Add corpus inputs that reliably exercise code paths selected by specific OTLP semantic values:

 - [x] Known runtime metric names and associated datapoint attributes
 - [x] The `host.name` resource attribute
 - [x] Instrumentation scope metadata
 - [x] Resource attributes used for ECS Fargate detection

Add cases for:

- [x] Histogram aggregations
- [x] Histogram bucket counters
- [x] Histogram-to-sketch conversion
- [x] Runtime metric remapping
- [x] Per-resource host attribution

Determine whether the comparison representation also needs to include OTLP origin metadata. `stele` currently compares metric names, tags, hosts, types, and values, but does not appear to compare the origin metadata carried by metric payloads.

## Implementation options

The missing input shapes could be added to the upstream Lading OTLP generator, or Millstone could generate or augment deterministic OTLP payloads using the in-repo `otlp-protos` crate. Choose the approach after confirming the required coverage and expected ownership.

## Initial tests

- `otlp-metrics-histograms`: cover delta and cumulative Histograms, histogram aggregations, bucket counters, and histogram-to-sketch conversion.
- `otlp-metrics-host-attribution`: send metrics with a deterministic `host.name` resource attribute and compare the resulting series host. Source inspection suggests ADP may not preserve this per-resource host, but the test should establish whether there is an observable parity gap.

## Follow-up investigation

Evaluate additional cases for:

- Delta Exponential Histograms
- Summaries
- Runtime metric remapping
- ECS Fargate resource detection
- Origin metadata

Open implementation issues for any ADP parity gaps exposed by these tests.

lucastemb:
## Summary

This aims to add the remaining correctness/ e2e test coverage for more niche OTLP metrics ingest features. Base64 encodings were generated to supply payloads for the test cases since they're _mostly_ static in nature, so adding this to lading, which randomly and deterministically generates sound output, would likely **not** be the best fit.

These correctness tests exist to verify behavior but also partially to detect any regressions down the line. 

Additionally, we do guard certain remapping functionality behind a boolean since the Core Agent seems to now gate runtime remapping behind a feature flag. If any issue arise, we can simply flip that boolean to false. It should be noted that customers cannot externally configure that value for the OTEL Collector / Core Agent so this design seems the most appropriate. 

 
## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Correctness test

## References

- Closes: #2006


test/correctness/cases/otlp-metrics-cumulative-histogram/millstone.yaml
```diff
@@ -0,0 +1,8 @@
+seed: [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131]
+target: "grpc://$GROUP:54317/opentelemetry.proto.collector.metrics.v1.MetricsService/Export"
+aggregation_bucket_width_secs: 10
+volume: 1
+corpus:
+  size: 1
+  payload:
+    static_base64: "Cp0BCh0KGwoMc2VydmljZS5uYW1lEgsKCW90bHAtdGVzdBJ8EnoKGW90bHAuY3VtdWxhdGl2ZS5oaXN0b2dyYW1KXQpZGQDKmjsAAAAAIQMAAAAAAAAAKQAAAAAAABhAMhgBAAAAAAAAAAEAAAAAAAAAAQAAAAAAAAA6EAAAAAAAAPA/AAAAAAAAAEBZAAAAAAAA8D9hAAAAAAAACEAQAg=="
```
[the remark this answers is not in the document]
lucastemb: Fixed by [336a47c](https://github.com/DataDog/saluki/pull/2518/commits/336a47c09ddef7858a14ab7c5aa29f17a0b780ee)

test/correctness/cases/otlp-metrics-as-type-rate/millstone.yaml
```diff
@@ -0,0 +1,8 @@
+seed: [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131]
+target: "grpc://$GROUP:54317/opentelemetry.proto.collector.metrics.v1.MetricsService/Export"
+aggregation_bucket_width_secs: 10
+volume: 1
+corpus:
+  size: 1
+  payload:
+    static_base64: "Cl8KHQobCgxzZXJ2aWNlLm5hbWUSCwoJb3RscC10ZXN0Ej4SPAoNb3RscC5yYXRlLnN1bTorCiUZAMqaOwAAAAAhAAAAAAAAFEA6EQoHYXNfdHlwZRIGCgRyYXRlEAEYAQ=="
```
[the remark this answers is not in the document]
lucastemb: Fixed by [336a47c](https://github.com/DataDog/saluki/pull/2518/commits/336a47c09ddef7858a14ab7c5aa29f17a0b780ee)

test/correctness/cases/otlp-metrics-cumulative-histogram/millstone.yaml
```diff
@@ -0,0 +1,8 @@
+seed: [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131]
+target: "grpc://$GROUP:54317/opentelemetry.proto.collector.metrics.v1.MetricsService/Export"
+aggregation_bucket_width_secs: 10
+volume: 1
+corpus:
+  size: 1
+  payload:
+    static_base64: "CvsBCh0KGwoMc2VydmljZS5uYW1lEgsKCW90bHAtdGVzdBLZARLWAQoZb3RscC5jdW11bGF0aXZlLmhpc3RvZ3JhbUq4AQpZGQDKmjsAAAAAIQMAAAAAAAAAKQAAAAAAABhAMhgBAAAAAAAAAAEAAAAAAAAAAQAAAAAAAAA6EAAAAAAAAPA/AAAAAAAAAEBZAAAAAAAA8D9hAAAAAAAACEAKWRkArqaPAgAAACEHAAAAAAAAACkAAAAAAAAuQDIYAgAAAAAAAAADAAAAAAAAAAIAAAAAAAAAOhAAAAAAAADwPwAAAAAAAABAWQAAAAAAAOA/YQAAAAAAABBAEAI="
```
webern: **GPT Sol Finding**

The two points are sufficient to exercise cumulative differencing, but their summaries and buckets are internally inconsistent.

The first point can represent samples `[1, 2, 3]`: count 3, sum 6, buckets `[1, 1, 1]`, min 1, max 3. The second point implies four additional samples because its cumulative buckets change from `[1, 1, 1]` to `[2, 3, 2]`. Since the cumulative minimum also changes to 0.5, the new first-bucket sample must be 0.5. The remaining samples are bounded by 2, 2, and 4, so their sum cannot produce the encoded delta of 9.

A coherent second cumulative point would have sum 14, representing four added samples such as `[0.5, 1.5, 2, 4]`. I also suggest setting the same nonzero start time on both cumulative points so this exercises the normal known-start path.

Replace the payload with:

```yaml
static_base64: "Co0CCh0KGwoMc2VydmljZS5uYW1lEgsKCW90bHAtdGVzdBLrARLoAQoZb3RscC5jdW11bGF0aXZlLmhpc3RvZ3JhbUrKAQpiEQDh9QUAAAAAGQDKmjsAAAAAIQMAAAAAAAAAKQAAAAAAABhAMhgBAAAAAAAAAAEAAAAAAAAAAQAAAAAAAAA6EAAAAAAAAPA/AAAAAAAAAEBZAAAAAAAA8D9hAAAAAAAACEAKYhEA4fUFAAAAABkArqaPAgAAACEHAAAAAAAAACkAAAAAAAAsQDIYAgAAAAAAAAADAAAAAAAAAAIAAAAAAAAAOhAAAAAAAADwPwAAAAAAAABAWQAAAAAAAOA/YQAAAAAAABBAEAI="
```

test/correctness/cases/otlp-metrics-exponential-histogram/millstone.yaml
```diff
@@ -0,0 +1,8 @@
+seed: [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131]
+target: "grpc://$GROUP:54317/opentelemetry.proto.collector.metrics.v1.MetricsService/Export"
+aggregation_bucket_width_secs: 10
+volume: 1
+corpus:
+  size: 1
+  payload:
+    static_base64: "CncKHQobCgxzZXJ2aWNlLm5hbWUSCwoJb3RscC10ZXN0ElYSVAoab3RscC5leHBvbmVudGlhbC5oaXN0b2dyYW1SNgoyGQDKmjsAAAAAIQMAAAAAAAAAKQAAAAAAABhAQgMSAQNhAAAAAAAA8D9pAAAAAAAACEAQAQ=="
```
webern: **GPT Sol Finding**

This payload is both internally inconsistent and too narrow to meaningfully exercise exponential-histogram conversion.

At scale 0, positive bucket index 0 represents values in `(1, 2]`. The payload places all three observations in that one bucket while reporting min 1 and max 3, neither of which is compatible with those bucket bounds. It also leaves the scale, offset, negative-store, zero-count, and multi-bin conversion paths effectively untested.

I suggest using two coherent datapoints at scale 1: one positive scenario with multiple occupied bins and a zero, and one mixed scenario with positive, negative, zero, and an empty interior bin. The mixed scenario intentionally omits `sum`, as required for a histogram containing negative observations.

Replace the payload with:

```yaml
static_base64: "CoQCCh0KGwoMc2VydmljZS5uYW1lEgsKCW90bHAtdGVzdBLiARLfAQoab3RscC5leHBvbmVudGlhbC5oaXN0b2dyYW1SwAEKYAoWCghzY2VuYXJpbxIKCghwb3NpdGl2ZREA4fUFAAAAABkAypo7AAAAACEEAAAAAAAAACkAAAAAAAAWQDACOQEAAAAAAAAAQgUSAwEBAWEAAAAAAAAAAGkAAAAAAAAEQApaChMKCHNjZW5hcmlvEgcKBW1peGVkEQDh9QUAAAAAGQDKmjsAAAAAIQUAAAAAAAAAMAI5AQAAAAAAAABCBRIDAQABSgQSAgEBYQAAAAAAAPy/aQAAAAAAAARAEAE="
```

test/correctness/cases/otlp-metrics-histogram-distributions/millstone.yaml
```diff
@@ -0,0 +1,8 @@
+seed: [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131]
+target: "grpc://$GROUP:54317/opentelemetry.proto.collector.metrics.v1.MetricsService/Export"
+aggregation_bucket_width_secs: 10
+volume: 1
+corpus:
+  size: 1
+  payload:
+    static_base64: "Cp8BCh0KGwoMc2VydmljZS5uYW1lEgsKCW90bHAtdGVzdBJ+EnwKG290bHAuZGlzdHJpYnV0aW9uLmhpc3RvZ3JhbUpdClkZAMqaOwAAAAAhBAAAAAAAAAApAAAAAAAAJEAyGAEAAAAAAAAAAgAAAAAAAAABAAAAAAAAADoQAAAAAAAA8D8AAAAAAAAEQFkAAAAAAADgP2EAAAAAAAAIQBAB"
```
webern: **GPT Sol Finding**

This input reaches the distribution conversion path, but its histogram summary cannot describe its buckets.

With bounds `[1, 2.5]`, bucket counts `[1, 2, 1]`, min 0.5, and max 3, the maximum possible sum is `0.5 + 2.5 + 2.5 + 3 = 8.5`. The encoded sum of 10 is therefore impossible.

A coherent sample set such as `[0.5, 1.5, 2, 3]` preserves the existing count, buckets, min, and max while producing a sum of 7. I also suggest providing a nonzero start time for the delta point.

Replace the payload with:

```yaml
static_base64: "CqoBCh0KGwoMc2VydmljZS5uYW1lEgsKCW90bHAtdGVzdBKIARKFAQobb3RscC5kaXN0cmlidXRpb24uaGlzdG9ncmFtSmYKYhEA4fUFAAAAABkAypo7AAAAACEEAAAAAAAAACkAAAAAAAAcQDIYAQAAAAAAAAACAAAAAAAAAAEAAAAAAAAAOhAAAAAAAADwPwAAAAAAAARAWQAAAAAAAOA/YQAAAAAAAAhAEAE="
```

test/correctness/cases/otlp-metrics-azure-host/millstone.yaml
```diff
@@ -0,0 +1,8 @@
+seed: [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131]
+target: "grpc://$GROUP:54317/opentelemetry.proto.collector.metrics.v1.MetricsService/Export"
+aggregation_bucket_width_secs: 10
+volume: 1
+corpus:
+  size: 1
+  payload:
+    static_base64: "CqgBCnkKGwoMc2VydmljZS5uYW1lEgsKCW90bHAtdGVzdAoZCg5jbG91ZC5wcm92aWRlchIHCgVhenVyZQoZCglob3N0Lm5hbWUSDAoKYWtzLW5vZGUtMQokChBjbG91ZC5hY2NvdW50LmlkEhAKDm15LWFrcy1jbHVzdGVyEisSKQoRb3RscC5henVyZS5tZXRyaWMqFAoSGQDKmjsAAAAAIQAAAAAAAPA/"
```
webern: I would say this one can be removed. Per the analysis below, deleting the code it is meant to check would still result in the test passing. i.e. this is no different than checking `host.name=aks-node-1`.

**GPT Sol Finding**

This payload does not exercise behavior that is specific to Azure.

Given `cloud.provider=azure` and only `host.name=aks-node-1`, the Azure resolver selects `aks-node-1`. If the Azure-specific branch were removed, however, the generic `host.name` fallback would select the same value. `cloud.account.id=my-aks-cluster` does not participate in Azure hostname resolution either.

This consequently cannot catch a regression in the Azure-specific logic, and the generic resource-host behavior is already covered by `otlp-metrics-host`, including existing Azure resource cases.

**Suggested change:** remove the `otlp-metrics-azure-host` correctness case. If there is a distinct Azure contract this is intended to protect, replace the payload with one whose expected output differs from generic `host.id`/`host.name` fallback.

test/correctness/cases/otlp-metrics-runtime-remapping/millstone.yaml
```diff
@@ -0,0 +1,25 @@
...
+    #    - value=5                     -> runtime.go.num_goroutine
+    #
+    # 3. process.runtime.dotnet.gc.heap.size (Histogram, delta) with two datapoints:
+    #    - generation=gen0, count=2    -> runtime.dotnet.gc.size.gen0 (generation stripped)
+    #    - generation=gen1, count=1    -> runtime.dotnet.gc.size.gen1 (generation stripped)
+    #
+    # Generated by the `generate_runtime_remapping_payload` ignored test in
+    # lib/saluki-components/src/sources/otlp/metrics/translator.rs.
+    static_base64: "CtEDCh0KGwoMc2VydmljZS5uYW1lEgsKCW90bHAtdGVzdBKvAxKqAQogcHJvY2Vzcy5ydW50aW1lLmp2bS5tZW1vcnkudXNhZ2UqhQEKIhkAypo7AAAAACEAAAAAAABZQDoOCgR0eXBlEgYKBGhlYXAKJhkAypo7AAAAACEAAAAAAABJQDoSCgR0eXBlEgoKCG5vbl9oZWFwCjcZAMqaOwAAAAAhAAAAAAAAaUA6DgoEdHlwZRIGCgRoZWFwOhMKBnJlZ2lvbhIJCgd1cy1lYXN0EjUKHXByb2Nlc3MucnVudGltZS5nby5nb3JvdXRpbmVzKhQKEhkAypo7AAAAADEFAAAAAAAAABLIAQojcHJvY2Vzcy5ydW50aW1lLmRvdG5ldC5nYy5oZWFwLnNpemVKoAEKTRkAypo7AAAAACECAAAAAAAAACkAAAAAAAAYQDIQAQAAAAAAAAABAAAAAAAAADoIAAAAAAAAJEBKFAoKZ2VuZXJhdGlvbhIGCgRnZW4wCk0ZAMqaOwAAAAAhAQAAAAAAAAApAAAAAAAACEAyEAEAAAAAAAAAAAAAAAAAAAA6CAAAAAAAACRAShQKCmdlbmVyYXRpb24SBgoEZ2VuMRAB"
```
webern: It's telling us to remove this test, but... it says

> language-beacon emission, and those `datadog.*` metrics are filtered by the correctness analyzer.

Which makes me wonder if we could add correctness analyzer assertions for these language-beacon emissions?

**GPT Sol Finding**

This test does not exercise runtime remapping because this PR sets `RUNTIME_REMAPPING_ENABLED` to `false`. The JVM, Go, and .NET metrics therefore pass through the ordinary gauge and histogram paths under their original names.

The payload comments claim that these inputs produce names such as `jvm.heap_memory`, `runtime.go.num_goroutine`, and `runtime.dotnet.gc.size.gen0`, but none of those mappings execute. The only remaining runtime-specific behavior is language-beacon emission, and those `datadog.*` metrics are filtered by the correctness analyzer.

Removing the runtime mapping tables and their attribute-selection logic would leave the compared output unchanged, so this cannot guard the remapping algorithm against regressions.

**Suggested change:** remove `otlp-metrics-runtime-remapping` until remapping can be enabled on both targets. If the intended contract is instead that runtime metrics remain unmodified, rename the case to `otlp-metrics-runtime-passthrough` and rewrite the payload comments to describe that negative behavior.

webern (approved):
Had to lean on the robots here. Overall it looks like a couple could be removed as they cannot actually catch a regression, and a few could be modified to be stronger. See what you think.

lucastemb:
> Had to lean on the robots here. Overall it looks like a couple could be removed as they cannot actually catch a regression, and a few could be modified to be stronger. See what you think.

@webern A lot of the comments were valid. I realized now, on second glance, for some reason, the agent assumed that there was no testing coverage of histograms or exponential histograms in lading _which just isn't true - especially considering how much I had to wrestle with it at the beginning to diagnose failures_ (🙄). 

I figured it would be best to drop all the correctness tests you commented on.

PR #2522, opened 2026-09-02
chore(config): app memory settings from typed config

webern:
## Human Summary

Found a couple of untracked saluki only configuration keys and added them to the inventory and documentation. Uneventful. LGTM.

## AI Summary

- Build memory bounds from typed control settings.
- Preserve cgroup limit detection, slop validation, limiter defaults, and validation modes.
- Add typed Saluki-only settings for memory mode and the global limiter.
- Remove the unused raw configuration input from app bootstrap.

## Change Type

- [x] Non-functional (chore, refactoring, docs)

## How did you test this PR?

- `make fmt`
- `make check-docs`
- `make check-clippy`
- `cargo check --workspace --tests`
- Targeted `cargo nextest` runs for affected packages

## References

- Progresses: #2169

PR #2521, opened 2026-09-02
chore(config): consume typed logging settings

webern:
## AI Summary

- Build logging settings from the typed control configuration.
- Drive runtime log-level changes through a narrow live value.
- Preserve platform log-file and rotation defaults with configuration provenance.
- Remove migrated raw parsing and update config inventory annotations.

## Change Type

- [x] Non-functional (chore, refactoring, docs)

## How did you test this PR?

- `make fmt`
- `make check-docs`
- `cargo check --workspace --tests`
- Targeted `cargo nextest` runs for affected packages and logging tests

## References

- Progresses: #2169

PR #2520, opened 2026-09-02
chore(config): use typed bootstrap settings

webern:
## Human Summary

LGTM

## Summary

- Read DogStatsD replay socket and pipe settings from typed configuration.
- Seed the internal metrics level with its typed `info` default before bootstrap.
- Remove the migrated raw configuration reads.

## Change Type

- [x] Non-functional (chore, refactoring, docs)

## How did you test this PR?

- `make fmt`
- `make check-docs`
- `make check-clippy`
- Targeted `cargo nextest` runs for `agent-data-plane`, `agent-data-plane-config`, and `agent-data-plane-config-system`

## References

- Progresses: #2169

PR #2523, opened 2026-09-02
chore(config): type environment detection inputs

webern:
## Human Summary

Larger diff but ultimately there is no real need for saluki-env to read what it needs from the configuration map. It can take the few, scalar parameters it needs as arguments when constructing types.

## AI Summary

- Model host identity, container roots, and containerd settings in typed configuration.
- Preserve source awareness so schema defaults do not suppress runtime discovery.
- Pass concrete paths and timeouts into environment providers.
- Reject negative CRI timeouts and remove migrated raw config reads from saluki-env.

## Change Type

## How did you test this PR?

- `make fmt`
- `make check-docs`
- Targeted `cargo nextest` and clippy runs for affected packages

## References

- Progresses: #2169

### DIFF
--- a/lib/agent-data-plane-config/src/shared.rs
+++ b/lib/agent-data-plane-config/src/shared.rs
@@ -55,7 +55,13 @@ pub struct SharedConfiguration {
 /// Host identity and container runtime discovery inputs.
 #[derive(Clone, Debug, Default, PartialEq, Serialize)]
 pub struct Environment {
-    /// Hostname override. A defaulted empty value is treated as absent.
+    /// Hostname reported for all emitted data.
+    ///
+    /// Only read in standalone mode, where it is reported verbatim. In connected mode the hostname comes from the
+    /// Datadog Agent and this value is ignored.
+    ///
+    /// Defaults to empty, and a defaulted or empty value is treated as absent. Operators running standalone must set
+    /// it explicitly; startup fails otherwise.
     pub hostname: ConfigValue<String>,
 
     /// containerd runtime discovery and client timeouts.
--- a/lib/agent-data-plane-config/src/shared.rs
+++ b/lib/agent-data-plane-config/src/shared.rs
@@ -55,7 +55,7 @@ pub struct SharedConfiguration {
 /// Host identity and container runtime discovery inputs.
 #[derive(Clone, Debug, Default, PartialEq, Serialize)]
 pub struct Environment {
-    /// Hostname override, with provenance to distinguish the schema default.
+    /// Hostname override. A defaulted empty value is treated as absent.
     pub hostname: ConfigValue<String>,
 
     /// containerd runtime discovery and client timeouts.
@@ -68,23 +68,23 @@ pub struct Environment {
 /// containerd runtime discovery and client timeouts.
 #[derive(Clone, Debug, Default, PartialEq, Serialize)]
 pub struct Containerd {
-    /// containerd gRPC socket, with provenance to control path detection.
+    /// containerd gRPC socket. A defaulted empty value enables path probing.
     pub socket_path: ConfigValue<PathBuf>,
 
-    /// Timeout for establishing a containerd gRPC connection.
+    /// Timeout for establishing a containerd gRPC connection. Defaults to 1 second; `0` never connects.
     pub connection_timeout: Duration,
 
-    /// Per-RPC timeout for containerd API calls.
+    /// Per-RPC timeout for containerd API calls. Defaults to 5 seconds; `0` fails every call.
     pub query_timeout: Duration,
 }
 
 /// Filesystem roots describing container workloads.
 #[derive(Clone, Debug, Default, PartialEq, Serialize)]
 pub struct ContainerRoots {
-    /// procfs root, with provenance for host-mount detection.
+    /// procfs root. Defaults to `/host/proc`, which is only used when set explicitly.
     pub proc_root: ConfigValue<PathBuf>,
 
-    /// cgroupfs root, with provenance for host-mount detection.
+    /// cgroupfs root. Defaults to `/host/sys/fs/cgroup/`, which is only used when set explicitly.
     pub cgroup_root: ConfigValue<PathBuf>,
 }
 
--- a/lib/agent-data-plane-config/src/shared.rs
+++ b/lib/agent-data-plane-config/src/shared.rs
@@ -37,6 +37,9 @@ pub struct SharedConfiguration {
     /// Secrets management, read by the Datadog intake forwarders.
     pub secrets: Secrets,
 
+    /// Host and container runtime discovery, read by the environment providers.
+    pub environment: Environment,
+
     /// Verbosity of the internal telemetry emitted about the runtime itself. (not in Datadog Agent
     /// config schema)
     pub metrics_level: String,
@@ -47,6 +50,42 @@ pub struct SharedConfiguration {
     /// DogStatsD context dumps. Defaults to unset when configuration does not provide a concrete
     /// `run_path`.
     pub run_path: Option<PathBuf>,
+}
+
+/// Host identity and container runtime discovery inputs.
+#[derive(Clone, Debug, Default, PartialEq, Serialize)]
+pub struct Environment {
+    /// Hostname override, with provenance to distinguish the schema default.
+    pub hostname: ConfigValue<String>,
+
+    /// containerd runtime discovery and client timeouts.
+    pub containerd: Containerd,
+
+    /// Filesystem roots describing container workloads.
+    pub container_roots: ContainerRoots,
+}
+
+/// containerd runtime discovery and client timeouts.
+#[derive(Clone, Debug, Default, PartialEq, Serialize)]
+pub struct Containerd {
+    /// containerd gRPC socket, with provenance to control path detection.
+    pub socket_path: ConfigValue<PathBuf>,
+
+    /// Timeout for establishing a containerd gRPC connection.
+    pub connection_timeout: Duration,
+
+    /// Per-RPC timeout for containerd API calls.
+    pub query_timeout: Duration,
+}
+
+/// Filesystem roots describing container workloads.
+#[derive(Clone, Debug, Default, PartialEq, Serialize)]
+pub struct ContainerRoots {
+    /// procfs root, with provenance for host-mount detection.
+    pub proc_root: ConfigValue<PathBuf>,
+
+    /// cgroupfs root, with provenance for host-mount detection.
+    pub cgroup_root: ConfigValue<PathBuf>,
 }
 
 /// Inputs used to derive deployment-wide static tags.
--- a/lib/datadog-agent/config-testing/src/smoke_test.rs
+++ b/lib/datadog-agent/config-testing/src/smoke_test.rs
@@ -313,12 +313,12 @@ mod tests {
 
     #[tokio::test]
     async fn flags_a_supported_key_that_never_changes_the_struct() {
-        // `CONTAINERD_CONFIGURATION` has registered (supported) keys. A factory that ignores the config
+        // `TYPED_CONFIG_SYSTEM` has registered (supported) keys. A factory that ignores the config
         // entirely means each supported key "produces the default struct", violating the supported-key
         // guarantee, which the harness must flag.
         let outcome = tokio::spawn(async {
             run_config_smoke_tests(
-                structs::CONTAINERD_CONFIGURATION,
+                structs::TYPED_CONFIG_SYSTEM,
                 &[],
                 json!({}),
                 |_cfg: GenericConfiguration| json!({}),
--- a/lib/datadog-agent/config-testing/src/smoke_test.rs
+++ b/lib/datadog-agent/config-testing/src/smoke_test.rs
@@ -313,9 +313,10 @@ mod tests {
 
     #[tokio::test]
     async fn flags_a_supported_key_that_never_changes_the_struct() {
-        // `TYPED_CONFIG_SYSTEM` has registered (supported) keys. A factory that ignores the config
-        // entirely means each supported key "produces the default struct", violating the supported-key
-        // guarantee, which the harness must flag.
+        // This case needs a consumer with registered (supported) keys. No struct-deserializing consumer has
+        // any left, so it uses the `TYPED_CONFIG_SYSTEM` sentinel. A factory that ignores the config entirely
+        // means each supported key "produces the default struct", violating the supported-key guarantee, which
+        // the harness must flag.
         let outcome = tokio::spawn(async {
             run_config_smoke_tests(
                 structs::TYPED_CONFIG_SYSTEM,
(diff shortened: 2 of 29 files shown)


webern:
* **#2523** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2523?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2523?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2512** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2512?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lib/datadog-agent/config-testing/src/smoke_test.rs
```diff
@@ -313,12 +313,12 @@ mod tests {
 
     #[tokio::test]
     async fn flags_a_supported_key_that_never_changes_the_struct() {
-        // `CONTAINERD_CONFIGURATION` has registered (supported) keys. A factory that ignores the config
+        // `TYPED_CONFIG_SYSTEM` has registered (supported) keys. A factory that ignores the config
```
webern: This looks weird. Don't we normally delete these? Does this work?

lib/saluki-env/src/features/containerd.rs
```diff
@@ -16,33 +15,21 @@ const DEFAULT_CONTAINERD_SOCKET_PATH_LINUX: &str = "/var/run/containerd/containe
...
 
 impl ContainerdDetector {
-    /// Tries to detect the containerd gRPC socket path.
-    ///
-    /// The socket path can be specified in the configuration, or if it's not present, default paths will be checked for
-    /// the presence of the socket path.
-    ///
-    /// If the socket path is configured or detected, and is a valid Unix domain socket, `Some` is returned with the
-    /// socket path. Otherwise, `None` is returned.
+    /// Returns a non-empty containerd socket override or probes well-known Unix sockets.
```
webern: This should explain the input parameter.

I think we can drop the `configured_` prefix for an optional parameter here and elsewhere. Instead the function documentation should just say it will use the path if given otherwise it will use well known paths.

lib/saluki-env/src/features/detector.rs
```diff
@@ -18,16 +19,16 @@ pub struct FeatureDetector {
 }
 
 impl FeatureDetector {
-    /// Creates a new `FeatureDetector` that checks for all possible features.
-    pub fn automatic(config: &GenericConfiguration) -> Self {
-        Self::for_feature(config, Feature::all_bits())
+    /// Detects all features, probing containerd when no socket is configured.
+    pub fn automatic(configured_containerd_socket_path: Option<PathBuf>) -> Self {
```
webern: I think we can drop the `configured_` prefix for an optional parameter here and elsewhere. Instead this is just an optional parameter and explain it as such.

lib/saluki-env/src/host/providers/fixed.rs
```diff
@@ -12,17 +11,9 @@ pub struct FixedHostProvider {
 }
 
 impl FixedHostProvider {
```
webern, line 13: Interesting. Should the whole `FixedHostProvider` abstraction go away? Or is it doing something useful still?

lib/saluki-env/src/workload/helpers/containerd/mod.rs
```diff
@@ -99,14 +67,16 @@ pub struct ContainerdClient {
...
     /// ## Errors
     ///
-    /// If the containerd socket path wasn't present in the configuration or couldn't be detected, or if the gRPC
-    /// transport to containerd couldn't be created, an error will be returned.
-    pub async fn from_configuration(config: &GenericConfiguration) -> Result<Self, GenericError> {
-        let socket_path = ContainerdDetector::detect_grpc_socket_path(config)
+    /// If the containerd socket path wasn't configured or couldn't be detected, or if the gRPC transport to containerd
+    /// couldn't be created, an error will be returned.
+    pub async fn new(
+        configured_socket_path: Option<PathBuf>, containerd_config: &ContainerdConfiguration,
```
webern: drop `configured_` and explain the None and Some behavior as elsewhere.

lib/datadog-agent/config-testing/src/smoke_test.rs
webern: > GPT Sol: This is the smoke-test harness's own meta-test, not a component smoke test. The real containerd invocation and `ContainerdConfiguration` consumer entry were deleted. This test deliberately supplies a broken factory and verifies that the harness detects a supported key which never changes its consumer. It now uses `TYPED_CONFIG_SYSTEM` because no struct-deserializing consumers remain in the registry; the test passes. I clarified the comment in [797e439fe4](https://github.com/DataDog/saluki/commit/797e439fe4).
>
> The broader harness now has no production callers and may be removable, but deleting the harness and its registry metadata is larger than this migration PR.

lib/saluki-env/src/features/containerd.rs
webern: > GPT Sol: Done. Dropped `configured_` and documented that a supplied socket path is used directly while `None` probes the well-known paths. Applied consistently across the detector, feature detector, client, and collector APIs in [797e439fe4](https://github.com/DataDog/saluki/commit/797e439fe4).

lib/saluki-env/src/workload/helpers/cgroups.rs
```diff
@@ -43,45 +42,36 @@ pub struct CgroupsConfiguration {
...
-                    PathBuf::from(DEFAULT_HOST_MAPPED_CGROUPFS_ROOT)
-                } else {
-                    // TODO: Consider if we need to do anything specific for Amazon Linux [1] or does the referenced code only
-                    // matter for cgroups v1?
-                    //
-                    // [1]: https://github.com/DataDog/datadog-agent/blob/fe75b815c2f135f0d2ea85d7a57a8fc8cbf56bd9/pkg/config/setup/config.go#L1172-L1173
-                    PathBuf::from(DEFAULT_CGROUPFS_ROOT)
-                }
+        let cgroupfs_root = cgroupfs_root.unwrap_or_else(|| {
+            if host_mapped_procfs {
```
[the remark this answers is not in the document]
webern: > GPT Sol: The mount layout described is genuinely mishandled, but this PR did not introduce it. On the parent branch, `CgroupsConfiguration::from_configuration` selected both roots using `HostMappedProcfs`; this migration moved that behavior verbatim. `HostMappedCgroupfs` is detected but currently has no consumer anywhere in the workspace.
>
> I did not change the selection logic here because this cutover is intended to preserve behavior, and changing the feature bit needs dedicated coverage for cgroup-only host mounts. I did document the existing coupling in [797e439fe4](https://github.com/DataDog/saluki/commit/797e439fe4). The behavioral correction is warranted as a separate follow-up.

lib/agent-data-plane-config/src/shared.rs
```diff
@@ -49,6 +52,42 @@ pub struct SharedConfiguration {
     pub run_path: Option<PathBuf>,
 }
 
+/// Host identity and container runtime discovery inputs.
+#[derive(Clone, Debug, Default, PartialEq, Serialize)]
+pub struct Environment {
+    /// Hostname override. A defaulted empty value is treated as absent.
+    pub hostname: ConfigValue<String>,
```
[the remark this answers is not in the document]
webern: > GPT Sol: Documented that this field is consumed only in standalone mode, that connected mode gets the hostname from the Agent, and that a defaulted or empty standalone hostname fails startup. Fixed in [797e439fe4](https://github.com/DataDog/saluki/commit/797e439fe4).

lib/saluki-env/src/features/detector.rs
webern: > GPT Sol: Done. Renamed the optional parameter in `automatic`, `for_feature`, and the private detection path, and documented its supplied-versus-absent behavior consistently. Fixed in [797e439fe4](https://github.com/DataDog/saluki/commit/797e439fe4).

lib/saluki-env/src/host/providers/fixed.rs
webern: > GPT Sol: It still has one concrete purpose: standalone mode has no Agent connection, so it cannot use `RemoteAgentHostProvider`, but the environment still needs a `HostProvider` implementation. `FixedHostProvider` supplies the explicitly configured hostname through that common interface. The raw-map constructor was the migration residue and is gone; the remaining implementation is configuration-independent. Removing it would mean removing or redesigning standalone mode, which the antithesis harness still exercises, so I left it in place.

lib/saluki-env/src/workload/helpers/containerd/mod.rs
webern: > GPT Sol: Done. Dropped the prefix in `ContainerdClient::new` and its collector pass-through, documented the `Some`/`None` behavior, and named the resolved local `detected_socket_path` to avoid shadowing. Fixed in [797e439fe4](https://github.com/DataDog/saluki/commit/797e439fe4).

lib/datadog-agent/config/schema/schema_overlay.yaml
```diff
@@ -575,7 +601,20 @@ inventory:
     pipelines: [cross_cutting]
     description: "CRI container runtime query timeout (s)"
     test_support:
-      used_by: [ContainerdConfiguration]
+      used_by: [TypedConfigSystem]
+      additional_attributes:
+        config_registry_filename: containerd.rs
+
+  cri_socket_path:
+    support: full
```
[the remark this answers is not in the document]
webern: > Opus: Good catch, and confirmed. `ContainerdDetector::detect_grpc_socket_path` only accepts a path containing `containerd`, so `/var/run/crio/crio.sock` leaves containerd undetected and no container runtime metadata collector starts. ADP's `Feature` enum has no CRI-O or generic CRI variant. The core Agent, by contrast, treats this key as the entry point for both integrations: `detectCriRuntimes` selects `Containerd` for a `containerd` path, `Crio` for a `crio` path, and `NonstandardCRIRuntime` otherwise, and its default probe list includes the CRI-O socket. So a user can reasonably expect a CRI-O socket to work here.
>
> Reclassified the key as `support: partial` and documented the divergence, which moves it out of the transparent-settings table into "Behavioral Differences" with a detail block. Runtime behavior is unchanged. Fixed in [4f6e6cafd4](https://github.com/DataDog/saluki/commit/4f6e6cafd499a2cd077e2b66c7e179c61237bf24).
>
> Whether ADP should actually detect and support CRI-O endpoints is a real question that needs more than a classification change, so I filed a follow-up to investigate: #2551.

aqian01 (commented):
I think there might be an annoying edge case.

Core Agent detects /cgroup/:
https://github.com/DataDog/datadog-agent/blob/5a5e02261d33e02497cff1665fdf1647615aade6/pkg/config/setup/common_settings.go#L1377-L1384

SetDefault records SourceDefault:
https://github.com/DataDog/datadog-agent/blob/5a5e02261d33e02497cff1665fdf1647615aade6/pkg/config/viperconfig/viper.go#L151-L155

Agent sends the value and source:
https://github.com/DataDog/datadog-agent/blob/5a5e02261d33e02497cff1665fdf1647615aade6/comp/core/configstream/impl/configstream.go#L291-L296

Saluki maps "default" to Provenance::Default:
https://github.com/DataDog/saluki/blob/7f8180075d9da1977314b017fb8e2b3491fe1d53/bin/agent-data-plane/src/internal/remote_agent.rs#L341-L359

is_explicit().then(...) discards the default value:
https://github.com/DataDog/saluki/blob/7f8180075d9da1977314b017fb8e2b3491fe1d53/bin/agent-data-plane/src/internal/env/workload/mod.rs#L126-L130

Saluki falls back to only the standard paths:
https://github.com/DataDog/saluki/blob/7f8180075d9da1977314b017fb8e2b3491fe1d53/lib/saluki-env/src/workload/helpers/cgroups.rs#L60-L68

Collector initialization fails if no hierarchy is found:
https://github.com/DataDog/saluki/blob/7f8180075d9da1977314b017fb8e2b3491fe1d53/lib/saluki-env/src/workload/collectors/cgroups.rs#L58-L66

webern:
> I think there might be an annoying edge case.

@aqian01 very good catch, thank you for the independent audit. As I understand it, the default is not detectable as a default in this case because of how the Agent detects it and sends it, so we need to do the detection here as well. This appears to fix it.

> Opus:
>
> On `main` the raw-map read of `container_cgroup_root` takes
> the Agent's value regardless of source, so this branch is where the
> `/cgroup` layout would have started failing. The cgroups collector error
> is fatal to startup, so the blast radius on an affected host is ADP not
> running at all.
>
> Fixed in https://github.com/DataDog/saluki/pull/2523/commits/c4497640c7294f5327a7a01ec64cc575063090cb
> by detecting the layout ourselves instead of consuming the Agent's
> default-sourced value:
>
> - New `Feature::LegacyCgroupfsRoot`, detected by looking for
>   `/cgroup/memory/memory.stat` — the same marker `common_settings.go`
>   uses, including its handling of a stat that fails for a reason other
>   than the file being absent.
> - `CgroupsConfiguration::new` now resolves the cgroupfs root as
>   explicit `container_cgroup_root`, then host-mapped
>   `/host/sys/fs/cgroup`, then legacy `/cgroup`, then `/sys/fs/cgroup`.
>   Host-mapped stays ahead of the legacy root because it implies we are
>   containerized, which is the branch the Agent settles before it gets to
>   its Amazon Linux check.
> - Unit tests for the new fallback and for that precedence, plus the
>   overlay documentation for `container_cgroup_root`.
>
> The general problem is worth naming, since it will come up again: on the
> wire, an Agent-side detection *result* and a static schema default both
> arrive as `source: default`, so we cannot distinguish them. We drop
> default-sourced values on purpose — ADP is the process that has to open
> these paths, so its own view of the filesystem has to win over a path
> the Agent resolved against its own — but the price is that any detection
> we want to honor has to be repeated on our side. Worth auditing the
> other keys where the Agent computes a default from the host rather than
> the schema.

webern:
> > I think there might be an annoying edge case.
> 
> @aqian01 very good catch, thank you for the independent audit. As I understand it, the default is not detectable as a default in this case because of how the Agent detects it and sends it, so we need to do the detection here as well. This appears to fix it.
> 
> > Opus:
> > On `main` the raw-map read of `container_cgroup_root` takes
> > the Agent's value regardless of source, so this branch is where the
> > `/cgroup` layout would have started failing. The cgroups collector error
> > is fatal to startup, so the blast radius on an affected host is ADP not
> > running at all.
> > Fixed in [c449764](https://github.com/DataDog/saluki/commit/c4497640c7294f5327a7a01ec64cc575063090cb)

A further audit yielded this https://github.com/DataDog/saluki/pull/2523/changes/23202fdc4a6ab0c543fc16cf878b3a97d085b305 to ensure it only happens when not-containerized.

aqian01 (approved):

PR #2533, opened 2026-09-02
fix(docs): resolve vale lint failures in doc comments

jszwedko:
## Summary

`make check-docs` was failing because Vale's doc-comment extraction doesn't exclude fenced code blocks from prose rules, so Rust keywords, identifiers, and code punctuation inside rustdoc examples were being flagged as spelling and quoting errors; this suppresses those rules around the affected examples and fixes a few genuine prose typos and a doc example with a missing closing code fence found along the way.

## Test plan

- [x] `make check-all` passes
- [x] `cargo test --doc` passes for all crates with edited doc examples, including a previously-broken example in `saluki-common::resource_tracking::groups` that was missing a closing code fence

### DIFF
--- a/bin/correctness/panoramic/src/dynamic_vars.rs
+++ b/bin/correctness/panoramic/src/dynamic_vars.rs
@@ -11,6 +11,7 @@
 //! In a test's `config.yaml`, add a `PANORAMIC_DYNAMIC_<KEY>` env var whose value is a shell
 //! command. Reference the resolved value anywhere in the config with `{{PANORAMIC_DYNAMIC_<KEY>}}`:
 //!
+//! <!-- vale off -->
 //! ```yaml
 //! container:
 //!   env:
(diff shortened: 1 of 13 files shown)


bin/correctness/panoramic/src/dynamic_vars.rs
```diff
@@ -11,6 +11,7 @@
 //! In a test's `config.yaml`, add a `PANORAMIC_DYNAMIC_<KEY>` env var whose value is a shell
 //! command. Reference the resolved value anywhere in the config with `{{PANORAMIC_DYNAMIC_<KEY>}}`:
 //!
+//! <!-- vale off -->
```
jszwedko, line 14: Apparently vale doesn't handle embedded codeblocks in documentation well. It tries to evaluate them as prose.

lucastemb (approved):

tobz (approved):

PR #2548, opened 2026-09-04
feat(otlp): derive peer tag key set from semantic registry

### LINKED ISSUE #2421: Traces(otlp+dd): Derive the peer tag key set from the semantic registry
## Background

Peer tags identify the remote thing a span talked to, and their hash is part of the stats aggregation key. If ADP's key set is ordered differently from the Agent's, the same span hashes differently and its stats land in a different bucket, so a cluster running both produces split aggregates. A static list also cannot pick up new peer attributes shipped remotely.

## Agent behavior

`basePeerTags` (`peer_tags.go:65-77`) walks sixteen peer concepts, collects every attribute name from `GetAttributePrecedence`, and sorts the result. `preparePeerTags` (`peer_tags.go:79-90`) appends the operator's `peer_tags`, deduplicates, and sorts again. `AgentConfig.PeerTagsCache` (`peer_tags.go:53-62`) snapshots that set with the registry's content hash, and `getPeerTagKeys` (`concentrator.go:98-108`) rebuilds the snapshot whenever the live hash differs, which is how remote semantic updates reach the concentrator.

## ADP behavior

`SpanConcentrator::new` (`span_concentrator.rs:142-149`) starts from the static `BASE_PEER_TAGS` table at `span_concentrator.rs:22`, appends configured tags it has not already seen, and never sorts. There is no registry, no content hash, and no rebuild, so the key set is fixed for the life of the process.

## Acceptance

- [ ] Peer tag keys come from the registry's precedence lists, not a static list
- [ ] The final key set is sorted and deduplicated
- [ ] A registry content-hash change rebuilds the key set
- [ ] The rebuild happens off the per-span path

## References

Depends on: #2413

lucastemb:
## Summary

Adds support for peer tag key set from the by extracting from the registry (including custom keys), deduplicating, and sorting to ensure canonical upstream ordering for proper aggregation. 

This prevents a pre-existing bug where stats are aggregated incorrectly due to different ordering in mixed Core Agent and ADP clusters.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests 
## References

- Closes: #2421


webern (commented):
This does not close the tracking issue, I updated the PR description to say "Progresses" and you should add the issue number that it actually does close there.

> Clanker: So my verdict is: the actual key derivation matches the Agent, but the claimed dynamic-update parity does not work. Lucas should either remove the refresh machinery and explicitly scope this to the embedded
 startup registry, or wire it to a genuinely live registry and refresh before span processing.

lucastemb:
> This does not close the tracking issue, I updated the PR description to say "Progresses" and you should add the issue number that it actually does close there.
> 
> > Clanker: So my verdict is: the actual key derivation matches the Agent, but the claimed dynamic-update parity does not work. Lucas should either remove the refresh machinery and explicitly scope this to the embedded
> > startup registry, or wire it to a genuinely live registry and refresh before span processing.

@webern Ha whoops. That was an oversight on my part. Fixed it. The incorrect numbering of the branch tripped me up.

webern:
> > This does not close the tracking issue, I updated the PR description to say "Progresses" and you should add the issue number that it actually does close there.
> > > Clanker: So my verdict is: the actual key derivation matches the Agent, but the claimed dynamic-update parity does not work. Lucas should either remove the refresh machinery and explicitly scope this to the embedded
> > > startup registry, or wire it to a genuinely live registry and refresh before span processing.
> 
> @webern Ha whoops. That was an oversight on my part. Fixed it. The incorrect numbering of the branch tripped me up.

@lucastemb There is some kind of finding there about this not being refreshed, i.e. it's not dynamic. Could you look into that?

lucastemb:
> > > This does not close the tracking issue, I updated the PR description to say "Progresses" and you should add the issue number that it actually does close there.
> > > > Clanker: So my verdict is: the actual key derivation matches the Agent, but the claimed dynamic-update parity does not work. Lucas should either remove the refresh machinery and explicitly scope this to the embedded
> > > > startup registry, or wire it to a genuinely live registry and refresh before span processing.
> > 
> > 
> > @webern Ha whoops. That was an oversight on my part. Fixed it. The incorrect numbering of the branch tripped me up.
> 
> @lucastemb There is some kind of finding there about this not being refreshed, i.e. it's not dynamic. Could you look into that?

@webern It took me a little bit to understand what was going on, but I added some scaffolding / infrastructure on the ADP side to be able to receive registry updates using the existing IPC infrastructure. Eventually, I'll have to open a PR against the Core Agent so that we can actually send that information here.

bin/agent-data-plane/src/internal/remote_agent.rs
```diff
@@ -344,6 +346,47 @@ const AGENT_DEFAULT_SOURCE: &str = "default";
 const AGENT_DECLARED_ONLY_SOURCE: &str = "schema";
 const AGENT_UNSET_SOURCES: [&str; 2] = [AGENT_DEFAULT_SOURCE, AGENT_DECLARED_ONLY_SOURCE];
 
+/// Reserved config-stream key for semantic registry JSON.
+const SEMANTIC_REGISTRY_SETTING_KEY: &str = "remote_configuration.apm_semantics.registry";
+
```
webern: What is this and where is it coming from? It is not part of the datadog schema as far as I can tell. My coding agent is having a really hard time figuring it out. It thinks you invented it (obviously you did not).

My concern is it looks like a rogue, un-tracked config key. If that's what it is, we need to upstream its existence to the Datadog config schema.

@ajgajg1134 are you aware of what this is?

Depending on the answer here, I can perhaps unblock with proper TODOs and tracking issues, but we can't be adding rogue untracked keys to be dug out of the config map like this as an accepted solution.

Edit: also, apologies if this is a result of my previous comment requesting the dynamic behavior and the "registry updates using the existing IPC" you added as a result. Basically we should figure out how to unblock rather than bogging this down with the problems we are sort of discovering... let's chat about it.

webern (changes requested):

bin/agent-data-plane/src/internal/remote_agent.rs
ajgajg1134: hmm yeah as far as I can tell that specific sub-key doesn't exist in the agent code? We do have `remote_configuration.apm_semantics` which controls if we use the remote apm_semantics library to do semantic mapping of span tags. (I can find more details about that if needed here?)

lucastemb: The idea here is that this was added in anticipation of a Core side Agent update where we propagate an updated registry over IPC to hot swap the reference registry in memory here on the Saluki side.

webern (approved):
Sorry for the rabbit hole on the future work to make this dynamically updatable.

PR #2550, opened 2026-09-04
fix(cgroups): resolve own container by controller inode

jszwedko:
## Summary

Agent Data Plane couldn't identify its own container when running in its own cgroup namespace — the default under control groups v2 with containerd or CRI-O. Telemetry about ADP itself, such as the liveness service check, went out untagged, and it failed silently. This resolves the container by its cgroup controller inode when the cgroup path can't name it.

## Problem

`get_self_container_id` reads `/proc/self/cgroup` and extracts a container ID from the path. With a private cgroup namespace, the process's own cgroup *is* the namespace root, so that file reads:

```
0::/
```

The line parses fine, but `Path::new("/").file_name()` is `None`, so no container ID comes out and `resolve_self_container` returns `None`. `get_self_container_tags` then short-circuits, and the liveness source emits its service check with no container tags.

## Verified on staging

Checked three ADP pods on `sasquatch.us1.staging.dog`. Every one is affected:

| Check | Result |
|---|---|
| `/proc/self/cgroup` | `0::/` on all three |
| cgroup namespace | `4026532806`, `4026532578`, `4026532772` — none is the host's `4026531835` |
| control groups version | v2 |

## Why not the Agent's mountinfo fallback

The core Agent handles this case by parsing `/proc/self/mountinfo`. That approach was the original plan here, and checking it against a real pod is what ruled it out — the only `hostname` mount on these nodes is:

```
/var-lib-containerd/io.containerd.grpc.v1.cri/sandboxes/c350804f…/hostname
```

That is the **pod sandbox** ID, not ADP's container, and the Agent deliberately rejects it (`matches[1] != containerdSandboxPrefix`). Implementing that parser would have returned nothing on exactly the deployment it was meant to fix. It's also what the Agent's own comment means by "not always reliable in Kubernetes+containerd (at least)".

## What does work

The Agent's *first* tier, `getSelfContainerIDFromInode`. Inodes are identical on both sides of a namespace boundary, so the namespace root's inode is the same one a walk of the host's hierarchy sees for that cgroup. Confirmed on a pod whose container ID is `8692402c…`:

```
$ stat -c %i /sys/fs/cgroup                                                   # inside the container
8578
$ stat -c %i /host/sys/fs/cgroup/kubepods.slice/…/cri-containerd-8692402c….scope
8578
```

ADP already builds exactly this mapping — `CgroupsMetadataCollector` registers `ContainerInode(8578) -> Container("8692402c…")` on every poll. The alias was already there; nothing was asking for it.

## What changed

`resolve_self_container` now tries two things in order:

1. **The cgroup path**, as before, yielding `EntityId::Container`.
2. **The controller inode** of `/sys/fs/cgroup`, yielding `EntityId::ContainerInode`, which the tag store resolves through the collector's alias.

The two cover disjoint cases, which is why both are kept:

| cgroup namespace | controller inode | `/proc/self/cgroup` |
|---|---|---|
| private | the container's own cgroup | `0::/`, names nothing |
| host | the hierarchy root, rejected as reserved | full path, names the container |

The inode fallback is additionally gated on the **unified** hierarchy. The alias map is keyed on a bare inode with no device number, which is only sound while every inode in it comes from one filesystem. Under cgroups v2 that holds — `/sys/fs/cgroup` is a cgroup in the same filesystem the collector walks, confirmed on staging where both it and the host-mapped view report device `32` and `cgroup2fs`. Under cgroups v1, `/sys/fs/cgroup` is the `tmpfs` the per-controller filesystems are mounted into, so an inode from it identifies nothing in the hierarchy and could collide with another container's. This PR is what first admits a foreign-filesystem inode into that keyspace, so the guard belongs here.

**Order is deliberate, and inverted relative to the Agent.** The path is self-verifying — it either names a container or it doesn't. The inode isn't: it's only useful if the collector has aliased it, and the resolver has no way to check that. Trying the inode first would risk returning an entity that resolves to nothing while masking a path lookup that would have worked. The host-namespace case is additionally protected by the reserved-inode guard, since the hierarchy root is inode 1 — verified on the same cluster — but ordering makes correctness independent of that holding on every kernel.

This also never consults the cgroup-namespace inode, so it sidesteps the caveat in the Agent's `namespace_linux.go` that host-namespace detection "does not work when running in DinD".

## Note for reviewers: the entity kind now varies

`resolve_self_container` returns a `Container` entity or a `ContainerInode` entity depending on which method succeeded. Both current callers feed it straight to `get_entity_tags`, which follows aliases, so they get the same answer either way. Documented on both the method and its wrapper. A future caller needing the container ID *string* would have to resolve the alias — worth knowing before adding one.

One behavioral consequence: the inode entity only resolves once the cgroups collector's first poll reaches the store, so the first liveness service check after startup may still go out untagged. Today it is never tagged on these hosts, so this is strictly an improvement, but it converges rather than being correct immediately.

## Test plan

- [x] Unit tests for the inode helper: a real directory yields its inode and passes the usable-inode check; a missing path yields `None`.
- [x] Reserved inodes are already covered by the existing `is_usable_controller_inode` tests, which is what rejects the hierarchy root in the host-namespace case.
- [x] `cargo test -p saluki-env` (68 tests), `cargo clippy --all-targets -p saluki-env -- -D warnings`, and `cargo check --workspace --tests` in a Linux container.
- [x] `cargo check -p saluki-env --tests` on macOS and for `x86_64-pc-windows-msvc`, since the new helper lives in a Linux-gated module but is referenced from code that is not.
- [x] `make check-release-notes` and `reno lint` clean.
- [ ] Worth confirming on staging after deploy that the liveness service check carries container tags.

Found while auditing ADP's cgroups implementation against `pkg/util/cgroups`; the last of the gaps that audit turned up.


lib/saluki-env/src/workload/helpers/cgroups.rs
```diff
@@ -622,6 +623,43 @@ pub(crate) fn get_self_container_id(interner: &GenericMapInterner) -> Option<Met
...
+/// identifies the container without needing a path that names it. Inodes are the same on both sides of a namespace
+/// boundary, so the value matches what a traversal of the host's hierarchy reports for the same cgroup.
+///
+/// Returns `None` when the process shares the host's cgroup namespace, since the namespace root is then the hierarchy
+/// root rather than any particular container, and its inode is rejected as reserved.
+///
+/// Like [`get_self_container_id`], this intentionally reads the process's own `/sys/fs/cgroup` rather than a configured
+/// cgroupfs root, which may refer to the host.
+pub(crate) fn get_self_cgroup_controller_inode() -> Option<u64> {
+    cgroup_controller_inode(Path::new(SELF_CGROUPFS_PATH))
```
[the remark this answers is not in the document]
tobz: This... seems plausible? Don't know enough, personally, to say if Autotest is 100% right here or not.

tobz (approved):

lib/saluki-env/src/workload/on_demand_pid/linux.rs
```diff
@@ -105,8 +107,37 @@ impl ResolverImpl {
...
+
+        // In our own cgroup namespace, `/proc/self/cgroup` reads `0::/` and names nothing, but the namespace root is
+        // our own cgroup. Its inode is the same one a traversal of the host's hierarchy reports for that cgroup, so
+        // handing it back lets the alias the collector registered resolve us.
+        //
+        // Only under the unified hierarchy, though. There, `/sys/fs/cgroup` is a cgroup in the same filesystem the
+        // collector walked, so the inode identifies the same object. Under cgroups v1 it's the tmpfs those controllers
+        // are mounted into, and an inode from a different filesystem means nothing to a map keyed on inode alone --
+        // at best it resolves nothing, at worst it collides with another container's.
+        if self.cgroups_reader.is_unified() {
```
webern, line 132: This sounds like a nit given, "On every deployment I can think of the two agree, so this may well be fine in practice."

> AI
>
> I think this gate may be checking a different filesystem than the one it's protecting. `is_unified()` describes the hierarchy the *collector* walks, which is `/host/sys/fs/cgroup` whenever `HostMappedCgroupfs` is detected, whereas `get_self_cgroup_controller_inode()` stats our own `/sys/fs/cgroup`. The invariant we actually need is "our own `/sys/fs/cgroup` is a cgroup2 mount", and this establishes it about a second, different mount.
>
> On every deployment I can think of the two agree, so this may well be fine in practice. The case I couldn't rule out is a sandboxed runtime with an emulated sysfs on a v2 node: our own `/sys/fs/cgroup` isn't cgroup2, the gate passes on the strength of the host hierarchy, and we hand a foreign inode into an alias map that is now fully populated with real v2 inodes — which seems like a somewhat better chance of colliding with the wrong container than the v1 case this gate was added for.
>
> Could we check our own view directly instead, using the same signal `HierarchyReader` already uses to identify a v2 mount?
>
> ```rust
> /// Returns whether the process's own cgroupfs mount is the cgroups v2 unified hierarchy.
> ///
> /// Every cgroup in the unified hierarchy exposes `cgroup.controllers`, so its presence at the root of our own mount
> /// establishes that an inode read from there belongs to the same filesystem the collector walks. Under cgroups v1
> /// that path is the `tmpfs` the controllers are mounted into, and the file isn't there.
> fn self_cgroupfs_is_unified() -> bool {
>     Path::new(SELF_CGROUPFS_PATH).join(CGROUPS_V2_CONTROLLERS_FILE).exists()
> }
> ```
>
> That would also make `get_self_cgroup_controller_inode` self-verifying rather than carrying the "callers **MUST** check `CgroupsReader::is_unified`" contract in its doc comment — which is the property the description credits the cgroup path with having, and the reason it's tried first. Happy to be told the divergence isn't worth the extra `stat`, though; `is_unified()` is a fine proxy for anything running under a normal kernel.

webern (approved):
I think the comment below is a nit and it's the worst thing Opus could find. 👍

lib/saluki-env/src/workload/on_demand_pid/linux.rs
tobz: Hmmm, yeah, I agree with this.

It's just simpler to reason about, even if the two paths end up being the same 99.999999% of the time.

lib/saluki-env/src/workload/helpers/cgroups.rs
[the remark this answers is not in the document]
jszwedko: Updated in [fix(cgroups): restrict the self inode fallback to the unified hierarchy](https://github.com/DataDog/saluki/pull/2550/changes/7cff2c99b8f68c2bec2fe9dba6826052c3d14875)

lib/saluki-env/src/workload/on_demand_pid/linux.rs
jszwedko: Mmm, yeah, good call. I updated it in [fix(cgroups): verify our own cgroupfs mount is unified](https://github.com/DataDog/saluki/pull/2550/changes/9f6057ff0951cf38b186d6586b7a1f2e423b38d0)

tobz (approved):

webern (approved):
LGTM

Issue #2560, opened 2026-09-08
Honor `forwarder_stop_wait_for_inflight` during shutdown.


webern:
## Background

The core agent's `forwarder_stop_wait_for_inflight` setting selects how forwarder workers stop. Its default value, `false`, cancels in-flight HTTP requests; `true` waits for them to complete. The wait remains bounded by `forwarder_stop_timeout`.

ADP does not currently read this setting. Closing its forwarder handle stops new dispatch, waits for active request tasks to finish, and then flushes pending transactions to the retry queue. The topology-wide shutdown timeout bounds this drain.

- [Core agent shutdown selection](https://github.com/DataDog/datadog-agent/blob/281d921619d52ce7b99aef40607285992c9c2e89/comp/forwarder/defaultforwarder/impl/worker.go#L107-L156)
- [ADP forwarder shutdown](https://github.com/DataDog/saluki/blob/ce1a3aed1d9c6861b17d423a5a3a023f1be3f8fa/lib/saluki-components/src/common/datadog/io.rs#L209-L220)
- [ADP in-flight request drain](https://github.com/DataDog/saluki/blob/ce1a3aed1d9c6861b17d423a5a3a023f1be3f8fa/lib/saluki-components/src/common/datadog/io.rs#L767-L843)

## Problem

ADP always behaves like `forwarder_stop_wait_for_inflight: true`. An operator cannot select the core agent's immediate-cancellation behavior by setting the key to `false`. This can make ADP wait for active requests until they finish or until the topology-wide shutdown timeout expires.

## Proposal

Honor `forwarder_stop_wait_for_inflight` in the Datadog forwarder shutdown path:

- Preserve the current bounded drain when the setting is `true`.
- When it is `false`, cancel active HTTP requests without waiting for normal completion.
- Preserve or requeue interrupted transactions where possible, and test both shutdown modes under active requests.

The implementation should continue to respect ADP's topology-wide shutdown timeout.

## Alternatives and Discussion

ADP could retain its unconditional bounded-drain policy and document the difference. Supporting the setting instead gives operators the same shutdown-policy choice as the core agent.

jszwedko:
[...]

Issue #2559, opened 2026-09-07
Honor `otlp_config.metrics.infra_attributes.as_tags`.


webern:
## Human Summary

A few new configuration options that seem relevant to ADP were discovered when updating the config schema. Below is the AI summary.
Sub-issue of #2005.

## Background

The core agent enriches OTLP resources with tags resolved from workload entities and global tags. The new
`otlp_config.metrics.infra_attributes.as_tags` setting controls whether custom tagger-derived tags become
metric tags. It defaults to `false`: custom tags remain resource attributes and the metrics translator drops
ones outside its allowlist. When enabled, the infrastructure-attributes processor adds a
`datadog.container.tag.` copy that the translator promotes. If
`otlp_config.metrics.resource_attributes_as_tags` is enabled, that broader setting takes precedence because
it already promotes every resource attribute.

ADP implements the underlying workload-tag enrichment without the core agent's infrastructure-attributes
processor. Its [OTLP origin resolver][1] resolves the same entity families plus global tags. The [metrics
translator][2] adds the resolved tags directly to every emitted metric context. This behavior was introduced
by #2285 to close #2070.

In proxy mode, ADP forwards OTLP metrics to the core agent, which owns enrichment and honors the setting. The
gap applies to ADP's native OTLP metrics path.

## Problem

ADP does not read `otlp_config.metrics.infra_attributes.as_tags`. Native processing always promotes resolved
workload tags directly onto metrics:

- With the setting enabled, ADP generally already produces the intended custom metric tags, although through
  a different implementation.
- With the setting disabled, including its default, ADP still emits custom workload tags that the core agent
  leaves as resource attributes and then drops from metrics.

`otlp_config.metrics.tag_cardinality: none` suppresses all entity and global tag enrichment, so it is not an
equivalent opt-out for custom-tag promotion. The new setting therefore exposes a behavioral difference rather
than a wholly missing enrichment feature.

## Proposal

Honor `otlp_config.metrics.infra_attributes.as_tags` in the native OTLP metrics path:

- Preserve the existing entity resolution, cardinality, precedence, and global-tag behavior.
- When the setting is disabled, retain tags covered by the core agent's known Datadog and OpenTelemetry
  mappings while withholding custom tagger-derived tags from metric contexts.
- When enabled, promote custom tagger-derived tags as ADP does today.
- Preserve the core agent's precedence when `resource_attributes_as_tags` is enabled.
[...]

Issue #2558, opened 2026-09-07
Assess metric lookback compatibility with ADP.


webern:
## Human Summary

A few new configuration options that seem relevant to ADP were discovered when updating the config schema. Below is the AI summary.

## Background

The core agent's metric lookback feature retains selected metric points and sketches locally. An
optional monitor can trigger forwarding of retained ranges around a metric's threshold breach.

The [lookback component][1] supplies a DogStatsD adapter to the [core agent demultiplexer][2]. It
observes aggregated and timestamped no-aggregation metrics. It also provides a sender manager for
shadow checks executed by the core agent. ADP has no corresponding metric-lookback configuration
or retention/monitoring integration in its DogStatsD path.

The schema adds 13 settings under `metric_lookback`:

| Scope | Settings |
|---|---|
| Shared feature gate and retention | `metric_lookback.enabled`, `metric_lookback.capacity`, `metric_lookback.shard_count` |
| DogStatsD admission | `metric_lookback.dogstatsd.metric_names` |
| Monitoring | `metric_lookback.monitor.mode`, `metric_lookback.monitor.metric_name`, `metric_lookback.monitor.evaluation_interval`, `metric_lookback.monitor.range_epsilon`, `metric_lookback.monitor.partition_tags` |
| Replay windows | `metric_lookback.egress.pre_trigger_window`, `metric_lookback.egress.post_recovery_window` |
| Core-agent shadow-check policy | `metric_lookback.enabled_checks`, `metric_lookback.collection_interval` |

The [shadow-check policy][3] consumes the last two settings to select checks and their execution
interval. ADP does not currently support checks.

## Problem

Moving DogStatsD processing to ADP may remove the observations feeding the core agent's lookback
rings and monitor. Classifying the whole namespace as unrelated to ADP would hide that potential
feature gap. Conversely, treating the check-scheduling settings as unsupported ADP functionality
would conflate core-agent execution policy with the data pipeline.

Shared settings can still affect core-agent shadow checks even when ADP does not implement their
DogStatsD behavior. Compatibility documentation needs to describe the affected path rather than
claim that the settings have no effect anywhere.

## Proposal

Investigate metric-lookback compatibility when DogStatsD processing is delegated to ADP:

- Trace whether any existing core-agent/ADP integration preserves DogStatsD observations in the
  lookback rings, including timestamped metrics and sketches.
- Establish whether DogStatsD lookback is required for ADP deployments and whether missing support
  warrants runtime work or a documented limitation.
[...]

Issue #2557, opened 2026-09-07
Clarify support for `dogstatsd_no_aggregation_pipeline_workers_count`.


webern:
## Human Summary

A few new configuration options that seem relevant to ADP were discovered when updating the config schema. Below is the AI summary.

## Background

The core agent uses `dogstatsd_no_aggregation_pipeline_workers_count` to size its separate
no-aggregation worker and serializer pool. The default is 1; nonpositive configured values are
[clamped to 1][1].

ADP already supports timestamped metrics without re-aggregation. It handles them through the
[aggregate transform's passthrough path][2], rather than a separately configurable no-aggregation
worker pool. ADP's `dogstatsd_workers_count` controls packet-decoding workers, not this stage.

## Problem

The compatibility documentation should explain why this worker-count setting has no effect in ADP
without implying that timestamped passthrough is unsupported. A rationale such as “ADP uses async
tasks” does not identify the architectural distinction or distinguish the two worker-count settings.

## Proposal

Treat this as a documentation and configuration-inventory fix:

- Document that ADP handles timestamped passthrough in the aggregate transform and has no separate
  no-aggregation worker pool to configure.
- State that the setting has no effect in ADP and is not an alias for `dogstatsd_workers_count`.
- Retain `support: none` and `severity: low`, and link this issue in the inventory.
- Keep `planned: false`: this issue does not plan runtime support for the setting.
- Regenerate the configuration documentation.

No new worker pool, configuration translation, or runtime behavior change is proposed. If a need for
independently configurable passthrough concurrency is established, track that as a separate feature.

Related: #1475, the earlier timestamped-passthrough support issue.

[1]: https://github.com/DataDog/datadog-agent/blob/281d921619d52ce7b99aef40607285992c9c2e89/comp/aggregator/demultiplexer/impl/demultiplexer.go#L124-L130
[2]: https://github.com/DataDog/saluki/blob/ce1a3aed1d9c6861b17d423a5a3a023f1be3f8fa/lib/saluki-components/src/transforms/aggregate/mod.rs#L547-L574

PR #2562, opened 2026-09-08
chore(config): update vendored datadog agent schema

webern:
## Human Summary

Updates the schema. We spent longer than we should have without doing this, hence the large diff.

When reviewing, start by focusing on what happens in `schema_overlay.yaml` to make sure we haven't excluded something you know to be important (I did batch handle these case-by-case). New issues were opened for configs that seem potentially relevant. There was some cleanup of Saluki-only keys promoted to Datadog keys, etc.

Introduces a skill, `/config-schema-update` so that someone else might be able to do this in the future aided by their LLM agent.

## AI Summary

Update the vendored Datadog Agent schema from `1b1d1834321d78aa450ac3efdf928c2a0176315d` to `281d921619d52ce7b99aef40607285992c9c2e89`.

- Review new and changed settings in the support inventory and regenerate the configuration model, environment bindings, classifier, registry, and documentation.
- Move `data_plane.stop_timeout` and `data_plane.serializer_zstd_compressor_level` to the Datadog configuration path while preserving precedence and explicit-versus-default behavior.
- Remove the retired `serializer_experimental_use_v3_api.series.endpoints` path in favor of the canonical `use_v3_api.series.*` controls.
- Support string-map deserialization and floating-point forwarder backoff settings. Keep schema placeholder defaults from replacing runtime-derived paths.
- Add a user-invoked schema update skill and warn against hand-editing the vendored snapshot.

## Change Type

## How did you test this PR?

Passed:

- `make build-schema-overlay`, `make fmt`, `cargo check --workspace`, `make check-docs`, `make check-clippy`, and `make check-deny`.
- 89 tests across `datadog-agent-config`, `datadog-agent-config-overlay-model`, and `datadog-agent-config-testing`.
- 109 focused tests for configuration loading, updates, and Datadog translation in `agent-data-plane-config-system`.
- New skill metadata, size, and Vale checks; `git diff --check`.

After rebasing onto the latest `main`, reran 24 focused tests covering string maps, classification, smoke tests, and default/provenance handling; all passed.

Direct lint of the existing `config-system` skill also flags six spelling errors outside the changed lines.

## References

- Related: #1788
- Related: #2555
- Related: #2556
- Related: #2557
- Related: #2558
- Related: #2559
- Related: #2560


jszwedko (commented):
Would it make any sense to have this run as part of the `bump-agent-version` workflow?

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -515,6 +513,28 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
         self.config.control.secure_api_listen_address = value;
     }
 
+    fn consume_data_plane_serializer_zstd_compressor_level(&mut self, value: i64) {
+        let provenance = self.sources.provenance("data_plane.serializer_zstd_compressor_level");
```
webern, line 517: I didn't catch this earlier in the provenance work. It's not ideal that we have to look these up by string key here. Surely this can be solved a better way. Opened #2564

docs/agent-data-plane/configuration/configuration.md
```diff
@@ -19,6 +19,36 @@ If you find an error on this page, please [open an issue].
...
+| `metric_lookback.enabled`                                        | Enable metric lookback                    | [#2558] |
+| `metric_lookback.monitor.evaluation_interval`                    | Lookback monitor evaluation window        | [#2558] |
+| `metric_lookback.monitor.metric_name`                            | Metric watched by lookback monitor        | [#2558] |
+| `metric_lookback.monitor.mode`                                   | Metric lookback monitor mode              | [#2558] |
+| `metric_lookback.monitor.partition_tags`                         | Lookback monitor tag partitions           | [#2558] |
+| `metric_lookback.monitor.range_epsilon`                          | Lookback monitor range threshold          | [#2558] |
+| `metric_lookback.shard_count`                                    | Metric lookback ring shards               | [#2558] |
+| `otlp_config.metrics.infra_attributes.as_tags`                   | Promote infrastructure attributes to tags | [#2559] |
+| `otlp_config.traces.infra_attributes.container_tag_promotion`    | OTLP trace container-tag promotion        | [#2338] |
```
jszwedko, line 47: I have to imagine we don't support _most_ of the `otlp_config` options. But maybe I'm wrong? Assuming I'm right, do we want to list them all here now? Or should we keep this focused on dogstatsd?

docs/agent-data-plane/configuration/configuration.md
```diff
@@ -670,14 +706,6 @@ Temporary development key for setting ADP's OTLP listen endpoints independently
 
 Temporary development key for setting ADP's OTLP listen endpoints independently from the Agent's.
 
-### `data_plane.serializer_zstd_compressor_level`
-
-ADP-specific zstd compression level, taking precedence over the Core Agent's `serializer_zstd_compressor_level`. When this key is unset, ADP falls back to `serializer_zstd_compressor_level` if you set that key explicitly, and otherwise uses its own default of 3. Level 3 achieves ~6% smaller payloads (65.3 MB vs 69.3 MB) without a net CPU increase, since ADP is more efficient than the Agent and can afford higher compression. Configure via `DD_DATA_PLANE_SERIALIZER_ZSTD_COMPRESSOR_LEVEL` or in ADP-specific configuration.
```
jszwedko, line 675: Did you mean to drop these descriptions of the options?

lib/datadog-agent/config/schema/core/admission_controller.yaml
```diff
@@ -1,5 +1,5 @@
 $schema: https://json-schema.org/draft/2020-12/schema
-$id: https://raw.githubusercontent.com/DataDog/schema/main/agent/admission_controller.yaml.schema.json
```
jszwedko, line 2: Suggestion: mark these vendored files as "machine generated" so that they are hidden on PRs by default.

lib/datadog-agent/config/src/generated/datadog_configuration.rs
```diff
@@ -144,7 +144,7 @@ pub struct DatadogConfiguration {
     #[serde(deserialize_with = "crate::cast_de::deserialize_bool")]
     pub dogstatsd_flush_incomplete_buckets: bool,
 
-    #[serde(default)]
+    #[serde(default = "defaults::datadog_configuration_dogstatsd_log_file")]
```
jszwedko, line 147: Nit: same here, suggestion to make these generated files as generated to hide them from GitHub diffs by default.

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
```diff
@@ -754,16 +777,47 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
         self.config.shared.endpoints.forwarder.apikey_validation_interval = value;
     }
 
-    fn consume_forwarder_backoff_base(&mut self, value: i64) {
-        self.config.shared.endpoints.forwarder.backoff_base = value as f64;
+    fn consume_forwarder_backoff_base(&mut self, value: f64) {
+        let value = if value <= 0.0 {
+            warn!("`forwarder_backoff_base` is not positive ({value}); using 2.");
+            2.0
```
jszwedko, line 785: Nit, should we use the default functions to source these values instead of hardcoding it in two places? Applies to following functions in this file too.

jszwedko (approved):
A few comments/questions/suggestions but nothing blocking

lib/datadog-agent/config/schema/core/admission_controller.yaml
webern: Agreed in principle. For the files we are actually generating *from* the schema, I agree these need to be less noisy and I will do so: created #2579 

For these schema files themselves. These should only change when we update the schema (*manually* in some sense of the word). I would prefer to see their diffs in PRs so that we can detect when contributors or LLMs try to patch them up. They shouldn't do that, and it would be good to see such a diff and flag it to see if there's a way we can keep our vendored copy pristine.

Resolving this comment as planned for follow-up.

lib/datadog-agent/config/src/generated/datadog_configuration.rs
webern: Agreed. Initially, what we were doing with the config system was too scary to not see these diffs, but we have since gained confidence.

Created #2579 

Resolving as planned for follow-up.

docs/agent-data-plane/configuration/configuration.md
webern: Hmmm, well, we should decide how we want to organize this file. Initially the file itself was called `dogstatsd.md`. When we started working on OTLP I named it `configuration.md` thinking it would be best to just lump all ADP configuration quirks into one documentation file.

Dividing them up into conceptually different buckets, OTLP vs Metrics, vs DogStatsD, or suppressing some and not others is messier than it sounds. Some things are shared, some are not. 

In other words, because this file is generated from the one source of truth `schema_overlay.yaml`, it's not 100% obvious how, or that we should, encode further documentation specific organizational principles into the one source of truth.

I'll keep it in mind, but I don't have an immediate solution at hand.

lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
webern: As usual, not simple. Will fix the inaccessibility of the defaults in #2580. Looking at starting to upstream schema fixes, like minimum value constraints as well. Added TODOs at the sites of the magic values for visibility and tracking.

docs/agent-data-plane/configuration/configuration.md
webern: No, I did a little reorganizing to get them back. Good catch.

PR #2563, opened 2026-09-08
feat(otlp): added ip quantization and aggregation by new key

### LINKED ISSUE #2422: Traces(otlp+dd): Quantize peer IP addresses before stats peer-tag aggregation
## Background

Peer tags become APM stats dimensions. A raw peer IP turns every host into its own aggregation key, so bucket cardinality grows with the fleet and per-service views fragment. The Agent collapses those values before hashing the peer tags.

## Agent behavior

`matchingPeerTags` (`span_concentrator.go:110-126`, and its sibling at 137) calls `QuantizePeerIPAddresses` on each matched peer-tag value before appending `key:value`. That helper splits on commas, runs `quantizeIP` per entry, and collapses duplicates. `quantizeIP` strips an `ip-` prefix or a `dnspoll`, `ftp`, `file`, `http`, or `https` scheme, parses host, port, and suffix, leaves `127.0.0.1`, `::1`, `169.254.169.254`, `fd00:ec2::254`, and `169.254.170.2` alone, and otherwise returns the prefix plus `blocked-ip-address`, the original port, and the suffix. Non-IP entries pass through.

## ADP behavior

`transforms/apm_stats/span_concentrator.rs` collects `matching_peer_tags` and uses the values as aggregation dimensions without inspecting them. No IP parsing, address allowlist, or `blocked-ip-address` placeholder exists anywhere in `lib/`.

## Acceptance

- [ ] A non-allowlisted peer IP becomes `blocked-ip-address` and keeps its port
- [ ] `127.0.0.1`, `::1`, and the cloud metadata addresses are unchanged
- [ ] An `ip-` prefix or a `scheme://` prefix survives quantization
- [ ] Comma-separated peer values are deduplicated after quantization
- [ ] Peer values that are not IP addresses are unchanged

## References

Depends on: #2421

lucastemb:
## Summary

This PR aims to add improved aggregation logic more inline with the upstream implementation. Currently, the aggregation key takes the site as is; however, this is a cardinality bomb. In an effort to combat this, we now quantize and obfuscate IPs, deduplicate, amongst other things.   

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests
## References

- Closes: #2422


lib/saluki-components/src/transforms/apm_stats/peer_ip_quantize.rs
```diff
@@ -0,0 +1,387 @@
...
+
+    for scheme in SCHEMES {
+        if let Some(scheme_index) = raw.find(scheme) {
+            let scheme_end = scheme_index + scheme.len() + 4;
+            if scheme_end < raw.len() && &raw[scheme_index + scheme.len()..scheme_end] == ":///" {
+                return (&raw[scheme_index..scheme_end], &raw[scheme_end..]);
+            }
+            let scheme_end = scheme_index + scheme.len() + 3;
+            if scheme_end < raw.len() && &raw[scheme_index + scheme.len()..scheme_end] == "://" {
+                return (&raw[scheme_index..scheme_end], &raw[scheme_end..]);
```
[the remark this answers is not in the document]
lucastemb: Text before a scheme is dropped, [just like in the upstream implementation](https://github.com/DataDog/datadog-agent/blob/main/pkg/obfuscate/ip_address.go#L45-L68).

lib/saluki-components/src/transforms/apm_stats/peer_ip_quantize.rs
```diff
@@ -0,0 +1,487 @@
...
+        _ => return None,
+    }
+
+    let bytes = s.as_bytes();
+    for &b in bytes.iter() {
+        match b {
+            b'.' | b'_' | b'-' => return parse_ipv4(s, b),
+            b':' => {
+                // IPv6: the whole remaining string must parse.
+                if s.parse::<Ipv6Addr>().is_ok() {
```
webern: I guess it caught that the `Ipv6Addr` parsing behavior does not match whatever the Agent is using. This is annoying and I'm not sure how important it is. I would prefer we not have to throw out `Ipv6Addr` and roll our own parsing for this. We probably need to look into this though and see what to do about it.

> GPT-5.4: It looks like this diverges from the Agent for scoped IPv6 addresses. `netip.ParseAddr` accepts values like `fe80::1%eth0`, while `Ipv6Addr` does not, so Saluki leaves `fe80::1%eth0` and `[fe80::1%eth0]:8080` unquantized. Could we preserve Agent parity and add both forms to the reference cases?

lib/saluki-components/src/transforms/apm_stats/span_concentrator.rs
```diff
@@ -286,6 +287,9 @@ impl SpanConcentrator {
         for key in keys_to_check {
             if let Some(value) = span.attributes.get(key.as_ref()).and_then(AttributeValue::as_string) {
                 if !value.is_empty() {
+                    // Quantize IP addresses before the value is hashed into the aggregation key, so
+                    // that per-host cardinality collapses to a single dimension.
+                    let value = quantize_peer_ip_addresses(value.as_ref());
```
webern, line 290: Nit: the agent caught some untested logic.

> GPT-5.4: Could we add a concentrator-level regression test for this integration point? The helper tests would still pass if this call were removed. A useful case would send otherwise-identical spans with different peer IPs, flush, and assert one grouped aggregation with combined hits and the `blocked-ip-address` peer tag.

lucastemb: Fixed in [a5526d1](https://github.com/DataDog/saluki/pull/2563/commits/a5526d1cbdf2b7f423dbb42e08ea9f4f263f1797)

lib/saluki-components/src/transforms/apm_stats/peer_ip_quantize.rs
lucastemb: Fixed in [a5526d1](https://github.com/DataDog/saluki/pull/2563/commits/a5526d1cbdf2b7f423dbb42e08ea9f4f263f1797)

lib/saluki-components/src/transforms/apm_stats/peer_ip_quantize.rs
```diff
@@ -0,0 +1,511 @@
...
+    for scheme in SCHEMES {
+        if let Some(scheme_index) = raw.find(scheme) {
+            let scheme_end = scheme_index + scheme.len() + 4;
+            if scheme_end < raw.len() && &raw[scheme_index + scheme.len()..scheme_end] == ":///" {
+                return (&raw[scheme_index..scheme_end], &raw[scheme_end..]);
+            }
+            let scheme_end = scheme_index + scheme.len() + 3;
+            if scheme_end < raw.len() && &raw[scheme_index + scheme.len()..scheme_end] == "://" {
+                return (&raw[scheme_index..scheme_end], &raw[scheme_end..]);
+            }
```
webern: Sorry... we found a panic case here.

The background is that, in Rust (unlike many other languages including Go) `String` is UTF-8 encoded and if you slice it somewhere that is not on a character boundary it can panic. So we should be careful about that if the input can be multibyte.

```rust
fn split_prefix(raw: &str) -> (&str, &str) {
    if let Some(rest) = raw.strip_prefix("ip-") {
        return ("ip-", rest);
    }

    for scheme in SCHEMES {
        let Some(scheme_index) = raw.find(scheme) else {
            continue;
        };

        // Slicing here is safe because `find` returns a character boundary and the schemes are
        // ASCII. Matching the separator with `strip_prefix` then avoids computing any further
        // index, which is what could otherwise land inside a multi-byte character and panic.
        let after_scheme = &raw[scheme_index + scheme.len()..];

        // `:///` is tried first because `://` is a prefix of it. The separator must be followed by
        // at least one character, matching the upstream implementation.
        for separator in [":///", "://"] {
            if let Some(rest) = after_scheme.strip_prefix(separator) {
                if !rest.is_empty() {
                    let scheme_end = scheme_index + scheme.len() + separator.len();
                    return (&raw[scheme_index..scheme_end], rest);
                }
            }
        }
    }

    ("", raw)
}
```

lucastemb: Fixed in [e7d8d3d](https://github.com/DataDog/saluki/pull/2563/commits/e7d8d3d8cd4084c6f0e73df5e12f039b7d98261e)

webern (approved):

PR #2565, opened 2026-09-08
fix(dogstatsd): exclude timestamped metrics from tag filterlist

olivielpeau:
## Summary

`dsd_tag_filterlist` ran before the aggregate transform's internal timestamp-based passthrough split, so it stripped tags from metrics that skip aggregation entirely. That's undesirable since filtering tags without aggregating can impact metric query results and pointless -- both halves of the filterlist (`metric_tag_filterlist` and `metric_tag_value_allowlist`) exist to merge samples that differ only in a filtered tag into a shared aggregated context, and metrics carrying an explicit client timestamp are never aggregated.

This extracts that split into a new `dsd_no_agg_split` transform placed ahead of `dsd_tag_filterlist`, so only metrics that actually get aggregated are subject to the filterlist. This matches the classic Datadog Agent's behavior, which does not apply its equivalent tag filterlist to metrics routed through its no-aggregation pipeline.

### Why a separate transform

Teaching `dsd_tag_filterlist` to skip timestamped metrics in place would have been a smaller change, but it can't be exact. A single `Metric` can hold both timestamped and non-timestamped values, and the filterlist rewrites the context those values share, so in-place skipping can only filter all of a mixed metric or none of it. Splitting the values into two metrics first gives each half the treatment it should get: the timestamped half keeps every tag, the remainder is filtered and aggregated as normal.

### Topology

`dsd_no_agg_split` is wired in only when `dogstatsd_no_aggregation_pipeline` is enabled (the default). With it disabled there is nothing to route out of the aggregation path, so the component is left out of the blueprint entirely rather than sitting in the hottest part of the process as a pure forwarder, and the metrics chain is exactly what it was before this PR.

When enabled, the chain becomes:

```
dsd_in.metrics → dsd_enrich → dsd_prefix_filter → dsd_no_agg_split → dsd_tag_filterlist → dsd_agg → dsd_post_agg_filter → metrics_enrich
                                                        └── .passthrough ─────────────────────────────→ dsd_post_agg_filter
```

The new transform forwards the incoming event buffer untouched when it holds no timestamped values — the common case — and uses a buffered dispatcher on the splitting path, so a full output buffer flushes and re-acquires rather than dropping events.

### Telemetry consequences

Three, all covered by the release note:

- The passthrough counters move with the split, so the `no_aggregation.processed` and `no_aggregation.flush` remapping rules now require `component_id:dsd_no_agg_split`. Those rules match on the emitting component's id, so leaving them pointed at `dsd_agg` would have dropped both metrics from the Agent telemetry payload. The metric names themselves are unchanged, to keep existing dashboards working.
- ADP's internal `adp.aggregate_passthrough_metrics_total` and `adp.aggregate_passthrough_flushes_total` are now tagged `component_id:dsd_no_agg_split`. Internal dashboards that group or filter by `component_id` need updating.
- `aggregator.processed` (`data_type:dogstatsd_metrics`) no longer counts timestamped samples, since they no longer reach the aggregator. They continue to be counted by `no_aggregation.processed`.

The passthrough batcher's `component_events_dropped_total` moves across with it, but retagged from `intentional:true` to `intentional:false` with `drop_reason:buffer_full`. Discarding a metric because the active buffer is full immediately after being flushed and replaced is a failure, not a deliberate shed like the aggregate transform's context-limit drops, which keep the `intentional:true` tag they had. The two other error paths in the new component are not counted as drops: a missing default output is a wiring bug, so acquiring the buffered dispatcher is an `expect` like every other caller in the tree, and a dispatch error means the downstream receiver is already gone during shutdown, which the framework accounts for via `events_discarded_total`.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

- New regression test (`cli::run::tests::no_agg_split_excludes_timestamped_metrics_from_tag_filterlist`) verifying a timestamped metric keeps tags that would otherwise be stripped, while a non-timestamped metric with the same tags still has them filtered before aggregation.
- New unit tests on `dsd_no_agg_split`'s timestamp-based split logic, including the mixed-timestamp case and the fast-path scan.
- New `state::metrics::tests::rar_rules_match_the_component_that_emits_them`, asserting the passthrough remapping rules match the component that emits those counters. The pre-existing rule test only checked help text, which is why the stale `component_id` wasn't caught.
- `cargo nextest run -p agent-data-plane -p saluki-components`: 1,248 tests passed.
- `make check-all`: Vale clean across 752 files, full-workspace Clippy and formatting clean, licenses/unused-deps/features clean. It stops on a pre-existing `cargo-deny` advisory for a yanked `der 0.8.0` in `Cargo.lock`, which this PR doesn't touch.
- `make check-release-notes` and `reno lint` pass.
- Correctness cases `dsd-timestamp-extension` (3,000 identical metrics) and `dsd-tag-filterlist` (178 identical metrics) both pass.


bin/agent-data-plane/src/cli/run.rs
```diff
@@ -1066,6 +1058,39 @@ async fn add_dsd_pipeline_to_blueprint(
...
+        blueprint.connect_components_in_order([
+            "dsd_in.metrics",
+            "dsd_enrich",
+            "dsd_prefix_filter",
+            "dsd_tag_filterlist",
+            "dsd_agg",
+            "dsd_post_agg_filter",
+            "metrics_enrich",
+        ])?;
+    }
```
tobz, line 1092: Just saying this out loud because it bothers me, not because this is anything you need to change: it's a bummer we have to be this verbose for an optional component.

Maybe one day I can figure out a pattern to allow more naturally declaring a component optionally so that it just gets ignored or optimized away from the graph if it's a no-op... 🤔

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -1443,6 +1467,143 @@ mod tests {
         .expect("DogStatsD retained identity test should complete without hanging");
     }
 
+    #[tokio::test]
+    async fn no_agg_split_excludes_timestamped_metrics_from_tag_filterlist() {
```
tobz, line 1471: Also commenting out loud here, but not stating anything that needs to change: we need a somewhat better answer for constructing these types of integration-esque tests, perhaps through the integration test harness.

It's very clunky to have it sitting around here as a unit test in a file that's really just wiring things together.

bin/agent-data-plane/src/components/dogstatsd_no_agg_split/telemetry.rs
```diff
@@ -0,0 +1,65 @@
+//! Telemetry for the DogStatsD no-aggregation-pipeline split transform.
+//!
+//! The passthrough metrics keep their original `aggregate_`-prefixed names, even though they're no longer emitted by
+//! the aggregate transform, so that existing dashboards and the Datadog Agent telemetry remappings keep working. The
+//! `component_id` tag on them does change, from `dsd_agg` to `dsd_no_agg_split`: see
+//! `crate::state::metrics::rules::get_aggregation_remappings`.
```
tobz, line 6: Nit: redundant comment that repeats what Git history will already tell us.

bin/agent-data-plane/src/state/metrics/mod.rs
```diff
@@ -569,6 +569,47 @@ mod tests {
         assert_eq!(find("transactions.requeued"), Some("Transaction requeue count"));
     }
 
+    #[test]
+    fn rar_rules_match_the_component_that_emits_them() {
```
tobz, line 573: Can we just change the test name slightly here to indicate we're specifically looking at no-agg remapping correctness?

bin/agent-data-plane/src/state/metrics/rules/aggregation.rs
```diff
@@ -29,16 +31,18 @@ pub fn get_aggregation_remappings() -> Vec<RemapperRule> {
         )
         .with_additional_tags(["data_type:dogstatsd_metrics"])
         .with_help_text("Amount of metrics/services_checks/events processed by the aggregator"),
+        // The passthrough counters are emitted by `dsd_no_agg_split`, not `dsd_agg`, since the timestamp-based
+        // split moved out of the aggregate transform.
```
tobz, line 35: Nit: redundant comment that repeats what Git history will already tell us.

bin/agent-data-plane/src/components/dogstatsd_no_agg_split/mod.rs
```diff
@@ -0,0 +1,446 @@
...
+//! Splitting first, rather than teaching `dogstatsd_tag_filterlist` to skip timestamped metrics in place, is what
+//! makes that exclusion exact. A single `Metric` can hold both timestamped and non-timestamped values, and the
+//! filterlist rewrites the context those values share, so in-place skipping can only choose between filtering all of
+//! a mixed metric or none of it. Splitting the values into two metrics first gives each half the treatment it should
+//! get: the timestamped half keeps every tag, and the remainder is filtered and aggregated as normal. That exactness
+//! is why this is a separate transform and not a few lines inside the filterlist.
+//!
+//! The transform is only wired into the topology when the Datadog `dogstatsd_no_aggregation_pipeline` key is enabled
+//! (the default). When it's disabled there is nothing to split out -- every metric is aggregated -- so the component
+//! is left out entirely rather than sitting in the hot path as a pure forwarder.
```
tobz, line 24: Nit: redundant comment that repeats (a lot of) what Git history will already tell us.

I think it would simply suffice to say what the transform does -- splits timestamped vs non-timestamped metrics to different outputs, with batching of timestamped metrics -- and roughly how we use it.

We don't need to explain why we created it, or when it gets added: Git history and the actual topology construction code already tell us that.

tobz (changes requested):
Functionally, everything looks and seems reasonable. Just needs a bit of polish and comment slop cleanup. 😅

olivielpeau:
@tobz thanks for the review! Addressed your comments

tobz (approved):

tobz (approved):

PR #2567, opened 2026-09-08
feat(otlp): trace metadata backfill and top-level fallback 

### LINKED ISSUE #2393: Traces(otlp+dd): Add trace metadata backfill and top-level fallback
## Background

ADP already groups OTLP spans by low 64 bits and its sampler finds roots and records sampling decision makers. It still lacks Agent-compatible origin backfill, full trace-ID consistency checks, and the chunk-wide top-level fallback when span-kind computation is disabled.

## Agent behavior

`receiveResourceSpansV2` groups spans by `OTelTraceIDToUint64`. Later, `GetRoot` selects the root, `setChunkAttributes` backfills origin and decision-maker, `SameTraceID` checks full IDs, and `ComputeTopLevel` marks roots, orphaned spans, and service-boundary spans when client computation is off.

## ADP behavior

`translator.rs` groups by low 64 bits and captures the high 64 bits. `TraceSampler` finds roots and later populates `Trace.decision_maker`; the encoder writes it. The OTLP path does not promote root `_dd.origin` or scan spans for `_dd.p.dm`, reject mixed full IDs in one group, or mark top-level spans chunk-wide when the flag is off.

## Acceptance

- [ ] OTLP spans with different full trace IDs are not accepted as one trace
- [ ] Origin and decision-maker metadata use the Agent's root/first-span rules
- [ ] The Agent root-selection rules identify the root for OTLP metadata backfill
- [ ] The Agent and ADP produce the same low-64-bit grouping key
- [ ] Disabled span-kind computation marks roots, orphans, and service boundaries top-level

## References

Depends on: #2383

lucastemb:
## Summary

This PR aims to add mechanisms to backfill metadata to distributed spans. 

By taking the bottom `128-bits` (`16 bytes`) of a `SpanID` we create a unique identifier to ensure that spans belongs to the same trace. Once we determine a parent-span, we use that to backfill and populate all related spans with the appropriate metadata regarding origin as well as _what_ influenced the decision to keep a trace (was it user configured, was it the result of sampling probability, etc). 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests
## References

- Closes: #2393


lib/saluki-components/src/common/otlp/traces/translator.rs
```diff
@@ -207,8 +211,16 @@ impl OtlpTracesTranslator {
                     priority: None,
                     trace_id_hex: None,
                     trace_id_high,
+                    rejected: false,
                 });
 
+                // Full trace ID consistency: spans in a group must agree with the group's first
+                // span on the high half of the trace ID. A zero high half (a 64-bit trace ID) is
+                // trusted on the low half alone.
+                if entry.trace_id_high != 0 && trace_id_high != 0 && entry.trace_id_high != trace_id_high {
```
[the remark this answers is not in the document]
lucastemb: Solved by [5fc9d12](https://github.com/DataDog/saluki/pull/2567/commits/5fc9d121aa54bbbb48a6f3172e2fbba07597e135)

lib/saluki-components/src/common/otlp/traces/translator.rs
```diff
@@ -275,6 +300,36 @@ impl Iterator for OtlpTraceEventsIter {
...
+                        .get(TAG_DECISION_MAKER)
+                        .and_then(AttributeValue::as_string)
+                });
+                if let Some(dm) = first_decision_maker {
+                    trace.decision_maker = Some(dm.clone());
+                }
+            }
+
+            if !self.compute_top_level_by_span_kind {
+                compute_top_level(trace.spans_mut());
```
webern: This first finding might be what @ajgajg1134 was mentioning where the order that things happen in is important. Perhaps this ordering issue predates the PR, not sure, but it is saying this:

Finding:

### 1. Metadata and top-level computation happen before span filtering

The translator computes the root, promotes metadata, and marks top-level spans at:

- `lib/saluki-components/src/common/otlp/traces/translator.rs:303-330`

Production then runs the OTTL filter, which can remove spans:

- `bin/agent-data-plane/src/cli/run.rs:743-748`
- `bin/agent-data-plane/src/components/ottl_filter_processor/mod.rs:126-131`

This can produce several incorrect results:

- If the root is filtered, `trace.origin` still comes from the removed root rather than the new Agent-selected root.
- If the first span carrying `_dd.p.dm` is filtered, its decision maker remains promoted.
- If a parent is filtered but its same-service child survives, the child becomes an orphan but remains unmarked as top-level.

The Agent performs span removal before normalization/root selection and computes top-level marks later in processing (`datadog-agent/pkg/trace/agent/agent.go:523-546,583-587`).

This is also the main performance issue: the translator discovers the root here, then the trace sampler repeats the same scan at `trace_sampler/mod.rs:304-314`. Moving backfill and fallback marking after filtering—probably into the sampler while it already has the root index—would fix both concerns.

lib/saluki-components/src/common/otlp/traces/translator.rs
webern: Also preexisting this PR? The way the bytes are being used from the ID is not fully consistent with Agent because, if I understand correctly, we aren't requiring the IDs to be exactly 16-bytes? Sounds like we are supposed to:
- accept empty
- accept exactly 16-bytes
- reject anything else
- also reject it if it equals zero

### 2. The “full trace ID” check accepts malformed, distinct OTLP IDs

Saluki projects arbitrary byte lengths onto two `u64`s:

- Low half: final eight bytes at `translator.rs:29-34`
- High half: first eight bytes only when length is at least 16 at `translator.rs:36-44`
- Consistency check: compares only those derived values at `translator.rs:207-220`

Consequently, distinct 9–15-byte IDs sharing the final eight bytes are accepted together. IDs longer than 16 can differ in their middle bytes and still compare equal. The new test at `translator.rs:607-624` also treats an 8-byte OTLP trace ID as valid and equivalent to its zero-padded form.

The Agent’s pdata decoder accepts only empty or exactly 16-byte trace IDs and rejects other lengths (`collector/pdata/internal/traceid.go:50-59`); its normalizer subsequently rejects a zero trace ID. Saluki should validate canonical 16-byte, nonzero OTLP IDs or retain and compare the complete byte sequence.

For valid 16-byte IDs, the implementation is correct.

lib/saluki-components/src/common/otlp/traces/translator.rs
```diff
@@ -275,6 +300,36 @@ impl Iterator for OtlpTraceEventsIter {
             trace.payload.tracer_version = self.resource_meta.tracer_version.clone();
             trace.attributes = Arc::clone(&self.resource_meta.attributes);
 
+            // Backfill trace-level metadata from the spans: origin comes from the root span's
+            // `_dd.origin` tag, and the decision maker is the first span in chunk order carrying
+            // `_dd.p.dm` - both only when not already set.
+            if let Some(root_idx) = get_root_span_index(trace.spans()) {
```
webern, line 306: This first finding might be what @ajgajg1134 was mentioning where the order that things happen in is important. Perhaps this ordering issue predates the PR, not sure, but it is saying this:

Finding:

> ### 1. Metadata and top-level computation happen before span filtering
> 
> The translator computes the root, promotes metadata, and marks top-level spans at:
> 
> - `lib/saluki-components/src/common/otlp/traces/translator.rs:303-330`
> 
> Production then runs the OTTL filter, which can remove spans:
> 
> - `bin/agent-data-plane/src/cli/run.rs:743-748`
> - `bin/agent-data-plane/src/components/ottl_filter_processor/mod.rs:126-131`
> 
> This can produce several incorrect results:
> 
> - If the root is filtered, `trace.origin` still comes from the removed root rather than the new Agent-selected root.
> - If the first span carrying `_dd.p.dm` is filtered, its decision maker remains promoted.
> - If a parent is filtered but its same-service child survives, the child becomes an orphan but remains unmarked as top-level.
> 
> The Agent performs span removal before normalization/root selection and computes top-level marks later in processing (`datadog-agent/pkg/trace/agent/agent.go:523-546,583-587`).
> 
> This is also the main performance issue: the translator discovers the root here, then the trace sampler repeats the same scan at `trace_sampler/mod.rs:304-314`. Moving backfill and fallback marking after filtering—probably into the sampler while it already has the root index—would fix both concerns.

lib/saluki-components/src/common/otlp/traces/translator.rs
```diff
@@ -207,8 +211,15 @@ impl OtlpTracesTranslator {
                     priority: None,
                     trace_id_hex: None,
                     trace_id_high,
+                    rejected: false,
                 });
 
+                // Full trace ID consistency: spans in a group must carry the same full trace ID
+                // as the group's first span.
+                if entry.trace_id_high != trace_id_high {
```
webern, line 219: Partially preexisting this PR, I guess: The way the bytes are being used from the ID is not fully consistent with Agent because, if I understand correctly, we aren't requiring the IDs to be exactly 16-bytes? Sounds like we are supposed to:
- accept empty
- accept exactly 16-bytes
- reject anything else
- also reject it if it equals zero

> ### 2. The “full trace ID” check accepts malformed, distinct OTLP IDs
> 
> Saluki projects arbitrary byte lengths onto two `u64`s:
> 
> - Low half: final eight bytes at `translator.rs:29-34`
> - High half: first eight bytes only when length is at least 16 at `translator.rs:36-44`
> - Consistency check: compares only those derived values at `translator.rs:207-220`
> 
> Consequently, distinct 9–15-byte IDs sharing the final eight bytes are accepted together. IDs longer than 16 can differ in their middle bytes and still compare equal. The new test at `translator.rs:607-624` also treats an 8-byte OTLP trace ID as valid and equivalent to its zero-padded form.
> 
> The Agent’s pdata decoder accepts only empty or exactly 16-byte trace IDs and rejects other lengths (`collector/pdata/internal/traceid.go:50-59`); its normalizer subsequently rejects a zero trace ID. Saluki should validate canonical 16-byte, nonzero OTLP IDs or retain and compare the complete byte sequence.
> 
> For valid 16-byte IDs, the implementation is correct.

lib/saluki-components/src/transforms/trace_sampler/mod.rs
```diff
@@ -472,18 +430,26 @@ impl TraceSampler {
         &self, trace: &mut Trace, keep: bool, priority: i32, decision_maker: &str, root_span_idx: usize,
     ) {
         let is_otlp = self.is_otlp_trace(trace, root_span_idx);
+        // Add tag for the decision maker: keep one backfilled at ingest (if the sampler itself
+        // has no decision to stamp) before falling back to the root span's tag.
+        let backfilled_decision_maker = if decision_maker.is_empty() {
```
webern, line 435: Sounds legit:

> ### 3. Promoted decision-maker metadata can overwrite the root span’s own value
> 
> The translator correctly promotes the first `_dd.p.dm` in chunk order. Later, `apply_sampling_metadata` writes that promoted value onto the root:
> 
> - `trace_sampler/mod.rs:433-470`
> 
> If an earlier child has `-8` and the root has `-9`, Saluki promotes `-8` and then replaces the root’s `-9` with `-8`. The Agent promotes the first value to chunk metadata but deliberately leaves each span’s metadata unchanged (`datadog-agent/pkg/trace/agent/normalizer.go:373-386`).
> 
> This requires conflicting input, but it is a concrete output-parity difference, and the Agent notes that downstream systems rely on the original span-level tag.

lib/saluki-components/src/common/datadog/mod.rs
```diff
@@ -126,3 +134,73 @@ pub fn get_trace_env(trace: &Trace, root_span_idx: usize) -> Option<&MetaString>
...
+    if spans.is_empty() {
+        return None;
+    }
+
+    let length = spans.len();
+    let mut parent_id_to_child: FastHashMap<u64, usize> = FastHashMap::default();
+
+    for i in 0..length {
+        let j = length - 1 - i;
+        if spans[j].parent_id() == 0 {
```
webern, line 154: I guess there is a step where the Agent fixes the ParentID when it seems to be wrong based on the condition below:

> The shared root finder itself matches Agent `GetRoot`. However, the Agent first repairs spans where `ParentID == TraceID == SpanID` by setting the parent to zero (`normalizer.go:255-257`). Saluki does not perform that repair, so malformed self-parented roots can cause origin, sampling, and top-level behavior to anchor to the wrong span.

webern (changes requested):
It seems we are stepping into a world of complexity here! Clean room audit found some things.

webern (approved):
Checks out!

PR #2572, opened 2026-09-09
fix(saluki-io): report HTTP/2 retry errors

### LINKED ISSUE #2568: Report structured HTTP/2 details in retry errors
## Human Intro

During an investigation of increased retries, we found that HTTP error details are absent from the logs (which would be really helpful to have).

> The following was written by GPT-5.6 Sol:

## Background

The standard HTTP retry lifecycle walks nested client errors before it writes a retry log. It has dedicated handling for `hyper_util::client::legacy::Error`, `rustls::Error`, and `std::io::Error`, but it does not recognize `h2::Error`. An HTTP/2 protocol failure therefore reaches the fallback and uses `h2::Error`'s `Display` output. In practice, that output may contain only `http2 error`.

The relevant code is [`CategorizedError::extract_nested`](https://github.com/DataDog/saluki/blob/c2ac928884b0ed5dfe024b2d5c815fc5b909bea6/lib/saluki-io/src/net/util/retry/lifecycle/http.rs#L76-L100). The fallback is rendered by [`CategorizedError::Other`](https://github.com/DataDog/saluki/blob/c2ac928884b0ed5dfe024b2d5c815fc5b909bea6/lib/saluki-io/src/net/util/retry/lifecycle/http.rs#L118-L129).

During an investigation, this hid the distinction among connection-level failures such as GOAWAY, stream resets, and protocol errors. The retry behavior was visible, but the logs did not contain enough information to identify the peer action or h2 reason code.

## Problem

Retry logs should report structured HTTP/2 error details when the nested cause is an `h2::Error`.

At minimum, the message should distinguish:

- GOAWAY from stream-level reset and other HTTP/2 failures;
- the h2 reason code, when present;
- whether the error originated from the remote peer, when the API exposes that information.

The request URI must remain sanitized as it is today.

## Implementation Ideas

We could add an HTTP/2 variant to `CategorizedError` and recognize `h2::Error` while walking the source chain. `saluki-io` may need a direct dependency on `h2` rather than relying on it transitively through Hyper.

The variant could retain a bounded set of structured fields such as the reason code and GOAWAY/reset flags, then format those fields in the existing warning. We should avoid logging debug output that may be unstable across dependency releases or include more data than intended.

It may also be useful to preserve the original error text as a fallback when no structured reason is available.

## Testing Ideas

A focused unit test could first reproduce the current result: a nested `h2::Error` formats as only `Request failed: http2 error`. The test should fail before the change.

After the change, table-driven tests could cover representative GOAWAY, reset, and reason-code cases. Each case should pass through the production `extract_nested` path rather than testing a test-only formatter. Existing tests should continue to verify URI sanitization and the Rustls-specific branches.

## Method

We traced sustained retry logs back to `StandardHttpRetryLifecycle::before_retry`. The source-chain walker reached an HTTP/2 error but did not have a matching downcast branch, so the fallback discarded the available protocol-specific context.

webern:
## Human Summary

TODO

## AI Summary

Walk HTTP retry error source chains and report structured HTTP/2 details for GOAWAY, stream reset, and other protocol failures. Add optional tracing fields for the error kind, reason code, and remote origin while preserving URI sanitization and existing HTTP, TLS, and client error messages.

Fixes #2568.

## Log output

Ignoring timestamp and target metadata, a retry warning previously reduced an HTTP/2 failure to:

```text
Request failed: http2 error. Retrying after 1s. error_count=4 request_uri=https://example.com/api/v1/series
```

A remote GOAWAY now produces:

```text
Request failed due to an HTTP/2 GOAWAY received from the remote peer (reason: detected excessive load generating behavior, code 11). Retrying after 1s. error_count=4 request_uri=https://example.com/api/v1/series http2.error_kind="go_away" http2.reason_code=11 http2.received_from_remote=true
```

A remote stream reset now produces:

```text
Request failed due to an HTTP/2 stream reset received from the remote peer (reason: refused stream before processing any application logic, code 7). Retrying after 1s. error_count=4 request_uri=https://example.com/api/v1/series http2.error_kind="reset" http2.reason_code=7 http2.received_from_remote=true
```

The exact field placement depends on the tracing formatter. The request URI remains sanitized.

`CategorizedError`'s `Display` implementation is covered with these exact outputs:

```text
Request failed due to an HTTP/2 GOAWAY received from the remote peer (reason: detected excessive load generating behavior, code 11)
Request failed due to an HTTP/2 stream reset received from the remote peer (reason: refused stream before processing any application logic, code 7)
Request failed due to an HTTP/2 error (reason: unexpected internal error encountered, code 2)
Request failed due to an HTTP/2 error (reason: unknown reason, code 9001)
```

## Change Type
- [x] Bug fix

## How did you test this PR?

- `make fmt`
- `make check-docs`
- `cargo check --workspace`
- Focused `cargo nextest` tests for the retry lifecycle
- `make check-all`
- `make check-release-notes`
- `.venv/bin/reno lint`

`make test-all` was started but stopped at operator request.

## References

- Closes: #2568

### DIFF
--- a/lib/saluki-io/src/net/util/retry/lifecycle/http.rs
+++ b/lib/saluki-io/src/net/util/retry/lifecycle/http.rs
@@ -1,4 +1,4 @@
-use std::{borrow::Cow, error::Error as _, fmt, time::Duration};
+use std::{borrow::Cow, fmt, time::Duration};
 
 use http::StatusCode;
 use tracing::{debug, warn};
@@ -23,7 +23,17 @@ where
         let request_uri = SanitizedRequestUri(req.uri());
         let categorized_error = CategorizedError::try_categorize(res);
 
-        warn!(error_count, %request_uri, "{}. Retrying after {:?}.", categorized_error, retry_backoff);
+        // The HTTP/2 fields are only present when the failure was an HTTP/2 error: `None` values are not recorded.
+        let http2_details = categorized_error.http2_details();
+
+        warn!(
+            error_count,
+            %request_uri,
+            http2_error_kind = http2_details.map(|details| details.kind.as_str()),
+            http2_reason_code = http2_details.and_then(|details| details.reason).map(u32::from),
+            http2_received_from_remote = http2_details.map(|details| details.received_from_remote),
+            "{}. Retrying after {:?}.", categorized_error, retry_backoff
+        );
     }
 
     fn after_success(&self, req: &http::Request<B>, _: &Result<http::Response<B2>, E>) {
@@ -55,9 +65,16 @@ impl fmt::Display for SanitizedRequestUri<'_> {
     }
 }
 
+/// Maximum number of errors visited when walking a source chain.
+///
+/// Chains are short in practice, so the limit only exists to bound the walk if an error ever reports itself, directly or
+/// indirectly, as its own source.
+const MAX_SOURCE_CHAIN_DEPTH: usize = 32;
+
 enum CategorizedError {
     Client(String),
     Tls(String),
+    Http2(Http2ErrorDetails),
     Http(StatusCode),
     Other(String),
 }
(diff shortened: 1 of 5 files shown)


lib/saluki-io/src/net/util/retry/lifecycle/http.rs
```diff
@@ -23,7 +23,17 @@ where
...
-        warn!(error_count, %request_uri, "{}. Retrying after {:?}.", categorized_error, retry_backoff);
+        // The HTTP/2 fields are only present when the failure was an HTTP/2 error: `None` values are not recorded.
+        let http2_details = categorized_error.http2_details();
+
+        warn!(
+            error_count,
+            %request_uri,
+            http2_error_kind = http2_details.map(|details| details.kind.as_str()),
+            http2_reason_code = http2_details.and_then(|details| details.reason).map(u32::from),
+            http2_received_from_remote = http2_details.map(|details| details.received_from_remote),
```
tobz: We should dot separate these for a little more structure: `http2.error_kind` instead of `http2_error_kind`, etc.

tobz (approved):
LGTM with one non-blocking nit.

lib/saluki-io/src/net/util/retry/lifecycle/http.rs
webern: Ah, makes sense: https://github.com/DataDog/saluki/pull/2572/commits/1713ab79d4e2f87ce5e6467830e88db39eccc237

PR #2582, opened 2026-09-10
fix(forwarder): count bytes after successful requests

### LINKED ISSUE #2581: Telemetry: total bytes sent overstated during retries.
Somehow this seems to have been missed among the issues I opened yesterday for the H2 retry storm thing.

Bytes sent are being accumulated in telemetry for sends that fail to reach the backend. Metrics sent (the count) however, are not (this is the correct behavior).

So the way we actually detected the retry storm was an apparent increase in "bytes per metric" i.e. total bytes / metric count. This was inflated by the retries due to incorrect accounting.

Fix this so that our bytes sent telemetry actually means bytes sent!

webern:
## Human Summary

Alters the behavior of telemetry accounting of bytes sent to not include failed transaction bytes as "sent". The background is that during a retry event, we saw an increase in bytes per metric, which was actually an artifact of the inflated counting of bytes that were never sent.

I am not entirely certain this is better than having `saluki_io::net::HttpClient` only count bytes from successful sends in the `with_bytes_sent_counter` but the models seem to agree that is not the right layer for this accounting.

## AI Summary

Move `component_bytes_sent_total` accounting from socket writes to successful forwarder transaction handling. Failed and retried attempts no longer inflate the counter.

Add regression coverage for two server-error retries followed by a successful response.

## Change Type

- [x] Bug fix

## How did you test this PR?

- `make fmt`
- `make check-docs`
- `make check-clippy`
- `cargo nextest run -p saluki-components 'common::datadog::io::tests'`
- `cargo nextest run -p saluki-components 'common::datadog::telemetry::tests'`
- `make check-release-notes`
- `.venv/bin/reno lint`

## References

- Closes: #2581

### DIFF
--- a/lib/saluki-components/src/common/datadog/io.rs
+++ b/lib/saluki-components/src/common/datadog/io.rs
@@ -65,6 +65,7 @@ type EndpointNameFn = dyn Fn(&Uri) -> Option<MetaString> + Send + Sync;
 
 struct InFlightTransaction<R> {
     metadata: Metadata,
+    body_size: u64,
     retry_counters: Option<TransactionRetryCounters>,
     result: R,
 }
@@ -120,6 +121,7 @@ async fn handle_in_flight_transaction_result<B>(
 {
     let InFlightTransaction {
         metadata,
+        body_size,
         retry_counters,
         result,
     } = match task_result {
@@ -144,7 +146,15 @@ async fn handle_in_flight_transaction_result<B>(
         // surfaced by the inspection layer in the service stack rather than here, since a retriable 403 becomes a
         // `Retry` result and never reaches this arm.
         Ok(http_response) => {
-            process_http_response(http_response, metadata, telemetry, endpoint_url, endpoint_domain).await
+            process_http_response(
+                http_response,
+                metadata,
+                body_size,
+                telemetry,
+                endpoint_url,
+                endpoint_domain,
+            )
+            .await
         }
 
         // The service itself encountered an error while sending the request or receiving the response:
@@ -368,7 +378,6 @@ where
             .with_min_tls_version(config.min_tls_version())
             .with_tls_handshake_timeout(config.tls_handshake_timeout())
             .with_http_protocol(config.http_protocol())
-            .with_bytes_sent_counter(telemetry.bytes_sent().clone())
             .with_endpoint_telemetry(
                 metrics_builder.clone(),
                 Some(move |uri: &Uri| endpoint_name_for_client(uri)),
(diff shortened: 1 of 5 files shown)


lib/saluki-components/src/common/datadog/io.rs
```diff
@@ -368,7 +378,6 @@ where
             .with_min_tls_version(config.min_tls_version())
             .with_tls_handshake_timeout(config.tls_handshake_timeout())
             .with_http_protocol(config.http_protocol())
-            .with_bytes_sent_counter(telemetry.bytes_sent().clone())
```
jszwedko, line 371: From what I can see, this function was only ever called here so we could drop it?

jszwedko (commented):
I'm of two minds on this one. I think it is actually useful to have a count of the raw bytes to understand the network traffic while it is _also_ useful to see just successfully sent bytes. I can see this PR aligns `component_bytes_sent_total` with the other `_sent_total` metrics. Maybe we could add a new metric for network bytes like `component_network_bytes_sent_total`?

Side-note that I wish we had went with Prometheus-style naming to have the unit last (i.e. `component_sent_bytes_total`) but alas.

lib/saluki-components/src/common/datadog/io.rs
webern: Yes it can be dropped 👍 

Edit: unless we are doing what you suggested in the main comment.

webern:
> I think it is actually useful to have a count of the raw bytes to understand the network traffic

I guess I don't know the use case on that one. I just saw the other side of it, where average bytes per metric was something someone was looking at and it was... wrong.

It depends on what the definition of sent is, I guess!

jszwedko:
> > I think it is actually useful to have a count of the raw bytes to understand the network traffic
> 
> I guess I don't know the use case on that one. I just saw the other side of it, where average bytes per metric was something someone was looking at and it was... wrong.
> 
> It depends on what the definition of sent is, I guess!

True! I'm ok with just adjusting it for now and we can wait until we have a use-case for "raw" sent bytes. The current implementation seems like a bug given the placement of the other counters.

webern:
> True! I'm ok with just adjusting it for now and we can wait until we have a use-case for "raw" sent bytes. The current implementation seems like a bug given the placement of the other counters.

@jszwedko Ok, I deleted that unused function here: https://github.com/DataDog/saluki/pull/2582/commits/96a21fc418a2a7eb7634efa83057ce9a1ac39f7e

jszwedko (approved):

PR #2585, opened 2026-09-10
feat(otlp): normalize span duration, start time, IDs, and type

### LINKED ISSUE #2382: Traces(otlp): Normalize OTLP span duration, start time, IDs, and type
## Background

A tracer that reports an end time before its start time makes ADP compute the duration by subtracting two `u64` timestamps, so the span carries a duration near `u64::MAX`. Other malformed fields also pass through: a zero span ID, a span that is its own parent, a `span.type` longer than the intake accepts.

## Agent behavior

`normalizeV1` in `pkg/trace/agent/normalizer.go` runs on every span of a trace chunk before the pipeline. It drops a span whose span ID is zero, and the chunk with it. It clears the parent ID when the parent ID equals the span ID. It then calls `validateAndFixDurationV1`, which returns zero for a negative duration or one where `start + duration` overflows; `validateAndFixStartTimeV1`, which replaces a start earlier than `Year2000NanosecTS` with `time.Now().UnixNano()` minus the duration; and `validateAndFixType`, which truncates `span.type` at `MaxTypeLen`, 100 bytes.

## ADP behavior

ADP has no equivalent step. `otel_to_dd_span_minimal` in `common/otlp/traces/transform.rs` computes the duration by subtracting `u64` timestamps, keeps whatever start time the span carries, and stores the parent ID from `convert_span_id` unchanged. `translate_spans` in `translator.rs` groups every span it receives, including one whose span ID converts to zero. Nothing truncates `span.type`.

## Acceptance

- [ ] An end time before the start time yields a zero duration, not a wrapped `u64`
- [ ] A start time before the year 2000 is replaced with now minus the duration
- [ ] A span whose span ID is zero is dropped, along with its trace
- [ ] A parent ID equal to the span ID is cleared to zero
- [ ] A `span.type` longer than 100 bytes is truncated

## References

Depends on: #2379

ajgajg1134:
Closes #2382

## Summary

OTLP spans converted to Datadog spans carried several malformed values that the trace-agent's normalizer would have fixed (#2382):

- A tracer reporting an **end time before its start time** wrapped the `u64` subtraction and produced a duration near `u64::MAX`
- A **start time before the year 2000** passed through untouched
- A span could be **its own parent**
- `span.type` could exceed the **100 bytes** the intake accepts
- Spans whose **span ID converts to zero** were kept, where the reference normalizer drops the span and the whole chunk containing it

There will be a follow up PR later to make this generic for the DD v1 traces pipeline as well (but that pipeline doesn't exist yet :P )

## Changes

Span-level rules are fixed inside `otel_to_dd_span_minimal`, which both the full and minimal OTLP-to-DD conversion paths funnel through, mirroring `validateAndFixDurationV1`, `validateAndFixStartTimeV1`, and `validateAndFixType` (`pkg/trace/agent/normalizer.go`):

- The duration fix deliberately uses **wrapping arithmetic** to match the reference's behavior when `start` exceeds `i64::MAX` (`saturating_sub` would invert the behavior on that boundary), and runs **before** the start fix, which uses the corrected duration — mirroring the reference's rule ordering.
- The zero-span-ID chunk drop lives in `translate_spans`, matching `normalizeTraceChunkV1`'s chunk granularity: the span is dropped along with the entire trace containing it, and the drop increments a new `component_events_dropped_total{intentional:true,drop_reason:span_id_zero}` counter.

New shared helpers (`validate_and_fix_duration`, `validate_and_fix_start_time`, `MAX_TYPE_LEN`, `YEAR_2000_NANOSEC_TS`) live in `common/otlp/traces/normalize.rs` with unit tests, including a regression test pinning the wrapping-arithmetic boundary.

## Testing

- `cargo nextest run -p saluki-components`: 1004/1004 passed (13 new tests covering all five acceptance criteria, the wrapping boundary, UTF-8-boundary truncation, chunk-drop semantics, and the drop counter's tag set)
- `make check-all`: clean (fmt, clippy, style guide, advisories, licenses, unused deps, docs, feature matrix)
- `make check-release-notes`: clean

### DIFF
--- a/lib/saluki-components/src/common/otlp/traces/normalize.rs
+++ b/lib/saluki-components/src/common/otlp/traces/normalize.rs
@@ -11,6 +11,19 @@ pub const MAX_SERVICE_LEN: usize = 100;
 pub const MAX_RESOURCE_LEN: usize = 5000;
 pub const MAX_TAG_LEN: usize = 200;
 
+/// Maximum length of `span.type`, mirroring `MaxTypeLen` (`pkg/trace/agent/normalizer.go`).
+pub const MAX_TYPE_LEN: usize = 100;
+
+// Above this, start+duration would overflow a signed 64-bit integer.
+const MAX_START_PLUS_DURATION: u64 = i64::MAX as u64;
+
+/// Below this, a start timestamp predates the reference's `Year2000NanosecTS` floor and is replaced.
+///
+/// `time.Date(2000, 1, 1, 0, 0, 0, 0, time.UTC).UnixNano()` (`pkg/trace/agent/normalizer.go`).
+/// Used as the fallback when the system clock is unavailable, so a rewritten start still clears the
+/// year-2000 floor it exists to enforce.
+pub const YEAR_2000_NANOSEC_TS: u64 = 946_684_800_000_000_000;
+
 // default service name we assign a span if it's missing and we have no reasonable fallback
 const DEFAULT_SERVICE_NAME: MetaString = MetaString::from_static("otlpresourcenoservicename");
 // default span name we assign a span if it's missing and we have no reasonable fallback
@@ -340,6 +353,48 @@ const fn is_valid_ascii_tag_char(c: char) -> bool {
     is_valid_ascii_start_char(c) || (c >= '0' && c <= '9') || c == '.' || c == '/' || c == '-'
 }
 
+/// Mirrors `validateAndFixDurationV1` (`pkg/trace/agent/normalizer.go`): a duration that would overflow
+/// a signed 64-bit integer when added to `start` becomes zero.
+///
+/// The reference computes `duration > math.MaxInt64 - uint64(start)` in wrapping `uint64` arithmetic.
+/// When `start > math.MaxInt64`, that subtraction wraps around to a very large number, so the
+/// comparison is false for any realistic duration and the duration survives untouched. This uses
+/// `wrapping_sub` rather than `saturating_sub` for exactly that reason: `saturating_sub` would clamp to
+/// zero for `start > i64::MAX`, making the comparison `duration > 0` and zeroing any non-zero
+/// duration—the opposite of the reference for that boundary. See the regression test at the boundary
+/// below.
+pub(super) fn validate_and_fix_duration(start: u64, duration: u64) -> u64 {
+    if duration > MAX_START_PLUS_DURATION.wrapping_sub(start) {
+        0
+    } else {
+        duration
+    }
+}
+
+/// Mirrors `validateAndFixStartTimeV1` (`pkg/trace/agent/normalizer.go`): a start timestamp before the
+/// year-2000 floor is replaced with the receive time minus `duration`, clamped to the receive time if
+/// that subtraction would go negative.
+pub(super) fn validate_and_fix_start_time(start: u64, duration: u64) -> u64 {
+    if start >= YEAR_2000_NANOSEC_TS {
+        return start;
+    }
+
+    let now = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| d.as_nanos() as u64)
+        // If the clock is unavailable or predates the epoch, fall back to the year-2000 floor itself
+        // rather than `0`, so the replacement still clears the floor it exists to enforce.
+        .unwrap_or(YEAR_2000_NANOSEC_TS);
+
+    let new_start = now.wrapping_sub(duration);
+    if new_start > now {
+        // Underflow: duration was larger than `now`.
+        now
+    } else {
+        new_start
+    }
+}
+
 /// Truncate string to `max_len` bytes, respecting UTF-8 boundaries.
 pub(super) fn truncate_utf8(s: &MetaString, max_len: usize) -> &str {
     truncate_utf8_str(s, max_len)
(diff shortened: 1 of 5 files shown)


lib/saluki-components/src/common/otlp/traces/normalize.rs
```diff
@@ -340,6 +353,48 @@ const fn is_valid_ascii_tag_char(c: char) -> bool {
     is_valid_ascii_start_char(c) || (c >= '0' && c <= '9') || c == '.' || c == '/' || c == '-'
 }
 
+/// Mirrors `validateAndFixDurationV1` (`pkg/trace/agent/normalizer.go`): a duration that would overflow
+/// a signed 64-bit integer when added to `start` becomes zero.
```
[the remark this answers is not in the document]
ajgajg1134: This seems wrong, there's a bunch of references in other code in this same module that links to the origin to provide info on parity of implementations.

lib/saluki-components/src/common/otlp/traces/transform.rs
```diff
@@ -366,9 +368,22 @@ pub fn otel_to_dd_span_minimal(
     let mut dd_span = DdSpan::default();
 
     let span_id = convert_span_id(&otel_span.span_id);
-    let parent_id = convert_span_id(&otel_span.parent_span_id);
+    let mut parent_id = convert_span_id(&otel_span.parent_span_id);
+
+    // Mirrors `normalizeV1` (`pkg/trace/agent/normalizer.go`): a span that is its own parent is
+    // treated as a root span.
+    if parent_id == span_id {
+        parent_id = 0;
```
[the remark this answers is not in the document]
ajgajg1134: This matches the trace-agent's existing behavior so should be kept for now.

lib/saluki-components/src/common/otlp/traces/normalize.rs
```diff
@@ -340,6 +353,48 @@ const fn is_valid_ascii_tag_char(c: char) -> bool {
...
+        // rather than `0`, so the replacement still clears the floor it exists to enforce.
+        .unwrap_or(YEAR_2000_NANOSEC_TS);
+
+    let new_start = now.wrapping_sub(duration);
+    if new_start > now {
+        // Underflow: duration was larger than `now`.
+        now
+    } else {
+        new_start
+    }
```
webern, line 395: nit: This could be a little easier to reason about like this:

 ```rust
   match now.checked_sub(duration) {
       Some(new_start) => new_start,
       None => now,
   }
 ```

Or

 ```rust
   now.checked_sub(duration).unwrap_or(now)
 ```

releasenotes/notes/otlp-traces-span-normalization-6e1d337695a39f60.yaml
```diff
@@ -0,0 +1,9 @@
+enhancements:
```
webern, line 1: nit: technically I think our stance is that we are not including these for features of the system that are not yet customer-facing, and we could remove this and use the no-changelog label instead. I'm not sure if that's a directive, like this *shouldn't* be in the changelog. Maybe it is more of a "you don't have to" guidance. Not sure, this is new in the repo.

lib/saluki-components/src/common/otlp/traces/translator.rs
```diff
@@ -233,6 +243,16 @@ impl OtlpTracesTranslator {
                     entry.priority = Some(priority as i32);
                 }
 
+                // A span ID of zero is malformed: drop the span and, like the reference normalizer,
+                // the entire chunk containing it.
+                if dd_span.span_id() == 0 {
+                    if !entry.dropped {
+                        entry.dropped = true;
+                        metrics.traces_dropped_span_id_zero().increment(1);
```
webern, line 251: > Claude Opus 5: Could this count the number of spans discarded, or otherwise use a trace-specific metric? `component_events_received_total{message_type:otlp_spans}` counts individual spans above, while this new `component_events_dropped_total` increments once per dropped trace, so the received and dropped event counters use different units. The Go agent records both one `TracesDropped.SpanIDZero` and `SpansDropped += len(chunk)`; as written here, dropping a 500-span trace would report one dropped event.

lib/saluki-components/src/common/otlp/traces/transform.rs
```diff
@@ -366,9 +368,22 @@ pub fn otel_to_dd_span_minimal(
     let mut dd_span = DdSpan::default();
 
     let span_id = convert_span_id(&otel_span.span_id);
-    let parent_id = convert_span_id(&otel_span.parent_span_id);
+    let mut parent_id = convert_span_id(&otel_span.parent_span_id);
+
+    // Mirrors `normalizeV1` (`pkg/trace/agent/normalizer.go`): a span that is its own parent is
+    // treated as a root span.
+    if parent_id == span_id {
```
webern, line 375: > Claude Opus 5: This matches `normalizeV1`, but it appears to differ from the path that Go OTLP spans actually take. `OTLPReceiver.convertSpan` produces a `pb.Span`, whose `normalize` check only clears the parent when `ParentID == TraceID && ParentID == SpanID`; the two-way check is used by the separate `normalizeV1` path. Could we confirm that adopting the V1 behavior here is intentional? This will root some spans that the current Go OTLP path leaves parented.

lib/saluki-components/src/common/otlp/traces/transform.rs
```diff
@@ -366,9 +368,22 @@ pub fn otel_to_dd_span_minimal(
...
+        parent_id = 0;
+    }
+
+    // Mirrors `validateAndFixDurationV1` and `validateAndFixStartTimeV1`, in that order: the duration
+    // is fixed first using the original start, then the start is fixed using the corrected duration.
+    // Without the duration fix, a tracer reporting an end time before its start time wraps the `u64`
+    // subtraction and the span carries a duration near `u64::MAX`.
     let start = otel_span.start_time_unix_nano;
```
webern, line 383: > Claude Opus 5: It appears that putting these fixes in the minimal conversion also changes the stats/concentrator path. The Go agent's `OtelSpanToDDSpanMinimal` passes through the raw duration, parent ID, and type; only its full ingest path subsequently runs the normalizer. Normalizing stats input may well be preferable, but could we call out or confirm this deliberate difference from the Go behavior?

webern (approved):
You now a lot more about this than I do. I left a couple of nits by hand and asked opus for an audit. You know Opus, it always finds stuff. But I leave it up to see if those are important or not.

lib/saluki-components/src/common/otlp/traces/transform.rs
ajgajg1134: Yeah this is because we haven't moved the OTLP code in the trace-agent to fully converting to the v1 pipeline but we do plan to so I think this is OK for now.

lib/saluki-components/src/common/otlp/traces/transform.rs
ajgajg1134: ooo this is interesting (and probably an oversight in the Go agent implementation). It's unlikely to be a "real" issue as it would only be triggered by a significant error in instrumentation (or malformed payloads). But I'll take a poke at it in the go trace-agent to see if it's an easy fix there

PR #2586, opened 2026-09-10
enhancement(forwarder): report http retry causes

### LINKED ISSUE #2570: Expose bounded HTTP retry causes in telemetry
## Human Intro

I feel like the model overreached a little bit below with its implementation and design ideas. But the main issue here is that our telemetry metrics do not seem to carry information as to what caused an HTTP retry.

> The following was written by GPT-5.6 Sol:

## Background

The HTTP retry lifecycle categorizes failures before it writes a warning. [`StandardHttpRetryLifecycle::before_retry`](https://github.com/DataDog/saluki/blob/c2ac928884b0ed5dfe024b2d5c815fc5b909bea6/lib/saluki-io/src/net/util/retry/lifecycle/http.rs#L15-L28) can distinguish HTTP responses, TLS failures, client failures, and uncategorized errors.

The existing endpoint telemetry exposes broad request-error buckets through `network_http_requests_errors_total`, including DNS, connection, TLS, request-write, and send failures. These registrations are defined in [`HttpTransactionErrorTelemetry`](https://github.com/DataDog/saluki/blob/c2ac928884b0ed5dfe024b2d5c815fc5b909bea6/lib/saluki-io/src/net/client/http/telemetry.rs#L40-L84). HTTP statuses receive a status-code tag through [`increment_http_error`](https://github.com/DataDog/saluki/blob/c2ac928884b0ed5dfe024b2d5c815fc5b909bea6/lib/saluki-io/src/net/client/http/telemetry.rs#L259-L270).

However, telemetry does not tie a retry to its cause. For example, HTTP/2 protocol failures, request timeouts, and connection resets can collapse into a broad send-error bucket even though the retry log has a more specific message.

## Problem

Operators can see that retries increased, but cannot use metrics to determine why. Historical investigation then depends on retaining and searching high-volume retry logs.

We should expose a bounded retry-cause dimension that answers whether retries came from HTTP status responses, TLS, HTTP/2, timeout, connection closure/reset, DNS, or another stable category. The design must avoid unbounded labels derived from error strings.

This issue concerns retry causes. Distinguishing actual retries from first attempts displaced by input overflow is tracked separately.

## Implementation Ideas

We could define a reusable, low-cardinality retry-cause enum and have both retry logging and telemetry consume it. This would reduce the risk that logs and metrics classify the same failure differently.

Possible emission points include the retry lifecycle, where the response or error is still available, or the forwarder queue path if the cause is carried with the requeued transaction. Emitting at classification time avoids reconstructing the cause after enqueue, but may require a telemetry hook that keeps the generic retry utility independent from a specific metrics backend.

The existing `domain` and `endpoint` tags are useful dimensions. The cause values should remain fixed and documented; raw status codes may remain a separate bounded tag.

## Testing Ideas

A regression test could first demonstrate that two distinct retryable failures, such as an HTTP/2 error and a timeout, increment only the same broad telemetry bucket.

Table-driven tests could then pass representative response and error chains through the production classifier/lifecycle path and assert the expected cause. Cases should include retryable HTTP statuses, non-retryable statuses, TLS, HTTP/2, timeout, connection reset, and an unknown fallback.

A cardinality test should verify that unknown error strings do not become label values.

## Method

During analysis of an HTTP retry increase, metrics established when and where retries occurred, but only retained logs showed the transport error text. The metrics could not distinguish the dominant HTTP/2 failures from a small number of request and TLS timeouts.

webern:
## Human Summary

This enhances the data that we can see when retries increase by tagging reasons. Below in the summary you can see exactly how the tags would have appeared to us during the increased `h2` retries that we investigated.

The cardinality is bounded by enums, so I think it should not be too costly. I tried to include in enough information for reviewers to understand the tag cardinality cost in this PR.

## AI Summary

Adds `network_http_requests_retry_causes_total`, tagged by domain and a bounded retry cause. HTTP/2 retries also include the frame, reason, and whether the local client or remote endpoint ended the request.

The retry lifecycle uses one error-chain classification for both logs and telemetry. It also distinguishes request timeouts, connection timeouts, and common connection failures.

### What this would have shown during the ADP 1.6.0 retry storm

The failure was h2 0.4.16 charging legitimate small DATA frames against its abuse budget, so
our own client sent GOAWAY and the interrupted requests were retried.

#### Series that would have appeared

```
network_http_requests_retry_causes_total{
  domain:https://<intake>,
  cause:http2,
  frame:go_away,
  initiator:local,
  reason:enhance_your_calm
}
```

Rising with the retry rate from the moment 1.6.0 rolled out, and stopping on rollback to 1.5.2.

The other causes stay flat, which is the part that was not visible at the time:
`cause:timeout` (`reason:request`, `reason:connect`) and `cause:tls` accounted for the small
number of unrelated failures, and `cause:http_status` for ordinary 5xx retries.

#### Questions, before and after

| Question | Before | After |
| --- | --- | --- |
| Are retries up? | `network_http_requests_retries_total` | unchanged |
| Was it us or the intake? | backend metrics, checked by hand | `initiator:local` |
| Which protocol layer failed? | retained logs | `cause:http2` |
| Why? | not available; logs said `http2 error` | `reason:enhance_your_calm` |
| Connection or single stream? | not available | `frame:go_away` |

#### Log line, same classification

Before (this PR):

```
Request failed: http2 error. Retrying after 1.5s.
```

Now (with this PR):

```
Request failed due to an HTTP/2 GOAWAY sent by this client
(reason: detected excessive load generating behavior, code 11). Retrying after 1.5s.
```

#### Limits

- The metric is per domain, so retries for `/api/v2/series` and `/api/beta/sketches` share one
  series. Route attribution stays on `network_http_requests_retries_total{endpoint}`.
- Requests already in flight when a connection is torn down can surface as hyper's
  "connection closed before message completed" instead of the h2 error, landing in `cause:client`.
  The `initiator:local` signal appears on the subset that carries the h2 error, which was enough
  here but is not a 1:1 count of affected requests.

### Tag cardinality

Series are registered lazily, on the retry path only, so a process that never retries
registers nothing.

Tags: `domain` plus `cause`; `frame` and `initiator` on `cause:http2`; `reason` on
`cause:http2` (when h2 names the code), `cause:timeout`, and `cause:connection`.

#### Common case

- No retries: 0 series.
- Occasional 5xx retries: 1 series per domain (`cause:http_status`).
- A single failure mode, such as the HTTP/2 retry storm this came from: 1–3 series per domain.

#### Worst case, per domain

| `cause` | extra tags | series |
| --- | --- | --- |
| `http_status` | — | 1 |
| `tls` | — | 1 |
| `client` | — | 1 |
| `other` | — | 1 |
| `timeout` | `reason`: `request`, `connect` | 2 |
| `connection` | `reason`: 7 I/O error kinds | 7 |
| `http2` | `frame` × `initiator` × `reason` | 75 |
| **total** | | **88** |

The HTTP/2 figure is 5 reachable `frame`/`initiator` pairs (`initiator` is `unknown` exactly
when no frame is attributable) × 15 `reason` states (14 named HTTP/2 codes, plus the tag being
absent). Multiplying the tag values without that correlation gives 148, which is the number to
use if you want a hard upper bound.

Multiply by the number of distinct intake domains: 1 for a normal deployment, 2–3 with dual
shipping or multi-region failover. So the practical ceiling is roughly 100–300 series, and only
if every failure mode occurs.

#### What keeps it bounded

- The counter map is keyed by an enum of enums, so the tag combinations are fixed at compile time.
- HTTP/2 reason codes that h2 does not name are omitted from the tags rather than stringified;
  the numeric code stays in the retry log.
- Error text, GOAWAY debug data, rustls error variants, and response status codes are never tags.
  Status code breakdowns already exist on `network_http_requests_errors_total`.

## Change Type

- [x] New feature

## How did you test this PR?

- `make fmt`
- `make check-docs`
- `make check-clippy`
- `make check-release-notes`
- `.venv/bin/reno lint`
- `cargo nextest run -p saluki-io -p saluki-components`

Tests reproduce a locally initiated HTTP/2 `GOAWAY(ENHANCE_YOUR_CALM)` by exhausting h2's small-DATA-frame budget and verify the resulting metric tags.

## References

- Closes: #2570


lib/saluki-io/src/net/util/retry/lifecycle/http.rs
```diff
@@ -186,54 +272,227 @@ impl Http2ErrorKind {
...
+        match kind {
+            io::ErrorKind::ConnectionRefused => Some(Self::Refused),
+            io::ErrorKind::ConnectionReset => Some(Self::Reset),
+            io::ErrorKind::ConnectionAborted => Some(Self::Aborted),
+            io::ErrorKind::BrokenPipe => Some(Self::BrokenPipe),
+            io::ErrorKind::NotConnected => Some(Self::NotConnected),
+            io::ErrorKind::UnexpectedEof => Some(Self::UnexpectedEof),
+            io::ErrorKind::HostUnreachable | io::ErrorKind::NetworkUnreachable | io::ErrorKind::NetworkDown => {
+                Some(Self::Unreachable)
+            }
```
jszwedko: I think it might actually be useful to distinguish these three 🤔 For example, the host network being down vs. some network interrupt in the middle would be useful to distinguish when troubleshooting.

lib/saluki-io/src/net/util/retry/lifecycle/http.rs
```diff
@@ -29,11 +49,15 @@ where
         warn!(
             error_count,
             %request_uri,
-            http2.error_kind = http2_details.map(|details| details.kind.as_str()),
+            http2.frame = http2_details.map(|details| details.frame.as_str()),
```
jszwedko, line 52: Does this still preserve all of the information we previously had in the log? It's a bit unclear to me what `frame` and `initiator` render as.

lib/saluki-io/src/net/util/retry/lifecycle/telemetry.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /// Counts a single retry.
+    pub(super) fn increment(&self, cause: RetryCause) {
+        let mut counters = self.counters.lock().unwrap();
+        counters
+            .entry(cause)
+            .or_insert_with(|| {
+                self.builder
+                    .register_counter_with_tags(NETWORK_HTTP_REQUESTS_RETRY_CAUSES_TOTAL, cause.tags())
```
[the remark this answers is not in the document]
jszwedko: I could see an argument to pass this through RAR so that it ends up in the Core Agent `/telemetry` too, but I don't think we need to add it to `/compat/metrics` since this is a net-new metric.

jszwedko (commented):
A couple of comments, otherwise 👍

lib/saluki-io/src/net/util/retry/lifecycle/http.rs
webern: Yes. By the way, the old version of that line is from a PR this week as well, but yes, preserved and enhanced!

- `http2.frame`: `"go_away"`, `"reset"`, or `"other"`
- `http2.initiator`: `"remote"`, `"local"`, or `"unknown"`

### Remote GOAWAY

**Before**

```text
WARN error_count=1 request_uri=https://example.com/api/v2/series http2.error_kind="go_away" http2.reason_code=11 http2.received_from_remote=true Request failed due to an HTTP/2 GOAWAY received from the remote peer (reason: detected excessive load generating behavior, code 11). Retrying after 1ms.
```

**After**

```text
WARN error_count=1 request_uri=https://example.com/api/v2/series http2.frame="go_away" http2.reason_code=11 http2.initiator="remote" Request failed due to an HTTP/2 GOAWAY received from the remote peer (reason: detected excessive load generating behavior, code 11). Retrying after 1ms.
```

### Locally initiated GOAWAY

**Before**

```text
WARN error_count=1 request_uri=https://example.com/api/v2/series http2.error_kind="go_away" http2.reason_code=11 http2.received_from_remote=false Request failed due to an HTTP/2 GOAWAY (reason: detected excessive load generating behavior, code 11). Retrying after 1ms.
```

**After**

```text
WARN error_count=1 request_uri=https://example.com/api/v2/series http2.frame="go_away" http2.reason_code=11 http2.initiator="local" Request failed due to an HTTP/2 GOAWAY sent by this client (reason: detected excessive load generating behavior, code 11). Retrying after 1ms.
```

### HTTP/2 error with no frame attribution

**Before**

```text
WARN error_count=1 request_uri=https://example.com/api/v2/series http2.error_kind="other" http2.reason_code=2 http2.received_from_remote=false Request failed due to an HTTP/2 error (reason: unexpected internal error encountered, code 2). Retrying after 1ms.
```

**After**

```text
WARN error_count=1 request_uri=https://example.com/api/v2/series http2.frame="other" http2.reason_code=2 http2.initiator="unknown" Request failed due to an HTTP/2 error (reason: unexpected internal error encountered, code 2). Retrying after 1ms.
```

lib/saluki-io/src/net/util/retry/lifecycle/http.rs
webern: Good idea: https://github.com/DataDog/saluki/pull/2586/commits/f6b8e41f6842f7052d3115fd2ecd813509c6657e

lib/saluki-io/src/net/util/retry/lifecycle/http.rs
jszwedko: Ah, great, thanks! The example log lines are helpful.

jszwedko (approved):

PR #2587, opened 2026-09-10
chore(config): expose datadog schema defaults

### LINKED ISSUE #2580: Config: make Datadog schema defaults accessible
## Human Intro

We do not have a `pub` way to access the default of a `DatadogConfiguration` field. Well, technically we could construct `DatadogConfiguration::default()` and get whatever we want from it, but it is a large object and we would not want to do that a bunch of times during translation or just to get one field from it.

The proposal below is to expose a pub static const `SCHEMA_DEFAULTS` which is just a `DatadogConfiguration::defualt()` lazy-locked value that you can get `&whatever.you.want` from.

Not a bad idea.

> The following was written by GPT 5.4 Sol:

## Background

`DatadogConfiguration` is generated from the vendored Datadog configuration schema. Its `Default` implementation therefore provides typed schema-derived defaults and already delegates to the generated default helpers: [`DatadogConfiguration::default()`](https://github.com/DataDog/saluki/blob/faebdbd0af8f3b51cd3b629e9bcc30bf7af3e0d3/lib/datadog-agent/config/src/generated/datadog_configuration.rs#L570-L713).

Translation sometimes needs the schema default when recovering from an invalid explicit value. The generated per-field helpers exist, but they are typify implementation details with restricted visibility and inconsistent shapes: [generated defaults](https://github.com/DataDog/saluki/blob/faebdbd0af8f3b51cd3b629e9bcc30bf7af3e0d3/lib/datadog-agent/config/src/generated/datadog_configuration.rs#L2068-L2110).

## Problem

Callers can construct `DatadogConfiguration::default()` whenever they need one field, but this allocates and initializes the full model each time. Exposing typify's helper functions would couple callers to generated implementation details. Without a stable accessor, translation code may instead repeat schema-owned values as literals, which can drift when the vendored schema changes.

We should provide a reusable, typed, immutable view of the generated defaults.

## Implementation Ideas

A small hand-written module in `datadog-agent-config` could lazily construct one default model and expose it through an inherent method:

```rust
use std::sync::LazyLock;

use crate::DatadogConfiguration;

static SCHEMA_DEFAULTS: LazyLock<DatadogConfiguration> =
    LazyLock::new(DatadogConfiguration::default);

impl DatadogConfiguration {
    /// Returns configuration populated with defaults from the vendored schema.
    pub fn schema_defaults() -> &'static Self {
        &SCHEMA_DEFAULTS
    }
}
```

Translation code could then read the field it needs without repeating its value:

```rust
let default = DatadogConfiguration::schema_defaults().forwarder_backoff_base;

let value = if value <= 0.0 {
    warn!("`forwarder_backoff_base` is not positive ({value}); using {default}.");
    default
} else {
    value
};
```

The accessor should remain outside generated code. This keeps typify's helpers private, gives callers one stable API, and constructs the generated default model once. Callers that need owned strings or collections can clone only the selected field.

## Testing Ideas

- Deserialize an empty object and compare its serialized value with `DatadogConfiguration::schema_defaults()` to verify that both paths resolve the same defaults.
- Verify repeated calls return the same reference with `std::ptr::eq`.
- Update translation fallback tests to derive expected values from `schema_defaults()` rather than repeating schema-owned literals.

webern:
## Human Summary

A quick follow up to #2562 to use the schema-generated default mechanism instead of magic numbers for clamping fallback values in the translation layer. 

## AI Summary

Expose a shared `DatadogConfiguration` populated from generated schema defaults. Use those defaults when translation rejects invalid forwarder backoff values.

## Change Type

## How did you test this PR?

- `make fmt`
- `make check-docs`
- `make check-clippy`
- `cargo nextest run -p datadog-agent-config -p agent-data-plane-config-system`

## References

- Closes: #2580

### DIFF
--- a/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
+++ b/lib/agent-data-plane-config-system/src/translators/datadog_translator.rs
@@ -778,11 +778,12 @@ impl DatadogConfigWitness for DatadogTranslator<'_> {
     }
 
     fn consume_forwarder_backoff_base(&mut self, value: f64) {
+        // TODO(#2580): the minimum is a literal here; the schema should carry it, ideally as a
+        // mechanical validation
         let value = if value <= 0.0 {
-            warn!("`forwarder_backoff_base` is not positive ({value}); using 2.");
-            // TODO(#2580): either use the schema default here, or upstream a validation to the
-            // schema itself
-            2.0
+            let default = DatadogConfiguration::schema_defaults().forwarder_backoff_base;
+            warn!("`forwarder_backoff_base` is not positive ({value}); using {default}.");
+            default
         } else {
             value
         };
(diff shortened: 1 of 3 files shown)


jszwedko (approved):
Nice!

PR #2597, opened 2026-09-14
chore(config): reduce generated diff churn

### LINKED ISSUE #2579: Config: make generated codefile diffs less noisy in PRs
Initially, what we were doing with the config system was too scary to not see the generated code diffs each time they changed, but we have since gained confidence. It is probably time to move some of code files that generated for config support either:
- out of the tree entirely (by gitignore or by generating out-of-tree), or
- Marked as machine generated so that GitHub does not show them in the PR diff automatically

Probably some combination of both of those things depending on the codefile and how useful looking at it might be when determining if a diff is sound.

PR feedback:

> Nit: same here, suggestion to make these generated files as generated to hide them from GitHub diffs by default.

_Originally posted by @jszwedko in https://github.com/DataDog/saluki/pull/2562#discussion_r3971814691_

webern:
## Human Summary

In #2562 @jszwedko noted that we should cut down on diff noise from machine generated files. I have looked at this and decided on a few things.
- I think it is important that we see schema diffs since the schema should not be hand-edited and it only diffs when we update it, which is in a PR scoped to that purpose.
- I also want to see `datadog_configuration.rs` which is equally crucial and should only change when the schema changes or we change the generating code.

The following are marked as machine generated to suppress diffs in PRs, but still tracked in tree so that we can see diffs locally or expand the diff if needed in a PR:
- `witness.rs`
- `classifier_data.rs`
- `env_keys.rs`

The config registry is now `.gitignore`d and generated into a `generated/` dir/mod. It's there for us to look at in editors and not off in build dirs, but we don't need to see the diff. Note the registry is probably not long for this word. Early on, in #1788 I described its removal. It's hanging on only because I want to make sure I think about what it was providing for us and replicating it's value somehow before forgetting about it forever. 

Overall the diffs touching config should be a lot less noisy after this PR.

## AI Summary

Reduce generated configuration diff churn while keeping schema and configuration model changes visible.

- Mark the generated witness, classifier data, and environment key tables as generated for GitHub diffs.
- Generate config registry tables under an ignored `generated/` module instead of committing them.
- Move the generated registry schema from `OUT_DIR` into the same ignored module.
- Format generated Rust before comparing it with tracked output, then watch tracked output so warm builds repair edits and expose drift to the existing CI check.

## Change Type

## How did you test this PR?

Passed:

- `make build-schema-overlay`
- `make fmt`
- `make check-docs`
- `make check-clippy`
- 89 tests across `datadog-agent-config`, `datadog-agent-config-overlay-model`, and `datadog-agent-config-testing`

Also verified that repeated builds do not rewrite generated files, deleted ignored output regenerates, and a hand edit to tracked generated output is repaired with a warm Cargo target.

## References

- Closes: #2579


.agents/skills/config-system/SKILL.md
```diff
@@ -106,6 +106,18 @@ Use the following command to regenerate it:
 make build-schema-overlay
 ```
 
+How much of that output a reviewer sees varies by file:
```
webern: Delete. There should be no diff to this skill in this PR.

jszwedko (approved):
Thanks for the clean-up!

Issue #2438, opened 2026-08-21
Traces: OTLP and Datadog-native trace parity


webern:
This is a tracking issue for tracing signals. We want ADP to ingest traces at transparent feature parity with the Datadog Agent, for both OTLP clients and Datadog-native traces. 

It is a sub-issue of #2005, which tracks OTLP ingest as a whole. TBD how we should break trackers down between OTLP and DD-native.

These are listed in a rough implementation order that should plausibly work, and groupings may plausibly be parallelisable (hard to say but I tried to make that work).

## Traces

- [ ] #2264
- [ ] #2268
- [ ] #2278

Emit OTLP http.status_code as a numeric span attribute:

- [x] #2379
- [x] #2413
- [ ] #2382
- [x] #2421
- [ ] #2383
- [x] #2422
- [x] #2393
- [x] #2394
- [ ] #2395
- [ ] #2414
- [ ] #2403

Match credit-card range, allowlist, and span-event coverage:

- [ ] #2402
- [ ] #2404
- [ ] #2405
- [ ] #2410
- [ ] #2411
- [ ] #2412

Support probabilistic hash seeds and full trace IDs:

- [ ] #2406
- [ ] #2407
- [ ] #2408
- [ ] #2409

Report panics as a metric and throttle repeated writer logs:

- [ ] #2415
- [ ] #2416
- [ ] #2417

Re-resolve the API key from the secrets backend after a 403:

- [ ] #2418
- [ ] #2419
- [ ] #2425

Wire OTLP span naming and validation settings:

- [ ] #2331
- [ ] #2332

Complete OTLP trace resource metadata extraction:

- [ ] #2334
- [ ] #2335
- [ ] #2338
- [ ] #2396
- [ ] #2397
- [ ] #2398
- [ ] #2399
- [ ] #2400

Honor APM writer tuning in trace and stats writers:

- [ ] #2340
- [ ] #2374

Tag onboarding metadata only on traces the sampler keeps:

- [ ] #2375
- [ ] #2376

No dependencies:

- [ ] #2401
- [ ] #2330
- [ ] #2420
- [ ] #2426
- [ ] #2391
- [ ] #2392
- [ ] #2423

## Integration

Add multi-region failover for APM payloads:

- [ ] #2429
- [ ] #2430

Read and normalize `apm_config.mode` and `apm_config.features`:

- [ ] #2427
- [ ] #2428

Build the remote-config client and product gating:

- [ ] #2431
- [ ] #2432
[...]

Issue #2528, opened 2026-09-02
Add windows support to integration tests that run datadog-intake


webern:
**<sub><sub>![P2 Badge](https://img.shields.io/badge/P2-yellow?style=flat)</sub></sub>  Restrict intake sidecars to Linux targets**

When a test enables `intake` and opts into the `windows` runtime, this starts `DriverConfig::datadog_intake`, whose default container OS creates the isolation-group network with Docker's Linux `bridge` driver. `build_driver_config` subsequently configures the Windows target for the same group, and Airlock requires its `nat` driver; because the bridge network already exists, the Windows target cannot join it. Reject `intake.enabled` for Windows (as the Unix runner does for macOS), or provide a Windows-compatible sidecar/network path.

Useful? React with 👍 / 👎.

_Originally posted by @chatgpt-codex-connector[bot] in https://github.com/DataDog/saluki/pull/2506#discussion_r3895841077_

Issue #2405, opened 2026-08-21
Traces(otlp+dd): Obfuscate malformed JSON and preserve key order and formatting


webern:
## Background

This obfuscator serves Elasticsearch, OpenSearch, MongoDB, and execution plans. Malformed JSON is returned untouched today, so raw query values reach the backend. Valid JSON is re-serialized through a `BTreeMap`, so keys come back alphabetized and whitespace is lost, which changes the resource string that stats aggregate on.

## Agent behavior

`json_scanner.go:40-96` is a per-byte state machine over strings, numbers, literals, and escapes. `jsonObfuscator.obfuscate` (`json.go:129-233`) copies input bytes to the output as it scans, keeps whitespace, replaces each non-key value with `?`, keeps the whole subtree under a `keepKeys` key, and passes the string value after a `transformKeys` key through the SQL transformer, which substitutes a fixed failure message when obfuscation fails. On a scan error, or an error at EOF, it appends `...` and returns what it has.

## ADP behavior

`JsonObfuscator::obfuscate` (`json.rs:30-42`) parses with `serde_json::from_str`, rewrites the `Value` tree, and re-serializes. A parse error returns the input unchanged, so nothing is obfuscated. `preserve_order` is off in the workspace, so objects round-trip through a `BTreeMap` and keys are alphabetized, and a failed `obfuscate_sql_string` falls through to `?`.

## Acceptance

- [ ] Malformed JSON returns obfuscated output up to the error, ending in `...`
- [ ] Object key order and input whitespace survive obfuscation
- [ ] A `keep_values` key keeps its whole subtree unobfuscated
- [ ] A failed SQL transform emits the Agent's failure message instead of `?`

## References

Depends on: #2404

PR #2599, opened 2026-09-14
fix(trace-obfuscation): match reference HTTP URL scrubbing

### LINKED ISSUE #2404: Traces(otlp+dd): Match Agent handling of encoded URL digits and parse failures
## Background

Query, path, and userinfo scrubbing already match for absolute URLs. Three edge cases differ: a percent-encoded digit escapes redaction, an unparseable URL is scrubbed more than the Agent scrubs it, and a relative `http.url` such as `/users/123/profile` is replaced wholesale with `?`, so the endpoint is lost from the trace.

## Agent behavior

`ObfuscateURLString` (`http.go:23-58`) calls `obfuscateUserInfo` when neither option is set, which returns the input unchanged if `url.Parse` fails; with an option set, a parse failure returns `?`. Otherwise it clears userinfo, force-adds `?` when dropping a non-empty query, and splits the decoded `u.Path`, replacing any segment that holds an ASCII digit with a `/REDACTED/` placeholder that is rewritten to `?`. Go's `url.Parse` accepts relative references such as `/users/123`.

## ADP behavior

`obfuscate_url` (`http.rs:10-45`) already matches on userinfo clearing, query removal, and returning `?` when parsing fails with an option enabled. `has_non_encoded_digit` (`http.rs:75-87`) skips `%XX` triplets in the percent-encoded `url.path()`, `obfuscate_userinfo_fallback` (`http.rs:49-58`) strips userinfo from unparseable values, and `Url::parse` rejects relative URLs.

## Configuration keys

- [ ] apm_config.obfuscation.http.remove_paths_with_digits
- [ ] apm_config.obfuscation.http.remove_query_string

## Acceptance

- [ ] A percent-encoded digit such as `%32` in a path segment triggers redaction
- [ ] An unparseable URL is unchanged when only userinfo removal applies
- [ ] A relative URL redacts its digit segments instead of becoming `?`
- [ ] Absolute-URL query, userinfo, and digit behavior is unchanged

## References

Depends on: #2402

webern:
## Human Summary

This is a first case of using `libdatadog` to deduplicate shared algorithms. In this case, the HTTP URL scrubber. At first this PR was blocked by its introduction of `serde feature="preserve_order"` which would have affected our workspace. Fortunately it was not needed and upstream is responsive so I fixed that with PR https://github.com/DataDog/libdatadog/pull/2529.

There are a few outstanding findings where the upstream algorithm differs from the Agent, and those are tracked with issues and I think I have PRs for them already. I think this PR can be merged and we will pick up those fixes when they land.

Speaking of which, I checked and apparently Renovate will bump these shas for us when it does its updates. That is both good and bad:
- Good: we won't be stale
- Bad: there could be breaking changes at any time. Later we may want to lock to versioned libdatadog releases, but for now we are landing related changes and moving fast.

## AI Summary

This replaces Saluki's HTTP URL scrubber with a small adapter around `libdd_trace_obfuscation::http`. Using the shared implementation fixes two gaps in the current code:

- Relative URLs such as `/users/123/profile` retain the endpoint and become `/users/?/profile`.
- Percent-encoded digits are decoded before path scanning, so `%32` triggers digit redaction.

Saluki keeps its existing safety behavior for an unparseable URL when both options are disabled: credentials are removed while the rest of the value is retained. This deliberately differs from the Datadog Agent, which returns that URL unchanged, including its credentials.

The span transform first uses libdatadog's allocation-free precheck on the borrowed URL. It copies and rewrites the value only when the precheck finds a possible trigger, and it updates the span only when the output differs.

The pinned libdatadog revision has three known URL differences from the Agent. Tests document the current outputs until the upstream fixes are available:

- [DataDog/libdatadog#2522](https://github.com/DataDog/libdatadog/issues/2522): percent-escaped unreserved path characters are decoded.
- [DataDog/libdatadog#2523](https://github.com/DataDog/libdatadog/issues/2523): brackets in a path or fragment can collapse the URL to `?` ([draft fix](https://github.com/DataDog/libdatadog/pull/2527)).
- [DataDog/libdatadog#2524](https://github.com/DataDog/libdatadog/issues/2524): the precheck can miss an unparseable authority that the rewrite would replace with `?` ([draft fix](https://github.com/DataDog/libdatadog/pull/2528)).

The dependency is pinned to `e0a0134b` with default features disabled, so it does not add another TLS stack. It adds 19 lockfile entries and requires git-source and `MIT-0` license allowances. It also enables `serde_json/preserve_order` across the workspace. That changes `serde_json::Map` from key-sorted to insertion-ordered and affects JSON output outside URL scrubbing. A test makes the ordering change visible, but this PR remains blocked on [DataDog/libdatadog#2525](https://github.com/DataDog/libdatadog/issues/2525) and its [draft fix](https://github.com/DataDog/libdatadog/pull/2529) so the feature can be removed upstream before this dependency lands.

## Change Type

- [x] Bug fix

## How did you test this PR?

Unit tests cover both parity fixes, the deliberate unparseable-URL behavior, existing absolute URLs, unchanged-output suppression, the three known libdatadog gaps, and the workspace JSON key-order effect.

The following checks pass after rebasing onto current `main`:

- `make fmt`
- `cargo check --workspace`
- `make check-docs`
- `make check-clippy`
- `cargo nextest run -p saluki-components trace_obfuscation::http` (20 tests)
- `make check-release-notes`
- `make check-deny`

## References

- Closes: #2404
- Progresses: #2438


webern:
> AI finding 1:
>
> ## Path percent-escapes are decoded, so URLs the Agent leaves byte-identical get rewritten
>
> **Severity: blocking (parity regression, all four config combinations, including the default).**
>
> `libdd_trace_obfuscation::http::obfuscate_url_string` runs `normalize_pct_encoded_unreserved` on the parsed path
> unconditionally, at both emit sites (`libdd-trace-obfuscation/src/http.rs:283` and `:298` at the pinned rev
> `e0a0134b`), decoding `%XX` whenever the byte is `A-Za-z0-9`, `.`, `_` or `~`:
>
> ```rust
> // libdd-trace-obfuscation/src/http.rs:42-66 (pinned rev)
> fn normalize_pct_encoded_unreserved(path: &str) -> String { ... }
> ```
>
> The reference implementation does not do this. `url.Parse` keeps the raw path, and `u.String()` emits it verbatim
> whenever it is a valid encoding of the decoded path; it re-escapes from the decoded path only when the raw form is
> unusable (non-ASCII or Go "category 1" bytes) or when the path was rewritten by the redaction loop
> (`net/url`: `setPath`, `EscapedPath`, `validEncoded`, driven from `pkg/obfuscate/http.go:23-58`).
>
> Differential run (a verbatim Go port of `pkg/obfuscate/http.go` on go1.26.1, versus the pinned libdd rev composed
> exactly as this PR composes it: the `mod.rs` screen, then `obfuscate_url`, then the `obfuscated == val` suppression):
>
> ```text
> config             input                                     Agent                                     this PR
> qs=F dg=F (default) https://example.com/a/%2E%2E/b           https://example.com/a/%2E%2E/b            https://example.com/a/../b
> qs=F dg=F (default) https://example.com/%7Euser/index.html   https://example.com/%7Euser/index.html    https://example.com/~user/index.html
> qs=F dg=F (default) https://example.com/path/%41BC/x         https://example.com/path/%41BC/x          https://example.com/path/ABC/x
> qs=T dg=F           https://example.com/users/%32/profile    https://example.com/users/%32/profile     https://example.com/users/2/profile
> ```
>
> The pre-change implementation matched the Agent on all four (`Url::parse` -> `changed == false` -> `None`), so this is
> a regression introduced here, not a pre-existing gap.
>
> Impact:
>
> - Under the default configuration, where the only documented effect is userinfo removal, ADP now overwrites
>   `http.url` on any span whose URL contains a percent-escaped unreserved character.
> - `%2E%2E` becomes `..`, so path-traversal probe traffic is silently normalized before it reaches the backend.
> - With `remove_paths_with_digits` off, `%32` is decoded to `2`: the change reveals digits rather than redacting them,
>   which is the opposite direction from the rest of this PR.
>
> Same family, same root cause: the pre-encode pass at `libdd-trace-obfuscation/src/http.rs:233-241` escapes category-1
> characters across the whole path region including the authority, where Go escapes only the path. With both options
> off, `http://foo bar.com/x` becomes `http://foo%20bar.com/x`; the Agent returns it unchanged.
>
> Remediation: fix it upstream and re-pin, that is, leave the path untouched when the redaction loop did not change it
> and the raw path is a valid encoding (Go's `validEncoded` rule). Until the pin moves, this divergence belongs in the
> PR description and the release note as a known gap, with a test that pins the current output so the eventual upstream
> fix is visible in the diff.

webern:
> AI finding 2:
>
> ## A `[` or `]` in the path or fragment now destroys the whole URL
>
> **Severity: blocking (over-redaction regression, whenever either option is on).**
>
> The pre-encode pass only escapes `[` and `]` when the path already contains a category-1 or non-ASCII character
> (`needs_full_path`, `libdd-trace-obfuscation/src/http.rs:23-30`, `:233-241`). Otherwise the brackets reach
> `UriRef::parse`, which rejects them (RFC 3986 forbids `[`/`]` outside an IP-literal host), so the parse-failure branch
> at `libdd-trace-obfuscation/src/http.rs:260-266` returns `"?"` whenever `remove_query_string` or
> `remove_paths_with_digits` is on.
>
> ```text
> config     input                                    Agent                             this PR   pre-PR
> qs=T dg=F  https://example.com/api/items[1]/detail  https://example.com/api/items[1]/detail  ?    https://example.com/api/items[1]/detail
> qs=F dg=T  https://example.com/api/items[1]/detail  https://example.com/api/?/detail         ?    https://example.com/api/?/detail
> qs=T dg=T  https://example.com/api/items[1]/detail  https://example.com/api/?/detail         ?    https://example.com/api/?/detail
> qs=T dg=F  http://foo.com/x#a[b                     http://foo.com/x#a[b                     ?    http://foo.com/x#a[b
> ```
>
> Go's `url.Parse` accepts brackets in a path and a fragment, and `validEncoded` explicitly allows them
> ("not specified in RFC 3986 but left alone by modern browsers"), so the Agent keeps the endpoint and redacts only the
> digit segment. The pre-change ADP code did the same. This PR loses the endpoint for the whole class, which is the
> exact failure mode #2404 was filed to remove ("a relative URL ... is replaced wholesale, so the endpoint is lost").
> Bracketed path segments and bracketed SPA fragments are ordinary production URLs, so this is not a fuzz-only case.
>
> Remediation: upstream, percent-encode every character `UriRef` rejects before parsing (brackets included) rather than
> only the category-1/category-2 sets, then re-pin. A regression test on
> `https://example.com/api/items[1]/detail` under all four option combinations would have caught it.

webern:
> AI finding 3:
>
> ## The precheck's documented guarantee does not hold, and malformed authorities escape the Agent's fail-safe
>
> **Severity: should fix (false invariant in a doc comment; fail-safe parity gap whenever either option is on).**
>
> `lib/saluki-components/src/transforms/trace_obfuscation/http.rs:12-19` states:
>
> > The scan is a conservative superset of the rewrite's triggers: `true` does not promise that the URL changes, but
> > `false` does promise that it does not.
>
> The pinned revision does not provide that. `should_obfuscate_url`
> (`libdd-trace-obfuscation/src/http.rs:126-197`) has no trigger for an authority that fails to parse, while
> `obfuscate_url_string` replaces exactly those inputs with `"?"`:
>
> ```text
> should_obfuscate_url("https://example.com:port/x", true, false)  == false
> obfuscate_url_string("https://example.com:port/x", true, false)  == "?"
> ```
>
> Same for `http://foo:bar.com/x` and the bare `":"`. Because `mod.rs:104` skips on `false`, and `libdd::obfuscate_url`
> applies the same screen internally, ADP forwards these URLs untouched:
>
> ```text
> config     input                        Agent  this PR                      pre-PR
> qs=T dg=F  https://example.com:port/x   ?      https://example.com:port/x   ?
> qs=F dg=T  http://foo:bar.com/x         ?      http://foo:bar.com/x         ?
> qs=T dg=T  :                            ?      :                            ?
> ```
>
> The `"?"` is a deliberate fail-safe in the reference implementation ("should not happen for valid URLs, but better
> obfuscate everything rather than expose sensitive information when this option is on",
> `pkg/obfuscate/http.go:29-33`), and the pre-change code honored it. Fuzzing found the same shape for authorities
> containing a space, `^` or `|`.
>
> Remediation: fix the screen upstream so it admits anything the transform would rewrite (a `:` in the authority
> region is enough), then re-pin. Until then, do not document a guarantee the dependency does not make: say that
> `false` means "no trigger byte found", and note that the two libdd entry points disagree on this class.

webern:
> AI finding 4:
>
> ## The unparseable-URL case is a deliberate libdd divergence, not Agent parity, and #2404's acceptance item is not met
>
> **Severity: non-blocking, but the release note and the commit message tell users the opposite of what ships.**
>
> With both options off, the reference implementation returns the input verbatim on a parse failure, credentials
> included (`pkg/obfuscate/http.go:16-21`: `if err != nil { return val }`). Verified against a verbatim Go port
> (go1.26.1):
>
> ```text
> input:  http://user:password@foo.com/%
> Agent:  ["http://user:password@foo.com/%", "?", "?", "?"]   # (qs=F dg=F), (qs=T dg=F), (qs=F dg=T), (qs=T dg=T)
> this PR ["http://foo.com/%",               "?", "?", "?"]
> ```
>
> The stripping comes from libdd's `strip_userinfo_best_effort` (`libdd-trace-obfuscation/src/http.rs:69-99`), added for
> APMSP-3086 precisely because credentials surviving that fallback was considered a bug in the reference behavior. Its
> own tests name it a regression fix, not parity.
>
> I think the shipped behavior is the right one: credentials should not survive obfuscation. The problem is that
> everything around it claims parity:
>
> - the release note says the unparseable URL "keeps everything but its credentials" as one of three "matches the
>   Datadog Agent" cases (`releasenotes/notes/fix-http-url-obfuscation-parity-de993f1778594148.yaml`);
> - the commit message and PR body list it as a case that "differed from the reference implementation";
> - #2404's acceptance item reads "An unparseable URL is unchanged when only userinfo removal applies", which the Agent
>   satisfies by keeping the credentials. Ticking that box as written is not accurate.
> - the test name `unparseable_url_only_loses_userinfo` documents libdd behavior as if derived from the Agent
>   (`lib/saluki-components/src/transforms/trace_obfuscation/http.rs:64-82`).
>
> Remediation: keep the behavior, restate it. Call it out as an intentional, safer divergence in the release note and in
> the PR body, and close the #2404 item with that wording so the next parity audit does not read the box as "verified
> identical to the Agent".

webern:
> AI finding 5:
>
> ## The new dependency turns on `serde_json/preserve_order` for the entire workspace
>
> **Severity: blocking until deliberately accepted; the change is repo-wide, invisible in the diff, and untested.**
>
> `libdd-trace-obfuscation` requires `serde_json` with `preserve_order`:
>
> ```toml
> # libdd-trace-obfuscation/Cargo.toml:17 (pinned rev e0a0134b)
> serde_json = { workspace = true, features = ["preserve_order"] }
> ```
>
> Cargo unifies features, so every crate in this workspace now gets it:
>
> ```text
> $ cargo tree -e features -p saluki-components
> ...
> |   +-- serde_json feature "preserve_order"
> |   |   +-- serde_json v1.0.151
> |   |   +-- serde_json feature "indexmap"
> ```
>
> and `Cargo.lock` confirms it (this PR adds `indexmap` to `serde_json`'s dependency list; it is absent on `main`).
>
> Effect: `serde_json::Map` becomes an insertion-ordered `IndexMap` instead of a key-sorted `BTreeMap` for all of ADP.
> Every place that builds or round-trips a JSON object changes its output byte order, for example:
>
> - `lib/saluki-components/src/transforms/trace_obfuscation/json.rs:36-42`, the Mongo/Elasticsearch/OpenSearch query
>   obfuscator, parses and re-serializes the document, so obfuscated query strings on spans change key order;
> - `lib/saluki-components/src/sources/otlp/logs/transform.rs:93` builds a `serde_json::Map` for log attributes;
> - `lib/saluki-config/src/lib.rs:531-558` and the config env providers build `Value::Object` trees that are serialized
>   for snapshots and API output.
>
> Nothing in the test suite catches it: `json.rs`'s tests compare parsed `Value`s, which are order-insensitive. Lookup
> complexity and allocation behavior of `Map` change too (hash map plus vector instead of a B-tree).
>
> This may well be harmless or even desirable, but it is a workspace-wide semantic change arriving as a side effect of
> a URL-scrubbing fix, and no reviewer would see it in the diff.
>
> Remediation: decide it explicitly. Either state it in the PR body and the release note (and add a test that pins the
> ordering ADP now emits), or ask upstream to make `preserve_order` an optional feature of
> `libdd-trace-obfuscation`, or keep the ~250 lines of URL logic in ADP and skip the dependency.

webern:
> AI finding 6:
>
> ## One 250-line function pulls in a second, unused copy of the whole trace obfuscator
>
> **Severity: non-blocking, but it is the reviewer's call, and the data to make it is not in the PR.**
>
> `libdd-trace-obfuscation` is not a leaf crate for `http`. At the pinned revision it ships
> `credit_cards.rs`, `ip_address.rs`, `json/`, `memcached.rs`, `redis.rs`, `redis_tokenizer.rs`, `replacer.rs`,
> `sql.rs` and `obfuscation_config.rs`, and it depends on `libdd-trace-protobuf`, `libdd-trace-utils`,
> `libdd-trace-normalization`, `libdd-common`, `libdd-capabilities`, `libdd-tinybytes`, `rmpv`, `serde-transcode`,
> `nix` and `thin-vec` (`Cargo.lock`, 19 new packages). ADP already has its own implementations of every one of those
> obfuscators in `lib/saluki-components/src/transforms/trace_obfuscation/`, so the tree now carries two copies of the
> same logic, one of them unreferenced.
>
> The PR body says "Binary size impact is not measured". Given that `.agents/skills/binary-size-analysis/SKILL.md`
> exists precisely for this question, measuring it is cheap next to the risk of finding out later.
>
> There is also a direction question worth answering once, in this PR, rather than per-function later: is the plan to
> migrate ADP's trace obfuscation to `libdd-trace-obfuscation` wholesale (in which case a shared crate for one function
> is a reasonable first step and the duplication is temporary), or to keep ADP's own implementations (in which case
> vendoring the ~250 lines of `http.rs` costs less than a pinned git dependency, its 19 crates, its
> `serde_json/preserve_order` requirement, and the `deny.toml` exception)?
>
> Remediation: run the binary-size comparison against `main` and put the delta in the PR body, and state the intended
> end state for the other obfuscators.

webern:
> AI finding 7:
>
> ## Minor: the screening scan runs twice, and the equality-suppression branch has no test
>
> **Severity: preference, non-blocking.**
>
> `libdd::obfuscate_url` already runs `should_obfuscate_url` internally
> (`libdd-trace-obfuscation/src/http.rs:201-215`), so after the screen in
> `lib/saluki-components/src/transforms/trace_obfuscation/mod.rs:104` every URL that is admitted is scanned twice.
> Measured on 200 synthetic API URLs at the pinned rev, the scan is about 50 ns of a roughly 530 ns obfuscation, so
> around 10% of the work on the path that matters. Calling `http::obfuscate_url_string` from the wrapper instead of
> `http::obfuscate_url` keeps one scan and the existing `obfuscated == val` check still supplies the `None`.
>
> For what it is worth, the screen itself is a clear win where it rejects: with both options off, the same 200 URLs go
> from about 220 ns (pre-PR) to about 63 ns, because nothing is admitted.
>
> Second, the `obfuscated == val` branch at
> `lib/saluki-components/src/transforms/trace_obfuscation/http.rs:32-34` is the one piece of logic this module adds on
> top of the dependency, and no test exercises it: every `None` assertion in the test module is produced by libdd's
> internal screen instead. An input that is admitted by the screen and comes back unchanged, for example
> `http://foo.com/foo%20bar/` (the `%` admits it, `%20` is not decoded), pins that branch.

webern:
> AI finding 8:
>
> ## Correction to AI finding 6: CI has already measured the binary size
>
> **Severity: informational; this supersedes the "measure it" request in AI finding 6.**
>
> I missed that the Binary Size Analysis comment on this PR already carries the number:
> **+439.53 KiB, +1.10%** (39.20 MiB baseline vs 39.63 MiB), inside the +5% threshold. Its module table also
> corroborates AI finding 5's footprint claim independently of `Cargo.lock`:
>
> ```text
> indexmap                 +36.67 KiB   (111 symbols)   <- serde_json/preserve_order pulled in
> hashbrown                +36.38 KiB  (1809 symbols)
> regex_syntax             +55.68 KiB   (341 symbols)
> libdd_trace_obfuscation  +18.51 KiB     (5 symbols)
> ```
>
> Note the shape of that: the `http` module ADP actually calls accounts for 18.51 KiB of it, while the dependency's
> incidental baggage (`indexmap`, `hashbrown`, `regex_syntax`) accounts for well over 100 KiB. Dead-code elimination
> does drop libdd's unused obfuscators, so the duplicate-implementation concern in AI finding 6 is a maintainability
> question rather than a size one.
>
> The remaining actionable part of AI finding 6 stands: quote the +1.10% in the PR body next to "Binary size impact is
> not measured", and state whether ADP is heading toward `libdd-trace-obfuscation` for the other obfuscators or
> keeping its own.

Cargo.toml
```diff
@@ -102,6 +102,10 @@ protobuf = { version = "3.7", default-features = false, features = [
 protobuf-codegen = { version = "3.7", default-features = false }
 piecemeal = { git = "https://github.com/tobz/piecemeal" }
 piecemeal-build = { git = "https://github.com/tobz/piecemeal" }
+# Datadog's shared trace obfuscation crate, consumed from GitHub because the version we need is not published to
+# crates.io yet, and pinned to a specific revision so upstream changes never land implicitly. `default-features =
+# false` drops the `https` feature, which would otherwise pull a second TLS stack in through `libdd-common`.
```
webern: Delete this comment.

```suggestion
```

releasenotes/notes/fix-http-url-obfuscation-parity-de993f1778594148.yaml
webern, line 1: Delete this file, not needed for unpublished functionality.

webern:
> > AI finding 1:
> > ## Path percent-escapes are decoded, so URLs the Agent leaves byte-identical get rewritten

@webern created https://github.com/DataDog/libdatadog/issues/2522 and will track it in our own project tracking, potentially for an upstream PR.

webern:
> AI
>
> > AI finding 2:
> >
> > ## A `[` or `]` in the path or fragment now destroys the whole URL
>
> This one looks like it has to be fixed in the dependency, so I filed it upstream as
> [DataDog/libdatadog#2523](https://github.com/DataDog/libdatadog/issues/2523), "Obfuscation: a `[` or `]` in a URL
> path or fragment makes `obfuscate_url_string` return `?`". Remediation is tracked there, and the issue is a
> sub-issue of #2438.
>
> In this PR I only pinned the current output in a test, so the upstream fix shows up as a diff when the pin moves.

webern:
> AI
>
> > AI finding 3:
> >
> > ## The precheck's documented guarantee does not hold, and malformed authorities escape the Agent's fail-safe
>
> The screen itself has to change in the dependency, so I filed that half upstream as
> [DataDog/libdatadog#2524](https://github.com/DataDog/libdatadog/issues/2524), "Obfuscation:
> `should_obfuscate_url` returns false for URLs `obfuscate_url_string` replaces with `?`". Remediation is tracked
> there, and the issue is a sub-issue of #2438.
>
> Over the audit corpus this class is 267 of 160,484 input/config combinations, and every one of them is an
> authority that fails to parse (`:`, `http://foo.com:abc/x`, `https://example.com:port/x`), so I think the shape
> upstream is exactly what the finding describes.
>
> In this PR I dropped the doc comment's guarantee and pinned the gap in a test.

webern:
> AI
>
> > AI finding 5:
> >
> > ## The new dependency turns on `serde_json/preserve_order` for the entire workspace
>
> I filed the "ask upstream" option as
> [DataDog/libdatadog#2525](https://github.com/DataDog/libdatadog/issues/2525),
> "`libdd-trace-obfuscation` forces `serde_json/preserve_order` on every dependent workspace", a sub-issue of #2438:
> nothing in that crate's `src/` mentions `serde_json::Map`, `Value::Object` or `preserve_order`, so I think the
> feature may not be needed at all there.
>
> Should we accept it in the meantime? For the one obfuscator it touches, insertion order is arguably the more
> faithful behavior: `pkg/obfuscate/json.go` streams over the raw JSON and never reorders keys, so
> `trace_obfuscation/json.rs` emitting input order is closer to the Agent than the sorted order it emitted before.
> This PR now says so in the body and pins that ordering in a test.

webern:
> AI
>
> Fixed in [63afe08](https://github.com/DataDog/saluki/commit/63afe08d9a19a708b459339469a4d52a497277ed),
> pushed to this branch.
>
> - **Finding 4**: the release note and the PR body now call the unparseable-URL case a deliberate, safer divergence
>   rather than parity, and the test is renamed `unparseable_url_loses_userinfo_by_design`. #2404's acceptance item 2
>   still reads as parity in the issue text; I left the issue alone, so you may want to close that box with the
>   divergence wording.
> - **Finding 7**: `obfuscate_url` now calls `http::obfuscate_url_string`, so the scan runs once, and
>   `unchanged_rewrite_reports_no_change` covers the `obfuscated == val` branch with `http://foo.com/foo%20bar/`.
>   Across the 162,568 input/config combinations of the audit corpus the span path produces identical output before
>   and after that swap.
> - **Findings 1, 2, 3**: only the parts that live in this repo. The doc comment no longer claims a guarantee the
>   dependency does not make, each divergence has a pinning test, the first two are known issues in the release note,
>   and the PR body has a "Known gaps" section. The fixes belong upstream: DataDog/libdatadog#2522,
>   DataDog/libdatadog#2523, DataDog/libdatadog#2524.
> - **Finding 5**: `test_obfuscate_preserves_key_order` in `trace_obfuscation/json.rs` pins the ordering and the PR
>   body records the workspace-wide feature; DataDog/libdatadog#2525 carries the upstream ask.
>
> `make test` (2968 tests) passes, as do `check-fmt`, `check-clippy`, `check-docs`, `check-licenses`,
> `check-unused-deps` and `check-release-notes`. `make check-deny` fails on the rustls 0.23.43 advisory
> RUSTSEC-2026-0285, which fails the same way on `main`.

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
```diff
@@ -1,436 +1,374 @@
...
+/// Returns `Some(obfuscated)` if the URL changed, `None` if it did not. Userinfo is always removed; the query string
+/// and path digits are removed per `config`. A URL that cannot be parsed loses its userinfo when both `config` options
+/// are off, and is replaced with `?` when either option is on.
+///
+/// # Design
+///
+/// This calls the rewrite directly rather than `libdd_trace_obfuscation::http::obfuscate_url`, which screens with
+/// [`should_obfuscate_url`] first: the span path has already screened by the time it gets here, and the equality check
+/// below is exact where the screen is approximate.
+pub fn obfuscate_url(val: &str, config: &HttpObfuscationConfig) -> Option<MetaString> {
```
webern, line 17: > AI finding 1:
>
> ## Path percent-escapes are decoded, so URLs the Agent leaves byte-identical get rewritten
>
> **Severity: blocking (parity regression, all four config combinations, including the default).**
>
> `libdd_trace_obfuscation::http::obfuscate_url_string` runs `normalize_pct_encoded_unreserved` on the parsed path
> unconditionally, at both emit sites (`libdd-trace-obfuscation/src/http.rs:283` and `:298` at the pinned rev
> `e0a0134b`), decoding `%XX` whenever the byte is `A-Za-z0-9`, `.`, `_` or `~`:
>
> ```rust
> // libdd-trace-obfuscation/src/http.rs:42-66 (pinned rev)
> fn normalize_pct_encoded_unreserved(path: &str) -> String { ... }
> ```
>
> The reference implementation does not do this. `url.Parse` keeps the raw path, and `u.String()` emits it verbatim
> whenever it is a valid encoding of the decoded path; it re-escapes from the decoded path only when the raw form is
> unusable (non-ASCII or Go "category 1" bytes) or when the path was rewritten by the redaction loop
> (`net/url`: `setPath`, `EscapedPath`, `validEncoded`, driven from `pkg/obfuscate/http.go:23-58`).
>
> Differential run (a verbatim Go port of `pkg/obfuscate/http.go` on go1.26.1, versus the pinned libdd rev composed
> exactly as this PR composes it: the `mod.rs` screen, then `obfuscate_url`, then the `obfuscated == val` suppression):
>
> ```text
> config             input                                     Agent                                     this PR
> qs=F dg=F (default) https://example.com/a/%2E%2E/b           https://example.com/a/%2E%2E/b            https://example.com/a/../b
> qs=F dg=F (default) https://example.com/%7Euser/index.html   https://example.com/%7Euser/index.html    https://example.com/~user/index.html
> qs=F dg=F (default) https://example.com/path/%41BC/x         https://example.com/path/%41BC/x          https://example.com/path/ABC/x
> qs=T dg=F           https://example.com/users/%32/profile    https://example.com/users/%32/profile     https://example.com/users/2/profile
> ```
>
> The pre-change implementation matched the Agent on all four (`Url::parse` -> `changed == false` -> `None`), so this is
> a regression introduced here, not a pre-existing gap.
>
> Impact:
>
> - Under the default configuration, where the only documented effect is userinfo removal, ADP now overwrites
>   `http.url` on any span whose URL contains a percent-escaped unreserved character.
> - `%2E%2E` becomes `..`, so path-traversal probe traffic is silently normalized before it reaches the backend.
> - With `remove_paths_with_digits` off, `%32` is decoded to `2`: the change reveals digits rather than redacting them,
>   which is the opposite direction from the rest of this PR.
>
> Same family, same root cause: the pre-encode pass at `libdd-trace-obfuscation/src/http.rs:233-241` escapes category-1
> characters across the whole path region including the authority, where Go escapes only the path. With both options
> off, `http://foo bar.com/x` becomes `http://foo%20bar.com/x`; the Agent returns it unchanged.
>
> Remediation: fix it upstream and re-pin, that is, leave the path untouched when the redaction loop did not change it
> and the raw path is a valid encoding (Go's `validEncoded` rule). Until the pin moves, this divergence belongs in the
> PR description and the release note as a known gap, with a test that pins the current output so the eventual upstream
> fix is visible in the diff.

See https://github.com/DataDog/libdatadog/issues/2522

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
```diff
@@ -1,436 +1,374 @@
...
+    let obfuscated = http::obfuscate_url_string(val, config.remove_query_string, config.remove_paths_with_digits);
```
webern, line 18: > AI finding 2:
>
> ## A `[` or `]` in the path or fragment now destroys the whole URL
>
> **Severity: blocking (over-redaction regression, whenever either option is on).**
>
> The pre-encode pass only escapes `[` and `]` when the path already contains a category-1 or non-ASCII character
> (`needs_full_path`, `libdd-trace-obfuscation/src/http.rs:23-30`, `:233-241`). Otherwise the brackets reach
> `UriRef::parse`, which rejects them (RFC 3986 forbids `[`/`]` outside an IP-literal host), so the parse-failure branch
> at `libdd-trace-obfuscation/src/http.rs:260-266` returns `"?"` whenever `remove_query_string` or
> `remove_paths_with_digits` is on.
>
> ```text
> config     input                                    Agent                             this PR   pre-PR
> qs=T dg=F  https://example.com/api/items[1]/detail  https://example.com/api/items[1]/detail  ?    https://example.com/api/items[1]/detail
> qs=F dg=T  https://example.com/api/items[1]/detail  https://example.com/api/?/detail         ?    https://example.com/api/?/detail
> qs=T dg=T  https://example.com/api/items[1]/detail  https://example.com/api/?/detail         ?    https://example.com/api/?/detail
> qs=T dg=F  http://foo.com/x#a[b                     http://foo.com/x#a[b                     ?    http://foo.com/x#a[b
> ```
>
> Go's `url.Parse` accepts brackets in a path and a fragment, and `validEncoded` explicitly allows them
> ("not specified in RFC 3986 but left alone by modern browsers"), so the Agent keeps the endpoint and redacts only the
> digit segment. The pre-change ADP code did the same. This PR loses the endpoint for the whole class, which is the
> exact failure mode #2404 was filed to remove ("a relative URL ... is replaced wholesale, so the endpoint is lost").
> Bracketed path segments and bracketed SPA fragments are ordinary production URLs, so this is not a fuzz-only case.
>

See https://github.com/DataDog/saluki/issues/2438
> Remediation: upstream, percent-encode every character `UriRef` rejects before parsing (brackets included) rather than
> only the category-1/category-2 sets, then re-pin. A regression test on
> `https://example.com/api/items[1]/detail` under all four option combinations would have caught it.

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
```diff
@@ -1,436 +1,374 @@
...
 
-fn obfuscate_path_digits(path: &str) -> Option<String> {
```
webern: > AI finding 3:
>
> ## The precheck's documented guarantee does not hold, and malformed authorities escape the Agent's fail-safe
>
> **Severity: should fix (false invariant in a doc comment; fail-safe parity gap whenever either option is on).**
>
> `lib/saluki-components/src/transforms/trace_obfuscation/http.rs:12-19` states:
>
> > The scan is a conservative superset of the rewrite's triggers: `true` does not promise that the URL changes, but
> > `false` does promise that it does not.
>
> The pinned revision does not provide that. `should_obfuscate_url`
> (`libdd-trace-obfuscation/src/http.rs:126-197`) has no trigger for an authority that fails to parse, while
> `obfuscate_url_string` replaces exactly those inputs with `"?"`:
>
> ```text
> should_obfuscate_url("https://example.com:port/x", true, false)  == false
> obfuscate_url_string("https://example.com:port/x", true, false)  == "?"
> ```
>
> Same for `http://foo:bar.com/x` and the bare `":"`. Because `mod.rs:104` skips on `false`, and `libdd::obfuscate_url`
> applies the same screen internally, ADP forwards these URLs untouched:
>
> ```text
> config     input                        Agent  this PR                      pre-PR
> qs=T dg=F  https://example.com:port/x   ?      https://example.com:port/x   ?
> qs=F dg=T  http://foo:bar.com/x         ?      http://foo:bar.com/x         ?
> qs=T dg=T  :                            ?      :                            ?
> ```
>
> The `"?"` is a deliberate fail-safe in the reference implementation ("should not happen for valid URLs, but better
> obfuscate everything rather than expose sensitive information when this option is on",
> `pkg/obfuscate/http.go:29-33`), and the pre-change code honored it. Fuzzing found the same shape for authorities
> containing a space, `^` or `|`.
>
> Remediation: fix the screen upstream so it admits anything the transform would rewrite (a `:` in the authority
> region is enough), then re-pin. Until then, do not document a guarantee the dependency does not make: say that
> `false` means "no trigger byte found", and note that the two libdd entry points disagree on this class.

Cargo.toml
```diff
@@ -102,6 +102,10 @@ protobuf = { version = "3.7", default-features = false, features = [
...
+libdd-trace-obfuscation = { git = "https://github.com/DataDog/libdatadog.git", rev = "e0a0134bbf0896603db615344e9b2950e7a2b2b9", default-features = false }
```
webern: > AI finding 5:
>
> ## The new dependency turns on `serde_json/preserve_order` for the entire workspace
>
> **Severity: blocking until deliberately accepted; the change is repo-wide, invisible in the diff, and untested.**
>
> `libdd-trace-obfuscation` requires `serde_json` with `preserve_order`:
>
> ```toml
> # libdd-trace-obfuscation/Cargo.toml:17 (pinned rev e0a0134b)
> serde_json = { workspace = true, features = ["preserve_order"] }
> ```
>
> Cargo unifies features, so every crate in this workspace now gets it:
>
> ```text
> $ cargo tree -e features -p saluki-components
> ...
> |   +-- serde_json feature "preserve_order"
> |   |   +-- serde_json v1.0.151
> |   |   +-- serde_json feature "indexmap"
> ```
>
> and `Cargo.lock` confirms it (this PR adds `indexmap` to `serde_json`'s dependency list; it is absent on `main`).
>
> Effect: `serde_json::Map` becomes an insertion-ordered `IndexMap` instead of a key-sorted `BTreeMap` for all of ADP.
> Every place that builds or round-trips a JSON object changes its output byte order, for example:
>
> - `lib/saluki-components/src/transforms/trace_obfuscation/json.rs:36-42`, the Mongo/Elasticsearch/OpenSearch query
>   obfuscator, parses and re-serializes the document, so obfuscated query strings on spans change key order;
> - `lib/saluki-components/src/sources/otlp/logs/transform.rs:93` builds a `serde_json::Map` for log attributes;
> - `lib/saluki-config/src/lib.rs:531-558` and the config env providers build `Value::Object` trees that are serialized
>   for snapshots and API output.
>
> Nothing in the test suite catches it: `json.rs`'s tests compare parsed `Value`s, which are order-insensitive. Lookup
> complexity and allocation behavior of `Map` change too (hash map plus vector instead of a B-tree).
>
> This may well be harmless or even desirable, but it is a workspace-wide semantic change arriving as a side effect of
> a URL-scrubbing fix, and no reviewer would see it in the diff.
>
> Remediation: decide it explicitly. Either state it in the PR body and the release note (and add a test that pins the
> ordering ADP now emits), or ask upstream to make `preserve_order` an optional feature of
> `libdd-trace-obfuscation`, or keep the ~250 lines of URL logic in ADP and skip the dependency.

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
webern: Resolving as this is tracked upstream now and in our project (and as we are at parity with libdatadog)

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
webern: Resolved as tracked upstream and in our tracking issue.

Cargo.toml
webern: > AI
>
> Fixed in [f0136f8](https://github.com/DataDog/saluki/commit/f0136f8f7b02a7b449a1a866aa15ad10c62018eb).
>
> Removed the dependency comment.

releasenotes/notes/fix-http-url-obfuscation-parity-de993f1778594148.yaml
webern: > AI
>
> Fixed in [f0136f8](https://github.com/DataDog/saluki/commit/f0136f8f7b02a7b449a1a866aa15ad10c62018eb).
>
> Removed the release note and applied `changelog/no-changelog`.

Cargo.toml
webern: > AI
>
> Fixed in [4e8af05](https://github.com/DataDog/saluki/commit/4e8af053bf87b63a0ba843cfb11594a66659b25f), by way of upstream [DataDog/libdatadog#2529](https://github.com/DataDog/libdatadog/pull/2529), which is merged.
>
> Upstream dropped the feature from the crate rather than us accepting it here:
>
> ```diff
> -serde_json = { workspace = true, features = ["preserve_order"] }
> +serde_json.workspace = true
> ```
>
> and added a `[[bans.features]]` entry to libdatadog's `deny.toml` denying `serde_json/preserve_order`, so `cargo deny check bans` there fails if it comes back. This commit moves the pin from `e0a0134b` to `3c2afd0`, the merge commit of that PR.
>
> Verification in this branch:
>
> ```text
> $ cargo tree -e features -p saluki-components | grep -c preserve_order
> 0
> $ cargo tree -e features --workspace | grep -c preserve_order
> 0
> $ cargo tree -e features -p agent-data-plane | grep -c preserve_order
> 0
> ```
>
> `Cargo.lock`'s `serde_json` entry no longer lists `indexmap`, and is now byte-identical to `main`'s. `make check-deny` and `make check-licenses` are both clean; `LICENSE-3rdparty.csv` needs no change.
>
> One consequence worth calling out: `json.rs` parses into a `Map` and re-serializes, so its output goes back to sorted keys. `test_obfuscate_preserves_key_order` was pinning the insertion order the leaked feature produced, so it is now `pinned_gap_key_order_is_sorted`, written like the other known-gap tests and pointing at #2405, which tracks the scanner port that keeps the input's key order.
>
> Would it be worth adding the same `[[bans.features]]` guard to Saluki's `deny.toml`? Happy to do that here or in a follow-up if you want it.

lib/saluki-components/src/transforms/trace_obfuscation/mod.rs
```diff
@@ -94,11 +94,19 @@ impl TraceObfuscation {
...
+            Some(v) if !v.is_empty() => v,
             _ => return,
         };
 
+        // Screen the borrowed URL first: obfuscating it needs an owned copy to release the borrow on the span, and most
+        // URLs have nothing to obfuscate. The screen is approximate in one direction, so a URL whose authority cannot
+        // be parsed is forwarded instead of being redacted wholesale; see `http::should_obfuscate_url`.
+        if !self.obfuscator.should_obfuscate_url(url) {
+            return;
+        }
```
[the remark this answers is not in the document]
webern: > AI
>
> Good catch, agreed. Fixed in [7c185ca](https://github.com/DataDog/saluki/commit/7c185ca83f83fbce9bad27d0ef343a3283bc4fa1).
>
> The screen is gone; every span URL goes to the rewrite now, which is what `main` does today. The precheck was only there to avoid copying the URL out of the span, and that copy turned out to be unnecessary: the rewrite reads the borrowed URL and hands back an owned string, so the borrow ends before the insert needs the span mutably.
>
> Reproduced first at the span level, against the pushed head before the fix:
>
> ```text
> assertion `left == right` failed: expected wholesale redaction for "https://example.com:port/x"
>   left: "https://example.com:port/x"
>  right: "?"
> ```
>
> That test is now in `trace_obfuscation/mod.rs`: `unparseable_url_is_redacted_on_the_span` covers `https://example.com:port/x`, `http://foo:bar.com/x` and `:` across all three enabled option combinations, and `unparseable_url_is_kept_when_both_options_are_off` pins the both-off case, where only userinfo is stripped. `cargo nextest run -p saluki-components` is green (986 tests), along with `make fmt`, `make check-docs`, `make check-clippy` and `cargo check --workspace`.
>
> The upstream scan is still wrong and is tracked in [DataDog/libdatadog#2524](https://github.com/DataDog/libdatadog/issues/2524), with a draft fix in [#2528](https://github.com/DataDog/libdatadog/pull/2528). I did not want to wait on that, since not calling it is both correct and simpler. The cost is that a URL with nothing to redact now pays for a parse instead of a byte scan; that is what `main` did before this PR, and the optimization can come back once the upstream scan is a real superset of the rewrite's triggers.
>
> I also dropped the tests that asserted the scan's own behavior, since nothing in ADP calls it anymore, and kept their rewrite assertions.

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
```diff
@@ -1,436 +1,357 @@
...
+//! Two behaviors of the shared crate diverge from the reference implementation. Each one is tracked upstream and
+//! pinned by a test below, so the output moves visibly in a diff when the pinned revision moves:
+//!
+//! - A percent-escaped unreserved character in the path is decoded, so a URL the reference implementation emits
+//!   byte-identical is rewritten (<https://github.com/DataDog/libdatadog/issues/2522>).
+//! - A `[` or `]` in the path or fragment fails the crate's stricter parse, so the URL collapses to `?` whenever
+//!   either config option is on (<https://github.com/DataDog/libdatadog/issues/2523>).
+//!
+//! Removing the userinfo from a URL that cannot be parsed is a deliberate divergence, not a gap: the reference
+//! implementation returns such a URL verbatim, credentials included.
```
webern: Bad clankie.

```suggestion
```

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
```diff
@@ -1,436 +1,357 @@
...
+/// are off, and is replaced with `?` when either option is on.
+///
+/// # Design
+///
+/// This calls the rewrite directly rather than `libdd_trace_obfuscation::http::obfuscate_url`, which screens with
+/// `libdd_trace_obfuscation::http::should_obfuscate_url` first. That scan allocates nothing, but it is not a superset of
+/// the rewrite's triggers: it finds nothing to do in a URL whose authority cannot be parsed, such as
+/// `https://example.com:port/x`, which the rewrite replaces with `?` when either option is on and which the reference
+/// implementation redacts (<https://github.com/DataDog/libdatadog/issues/2524>). Screening with it would forward those
+/// URLs unchanged, so every URL reaches the rewrite and the exact equality check below.
```
webern: Bad clankie

```suggestion
```

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
```diff
@@ -1,436 +1,357 @@
...
-            }
-        }
-    }
-    None
-}
-
-fn obfuscate_path_digits(path: &str) -> Option<String> {
-    let needs_obfuscation = path.split('/').any(has_non_encoded_digit);
-    if !needs_obfuscation {
+    // The rewrite normalizes escaping even where nothing sensitive was found, so it can hand back a string equal to its
```
webern: Don't use the term "rewrite". It makes no sense. Reword this. The "implementation"? The "algorithm"? The rewritten URL?

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
```diff
@@ -1,436 +1,357 @@
...
+    // implementation emits them byte-identical. Pinned so the upstream fix shows up here as a diff.
     #[test]
-    fn test_both_no_digits_query_removed() {
-        let config = HttpObfuscationConfig {
-            remove_query_string: true,
-            remove_paths_with_digits: true,
-        };
-        let result = obfuscate_url("http://foo.com/name/id?query=search", &config);
-        assert_eq!(result.unwrap().as_ref(), "http://foo.com/name/id?");
+    fn pinned_gap_path_escapes_are_decoded() {
```
webern: Don't pin undesired behavior with a test. This is the upstream's concern.

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
```diff
@@ -1,436 +1,357 @@
...
+    // implementation keeps the endpoint and redacts only the digit segment. Pinned so the upstream fix is visible.
     #[test]
-    fn test_both_digits_and_query_removed() {
-        let config = HttpObfuscationConfig {
-            remove_query_string: true,
-            remove_paths_with_digits: true,
-        };
-        let result = obfuscate_url("http://foo.com/id/123/page/1?search=bar&page=2", &config);
-        assert_eq!(result.unwrap().as_ref(), "http://foo.com/id/?/page/??");
+    fn pinned_gap_brackets_collapse_the_url() {
```
webern: Do not test undesired behavior of the upstream. Delete this test.

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
```diff
@@ -1,436 +1,357 @@
...
+                assert_eq!(
+                    obfuscate(url, remove_query_string, remove_paths_with_digits).as_deref(),
+                    Some("?"),
+                    "expected wholesale redaction for {url:?}"
+                );
+            }
+        }
     }
 
+    // Every URL reaches the rewrite now, including ones a byte scan would not flag, so pin that the rewrite still finds
```
webern: "now"? Is this a news article? Do not "announce" stuff like this. Write stateless comments that describe what code does, not what's different about it from yesterday or your last push.

lib/saluki-components/src/transforms/trace_obfuscation/json.rs
```diff
@@ -100,6 +100,19 @@ mod tests {
...
+    // reference implementation streams over the raw JSON and keeps the input's key order. Key order is part of the
+    // output and every other test here compares parsed values, which are order-insensitive, so pin it here: porting
+    // the scanner closes this gap and the change shows up as a diff.
+    #[test]
+    fn pinned_gap_key_order_is_sorted() {
+        let json = r#"{"z": 1, "m": {"y": 2, "a": 3}, "a": 4}"#;
+        let result = obfuscate_json_string(json, &[], &[], &default_sql_config());
+
+        assert_eq!(result, r#"{"a":"?","m":{"a":"?","y":"?"},"z":"?"}"#);
+    }
```
webern: Delete this test. Do not pin undesired upstream behavior in a test.

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
webern: > AI
> 
> Fixed in [666b42b](https://github.com/DataDog/saluki/commit/666b42b854c701f7f67bf62df896e14adce2c573).
> 
> Deleted the `# Missing` section and the sentence after it.

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
webern: > AI
> 
> Fixed in [666b42b](https://github.com/DataDog/saluki/commit/666b42b854c701f7f67bf62df896e14adce2c573).
> 
> Deleted the `# Design` section.

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
webern: > AI
> 
> Fixed in [666b42b](https://github.com/DataDog/saluki/commit/666b42b854c701f7f67bf62df896e14adce2c573).
> 
> Reworded to "the algorithm" here, and swept the term out of the rest of the comments
> and the test names in `http.rs` and `mod.rs`.

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
webern: > AI
> 
> Fixed in [666b42b](https://github.com/DataDog/saluki/commit/666b42b854c701f7f67bf62df896e14adce2c573).
> 
> Deleted. The gap stays tracked upstream in
> [DataDog/libdatadog#2522](https://github.com/DataDog/libdatadog/issues/2522).

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
webern: > AI
> 
> Fixed in [666b42b](https://github.com/DataDog/saluki/commit/666b42b854c701f7f67bf62df896e14adce2c573).
> 
> Deleted the test. The gap stays tracked upstream in
> [DataDog/libdatadog#2523](https://github.com/DataDog/libdatadog/issues/2523).

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
webern: > AI
> 
> Fixed in [666b42b](https://github.com/DataDog/saluki/commit/666b42b854c701f7f67bf62df896e14adce2c573).
> 
> Reworded the comment to describe what the test checks, without announcing the change
> that motivated it.

lib/saluki-components/src/transforms/trace_obfuscation/json.rs
webern: > AI
> 
> Fixed in [666b42b](https://github.com/DataDog/saluki/commit/666b42b854c701f7f67bf62df896e14adce2c573).
> 
> Deleted the test. Key order stays with [#2405](https://github.com/DataDog/saluki/issues/2405).

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
```diff
@@ -1,436 +1,292 @@
 //! HTTP URL obfuscation.
+//!
+//! The algorithm lives in `libdd_trace_obfuscation::http`, a shared crate that tracks the reference implementation's
+//! URL handling: its escaping rules, its acceptance of relative references, and its decoding of percent-encoded path
+//! characters before the digit scan. This module adapts that leaf API to [`HttpObfuscationConfig`] and
+//! [`MetaString`].
```
webern:
```suggestion
//! characters before the digit scan.
```

lucastemb (approved):
Looks clean. The idea to use libdatadog for obfuscation was a good call out from Andrew. 👍

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
Eldolfin, line 1: I don't think it makes sense to keep all these tests here now that the implementation doesn't live in this repo anymore. If we break these on our side it would take a release and bump on your side to realize that.

Do these tests exist in libdatadog implementation ? If so they can be removed from here

Eldolfin (approved):
Cool to see a new user of libdatadog 😎

lib/saluki-components/src/transforms/trace_obfuscation/http.rs
webern: Makes sense, dropped those that seem duplicative here https://github.com/DataDog/saluki/pull/2599/commits/dd437d278622d82b5fdfe3273801b0be878c2bb8

Issue #2412, opened 2026-08-21
Traces(otlp+dd): Add the bounded SQL obfuscation result cache


webern:
## Background

A service usually sends the same few query shapes over and over, and SQL obfuscation is the most expensive step in the trace path. Without a result cache, ADP re-tokenizes every duplicate query, so CPU cost scales with span volume rather than with query variety.

## Agent behavior

`ObfuscateSQLStringWithOptions` (`sql.go:314-349`) hashes the query together with the serialized options, returns a cached `ObfuscatedQuery` on a hit, and stores each successful result with `oq.Cost()` as its weight. `cache.go:34-82` builds the cache only when it is enabled, bounds it by the configured maximum size, and reports hit and miss counts.

## ADP behavior

`transforms/trace_obfuscation/` calls `obfuscate_sql_string` for every SQL span with no memoization. There is no cache, no per-entry cost accounting, and no hit or miss telemetry, so identical resources are re-tokenized on every span.

## Configuration keys

- [ ] apm_config.obfuscation.cache.enabled
- [ ] apm_config.obfuscation.cache.max_size

## Acceptance

- [ ] Repeated identical queries are obfuscated once while cached
- [ ] The cache is bypassed entirely when `cache.enabled` is false
- [ ] Entries are evicted so the total byte cost stays under `max_size`
- [ ] Hit and miss counts are reported as metrics
- [ ] A changed obfuscation option produces a separate cache entry

## References

Depends on: #2411

Issue #2411, opened 2026-08-21
Traces(otlp+dd): Honor apm_config.obfuscation_mode for SQL obfuscation


webern:
## Background

Operators use `obfuscation_mode` to choose how much of a query survives: obfuscation alone, normalization alone, or both. ADP ignores the setting, so a customer who configured normalization sees the legacy tokenizer output instead, and query shapes differ between the Agent and ADP for the same traffic.

## Agent behavior

`ObfuscateSQLStringWithOptions` (`sql.go:314-349`) branches on `opts.ObfuscationMode`: a non-empty mode calls `ObfuscateWithSQLLexer` (`sql.go:470-539`), which uses go-sqllexer to obfuscate, normalize, or do both, while an empty mode falls back to `obfuscateSQLString` and the legacy tokenizer. The mode is validated when the config is loaded (`config.go:141-162`).

## ADP behavior

`transforms/trace_obfuscation/sql.rs` always runs its own tokenizer, the equivalent of the Agent's empty-mode path. No configuration field carries the mode, and there is no normalize-only or obfuscate-and-normalize behavior to select.

## Configuration keys

- [ ] apm_config.obfuscation_mode

## Acceptance

- [ ] An empty `obfuscation_mode` keeps the current tokenizer output
- [ ] `obfuscate_only` replaces literals without rewriting query shape
- [ ] `normalize_only` normalizes shape while leaving literals in place
- [ ] `obfuscate_and_normalize` applies both, in that order
- [ ] An unrecognized mode is rejected at startup, not per query

## References

Depends on: #2410

Issue #2410, opened 2026-08-21
Traces(otlp+dd): Collapse SQL value groups and retry the escape dialect


webern:
## Background

The obfuscated query becomes the span resource and the APM stats resource dimension. A multi-row insert therefore produces a distinct resource per row count, splitting one operation into many, and a query whose escape dialect is guessed wrong is discarded as non-parsable instead of being obfuscated.

## Agent behavior

`groupingFilter` (`sql.go:225-264`) drops repeated value groups so a multi-row `VALUES` clause reduces to one `( ? )`, and also collapses repeated `?,?,?` runs. `obfuscateSQLString` (`sql.go:351-367`) inspects a failure for backslash-related causes and retries once with `literal_escapes` flipped, returning the second result when it succeeds.

## ADP behavior

`obfuscate_sql_string` in `transforms/trace_obfuscation/sql.rs` runs one pass with `literal_escapes` fixed to true and has no retry, so a mis-guessed dialect yields an error and the caller marks the span non-parsable. Grouping is unfinished: the TODO at `sql.rs:289` leaves every row of a multi-row `VALUES` clause in the output.

## Acceptance

- [ ] A multi-row `VALUES` list collapses to a single `( ? )` group
- [ ] A backslash-related failure retries once with the opposite dialect
- [ ] The retry result, not the first error, reaches the span resource
- [ ] `sql.tables` still carries the comma-separated table list

## References

Depends on: #2405

PR #2601, opened 2026-09-14
fix(otlp): obfuscate malformed json queries and keep key order

### LINKED ISSUE #2405: Traces(otlp+dd): Obfuscate malformed JSON and preserve key order and formatting

webern:
## Human Summary

TODO: Add a human-written summary before requesting review.

## AI Summary

This replaces the `serde_json::Value` rewrite with a single-pass JSON scanner ported from the Datadog Agent. Malformed input is obfuscated up to the error and ends in `...` instead of being returned with raw values. Valid input keeps its key order and text; whitespace between tokens is dropped to match the Agent.

Values selected for SQL obfuscation are unescaped before transformation. SQL failures now use the Agent's failure message instead of `?`.

The implementation stays local because libdatadog's `JsonObfuscator` cannot carry `SqlObfuscationConfig` or report transformer failures. Its dependency also enables `serde_json/preserve_order` across the `agent-data-plane` dependency graph.

Differential testing found pre-existing SQL tokenizer gaps and one JSON-unescaping difference, recorded in the review. They are not introduced by the scanner change.

## Change Type

- [x] Bug fix

## How did you test this PR?

Unit tests cover key order, whitespace, nesting, malformed and truncated input, kept subtrees, SQL transformation, escaped keys, multibyte values, and multiple documents. A property test runs arbitrary strings through the scanner. Outputs were compared with the Agent implementation, including 85,000 differential cases.

The following checks pass:

- `make fmt`
- `cargo check --workspace`
- `make check-clippy`
- `make check-docs`
- `make check-deny`
- `make check-licenses`
- `make check-release-notes`
- `cargo nextest run -p saluki-components trace_obfuscation` (113 tests)

## References

- Closes: #2405
- Progresses: #2438


webern:
> AI
>
> ## AI finding 1: a SQL value that starts with `#` looks like it panics (pre-existing, not from this PR)
>
> While fuzzing the new transform path I hit a panic in the SQL tokenizer, not in this PR's code. I
> think the new `Err` branch at `json.rs:220-226` gives it a second route in, so it seemed worth
> raising here even though the bug predates the PR.
>
> `sql_tokenizer.rs:198` enters `scan_comment_to_eol()` after consuming a single `#`, but that
> function assumes two characters were consumed:
>
> ```rust
> // sql_tokenizer.rs:414-415
> fn scan_comment_to_eol(&mut self) -> (TokenKind, &'a [u8]) {
>     let start = self.pos - 2;
> ```
>
> The `--` caller advances twice first (`sql_tokenizer.rs:154-157`), so it is fine; the `#` caller does
> not. With `#` as the first character of the query, `self.pos == 1`, so `1 - 2` underflows. As far as I
> can tell that panics under either profile: `[profile.release]` has `overflow-checks = true`
> (`Cargo.toml:276`) so it trips on the subtraction, and `optimized-release` has them off
> (`Cargo.toml:283`) so it trips on the slice index on line 421 with
> `range start index 18446744073709551615 out of range`.
>
> Reproduced by driving the real `sql.rs`/`sql_tokenizer.rs` from a copy of this PR's `json.rs` outside
> the repo, with `obfuscate_sql_values: ["query"]`:
>
> ```
> {"query": "#"}              -> panic at sql_tokenizer.rs:421
> {"query": "# comment only"} -> panic at sql_tokenizer.rs:421
> {"query": " #"}             -> ok (pos is already 2)
> {"query": "select 1 # c"}   -> ok
> ```
>
> 126 of 6000 adversarial SQL values panicked and every one of them started with `#`. The Agent handles
> it: `scanCommentType1` (`pkg/obfuscate/sql_tokenizer.go:842-851`) uses its tracked token buffer
> rather than arithmetic, the result comes out empty, and `sql.go:450-451` turns that into the
> "Datadog-agent failed to obfuscate SQL string" message.
>
> Impact, if I have this right: any tracer value beginning with `#` (a MySQL comment) takes down the
> component. The default `obfuscate_sql_values` is `[]` (`apm_config.yaml:240-243`), so this PR's path
> only reaches it when an operator configures it, but `mod.rs:137` reaches the same function for every
> SQL span, so a `db.statement` of `# ...` would do it today.
>
> Could we fix it in the tokenizer, either by tracking the token start the way the Agent does or by
> passing the comment-marker width in, and add a regression test for `#` and `#>`/`#-` at position 0?
> Should this be its own issue rather than something for this PR to carry? I don't think #2410,
> #2411 or #2412 touch it.

webern:
> AI
>
> ## AI finding 2: `\/` in a SQL value makes us diverge from the Agent (in our favour, I think)
>
> `json.rs:219` unescapes the collected literal with serde:
>
> ```rust
> let query = serde_json::from_str::<String>(&self.buf).unwrap_or_else(|_| self.buf.clone());
> ```
>
> The Agent uses `strconv.Unquote` (`pkg/obfuscate/json.go:174-177`), which is Go string-literal syntax,
> not JSON. Go has no `\/` escape, so `Unquote` fails and the Agent falls back to the raw literal
> **with its surrounding quotes**. The SQL obfuscator then reads the whole thing as one double-quoted
> identifier and emits the value unobfuscated. Running `pkg/obfuscate` against a copy of this PR:
>
> ```
> in:  {"query": "select * from t where path = 'a\/b'"}
> go:  {"query":"select * from t where path = 'a/b'"}     <- literal survives
> rs:  {"query":"select * from t where path = ?"}
>
> in:  {"query": "select 10 \/ 2 from t"}
> go:  {"query":"select 10 / 2 from t"}
> rs:  {"query":"select ? / ? from t"}
> ```
>
> Same shape for a valid surrogate pair (`\ud83d\ude00`), which `Unquote` also rejects and the Agent
> then leaks as `ud83dude00`. Escapes that both accept (`\u0041`, `\t`, `\\`, `\"`) agree, and a
> non-string literal like `{"query": 42}` agrees too, so as far as I can tell `\/` and surrogate pairs
> are the whole divergence set. This is the only place in 85k differential cases where the JSON layer
> disagrees with the Agent; everything else matched byte for byte.
>
> Our behaviour looks strictly better to me, and `\/` is not exotic: PHP's `json_encode` escapes every
> `/` by default, so a PHP tracer sending an exec plan or an ES body would hit it. But the resource
> string then differs from the Agent's for those spans, which is the same stats-splitting problem the
> key-order fix in this PR is about, so I don't think it should be silent.
>
> Should we lock it in with a test next to the "where the two disagree, this follows the Agent" case, a
> comment on that line saying we deliberately do not reproduce `strconv.Unquote`'s rejection of `\/`,
> and a line in the PR body, so that "every expected value is what the Agent returns, character for
> character" is not read as covering this? Happy to be told this is too small to bother with.

webern:
> AI
>
> ## AI finding 3: two Ok/Err disagreements with the Agent's SQL obfuscator, now visible here and tracked nowhere
>
> Writing the Agent's failure message on `Err` (`json.rs:220-226`) is right, but it makes every
> Ok/Err disagreement between our `obfuscate_sql_string` and the Agent's show up in the body. I found
> two in differential testing, in opposite directions, and I could not find either in the acceptance
> lists of #2410, #2411 or #2412.
>
> **1. Empty result.** The Agent errors on an empty obfuscation result (`pkg/obfuscate/sql.go:450-451`,
> `errors.New("result is empty")`) and so writes the failure message; we return `Ok("")`:
>
> ```
> in:  {"query": ""}     go: {"query":"Datadog-agent failed to obfuscate..."}   rs: {"query":""}
> in:  {"query": "\t"}   go: {"query":"Datadog-agent failed to obfuscate..."}   rs: {"query":""}
> ```
>
> The PR body calls this out and points at #2410, but #2410's acceptance items are the `VALUES`
> grouping and the escape-dialect retry; nothing there would make an empty result an error. Could we
> add an acceptance item to #2410 (or file a small follow-up) so it does not fall through?
>
> **2. Non-ASCII identifiers.** `sql_tokenizer.rs:597-609` classifies identifier characters with
> `is_ascii_alphabetic`/`is_ascii_alphanumeric`, where the Agent uses `unicode.IsLetter`
> (`pkg/obfuscate/sql_tokenizer.go:914-920`). So a query with a non-ASCII identifier is a `LexError`
> for us and fine for the Agent:
>
> ```
> in:  {"query": "select * from café"}
> go:  {"query":"select * from café"}
> rs:  {"query":"Datadog-agent failed to obfuscate SQL string. Enable agent debug logs for more info."}
> ```
>
> A literal like `'José'` is fine, since it is inside quotes; it is identifiers, and any non-ASCII text
> outside a literal, that trip it. Before this PR those values came out as `"?"`; now a 79-character
> English sentence lands in the body and, for the exec-plan case, in the resource that stats aggregate
> on. The same gap already shows up as "Non-parsable SQL query" on the span path (`mod.rs:155-160`), so
> it is not this PR's doing, and the same lines also drop the Agent's `@` and `#` identifier characters.
>
> I don't think either of these should hold up this PR. Should they be tracked somewhere, though, since
> the SQL issues as written wouldn't catch them?

lib/saluki-components/src/transforms/trace_obfuscation/json.rs
```diff
@@ -27,56 +47,201 @@ impl JsonObfuscator {
...
+    }
+
+    /// Handles the end of a value: writes its SQL obfuscation if one was collected, or leaves a
+    /// kept subtree once the pass climbs back out of it.
+    fn finish_value(&mut self, depth: usize) {
+        if self.transforming {
+            // The collected characters are a JSON literal. A string is unescaped before
+            // obfuscation; anything else is passed on as written.
+            let query = serde_json::from_str::<String>(&self.buf).unwrap_or_else(|_| self.buf.clone());
+            let obfuscated = match obfuscate_sql_string(&query, &self.obfuscator.sql_config) {
```
webern, line 220: Owner instructions: We should open this as an issue upstream, track it in our saluki traces tracking issue, and reply to and resolve this comment with a link to the upstream issue.

> AI
>
> ## AI finding 1: a SQL value that starts with `#` looks like it panics (pre-existing, not from this PR)
>
> While fuzzing the new transform path I hit a panic in the SQL tokenizer, not in this PR's code. I
> think the new `Err` branch at `json.rs:220-226` gives it a second route in, so it seemed worth
> raising here even though the bug predates the PR.
>
> `sql_tokenizer.rs:198` enters `scan_comment_to_eol()` after consuming a single `#`, but that
> function assumes two characters were consumed:
>
> ```rust
> // sql_tokenizer.rs:414-415
> fn scan_comment_to_eol(&mut self) -> (TokenKind, &'a [u8]) {
>     let start = self.pos - 2;
> ```
>
> The `--` caller advances twice first (`sql_tokenizer.rs:154-157`), so it is fine; the `#` caller does
> not. With `#` as the first character of the query, `self.pos == 1`, so `1 - 2` underflows. As far as I
> can tell that panics under either profile: `[profile.release]` has `overflow-checks = true`
> (`Cargo.toml:276`) so it trips on the subtraction, and `optimized-release` has them off
> (`Cargo.toml:283`) so it trips on the slice index on line 421 with
> `range start index 18446744073709551615 out of range`.
>
> Reproduced by driving the real `sql.rs`/`sql_tokenizer.rs` from a copy of this PR's `json.rs` outside
> the repo, with `obfuscate_sql_values: ["query"]`:
>
> ```
> {"query": "#"}              -> panic at sql_tokenizer.rs:421
> {"query": "# comment only"} -> panic at sql_tokenizer.rs:421
> {"query": " #"}             -> ok (pos is already 2)
> {"query": "select 1 # c"}   -> ok
> ```
>
> 126 of 6000 adversarial SQL values panicked and every one of them started with `#`. The Agent handles
> it: `scanCommentType1` (`pkg/obfuscate/sql_tokenizer.go:842-851`) uses its tracked token buffer
> rather than arithmetic, the result comes out empty, and `sql.go:450-451` turns that into the
> "Datadog-agent failed to obfuscate SQL string" message.
>
> Impact, if I have this right: any tracer value beginning with `#` (a MySQL comment) takes down the
> component. The default `obfuscate_sql_values` is `[]` (`apm_config.yaml:240-243`), so this PR's path
> only reaches it when an operator configures it, but `mod.rs:137` reaches the same function for every
> SQL span, so a `db.statement` of `# ...` would do it today.
>
> Could we fix it in the tokenizer, either by tracking the token start the way the Agent does or by
> passing the comment-marker width in, and add a regression test for `#` and `#>`/`#-` at position 0?
> Should this be its own issue rather than something for this PR to carry? I don't think #2410,
> #2411 or #2412 touch it.

lib/saluki-components/src/transforms/trace_obfuscation/json.rs
```diff
@@ -27,56 +47,201 @@ impl JsonObfuscator {
...
+        self.wiped = false;
+    }
+
+    /// Handles the end of a value: writes its SQL obfuscation if one was collected, or leaves a
+    /// kept subtree once the pass climbs back out of it.
+    fn finish_value(&mut self, depth: usize) {
+        if self.transforming {
+            // The collected characters are a JSON literal. A string is unescaped before
+            // obfuscation; anything else is passed on as written.
+            let query = serde_json::from_str::<String>(&self.buf).unwrap_or_else(|_| self.buf.clone());
```
webern, line 219: Owner instructions: We should lock in the desirable divergence with a test that briefly describes in its documentation why we wrote the test, what the divergence is, and why we think it is better than the Agent's implementation.

> AI
>
> ## AI finding 2: `\/` in a SQL value makes us diverge from the Agent (in our favour, I think)
>
> `json.rs:219` unescapes the collected literal with serde:
>
> ```rust
> let query = serde_json::from_str::<String>(&self.buf).unwrap_or_else(|_| self.buf.clone());
> ```
>
> The Agent uses `strconv.Unquote` (`pkg/obfuscate/json.go:174-177`), which is Go string-literal syntax,
> not JSON. Go has no `\/` escape, so `Unquote` fails and the Agent falls back to the raw literal
> **with its surrounding quotes**. The SQL obfuscator then reads the whole thing as one double-quoted
> identifier and emits the value unobfuscated. Running `pkg/obfuscate` against a copy of this PR:
>
> ```
> in:  {"query": "select * from t where path = 'a\/b'"}
> go:  {"query":"select * from t where path = 'a/b'"}     <- literal survives
> rs:  {"query":"select * from t where path = ?"}
>
> in:  {"query": "select 10 \/ 2 from t"}
> go:  {"query":"select 10 / 2 from t"}
> rs:  {"query":"select ? / ? from t"}
> ```
>
> Same shape for a valid surrogate pair (`\ud83d\ude00`), which `Unquote` also rejects and the Agent
> then leaks as `ud83dude00`. Escapes that both accept (`\u0041`, `\t`, `\\`, `\"`) agree, and a
> non-string literal like `{"query": 42}` agrees too, so as far as I can tell `\/` and surrogate pairs
> are the whole divergence set. This is the only place in 85k differential cases where the JSON layer
> disagrees with the Agent; everything else matched byte for byte.
>
> Our behaviour looks strictly better to me, and `\/` is not exotic: PHP's `json_encode` escapes every
> `/` by default, so a PHP tracer sending an exec plan or an ES body would hit it. But the resource
> string then differs from the Agent's for those spans, which is the same stats-splitting problem the
> key-order fix in this PR is about, so I don't think it should be silent.
>
> Should we lock it in with a test next to the "where the two disagree, this follows the Agent" case, a
> comment on that line saying we deliberately do not reproduce `strconv.Unquote`'s rejection of `\/`,
> and a line in the PR body, so that "every expected value is what the Agent returns, character for
> character" is not read as covering this? Happy to be told this is too small to bother with.

lib/saluki-components/src/transforms/trace_obfuscation/json.rs
```diff
@@ -27,56 +47,201 @@ impl JsonObfuscator {
...
+            let obfuscated = match obfuscate_sql_string(&query, &self.obfuscator.sql_config) {
+                Ok(sql) => sql.query,
```
webern, line 221: Owner instructions: We should open this as an issue upstream, track it in our saluki traces tracking issue, and reply to and resolve this comment with a link to the upstream issue.

> AI
>
> ## AI finding 3: two Ok/Err disagreements with the Agent's SQL obfuscator, now visible here and tracked nowhere
>
> Writing the Agent's failure message on `Err` (`json.rs:220-226`) is right, but it makes every
> Ok/Err disagreement between our `obfuscate_sql_string` and the Agent's show up in the body. I found
> two in differential testing, in opposite directions, and I could not find either in the acceptance
> lists of #2410, #2411 or #2412.
>
> **1. Empty result.** The Agent errors on an empty obfuscation result (`pkg/obfuscate/sql.go:450-451`,
> `errors.New("result is empty")`) and so writes the failure message; we return `Ok("")`:
>
> ```
> in:  {"query": ""}     go: {"query":"Datadog-agent failed to obfuscate..."}   rs: {"query":""}
> in:  {"query": "\t"}   go: {"query":"Datadog-agent failed to obfuscate..."}   rs: {"query":""}
> ```
>
> The PR body calls this out and points at #2410, but #2410's acceptance items are the `VALUES`
> grouping and the escape-dialect retry; nothing there would make an empty result an error. Could we
> add an acceptance item to #2410 (or file a small follow-up) so it does not fall through?
>
> **2. Non-ASCII identifiers.** `sql_tokenizer.rs:597-609` classifies identifier characters with
> `is_ascii_alphabetic`/`is_ascii_alphanumeric`, where the Agent uses `unicode.IsLetter`
> (`pkg/obfuscate/sql_tokenizer.go:914-920`). So a query with a non-ASCII identifier is a `LexError`
> for us and fine for the Agent:
>
> ```
> in:  {"query": "select * from café"}
> go:  {"query":"select * from café"}
> rs:  {"query":"Datadog-agent failed to obfuscate SQL string. Enable agent debug logs for more info."}
> ```
>
> A literal like `'José'` is fine, since it is inside quotes; it is identifiers, and any non-ASCII text
> outside a literal, that trip it. Before this PR those values came out as `"?"`; now a 79-character
> English sentence lands in the body and, for the exec-plan case, in the resource that stats aggregate
> on. The same gap already shows up as "Non-parsable SQL query" on the span path (`mod.rs:155-160`), so
> it is not this PR's doing, and the same lines also drop the Agent's `@` and `#` identifier characters.
>
> I don't think either of these should hold up this PR. Should they be tracked somewhere, though, since
> the SQL issues as written wouldn't catch them?

webern (commented):
> AI
>
> Self-review found three SQL-obfuscation edge cases. They are pre-existing or intentional differences and do not appear to block this scanner change. Details are inline.

lib/saluki-components/src/transforms/trace_obfuscation/json.rs
webern: > AI
>
> Opened upstream as DataDog/libdatadog#2542. I checked the crate first: `libdd-trace-obfuscation` does not share the panic (`#`, `#>`, `#-`, `--`, `/*` at position 0 all obfuscate without panicking, in generic, MSSQL and PostgreSQL modes), so the issue asks for regression tests that pin that class, and records that a comment-only query obfuscates to an empty string where the Agent fails. The vendored tokenizer here still needs its own fix, tracked with the other SQL work.

lib/saluki-components/src/transforms/trace_obfuscation/json.rs
webern: > AI
>
> Opened upstream as DataDog/libdatadog#2541. Of the two disagreements, the empty-result one exists in the crate too (verified: `obfuscate_sql_string("\t")` returns `""` where the Agent errors and writes its failure message), so the issue asks whether the crate should expose the Agent's failure semantics for an empty result. The non-ASCII identifier one the crate already gets right (verified: `obfuscate_sql_string("select * from café")` keeps the identifier, like the Agent), and the issue asks for a regression test to pin it. The vendored tokenizer's `is_ascii_alphabetic` classification and the empty-result `Ok` are still ours to fix here, tracked with the other SQL work.

lib/saluki-components/src/transforms/trace_obfuscation/json.rs
webern: > AI
>
> Locked in with a test in [ffda0601](https://github.com/DataDog/saluki/commit/ffda060104bdbdf1e6a6c2e1e015030fec0a72de): `sql_value_with_json_only_escapes_is_still_obfuscated` sits next to the other SQL unescaping cases. Its documentation says why the expected values are deliberately not the Agent's: `strconv.Unquote` is Go string syntax, not JSON, so it rejects `\/` and surrogate pairs, falls back to the raw literal with its quotes, and the SQL tokenizer then reads the whole value as one quoted identifier and passes it through unobfuscated. We unescape with serde, so a JSON-valid value always reaches the SQL tokenizer and its literals are replaced, which matters for PHP tracers whose `json_encode` escapes every `/` as `\/` by default.

PR #2602, opened 2026-09-14
fix(traces): match Agent credit-card range, allowlist, and span-event coverage

### LINKED ISSUE #2402: Traces(otlp+dd): Match credit-card range, allowlist, and span-event coverage
## Background

Card detection is a false-positive-sensitive heuristic, and a miss writes a card number into storage. ADP treats 17-to-19-digit numbers as cards, scrubs five Databricks and Spark identifiers the Agent keeps, and never looks at span-event attributes, so cards inside span events reach the backend.

## Agent behavior

`IsCardNumber` (`credit_cards.go:89-163`) rejects fewer than 12 digits and aborts once the digit count passes 16, checking IIN prefixes as it scans and running the Luhn checksum when enabled. `ShouldObfuscateCCKey` skips `_`-prefixed keys, an allowlist that adds `databricks_job_id`, `databricks_job_run_id`, `databricks_task_run_id`, `config.spark_app_startTime`, and `config.spark_databricks_job_parentRunId`, and any configured `keep_values`. `obfuscateSpanEvent` and `ccObfuscateAttributeArray` (`obfuscate.go:238-291`) scrub string, int, and float event attributes, including array elements.

## ADP behavior

`credit_cards.rs:131` accepts a digit count of 12 through 19, and `ALLOWLISTED_TAGS` (`credit_cards.rs:9`) omits the five Databricks and Spark keys; `_`-prefixed keys and `keep_values` are already honored. The obfuscation transform walks only `Span::attributes`, never `Span::span_events`, so no event attribute or `AttributeValue::Array` element is scrubbed.

## Configuration keys

- [ ] apm_config.obfuscation.credit_cards.enabled
- [ ] apm_config.obfuscation.credit_cards.keep_values
- [ ] apm_config.obfuscation.credit_cards.luhn

## Acceptance

- [ ] A digit count above 16 is not a card-number candidate
- [ ] All five Databricks and Spark keys are exempt from card scrubbing
- [ ] Card numbers in span-event attributes are replaced with `?`
- [ ] Array-valued span-event attributes are scrubbed element by element
- [ ] Int and float event attributes are compared as their string form

webern:
## Human Summary

TODO: Add a human-written summary before requesting review.

## AI Summary

This aligns credit-card obfuscation with the Datadog Agent in three areas:

- Card detection now rejects values longer than 16 digits.
- Five Databricks and Spark identifiers are added to the key allowlist.
- Structured span-event attributes are scrubbed when they contain string, integer, float, or one-level array card values.

Detection uses libdatadog's `is_card_number`, removing the local digit scan, IIN table, and Luhn implementation. The dependency is pinned by revision with default features disabled; it adds 19 lockfile entries and requires the `MIT-0` license and git-source allowlists.

The span-event walk applies to ingest paths that populate `Span::span_events`, including the Datadog intake. OTLP events are marshalled into an `events` string attribute before trace obfuscation, as they are in the Agent, so this change does not inspect attributes inside that JSON blob. The review records open questions about nested typed values, numeric span attributes, and the dependency trade.

## Change Type

- [x] Bug fix

## How did you test this PR?

The Agent card corpus runs against libdatadog's detector. Additional unit tests cover the 16-digit limit, the five allowlisted keys, string, integer, float, and array event attributes, `_`-prefixed and configured keep-value keys, span attributes, and disabled credit-card obfuscation.

The following checks pass:

- `make fmt`
- `make check-clippy`
- `make check-docs`
- `make check-deny`
- `make check-licenses`
- `make check-unused-deps`
- `make check-features`
- `make test`

## References

- Closes: #2402
- Progresses: #2438


webern:
> AI finding 1
>
> **Is the new span-event walk reachable on the OTLP path?** I may be missing something, but I can
> only find one producer of `Span::span_events`: the Datadog intake decoder
> (`lib/saluki-components/src/decoders/datadog/mod.rs:287` reads field 12, `:301` calls
> `with_span_events`). On the OTLP path the events are marshalled into a string attribute long before
> the transform sees the span:
>
> ```rust
> // lib/saluki-components/src/common/otlp/traces/transform.rs:180
> if let Some(events_json) = marshal_events(&otel_span.events) {
>     attrs.insert(
>         MetaString::from_static("events"),
>         AttributeValue::String(events_json.into()),
>     );
> }
> ```
>
> and `otlp_traces_decode` is wired upstream of `traces_enrich` (which is where `trace_obfuscation`
> lives): `bin/agent-data-plane/src/cli/run.rs:1031` and `:656`. If that reading is right,
> `span.span_events()` is always empty for OTLP-originated spans, so the new walk only ever fires for
> the Datadog intake path, and #2402's acceptance line "card numbers in span-event attributes are
> replaced with `?`" isn't actually met for OTLP input - the card still ships inside the `events` JSON
> blob, which can never match `is_card_number` because it contains `{`.
>
> For what it's worth I don't think this is a divergence from the Agent: its OTLP path does the same
> thing (`pkg/trace/transform/transform.go:328-330` writes `Meta["events"] = MarshalEvents(...)` and
> nothing scrubs that blob afterwards). But the PR title, the issue labels and the tests (which build
> spans directly) all read as OTLP coverage, so a future reader has no way to see where the line is.
>
> Could we either say so explicitly - commit body plus a sentence on
> `obfuscate_credit_cards_in_span_events` about which ingest path populates `span_events` - or, if we
> want to be ahead of the Agent here, scrub the events in the OTLP transform before `marshal_events`
> runs?

webern:
> AI finding 2
>
> **Typed variants our own decoder can produce are skipped.** `obfuscate_credit_card_attribute` ends
> in `_ => false`
> (`lib/saluki-components/src/transforms/trace_obfuscation/mod.rs:145`), and the doc above it says the
> remaining variants "have no counterpart in the trace payloads this mirrors". That's true of
> `pb.SpanEvent` (string/bool/int/double plus an array of scalars, no nesting), but I think it isn't
> true of our model: the intake decoder can build three more shapes for an event attribute, and the
> encoder forwards all of them unchanged.
>
> - `AttributeValue::Bytes` - `lib/saluki-components/src/decoders/datadog/value.rs:136`
> - nested `AttributeValue::Array` - `value.rs:137`, recursive up to `MAX_ANY_VALUE_DEPTH = 64`
>   (`value.rs:16`)
> - `AttributeValue::KeyValueList` - `value.rs:151`
> - encoder side: `lib/saluki-components/src/encoders/datadog/traces/mod.rs:785-801`, reached for event
>   attributes at `:651-664`
>
> So an event attribute shaped like `[["4111111111111111"]]`, or a key-value list holding a card, looks
> to me like it reaches the intake with the card intact - the outer value is `Array`/`KeyValueList`, the
> walk descends one level at most (`mod.rs:113-118`), and the inner value falls into `_ => false`.
>
> Should the walk recurse into array elements and key-value-list values instead of stopping at one
> level? If you'd rather keep the reference's shape exactly, could we pin the decision with a test and
> reword the comment to say why (something like "our decoder accepts nesting; we deliberately match the
> reference and stop at one level"), so it doesn't read as "cannot happen"?
>
> `Bytes` I'm less sure about - the Agent would have base64'd such a value before this point, so
> nothing would have matched there either. Skipping it may well be the right call; it just isn't
> obviously the same reason as the others.

webern:
> AI finding 3
>
> **Span attributes and event attributes now disagree about numbers.** After this change an `Int` or
> `Float` event attribute is compared as its string form
> (`lib/saluki-components/src/transforms/trace_obfuscation/mod.rs:134-144`), but the span-level walk
> still only matches `AttributeValue::String` (`mod.rs:87-96`) - and the intake decoder produces typed
> span attributes too (`lib/saluki-components/src/decoders/datadog/mod.rs:284`, field 9 ->
> `read_attributes_map`). So `payment.card = 4111111111111111` sent as an integer is scrubbed inside a
> span event and forwarded when it sits directly on the span.
>
> I don't think we're diverging from the Agent: there the asymmetry is an artifact of Meta vs Metrics
> (`pkg/trace/agent/obfuscate.go:78-87` walks `Meta` only, and OTLP ints/doubles land in `Metrics` per
> `pkg/trace/transform/transform.go:307-317`), so a numeric span attribute was never a candidate there
> either. But in our typed model the two halves of the same span now behave differently for the same
> value, which surprised me while reading.
>
> Is that intended? A sentence in the doc comment on `obfuscate_credit_cards_in_span` - or a follow-up
> issue if we think the span-level walk should cover `Int`/`Float` as well - would save the next reader
> the trip.

webern:
> AI finding 4
>
> **The dependency trade may deserve a second opinion - non-blocking, and detection correctness is not
> the issue.** I checked the pinned revision rather than taking it on trust: I built a differential
> harness with the Go `IsCardNumber`/`luhnValid`/`validCardPrefix` copied verbatim from
> `pkg/obfuscate/credit_cards.go` on one side, and `is_card_number` from `libdd-trace-obfuscation` at
> `e0a0134` on the other, then ran 40,053 inputs through both with `luhn` off and on - random
> digit/space/dash strings, invalid characters, non-ASCII digits (Arabic-Indic, full-width, CJK), the
> 12-to-19 digit bands, IIN boundaries around 699999/700000, and 100k-character inputs. **Zero verdict
> differences.** Against the implementation this PR deletes, the only behavior change is the
> 17/18/19-digit band: 1,095 inputs stop being treated as cards, and nothing starts being treated as one.
> That is exactly what #2402 asked for.
>
> So my question is only about cost. The detection bug was the `12..=19` bound - one line - and the price
> of importing it is 19 new `Cargo.lock` entries for one leaf function: 7 crates from the git source
> (`libdd-trace-obfuscation`, `-utils`, `-protobuf`, `-common`, `-normalization`, `-tinybytes`,
> `-capabilities`), plus `fluent-uri`, `borrow-or-share`, `ref-cast(-impl)`, `const_format(+proc-macros)`,
> `konst(+macro_rules)`, `rmpv`, `serde-transcode`, `thin-vec`, and a second `nix` (0.29 through
> `libdd-common`, next to the 0.31.3 we already build through `vsock` on Linux). Plus the `MIT-0` policy
> line you flagged, and the allowlist stays duplicated here anyway because `should_obfuscate_cc_key` is
> private upstream (`libdd-trace-obfuscation/src/obfuscate.rs:388`).
>
> Two questions, since this sets a precedent wider than this PR:
>
> 1. If libdatadog is an acceptable dependency, should the hand-ported SQL work (#2410, #2411, #2412)
>    reconsider `libdd_trace_obfuscation::sql` rather than maintain a second port of the same Go code?
> 2. If we'd rather not carry the dependency, is upstreaming a `pub should_obfuscate_cc_key` (and keeping
>    our own ~60-line detector, now with the `12..=16` bound and a corpus test) the cheaper path?
>
> Either answer seems defensible to me - I'd just rather it were decided deliberately than as a
> side-effect of a two-line parity fix.

lib/saluki-components/src/transforms/trace_obfuscation/mod.rs
```diff
@@ -67,6 +70,7 @@ impl TraceObfuscation {
     fn obfuscate_span(&mut self, span: &mut Span) {
         if self.obfuscator.config.credit_cards.enabled {
             self.obfuscate_credit_cards_in_span(span);
+            self.obfuscate_credit_cards_in_span_events(span);
```
webern, line 73: > AI
>
> ## Finding 1
>
> **Is the new span-event walk reachable on the OTLP path?** I may be missing something, but I can
> only find one producer of `Span::span_events`: the Datadog intake decoder
> (`lib/saluki-components/src/decoders/datadog/mod.rs:287` reads field 12, `:301` calls
> `with_span_events`). On the OTLP path the events are marshalled into a string attribute long before
> the transform sees the span:
>
> ```rust
> // lib/saluki-components/src/common/otlp/traces/transform.rs:180
> if let Some(events_json) = marshal_events(&otel_span.events) {
>     attrs.insert(
>         MetaString::from_static("events"),
>         AttributeValue::String(events_json.into()),
>     );
> }
> ```
>
> and `otlp_traces_decode` is wired upstream of `traces_enrich` (which is where `trace_obfuscation`
> lives): `bin/agent-data-plane/src/cli/run.rs:1031` and `:656`. If that reading is right,
> `span.span_events()` is always empty for OTLP-originated spans, so the new walk only ever fires for
> the Datadog intake path, and #2402's acceptance line "card numbers in span-event attributes are
> replaced with `?`" isn't actually met for OTLP input - the card still ships inside the `events` JSON
> blob, which can never match `is_card_number` because it contains `{`.
>
> For what it's worth I don't think this is a divergence from the Agent: its OTLP path does the same
> thing (`pkg/trace/transform/transform.go:328-330` writes `Meta["events"] = MarshalEvents(...)` and
> nothing scrubs that blob afterwards). But the PR title, the issue labels and the tests (which build
> spans directly) all read as OTLP coverage, so a future reader has no way to see where the line is.
>
> Could we either say so explicitly - commit body plus a sentence on
> `obfuscate_credit_cards_in_span_events` about which ingest path populates `span_events` - or, if we
> want to be ahead of the Agent here, scrub the events in the OTLP transform before `marshal_events`
> runs?

lib/saluki-components/src/transforms/trace_obfuscation/mod.rs
```diff
@@ -93,6 +97,59 @@ impl TraceObfuscation {
...
+                formatted.clear();
+                let _ = write!(formatted, "{}", int_val);
+                self.obfuscator.is_credit_card_number(formatted)
+            }
+            AttributeValue::Float(float_val) => {
+                formatted.clear();
+                let _ = write!(formatted, "{}", float_val);
+                self.obfuscator.is_credit_card_number(formatted)
+            }
+            _ => false,
```
webern, line 145: > AI
>
> ## Finding 2
>
> **Typed variants our own decoder can produce are skipped.** `obfuscate_credit_card_attribute` ends
> in `_ => false`
> (`lib/saluki-components/src/transforms/trace_obfuscation/mod.rs:145`), and the doc above it says the
> remaining variants "have no counterpart in the trace payloads this mirrors". That's true of
> `pb.SpanEvent` (string/bool/int/double plus an array of scalars, no nesting), but I think it isn't
> true of our model: the intake decoder can build three more shapes for an event attribute, and the
> encoder forwards all of them unchanged.
>
> - `AttributeValue::Bytes` - `lib/saluki-components/src/decoders/datadog/value.rs:136`
> - nested `AttributeValue::Array` - `value.rs:137`, recursive up to `MAX_ANY_VALUE_DEPTH = 64`
>   (`value.rs:16`)
> - `AttributeValue::KeyValueList` - `value.rs:151`
> - encoder side: `lib/saluki-components/src/encoders/datadog/traces/mod.rs:785-801`, reached for event
>   attributes at `:651-664`
>
> So an event attribute shaped like `[["4111111111111111"]]`, or a key-value list holding a card, looks
> to me like it reaches the intake with the card intact - the outer value is `Array`/`KeyValueList`, the
> walk descends one level at most (`mod.rs:113-118`), and the inner value falls into `_ => false`.
>
> Should the walk recurse into array elements and key-value-list values instead of stopping at one
> level? If you'd rather keep the reference's shape exactly, could we pin the decision with a test and
> reword the comment to say why (something like "our decoder accepts nesting; we deliberately match the
> reference and stop at one level"), so it doesn't read as "cannot happen"?
>
> `Bytes` I'm less sure about - the Agent would have base64'd such a value before this point, so
> nothing would have matched there either. Skipping it may well be the right call; it just isn't
> obviously the same reason as the others.

webern (commented):
> AI
>
> Self-review found four scope and dependency questions. Details are inline.

lib/saluki-components/src/transforms/trace_obfuscation/mod.rs
```diff
@@ -93,6 +97,59 @@ impl TraceObfuscation {
...
+
+    /// Replaces `value` with `?` when it holds a card number, using `formatted` as scratch space.
+    ///
+    /// Only string, integer and floating-point attributes can hold a card number. Booleans cannot,
+    /// and the remaining variants have no counterpart in the trace payloads this mirrors, so they
+    /// are left alone. Nested arrays are left alone for the same reason.
+    fn obfuscate_credit_card_attribute(&self, value: &mut AttributeValue, formatted: &mut String) {
+        let is_card = match value {
+            AttributeValue::String(str_val) => self.obfuscator.is_credit_card_number(str_val.as_ref()),
+            AttributeValue::Int(int_val) => {
```
webern, line 135: > AI
>
> ## Finding 3
>
> **Span attributes and event attributes now disagree about numbers.** After this change an `Int` or
> `Float` event attribute is compared as its string form
> (`lib/saluki-components/src/transforms/trace_obfuscation/mod.rs:134-144`), but the span-level walk
> still only matches `AttributeValue::String` (`mod.rs:87-96`) - and the intake decoder produces typed
> span attributes too (`lib/saluki-components/src/decoders/datadog/mod.rs:284`, field 9 ->
> `read_attributes_map`). So `payment.card = 4111111111111111` sent as an integer is scrubbed inside a
> span event and forwarded when it sits directly on the span.
>
> I don't think we're diverging from the Agent: there the asymmetry is an artifact of Meta vs Metrics
> (`pkg/trace/agent/obfuscate.go:78-87` walks `Meta` only, and OTLP ints/doubles land in `Metrics` per
> `pkg/trace/transform/transform.go:307-317`), so a numeric span attribute was never a candidate there
> either. But in our typed model the two halves of the same span now behave differently for the same
> value, which surprised me while reading.
>
> Is that intended? A sentence in the doc comment on `obfuscate_credit_cards_in_span` - or a follow-up
> issue if we think the span-level walk should cover `Int`/`Float` as well - would save the next reader
> the trip.

Cargo.toml
```diff
@@ -155,6 +155,10 @@ rand = { version = "0.10", default-features = false }
 rand_distr = { version = "0.6", default-features = false }
 k8s-openapi = { version = "0.28", default-features = false, features = ["v1_33"] }
 kube = { version = "4.0", default-features = false }
+# Trace obfuscation primitives shared with the Datadog Agent. Pinned to a revision because the crate is consumed from
+# GitHub directly. `default-features = false` drops the `https` feature, which would otherwise pull a TLS stack and an
+# HTTP client in through `libdd-common`.
+libdd-trace-obfuscation = { git = "https://github.com/DataDog/libdatadog", rev = "e0a0134bbf0896603db615344e9b2950e7a2b2b9", default-features = false }
```
webern: > AI
>
> ## Finding 4
>
> **The dependency trade may deserve a second opinion - non-blocking, and detection correctness is not
> the issue.** I checked the pinned revision rather than taking it on trust: I built a differential
> harness with the Go `IsCardNumber`/`luhnValid`/`validCardPrefix` copied verbatim from
> `pkg/obfuscate/credit_cards.go` on one side, and `is_card_number` from `libdd-trace-obfuscation` at
> `e0a0134` on the other, then ran 40,053 inputs through both with `luhn` off and on - random
> digit/space/dash strings, invalid characters, non-ASCII digits (Arabic-Indic, full-width, CJK), the
> 12-to-19 digit bands, IIN boundaries around 699999/700000, and 100k-character inputs. **Zero verdict
> differences.** Against the implementation this PR deletes, the only behavior change is the
> 17/18/19-digit band: 1,095 inputs stop being treated as cards, and nothing starts being treated as one.
> That is exactly what #2402 asked for.
>
> So my question is only about cost. The detection bug was the `12..=19` bound - one line - and the price
> of importing it is 19 new `Cargo.lock` entries for one leaf function: 7 crates from the git source
> (`libdd-trace-obfuscation`, `-utils`, `-protobuf`, `-common`, `-normalization`, `-tinybytes`,
> `-capabilities`), plus `fluent-uri`, `borrow-or-share`, `ref-cast(-impl)`, `const_format(+proc-macros)`,
> `konst(+macro_rules)`, `rmpv`, `serde-transcode`, `thin-vec`, and a second `nix` (0.29 through
> `libdd-common`, next to the 0.31.3 we already build through `vsock` on Linux). Plus the `MIT-0` policy
> line you flagged, and the allowlist stays duplicated here anyway because `should_obfuscate_cc_key` is
> private upstream (`libdd-trace-obfuscation/src/obfuscate.rs:388`).
>
> Two questions, since this sets a precedent wider than this PR:
>
> 1. If libdatadog is an acceptable dependency, should the hand-ported SQL work (#2410, #2411, #2412)
>    reconsider `libdd_trace_obfuscation::sql` rather than maintain a second port of the same Go code?
> 2. If we'd rather not carry the dependency, is upstreaming a `pub should_obfuscate_cc_key` (and keeping
>    our own ~60-line detector, now with the `12..=16` bound and a corpus test) the cheaper path?
>
> Either answer seems defensible to me - I'd just rather it were decided deliberately than as a
> side-effect of a two-line parity fix.

PR #2604, opened 2026-09-14
chore(io): implement resource registry support for listeners/streams

tobz:
## Summary

This PR adds support to `Listener`/`ConnectionOrientedListener` to be used through the resource registry, which allows for maintaining bound listener sockets across component restarts.

In #2446, we added support for the "resource" registry, which is both a literal registry primitive but also, more than that, a conceptual idea of how we want topologies to function in the future. Currently, we can't support the restarting/rebuilding of an individual component because of issues around reacquiring the necessary resources -- sockets, dispatcher, consumer, etc -- that the old instance of a component held on to. The resource registry is aimed at solving at least part of that problem: register "resources" that need to be long-lived across a component lifecycle, such that a component "acquiring" a resource is the equivalent of leasing it temporarily, after which it's returned to the registry and can be reacquired by another component, without actually releasing the resource back to the operating system.

In this PR, we're introducing support to allow network listeners to be registered in the resource registry, which enforces a particular set of constraints, namely that the resource can be fully "recovered". Leasing a resource is a lot like holding a mutex guard, where holding the guard implies ownership. This is simple for network listeners using connection-oriented transports like TCP, because all you need is the listener and the values you get when accepting a new connection are standalone: they're not tied to the listener once they've been accepted. However, because we abstract that concept over _connectionless_ transports like UDP, we need a way to hide the fact that a "connection" returned by a UDP listener is actually borrowed, and then we would need to wait for something using a UDP listener to drop all of the "accepted" connections before the listener could actually be reused.

In order to accomplish this, we've introduced the notion of "subleases", which is really just a way for some resource to inform the registry that the resource itself will end up being subdivided, and that it's not available to be reacquired by another caller into all of those "subresources", each with their own "sublease", have been returned.

Beyond that, we've made minimal changes to the DogStatsD source to start using the resource registry for acquiring network listeners as a way to start testing out the feature end-to-end, and will replace other usages of network listeners in a follow-up PR.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

New and existing tests.

## References

DADP-2


tobz:
* **#2604** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2604?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2604?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lib/saluki-components/src/sources/dogstatsd/mod.rs
```diff
@@ -657,8 +658,11 @@ impl DogStatsDConfiguration {
         }
     }
 
-    /// Builds the appropriate `Listener` objects.
-    async fn build_listeners(&self) -> Result<Vec<Listener>, Error> {
+    /// Acquires the appropriate `Listener` objects from the resource registry.
+    ///
+    /// The listeners are leased, not owned: the registry keeps the bound sockets when this source is torn down, so a
```
webern, line 663: Cool idea.

lib/saluki-core/src/runtime/state/resources/mod.rs
```diff
@@ -361,19 +466,75 @@ impl Entry {
...
+            _ => unreachable!("a claimed entry is idle"),
         };
 
+        let mut value = match mem::replace(&mut self.state, EntryState::Leased(lease_info)) {
+            EntryState::Idle { value, .. } => value,
+            _ => unreachable!("a claimed entry is idle and holds its value"),
+        };
+
+        // Clear whatever the previous holder accumulated, now that it and its subleases are all gone.
+        (self.reset)(&mut *value);
```
webern: If this line panics, unwinding drops `value`, potentially closing the socket. Conceptual fix:

 ```rust
   let mut state = self.inner.lock().unwrap();
   let entry = state.entries.get_mut(&key).expect(...);

   let reset = entry.reset;
   let mut lease = entry.hand_over_without_reset::<S>(self, &key);

   drop(state);

   let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
       reset(&mut *lease);
   }));

   if let Err(payload) = result {
       lease.discard();
       std::panic::resume_unwind(payload);
   }

   return Ok(lease);
 ```

webern (approved):
This is pretty cool. I scrubbed for any RAII holes and found one panic case that could leave state inconsistent. Flagged.

lib/saluki-core/src/runtime/state/resources/mod.rs
tobz: Good catch.

For now, I'm going to avoid doing any panic catching/unwinding because the clanker pointed out that there's a few other spots where this style of problem exists and I want to think of a more holistic solution, given that our current implementations of `ResourceSpecification` can't/won't panic in their `reset` calls.

I did update things in [00347fc](https://github.com/DataDog/saluki/pull/2604/commits/00347fc1204f6ab17aad186ea55cc9bf9d6f31c8) so that we call `reset` from outside of the lock, though, which ensures that we avoid poisoning the lock _and_ we drop the lease, while leaving the resource in whatever state it was left it after the `reset` call.

PR #2607, opened 2026-09-14
feat(dogstatsd): expose diagnostics through remote commands

thieman:
## Summary

Expose Agent Data Plane DogStatsD diagnostic tools through the Datadog Agent Remote Command Provider introduced by datadog-agent#55260.

## Key changes

- Vendor the `datadog.remoteagent.command.v1` protocol and generate tonic service types.
- Register a `dogstatsd` provider on ADP's existing privileged mTLS Remote Agent API.
- Expose `stats`, `capture`, `replay`, `top`, and `dump-contexts` under `agent remote dogstatsd`.
- Reuse the ADP command implementation for direct and remote execution while preserving direct CLI status logging.
- Validate remote argument names/types before side effects, stream stdout/stderr plus exit codes with bounded buffering, and cancel replay when the client disconnects.

## Test plan

- [x] `make fmt`
- [x] `cargo nextest run -p datadog-protos`
- [x] `cargo nextest run -p agent-data-plane`
- [x] `git diff --check`
- [ ] Cross-repository smoke test after an installable Datadog Agent build includes #55260:
  - `agent remote` discovery and command help
  - successful short command and invalid typed argument
  - replay cancellation cleanup

## Known limitation

This does not yet transparently route legacy `agent dogstatsd*` commands through the remote provider when ADP owns DogStatsD; that Core Agent compatibility change remains follow-up work.

## Validation note

`make check-all` and the pre-commit dependency audit are currently blocked by pre-existing `RUSTSEC-2026-0285` for the `rustls 0.23.43` lockfile entry already present on `main`. The feature-specific format, compile, and test checks above pass.


bin/agent-data-plane/src/cli/dogstatsd/top.rs
```diff
@@ -217,20 +266,78 @@ mod tests {
...
+        std::fs::write(&replacement, b"not-json").expect("replacement artifact should be written");
+        std::fs::rename(&replacement, artifact.path()).expect("artifact path should be replaced");
+
+        let rendered = render_report_from_file(artifact.path(), file, 10, 5).expect("opened artifact should render");
+
+        assert!(rendered.contains("original.metric"), "{rendered}");
+    }
+
+    #[tokio::test]
+    async fn dogstatsd_top_offline_rejects_a_non_regular_file() {
```
thieman: Is this test necessary? Or was this added as a regression test for the Windows problems we were having? If the latter, remove it

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -1,12 +1,17 @@
 use std::cmp::Reverse;
 use std::collections::{HashMap, HashSet};
+use std::future::Future;
+use std::io::Write;
+#[cfg(unix)]
+use std::os::unix::fs::OpenOptionsExt as _;
```
thieman, line 5: Why is this import necessary? Same question for the AsyncWriteExt scoped to only Windows below, and if this is common functionality can we extract it to a shared lib file used by the various commands?

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -168,139 +174,224 @@ struct StatsResponse<'a> {
 
 /// Entrypoint for the `dogstatsd` commands.
 pub async fn handle_dogstatsd_command(local_config: LoadedConfiguration, cmd: DogstatsdCommand) {
-    if let Err(error) = run_dogstatsd_command(&local_config, cmd).await {
+    let cancellation = CancellationToken::new();
+    let signal_task = matches!(&cmd.subcommand, DogstatsdSubcommand::Replay(_)).then(|| {
```
thieman, line 190: Why is this specific to replay?

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -168,139 +174,224 @@ struct StatsResponse<'a> {
...
+    }
+
+    if let Err(error) = result {
         error!("{:#}", error);
         std::process::exit(1);
     }
 }
 
-async fn run_dogstatsd_command(local_config: &LoadedConfiguration, cmd: DogstatsdCommand) -> Result<(), GenericError> {
+pub(crate) async fn run_dogstatsd_command(
```
thieman, line 211: Why do we need to carry this `stream_status` bool around, can we simplify?

bin/agent-data-plane/src/cli/dogstatsd/top.rs
thieman: [GPT-5.6 Terra (OpenAI)] Removed in 9187d4ce8d. The FIFO rejection and descriptor-replacement tests retain the relevant regular-file and TOCTOU coverage; this directory-only assertion was redundant and was the source of Windows-specific error-text handling.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] These remain scoped to replays platform-specific transport implementation: Unix OpenOptionsExt supplies O_NONBLOCK so FIFO-like inputs can be opened without blocking before descriptor validation, while Windows AsyncWriteExt writes replay payloads to named pipes. They are not shared command-framework behavior, so I kept them local rather than adding a shared abstraction.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] Replay is the only direct command that can intentionally run indefinitely (--loops 0) and owns a replay session that must be finished after interruption. The signal-driven cancellation token preserves that existing lifecycle; the other direct commands retain their prior behavior.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] Agreed. Reworked in 9187d4ce8d: the command runner no longer receives stream_status. Direct and remote callers now inject output adapters and cancellation tokens instead. Command code uses the same status/output and cancellation interfaces in both modes, while replay session cleanup remains an explicit resource-lifecycle concern.

bin/agent-data-plane/src/cli/utils.rs
```diff
@@ -23,19 +24,17 @@ pub struct DataPlaneAPIClient {
     authority: String,
 }
 
-/// Builds a data plane API client or exits after logging the error.
-pub(super) async fn get_api_client_or_exit(local_config: &LoadedConfiguration) -> DataPlaneAPIClient {
-    let config = local_config.local();
+/// Builds a data plane API client from the given typed configuration.
+pub(crate) async fn get_api_client(config: &SalukiConfiguration) -> Result<DataPlaneAPIClient, GenericError> {
```
thieman, line 28: Why is the new function needed here? Can we simplify back to one?

bin/agent-data-plane/src/dogstatsd_contexts/mod.rs
```diff
@@ -13,12 +15,19 @@ mod report;
 /// Agent-compatible command API route for requesting a DogStatsD context dump.
 pub(crate) const CONTEXT_DUMP_ROUTE: &str = "/agent/dogstatsd-contexts-dump";
 
+#[cfg(test)]
 pub(crate) fn read_report(path: &Path) -> Result<ContextReport, GenericError> {
```
thieman, line 19: Can we clean this up if it's no longer used in production code? Also does it make any sense to have this both pub(crate) and cfg(test)?

bin/agent-data-plane/src/internal/control_plane.rs
```diff
@@ -53,7 +53,7 @@ pub async fn create_control_plane_supervisor(
         logging_controller,
     ));
     supervisor.add_worker(ConfigWorker::new(config_system.raw_snapshot()));
-    supervisor.add_worker(ConfigRuntimeWorker::new(current_config));
+    supervisor.add_worker(ConfigRuntimeWorker::new(Arc::clone(&current_config)));
```
thieman, line 56: Why is Arc::clone needed here but not for the other workers? Can we scope this parameter down to just the configuration we actually need?

ci/tooling/check-deny.sh
```diff
@@ -36,7 +36,16 @@ advisory_fingerprints() {
         printf '%s\n' "${output}" >&2
         exit "${status}"
     fi
-    printf '%s\n' "${output}" | jq -r '
+    # Cargo-deny normally writes one JSON object per stderr line, but Cargo can add non-JSON
+    # warnings to the same stream in CI. Preserve those warnings in job output while only sending
+    # JSON diagnostics to the fingerprint parser.
+    while IFS= read -r line; do
```
thieman, line 42: Driveby fix resolving a CI failure in cargo-deny coming from some warnings about installing a toolchain that wasn't already on the CI box. Clanker, ignore this thread

lib/protos/datadog/proto/datadog-agent/datadog/remoteagent/command.proto
```diff
@@ -0,0 +1,70 @@
+syntax = "proto3";
```
thieman, line 1: Just confirming this is an exact copy of the protobuf definition from the previous datadog-agent PR?

lib/saluki-components/src/sources/dogstatsd/replay/reader.rs
```diff
@@ -37,20 +37,35 @@ impl TrafficCaptureReader {
     /// Detects zstd-compressed inputs by magic bytes and decompresses transparently. Validates the
     /// Datadog capture header and parses the file version.
     pub fn from_path(path: &Path) -> Result<Self, GenericError> {
-        let raw =
-            fs::read(path).map_err(|e| generic_error!("Failed to read capture file '{}': {}", path.display(), e))?;
+        let file = fs::File::open(path)
```
thieman, line 40: Explain why these changes were necessary, what's actually going on here?

lib/saluki-core/src/runtime/supervisor.rs
```diff
@@ -4056,6 +4056,14 @@ mod tests {
 
         let tree = parent_sup.tree_handle();
         let (tx, handle) = run_supervisor_with_trigger(parent_sup).await;
+        wait_until("the nested supervisor has started", || {
```
thieman, line 4059: Driveby for flake that happened while testing this PR. Clanker, ignore this thread

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: I still don't understand how they're actually being used, especially given this one is imported as `_`. Help me understand?

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -622,30 +820,460 @@ fn get_stylized_table() -> Table {
     table
 }
 
-async fn output_lines<I>(lines: I) -> io::Result<()>
+fn report_status(output: &mut dyn CommandOutput, message: String) -> std::io::Result<()> {
```
thieman: This doesn't seem that useful, can we inline this where it's used instead?

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -352,6 +457,95 @@ fn dogstatsd_replay_target(listeners: &Listeners) -> Result<ReplayTarget, Generi
...
+
+    run_cancellable_blocking(
+        cancellation,
+        tokio::task::spawn_blocking(load),
+        "DogStatsD replay capture loading",
+    )
+    .await
+}
+
+pub(super) async fn run_cancellable_blocking<T>(
```
thieman, line 552: If this is only used once, inline it

bin/agent-data-plane/src/internal/remote_agent.rs
```diff
@@ -509,6 +547,191 @@ impl StatusProvider for RemoteAgentImpl {
...
+                "session ID not set; must be registered with Core Agent",
+            ))?
+            .to_grpc_header_value();
+        let mut response = tonic::Response::new(value);
+        response.metadata_mut().append(SESSION_ID_METADATA_KEY, session_id);
+        Ok(response)
+    }
+}
+
+const MAX_REMOTE_COMMAND_OUTPUT_BYTES: usize = 16 * 1024 * 1024;
```
thieman: Do a review of gRPC settings both here and on the datadog-agent side and flag any potential problems. We need to ensure that the streaming implementation on the client can send back arbitrarily large amounts of data without blowing through the size limits (should be possible using streaming, right?)

bin/agent-data-plane/src/internal/remote_agent.rs
```diff
@@ -509,6 +547,191 @@ impl StatusProvider for RemoteAgentImpl {
...
+    }
+}
+
+impl std::io::Write for RemoteCommandOutput {
+    fn write(&mut self, buffer: &[u8]) -> std::io::Result<usize> {
+        let new_len = self.bytes.len().saturating_add(buffer.len());
+        if new_len > MAX_REMOTE_COMMAND_OUTPUT_BYTES {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::WriteZero,
+                "remote command output exceeds the 16 MiB limit",
```
thieman: Re: last comment on gRPC sizes, failing out here is not acceptable

bin/agent-data-plane/src/internal/remote_agent.rs
```diff
@@ -509,6 +547,191 @@ impl StatusProvider for RemoteAgentImpl {
...
+            ));
+        }
+        self.bytes.extend_from_slice(buffer);
+        Ok(buffer.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
```
thieman: Skeptical that making `flush` a no-op is right, can you explain?

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] `OpenOptionsExt as _` brings the Unix extension trait into scope only for method resolution: it enables `OpenOptions::custom_flags(libc::O_NONBLOCK)` without introducing a local name, hence `_`. `O_NONBLOCK` is needed so opening a FIFO-like remote-controlled path cannot block before we inspect its opened descriptor and reject it as non-regular. `AsyncWriteExt` similarly provides the `write_all` and `flush` extension methods used by the Windows named-pipe replay sender. Both imports are used only by those platform-specific branches.

bin/agent-data-plane/src/cli/utils.rs
thieman: [GPT-5.6 Terra (OpenAI)] The old helper logged and called `process::exit`, which cannot serve the remote path: RCP needs the connection error returned so it can write stderr and an exit-code frame. The shared runner therefore needs the fallible `get_api_client(&SalukiConfiguration) -> Result<_>` form. The direct CLI now owns the policy of logging/exiting after the shared runner returns, so we are back to one client-construction path rather than maintaining direct and remote variants.

bin/agent-data-plane/src/dogstatsd_contexts/mod.rs
thieman: [GPT-5.6 Terra (OpenAI)] It is now `#[cfg(test)]`, so it is absent from production builds. `pub(crate)` is needed only because tests in sibling crate modules exercise report parsing; the two modifiers describe different boundaries (build configuration versus crate visibility). I will verify whether sibling access remains necessary and narrow it further if not.

bin/agent-data-plane/src/internal/control_plane.rs
thieman: [GPT-5.6 Terra (OpenAI)] `RemoteCommandProviderImpl` retains `current_config` for request-time snapshots, while supervisor setup also continues using the original shared handle. `Arc::clone` is the explicit ownership split required to give both workers a handle; the other workers either consume their constructor input or do not retain this live configuration handle. I agree the provider should ideally receive the narrowest configuration view it needs, but it must load a fresh typed snapshot for each command today.

ci/tooling/check-deny.sh
thieman: [GPT-5.6 Terra (OpenAI)] Agreed that this is a drive-by robustness fix rather than Remote Command Provider behavior. It was needed to keep the CI deny job diagnosable when Cargo writes non-JSON toolchain warnings alongside cargo-deny JSON diagnostics. I am leaving this thread open while we decide whether to split it from the feature branch rather than treating it as feature scope.

lib/protos/datadog/proto/datadog-agent/datadog/remoteagent/command.proto
thieman: [GPT-5.6 Terra (OpenAI)] Confirmed against the PR source at `DataDog/datadog-agent@298b583b2f8dab28315ea78b8af871933420b164`: this is an exact byte-for-byte copy of `pkg/proto/datadog/remoteagent/command.proto` from datadog-agent PR 55260. It is vendored here so Saluki generates the same Rust bindings independently.

lib/saluki-components/src/sources/dogstatsd/replay/reader.rs
thieman: [GPT-5.6 Terra (OpenAI)] This is required for remote file safety. Previously `from_path` reopened a path internally; after the caller checked a path, a concurrent process could replace it before that second open. The CLI now opens once with nonblocking Unix flags, checks metadata on that opened descriptor, requires a regular file, and passes the same `File` to `TrafficCaptureReader::from_file`. The reader change separates safe descriptor acquisition from capture parsing and removes that TOCTOU window.

lib/saluki-core/src/runtime/supervisor.rs
thieman: [GPT-5.6 Terra (OpenAI)] Correct: this is an unrelated test-flake stabilization discovered during this PRs validation, not Remote Command Provider functionality. I am leaving it open while we decide whether to split it out rather than implicitly treating it as feature scope.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] Agreed this helper may not earn its abstraction. I am reviewing the remaining call site now; if it is still single-use after the output-interface extraction, I will inline it rather than keep a one-line wrapper.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] Agreed. This started as a testable cancellation boundary, but it is currently single-use. I am reviewing whether the inline form remains clear while retaining the important behavior: cancellation before file access and awaiting any started blocking load rather than detaching it.

bin/agent-data-plane/src/internal/remote_agent.rs
thieman: [GPT-5.6 Terra (OpenAI)] I reviewed the current design further. Individual stdout/stderr frames are UTF-8 chunked below grpc-gos default 4 MiB inbound message limit, but the command output is still accumulated in ADP first and capped at 16 MiB. So the streaming RPC does not yet provide arbitrarily large output in the way this thread expects. I agree that needs a follow-up design/change: incremental bounded frame transmission, no normal aggregate-output cap, and an explicit review of the Core Agent client receive-size settings. Leaving this open for collaboration rather than claiming this is solved.

bin/agent-data-plane/src/internal/remote_agent.rs
thieman: [GPT-5.6 Terra (OpenAI)] Agreed. The current 16 MiB aggregate failure path is not an acceptable long-term output policy. It was a defensive first implementation to avoid unbounded memory while output was buffered; the correct solution is to stop requiring full aggregation and forward bounded frames incrementally. I will keep this thread open for the streaming/output design work.

bin/agent-data-plane/src/internal/remote_agent.rs
thieman: [GPT-5.6 Terra (OpenAI)] In the current implementation `RemoteCommandOutput` is an in-memory `Vec<u8>` adapter, not a buffered OS/socket writer. `write` appends synchronously and there is no deferred resource state, so `flush()` has nothing to do; returning `Ok(())` satisfies `Write` without changing visibility. That said, this confirms the larger concern in the adjacent threads: output is buffered rather than live-streamed. When we replace this with incremental frame forwarding, flush semantics should be revisited as part of that design.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] Resolved in 42eea22617: I inlined `report_status` at every call site and removed the forwarding helper. `make fmt`, `make check-fmt`, and the 300-test ADP suite passed. I will make no further change for this thread.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] Resolved in 42eea22617: I inlined `load_replay_capture` at its only production call site. I retained `run_cancellable_replay_load` because it is the actual reusable lifecycle primitive and is directly tested for pre-cancellation and joining started blocking work. `make fmt`, `make check-fmt`, and the 300-test ADP suite passed. I will make no further change for this thread.

bin/agent-data-plane/src/dogstatsd_contexts/mod.rs
thieman: [GPT-5.6 Terra (OpenAI)] Verified: `read_report` is used by `dogstatsd_contexts::artifact_tests`, a sibling test module, so `pub(crate)` is required while `#[cfg(test)]` keeps it out of production builds. I will make no further change for this thread.

bin/agent-data-plane/src/cli/utils.rs
thieman: [GPT-5.6 Terra (OpenAI)] Correction to my earlier wording: `get_api_client` is the one fallible construction primitive used by the shared runner, while `get_api_client_or_exit` remains a thin direct-CLI policy wrapper for existing config/debug handlers. Keeping that wrapper avoids rewriting unrelated handlers; the remote path never uses it. I will make no further change for this thread.

bin/agent-data-plane/src/internal/control_plane.rs
thieman: [GPT-5.6 Terra (OpenAI)] The provider needs the live `ArcSwap<SalukiConfiguration>` so each remote request uses the current typed configuration; a static subset would become stale on config reload. The explicit `Arc::clone` is therefore necessary ownership transfer. I will make no further change for this thread.

ci/tooling/check-deny.sh
thieman: [GPT-5.6 Terra (OpenAI)] The change is technically unrelated but protects the required CI validation from Cargo toolchain warnings corrupting the JSON-only fingerprint parser. Do you want this robustness fix retained on this PR, or should I move it to a separate follow-up before final review?

lib/saluki-core/src/runtime/supervisor.rs
thieman: [GPT-5.6 Terra (OpenAI)] This is technically unrelated to remote commands. Do you want the supervisor flake fix retained here because it blocked validation, or should I move it to a separate follow-up before final review?

bin/agent-data-plane/src/internal/remote_agent.rs
thieman: [GPT-5.6 Terra (OpenAI)] The current design needs an incremental-output refactor; I will not close this thread as solved. Do you want that work completed in this PR before we proceed with further review, or do you prefer to land the current bounded implementation and track live/unbounded output separately?

bin/agent-data-plane/src/internal/remote_agent.rs
thieman: [GPT-5.6 Terra (OpenAI)] I will not retain the 16 MiB aggregate failure as the final policy if we implement the incremental streaming follow-up. Do you want that change made in this PR now, paired with the client-side receive-limit review, or tracked as a follow-up?

bin/agent-data-plane/src/internal/remote_agent.rs
thieman: [GPT-5.6 Terra (OpenAI)] The current no-op is correct for an in-memory `Vec<u8>` writer, so no standalone flush change is needed. If we implement incremental frame forwarding, I will replace this adapter and define real flush behavior as part of that change.

bin/agent-data-plane/src/internal/remote_agent.rs
thieman: Ugh yes the output needs to be incremental / streaming. Target low latency (say maximum of 100ms between flushes) and we need to make sure we never exceed the max size limits.

bin/agent-data-plane/src/internal/remote_agent.rs
```diff
@@ -509,6 +547,191 @@ impl StatusProvider for RemoteAgentImpl {
     }
 }
 
+pub(crate) struct RemoteCommandProviderImpl {
```
thieman: Would it be against standing patterns to move the command provider-specific code into its own file, just for organization purposes? This file is already pretty long

bin/agent-data-plane/src/internal/remote_agent.rs
thieman: Per other thread, need to do incremental streaming

bin/agent-data-plane/src/internal/remote_agent.rs
thieman: Per other thread, need to do incremental streaming

bin/agent-data-plane/src/internal/remote_agent.rs
thieman: [GPT-5.6 Terra (OpenAI)] Implemented in signed commits 80f3e37f90 and 809d9d077b. Remote stdout now flows through a bounded asynchronous event queue to a framing task instead of accumulating a transcript. It flushes explicitly and on a 100 ms timer; the timer has deterministic priority over continuously available input. Frames are UTF-8-safe and capped at 3 MiB, with tests including output above the old 16 MiB aggregate cap. I also reviewed Core Agent PR 55260: its registry client sets a 64 MiB receive limit, but `agent remote` uses grpc-gos default 4 MiB receive limit, so the 3 MiB frame cap remains required. I will make no further change for this thread.

bin/agent-data-plane/src/internal/remote_agent.rs
thieman: [GPT-5.6 Terra (OpenAI)] Resolved in signed commits 80f3e37f90 and 809d9d077b. The 16 MiB aggregate-output limit and its failure path are removed. Output is forwarded incrementally with bounded queue backpressure and ≤3 MiB frames, so normal command output is not rejected for exceeding a cumulative cap. I will make no further change for this thread.

bin/agent-data-plane/src/internal/remote_agent.rs
thieman: [GPT-5.6 Terra (OpenAI)] Resolved in signed commits 80f3e37f90 and 809d9d077b. `flush` is no longer a no-op: it queues an explicit flush event to the asynchronous framing task, which forwards pending output immediately; the task also flushes pending output on its 100 ms timer and at completion. I will make no further change for this thread.

bin/agent-data-plane/src/internal/remote_agent.rs
thieman: [GPT-5.6 Terra (OpenAI)] Implemented in signed commit 80f3e37f90. I moved RemoteCommandProvider-specific RPC bridging, bounded asynchronous output framing, stream-drop cancellation, and its tests into `internal/remote_command.rs`. `remote_agent.rs` now remains focused on remote-agent lifecycle/service construction. I will make no further change for this thread.

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -168,139 +178,211 @@ struct StatsResponse<'a> {
...
-    info!("Preparing DogStatsD replay from '{}'.", cmd.replay_file_path.display());
+    output
+        .write_status(&format!(
+            "Preparing DogStatsD replay from '{}'.",
+            cmd.replay_file_path.display()
+        ))
+        .await?;
 
     #[cfg(not(target_os = "linux"))]
     tracing::warn!(
```
thieman: This seems like it needs to go through `output`? Check for similar uses as well

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -314,6 +396,36 @@ async fn handle_dogstatsd_replay(
...
+{
+    let replay_result = if cancellation.is_cancelled() {
+        Ok(())
+    } else {
+        replay.await
+    };
+    let finish_result = finish(session_id).await;
+
+    (replay_result, finish_result)
+}
```
thieman, line 459: Can the two helpers above be inlined? Not sure they are helping much unless they are being used multiple times each

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -441,21 +631,21 @@ async fn run_dogstatsd_replay(
             return Ok(());
         }
         iteration = iteration.saturating_add(1);
+        reader.rewind();
```
thieman, line 660: This is new?

bin/agent-data-plane/src/cli/remote.rs
```diff
@@ -0,0 +1,289 @@
...
+///
+/// Output implementations apply backpressure asynchronously. Command execution must not use synchronous writers because
+/// a slow remote-command client must not block a Tokio worker thread or turn ordinary output into a write failure.
+#[async_trait::async_trait]
+pub(crate) trait CommandOutput: Send {
+    /// Writes a progress update for the command.
+    async fn write_status(&mut self, message: &str) -> std::io::Result<()>;
+
+    /// Writes UTF-8 report output for the command.
+    async fn write_report(&mut self, output: &str) -> std::io::Result<()>;
```
thieman: The status and report nouns are odd here, did you get them from somewhere? Do they correspond to anything more standard? Like are these equivalently stdout and stderr? Or what is the difference?

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -622,30 +816,459 @@ fn get_stylized_table() -> Table {
...
+        output.write_report(&line).await?;
+        output.write_report("\n").await?;
     }
-    stdout.flush().await?;
-    Ok(())
+    output.flush().await
+}
+
+pub(crate) const REMOTE_DOGSTATSD_COMMANDS: &[RemoteCommandDescriptor] = &[
+    RemoteCommandDescriptor {
```
thieman, line 856: Confirm the parameter names and helpers were all taken from the existing CLI implementation

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] Resolved in signed commit 9b8f388470. The non-Linux replay warning now goes through the injected command-progress output instead of directly to tracing. The direct adapter still logs it, while the remote adapter sends it as stdout progress. Added focused coverage for both routing paths; 303 ADP tests passed. I will make no further change for this thread.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] I reviewed these two helpers and will retain them. They encode the replay-session lifecycle invariant: do not issue a start request after cancellation, but once a start request is in flight, await its result so any created server-side session can always be finished. Their generic, isolated form allows direct cancellation/cleanup tests without an IPC mock or a real server. Inlining them would obscure that invariant and remove focused testability. I will make no further change for this thread.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] Yes, this is new and required for `replay --loops > 1`. `TrafficCaptureReader` reaches EOF after one iteration, so each subsequent loop must rewind before replaying packets again. Without it, every loop after the first would be empty. I will retain this behavior.

bin/agent-data-plane/src/cli/remote.rs
thieman: [GPT-5.6 Terra (OpenAI)] Resolved in signed commit 9b8f388470. I renamed the interface methods to `write_progress`, `write_stdout`, and `flush_stdout`. Progress is a human-facing lifecycle update: direct CLI logs it through tracing, while remote execution emits it as stdout. Stdout is the command report/table data written by both modes. Neither represents stderr; terminal failures are separately framed as stderr by the remote service. 303 ADP tests passed. I will make no further change for this thread.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] Confirmed. Every descriptor parameter name, short name, helper, type, and requiredness was transcribed from the existing Argh CLI definitions. The descriptor/parser tests assert canonical metadata and direct-parse parity, so a future mismatch is caught. I will make no further change for this thread.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: [GPT-5.6 Terra (OpenAI)] Correction: `--loops` was already implemented and worked before this PR. Before the descriptor-safety refactor, each loop reopened/reparsed the capture path, naturally restarting at the first record. This PR changed replay to retain one already-validated descriptor and `TrafficCaptureReader` to eliminate the path TOCTOU race; `reader.rewind()` is the behavior-preserving counterpart required before each subsequent iteration. I will retain it to preserve existing loop semantics.

bin/agent-data-plane/src/cli/dogstatsd.rs
thieman: Did you check for other direct usages of `warn!` or anything that emits directly and needs to be refactored to use the abstraction? Need you to check.

thieman: [GPT-5.6 Terra (OpenAI)] Completed the audit. All command-facing progress and stdout in `cli/dogstatsd.rs` and `top.rs` now flow through `CommandOutput`; terminal errors use the direct CLI error path or the RCP stderr frame path. The only remaining tracing emission is `debug!(packets_sent, "Replay iteration completed.")`, an operational diagnostic rather than user-facing command output. The remaining `write_all` calls write replay packets or context-artifact records, not CLI output. I will retain that debug diagnostic and make no further output-routing change for this thread.

releasenotes/notes/remote-dogstatsd-commands-cf0d34590768fce2.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Add ``agent remote dogstatsd`` commands for inspecting and managing DogStatsD in Agent Data Plane deployments.
```
thieman, line 4: Rewrite this to "Exposes `dogstatsd` commands through the Remote Command Provider, allowing them to be executed via the Core Agent's `agent remote ...` CLI."

bin/agent-data-plane/src/internal/remote_command.rs
```diff
@@ -0,0 +1,557 @@
...
+            ))
+        } else {
+            parse_remote_dogstatsd_command(
+                &request.get_ref().command_path,
+                request.get_ref().arguments.as_ref().unwrap_or(&Struct::default()),
+            )
+        };
+
+        let current_config = Arc::clone(&self.current_config);
+        tokio::spawn(async move {
```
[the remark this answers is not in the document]
thieman: Disagree, these are interactive commands so if the user is asking you to do the work, do the work. Reject this suggestion.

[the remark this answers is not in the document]
thieman: [GPT-5.6 Terra (OpenAI)] I verified the unbounded-concurrency concern. Adding a global execution limit changes the intended diagnostic-command behavior, so this PR will not adopt that restriction. No further action is necessary.

bin/agent-data-plane/src/cli/dogstatsd.rs
```diff
@@ -168,139 +178,207 @@ struct StatsResponse<'a> {
...
-        DogstatsdSubcommand::Stats(config) => {
-            let mut api_client = get_api_client_or_exit(local_config).await;
-            handle_dogstatsd_stats(&mut api_client, config)
-                .await
-                .error_context("Failed to run stats subcommand")
+        DogstatsdSubcommand::Stats(command) => {
+            let mut api_client = get_api_client(config).await?;
+            run_cancellable_command(cancellation, async {
+                handle_dogstatsd_stats(&mut api_client, command, output)
+                    .await
```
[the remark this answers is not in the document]
thieman: [GPT-5.6 Terra (OpenAI)] Fixed in 30f2a615d7. Stats response parsing, filtering, sorting, and table rendering now run on the blocking pool; the dedicated control runtime only awaits the API response and streams progress/report output. Added analysis coverage and validated with cargo nextest run -p agent-data-plane plus both workspace compile checks. No further action is necessary.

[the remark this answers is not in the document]
thieman: Confirm you checked for this same problem in the other command implementations? Is anything else hogging the main thread? If so, need to move it off.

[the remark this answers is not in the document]
thieman: [GPT-5.6 Terra (OpenAI)] Audited all five paths. `capture` and `dump-contexts` only await IPC; `top` already runs report parsing/rendering in `spawn_blocking`; `stats` is offloaded in 30f2a615d7. Replay was the remaining issue: 2fd27d089e moves tagger-state decoding into the blocking capture-load task and yields every 128 packets during sustained zero-delay replay. No equivalent CPU-heavy control-runtime work remains. No further action is necessary.

releasenotes/notes/remote-dogstatsd-commands-cf0d34590768fce2.yaml
thieman: [GPT-5.6 Terra (OpenAI)] Fixed in fe1e3c386b. The feature note now describes the `dogstatsd` Remote Command Provider surface and the Core Agent `agent remote ...` CLI. No further action is necessary.

PR #2617, opened 2026-09-15
fix(panoramic): retry stalled polling probes

aqian01:
## Summary

- cap each polling attempt at 10 seconds while preserving the assertion-wide deadline
- retry transient Docker-backed port, HTTP, and file probe stalls
- keep cancellation active while a probe attempt is pending
- add regression coverage for attempt deadlines, assertion deadlines, and cancellation

## Context

Windows endpoint and file assertions execute commands inside the target through the Docker Engine API. Each assertion checked its 60-second deadline only between attempts, while `create_exec`, `start_exec`, attached-output draining, and `inspect_exec` could remain pending for the Docker client default of 120 seconds or indefinitely after an upgraded attach response.

This has produced the same case-watchdog signature in three independent main-branch runs:

- [telemetry-endpoint on 2026-07-17](https://gitlab.ddbuild.io/DataDog/saluki/-/jobs/1868149389): 115.89 seconds pending in assertions before the 120-second case timeout
- [telemetry-endpoint on 2026-08-31](https://gitlab.ddbuild.io/DataDog/saluki/-/jobs/1998013223): 118.59 seconds pending in assertions before the 120-second case timeout
- [otlp-traces-enabled on 2026-09-15](https://gitlab.ddbuild.io/DataDog/saluki/-/jobs/2043239995): 118.15 seconds pending in assertions before the 120-second case timeout

A normal unavailable endpoint would return between attempts and fail at its configured 60-second assertion deadline. Remaining pending until the case-wide watchdog proves that one of the Docker-backed probe operations stalled instead. The same commits and endpoints pass surrounding Windows runs, so this is an intermittent harness failure rather than a product listener regression.

This change limits an individual polling attempt to 10 seconds and retries it within the original assertion budget. A brief Docker exec/attach stall is therefore discarded and retried, while a genuinely unavailable endpoint still fails at the configured assertion deadline.

## Test plan

- `make fmt`
- `cargo check -p panoramic --all-targets`
- `cargo nextest run -p panoramic polling::tests --no-fail-fast` (3 passed)
- `cargo nextest run -p panoramic --no-fail-fast` (119 passed)
- `make check-all`


bin/correctness/panoramic/src/assertions/file_contains.rs
```diff
@@ -95,10 +98,35 @@ impl Assertion for FileContainsAssertion {
...
+                PROBE_ATTEMPT_TIMEOUT,
+                &ctx.cancel_token,
+                &ctx.container_exit_token,
+                read_attempt,
+            )
+            .await
+            {
+                PollAttemptResult::Completed(result) => result,
+                PollAttemptResult::TimedOut => {
+                    trace!(path = %self.path, "File read attempt timed out, retrying...");
```
webern, line 119: This would give us something visual in CI to know when probe-retries were occurring in the failed test. It's not the prettiest, but it seems like a good idea to have that information when looking at the failure in CI.

> AI
>
> I think retrying a stalled probe is the right behavior, but could we preserve the timeout signal in the eventual assertion failure? As written, repeated Docker stalls end with the same message as a probe that completed normally and found the file missing.
>
> One relatively small option would be to count timed-out attempts and add the count to the existing failure message:
>
> ```rust
> let mut timed_out_attempts = 0;
>
> loop {
>     if Instant::now() > deadline {
>         let timeout_detail = match timed_out_attempts {
>             0 => String::new(),
>             1 => " (1 probe attempt timed out)".to_string(),
>             count => format!(" ({count} probe attempts timed out)"),
>         };
>
>         return AssertionResult {
>             name: self.name().to_string(),
>             passed: false,
>             message: match &self.pattern {
>                 None => format!(
>                     "File '{}' did not exist after {:?}{}.",
>                     self.path, self.timeout, timeout_detail
>                 ),
>                 Some(pattern) => format!(
>                     "Pattern '{}' not found in file '{}' after {:?}{}.",
>                     pattern, self.path, self.timeout, timeout_detail
>                 ),
>             },
>             duration: started.elapsed(),
>         };
>     }
>
>     // ...
>     match run_poll_attempt(/* ... */).await {
>         PollAttemptResult::TimedOut => {
>             timed_out_attempts += 1;
>             continue;
>         }
>         // ...
>     }
> }
> ```
>
> I think the same pattern would be useful for the HTTP and port probes. That would keep the retry behavior while letting someone reading the CI failure distinguish an ordinary unmet assertion from one affected by probe timeouts.

webern (commented):
I'm not super, super sure, because this is complicated to reason about; but I think this might make the high level diagnostic signal worse (even if it gives the test more of a chance of passing.

A proposed fix puts at least the probe timeout counts in the failure message so we know when looking at the failure that the harness was probably acting up. I think longer term we should have a more robust way of getting harness failures out in a way that is distinguishable from assertion failures.

Clankie explains:

```
Before this PR, a persistent stall produced:

 - whole-test outcome: TimedOut
 - active phase: assertions
 - evidence that a supposedly 60-second assertion was still running near 120 seconds

 That did not prove which Docker operation was stuck, but it strongly distinguished a harness-side hang from an ordinary assertion failure.

 After this PR, a persistent stall produces:

 - assertion outcome: Failed
 - the specific assertion and target, such as the file path or port
 - the same message produced when the target genuinely never becomes ready

 So localization is better, but root-cause classification is worse: “Docker repeatedly stalled” and “ADP never created the file” become indistinguishable in the normal report.
```

PR #2625, opened 2026-09-16
feat(agent-data-plane): add stateful metrics sender with persistent retries

aqian01:
## Summary

Add opt-in Foldspace metrics delivery through `data_plane.stateful_metrics_endpoint`. Count, rate, and gauge metrics use the new sender; other metric types keep their HTTP route.

```mermaid
flowchart TD
    subgraph ADP["ADP binary"]
        Enrich["metrics_enrich"] --> Router["stateful_metrics_route"]
        Router -->|"Other metric types"| HTTP["dd_metrics_encode → dd_out"]

        subgraph Sender["stateful_metrics destination — one worker task"]
            Convert["Convert ADP metrics to logical batches"]
            High["High-priority queue"]
            Low["Low-priority retry queue"]
            Disk["Disk storage"]
            Pump["pump(): select next logical batch"]
            Transport["gRPC transport"]

            Convert --> High
            High --> Pump
            Low --> Pump
            Low <--> Disk

            subgraph Foldspace["Foldspace library boundary — sans-I/O"]
                Core["StatefulMetricsClient"]
            end

            Pump -->|"push_batch()"| Core
            Core -->|"SendPayload"| Transport
            Transport -->|"ACK or failure"| Core
            Core -->|"Recovered logical batches"| Low
        end

        Router -->|"Count / rate / gauge"| Convert
    end

    Transport <-->|"gRPC stream"| Intake["Foldspace intake process"]
```

## Behavior

- **Opt-in routing:** Setting `data_plane.stateful_metrics_endpoint` routes count, rate, and gauge series through Foldspace. Other metric types retain their existing HTTP route. Without the setting, delivery is unchanged.
- **Ownership:** Each sender worker owns its own Foldspace core, gRPC stream, and retry queue. This PR starts one worker for one destination. ADP links Foldspace through Cargo without copying its source.
- **Responsibilities:** ADP owns transport, timers, credentials, and retry storage. Foldspace owns encoding, compression, dictionaries, and protocol state.
- **Flushing:** Batches flush at the configured series threshold, when the partial-batch timer expires, or during shutdown. New input does not extend an existing flush deadline. The sender allows up to eight unacknowledged payloads.
- **Retries:** Fresh input enters the high-priority queue; recovered batches and high-priority overflow enter the low-priority queue. Retry entries contain logical metrics and are encoded again against the current Foldspace state. Original ADP metric objects are not retained.
- **Failure handling:** Transient failures reconnect with exponential backoff from 250 ms to 30 seconds. Outstanding data times out after 30 seconds without acknowledgement progress. Authentication, capability, and protocol failures suspend delivery; invalid-argument failures abandon the affected unacknowledged batches. Delivery never automatically switches to HTTP.
- **Credential changes:** Updating the API key resets destination dictionary state, recovers outstanding logical work, and resumes suspended delivery.
- **Persistence and guarantees:** Optional disk persistence uses ADP’s existing retry-storage policy. Lost acknowledgements can cause duplicate delivery. Queue limits can cause drops, and crashes can lose data still held in memory.
- **Shutdown:** ADP attempts delivery for half the configured stop timeout, capped at 30 seconds. It then recovers buffered and unacknowledged work into retry storage and flushes the queue.
- **Current limits:** This is a plaintext integration path. TLS/proxy support, multiple workers, and explicit byte limits for Foldspace’s inflight and dictionary memory remain follow-ups.

## Testing

The tests run actual binaries:

```text
DogStatsD → ADP + embedded Foldspace core → gRPC → Foldspace intake → decoded JSONL
```

**TL;DR:** We sent **30,000 metric points** through the actual ADP and separate Foldspace intake binaries at a target **1,000 points/second**. Every decoded point matched the generated expectations for metric name, packet tag, timestamp, and value, with **no missing or extra points**, including a run that paused intake consumption for **12 seconds** and then drained the backlog.

For an independent HTTP comparison, we sent **six known gauge/counter points through ADP with Foldspace enabled**, then restarted ADP with Foldspace disabled and HTTP V3 enabled and replayed **the same inputs and timestamps**. The separate Foldspace intake and Saluki’s `datadog-intake` binary independently decoded the two outputs. **All six points matched** by metric name, tags, host, type, interval, timestamp, and value, including counter-to-rate conversion. The comparison ignored batching and ordering but preserved duplicate counts. This was separate from the 30,000-point load test, which compared against generated expectations. Resources beyond host and origin metadata were checked separately against explicit expectations because the HTTP comparison model omits them.

**All 11 binary scenarios passed**, covering metadata, timer/threshold/shutdown flushing, restarts, stream rejection, stalled consumption, and lost or timed-out acknowledgements. The ACK-failure cases produced the expected duplicates on replay. These are short local correctness tests, not a production-scale benchmark.

**All required source and test files are committed in this PR.** The binary results below used the same dependency, telemetry fix, and executable harness code now checked in at `1233fd5dcf`. The harness is experimental local integration-test tooling; it is not part of the ADP runtime or a CI job. ADP links Foldspace through Cargo; no source is copied into Saluki.

### Prerequisites

Validated on macOS ARM64 with Rust 1.94.0 and Python 3.14. The harness requires Python 3.11+, OpenSSL, and local socket access; building requires a C/C++ toolchain, CMake, pkg-config, and `protoc`. Install Git, GitHub CLI, and rustup as part of the normal development setup.

You need access to the private [Foldspace repository](https://github.com/DataDog/foldspace) and the Datadog testing registry at `depot-read-api-rust.us1.ddbuild.io` through the internal network/VPN. The registry configuration is already in each checkout.

Related changes:

- [Foldspace #80: caller-driven buffering and flushing](https://github.com/DataDog/foldspace/pull/80) — tested revision `9d4b57edee8095b2fc827b6480dcda39df749a4f`.
- [Foldspace #79: separate metrics intake and decoder](https://github.com/DataDog/foldspace/pull/79) — included in that revision.

Test tooling and documentation:

- [Binary harness](https://github.com/DataDog/saluki/blob/1233fd5dcf7551750b24b05ab232b274e6f6c45e/ci/tooling/test-stateful-metrics-binaries.py): starts the binaries, sends metrics, checks decoded output, and cleans up processes.
- [Fault proxy](https://github.com/DataDog/saluki/blob/1233fd5dcf7551750b24b05ab232b274e6f6c45e/ci/tooling/stateful_metrics_proxy.py): injects stream failures and pauses traffic or acknowledgements.
- [Python requirements](https://github.com/DataDog/saluki/blob/1233fd5dcf7551750b24b05ab232b274e6f6c45e/ci/tooling/stateful-metrics-requirements.txt): pins the proxy's HTTP/2 dependency.
- [Testing guide](https://github.com/DataDog/saluki/blob/1233fd5dcf7551750b24b05ab232b274e6f6c45e/docs/development/stateful-metrics.md#repeatable-binary-tests): usage, scenarios, artifacts, and limitations.

The commands below run the scripts directly from the checkout. No extraction or source patch is needed.

### Steps to reproduce

Run these blocks in the same shell. All paths are created below; no author-specific checkout paths are required.

1. Check tools and GitHub authentication. If `gh auth status` fails, run `gh auth login --hostname github.com` first.

   ```sh
   command -v git gh rustup cc cmake pkg-config protoc python3 openssl
   python3 -c 'import sys; assert sys.version_info >= (3, 11), "Python 3.11+ required"'
   gh auth status --hostname github.com
   gh auth setup-git --hostname github.com
   rustup toolchain install 1.94.0 --profile minimal --component clippy --component rustfmt
   FOLDSPACE_TEST_ROOT="$(mktemp -d /tmp/adp-foldspace-review.XXXXXX)"
   ```

2. Check out this PR and the exact Foldspace revision. Use [Foldspace PR #80](https://github.com/DataDog/foldspace/pull/80)'s revision for both the ADP library dependency and the intake binary; it already contains [Foldspace PR #79](https://github.com/DataDog/foldspace/pull/79), which adds the metrics intake.

   ```sh
   git clone https://github.com/DataDog/saluki.git "$FOLDSPACE_TEST_ROOT/saluki"
   cd "$FOLDSPACE_TEST_ROOT/saluki"
   gh pr checkout 2625
   git checkout --detach 1233fd5dcf

   git clone https://github.com/DataDog/foldspace.git "$FOLDSPACE_TEST_ROOT/foldspace"
   git -C "$FOLDSPACE_TEST_ROOT/foldspace" fetch origin 9d4b57edee8095b2fc827b6480dcda39df749a4f
   git -C "$FOLDSPACE_TEST_ROOT/foldspace" checkout --detach 9d4b57edee8095b2fc827b6480dcda39df749a4f
   ```

3. Build the actual binaries. The separate target directories make their paths explicit. Do not add `--offline` on a fresh checkout.

   ```sh
   cd "$FOLDSPACE_TEST_ROOT/foldspace"
   CARGO_TARGET_DIR="$FOLDSPACE_TEST_ROOT/foldspace-target" \
     cargo build --locked -p foldspace-grpc-server --bin foldspace-intake

   cd "$FOLDSPACE_TEST_ROOT/saluki"
   CARGO_TARGET_DIR="$FOLDSPACE_TEST_ROOT/saluki-target" \
     cargo build --locked --bin agent-data-plane --bin datadog-intake

   python3 -m venv "$FOLDSPACE_TEST_ROOT/venv"
   "$FOLDSPACE_TEST_ROOT/venv/bin/python" -m pip install \
     -r "$FOLDSPACE_TEST_ROOT/saluki/ci/tooling/stateful-metrics-requirements.txt"
   ```

4. Run the complete 11-scenario suite. This uses the five-second consumption pause from the original full run. Ports 2049/TCP and 9125/UDP must be available for the HTTP comparison; other listener ports are selected automatically.

   ```sh
   "$FOLDSPACE_TEST_ROOT/venv/bin/python" "$FOLDSPACE_TEST_ROOT/saluki/ci/tooling/test-stateful-metrics-binaries.py" \
     --adp "$FOLDSPACE_TEST_ROOT/saluki-target/debug/agent-data-plane" \
     --intake "$FOLDSPACE_TEST_ROOT/foldspace-target/debug/foldspace-intake" \
     --http-intake "$FOLDSPACE_TEST_ROOT/saluki-target/debug/datadog-intake" \
     --load-pause-seconds 5 \
     --output "$FOLDSPACE_TEST_ROOT/results"
   ```

   The harness generates the ADP configuration and IPC certificate, uses a dummy API key and loopback endpoints, starts intake in `INTAKE_MODE=metrics`, waits for readiness, sends timestamped DogStatsD packets, checks decoded output, injects failures, and stops the processes. The HTTP comparison replays the same inputs through ADP's HTTP V3 path and an independent decoder.

   Expect 11 `PASS` lines and exit status 0. Case names are `timer_and_metadata`, `http_comparison`, `threshold`, `shutdown`, `reconnect`, `disk_restart`, `rejected_stream`, `lost_ack`, `stalled_reads`, `ack_timeout`, and `sustained_load`. Use repeated `--case NAME` arguments to select cases. The ACK-timeout case deliberately waits about 30 seconds.

5. Repeat the load case with a 12-second pause to fill the inflight window and exercise queue growth, then inspect both result files:

   ```sh
   "$FOLDSPACE_TEST_ROOT/venv/bin/python" "$FOLDSPACE_TEST_ROOT/saluki/ci/tooling/test-stateful-metrics-binaries.py" \
     --adp "$FOLDSPACE_TEST_ROOT/saluki-target/debug/agent-data-plane" \
     --intake "$FOLDSPACE_TEST_ROOT/foldspace-target/debug/foldspace-intake" \
     --case sustained_load \
     --load-pause-seconds 12 \
     --output "$FOLDSPACE_TEST_ROOT/load-12s"

   python3 -m json.tool "$FOLDSPACE_TEST_ROOT/results/result.json"
   python3 -m json.tool "$FOLDSPACE_TEST_ROOT/load-12s/result.json"
   ```

   Expect all statuses to be `passed`. Each output directory contains configurations, process logs, decoded JSONL journals, telemetry, and binary SHA-256 hashes. Load cases also write `sustained_load/load-samples.json`. Use a new output directory for reruns; the harness refuses to overwrite an existing directory.

6. Run the focused Rust tests and workspace checks:

   ```sh
   cd "$FOLDSPACE_TEST_ROOT/saluki"
   cargo +1.94.0 install cargo-nextest --version 0.9.99 --locked
   export CARGO_TARGET_DIR="$FOLDSPACE_TEST_ROOT/saluki-target"

   cargo check --workspace --locked
   cargo check --workspace --tests --locked
   cargo nextest run --locked \
     -p agent-data-plane -p agent-data-plane-config-system \
     -E 'test(stateful_metrics)'
   cargo clippy --all-targets --workspace --locked -- -D warnings
   ```

   Expect 26 focused tests to pass. Repository formatting, documentation, unused-dependency, and release-note checks also passed locally. The aggregate dependency/license checks have the known blockers listed below.

### Results

**All 11 binary scenarios passed**, followed by a stronger load run:

| Test | Result |
|------|--------|
| Sparse input and dictionary reuse | Timer flush delivered correct values, timestamps, tags, resources, and origin. ACK counters survived an idle gap. |
| HTTP V3 comparison | All six points matched by name, tags, host, type, interval, timestamp, and value, independent of batching and ordering. |
| Threshold and shutdown flush | Two series flushed before a 60-second timer; a partial batch delivered on SIGTERM and ADP exited cleanly. |
| Intake and ADP restart | Delivery resumed after intake restart; persisted logical retries delivered after ADP restart. |
| Stream rejection and stalled consumption | Rejected streams recovered; all 600 points delivered after paused consumption resumed. |
| Lost ACK and ACK timeout | Both recovered by replaying the unacknowledged point, including the expected duplicate. The deadline fired after approximately 31 seconds. |
| Sustained load | 30,000 points at a target 1,000 points/second all decoded correctly, with no loss, duplicates, reported drops, or stream failures. |

The stronger load run paused intake consumption for **12 seconds**. Fresh-work backlog reached **four batches** and drained to zero. Peak sampled RSS was **68.3 MiB for ADP** and **22.6 MiB for intake**. The initial full suite used a five-second pause; the current harness defaults to 12 seconds and verifies backlog growth.

Testing found a telemetry bug introduced by this PR: the new sender used temporary counter handles, which ADP’s existing idle cleanup could evict. Commit `497aebf324` retains the handles on the worker and attaches component labels. The observed failure affected counter reporting; metric delivery succeeded. The actual-binary regression verifies that ACK counters survive an idle gap and continue increasing.

Other local checks passed: **26 focused Rust tests**, workspace checks including tests, formatting, workspace Clippy, documentation, unused dependencies, release-note validation, and Python syntax checks. Full `make test-all` was not run.

## Remaining blockers and scope

- Public Saluki CI cannot fetch internal Foldspace; local binary testing is the current validation path.
- `make check-all` stops at `check-deny` because the `patterns` snapshot lacks license metadata. `make check-licenses` also reports missing repository metadata.
- These are short macOS ARM64 debug-binary tests. The fault proxy verifies client recovery; production metrics-server load shedding and staging shadow comparison still need separate validation.
- Plaintext testing only. TLS/proxy support, sharding, and core memory byte limits remain future work. Disk persistence follows the existing spill/shutdown policy; crashes can lose in-memory data and lost acknowledgements can cause duplicates.

[Foldspace stack](https://github.com/DataDog/foldspace/pull/80) · [Metrics intake](https://github.com/DataDog/foldspace/pull/79)


bin/agent-data-plane/src/cli/run.rs
```diff
@@ -462,8 +468,34 @@ async fn add_baseline_metrics_pipeline_to_blueprint(
...
-        // Metrics, then forwarding.
-        .connect_components_in_order(["metrics_enrich", "dd_metrics_encode", "dd_out"])?;
+        .connect_components("dd_metrics_encode", "dd_out")?;
+
+    if let Some(endpoint) = &config.domains.stateful_metrics.endpoint {
+        if !shared.endpoints.additional_endpoints.is_empty() {
+            return Err(generic_error!(
+                "stateful metrics testing does not support additional_endpoints"
+            ));
+        }
```
tobz: We can't throw an error here since it would just blow up the process at startup for anyone using `additional_endpoints`.

Capture this boolean and fold it into the above check: `if stateful_metric_endpoint && has_no_additional_endpoints`, etc. Make it also log a warning that stateful metrics mode is incompatible with `additional_endpoints` being declared.

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -462,8 +468,34 @@ async fn add_baseline_metrics_pipeline_to_blueprint(
...
+        blueprint
+            .add_transform("stateful_metrics_route", StatefulMetricsRouterConfiguration)?
```
tobz: Can we call this `dd_stateful_metrics_router`?

bin/agent-data-plane/src/cli/run.rs
```diff
@@ -462,8 +468,34 @@ async fn add_baseline_metrics_pipeline_to_blueprint(
...
+            .add_destination(
+                "stateful_metrics",
```
tobz: And call this `dd_stateful_metrics_out`?

bin/agent-data-plane/src/components/stateful_metrics/conversion.rs
```diff
@@ -0,0 +1,93 @@
...
+    let points = points
+        .filter_map(|(timestamp, value)| {
+            let value = interval
+                .filter(|interval| !interval.is_zero())
+                .map_or(value, |interval| value / interval.as_secs_f64());
+            value
+                .is_finite()
+                .then(|| MetricPoint::new(timestamp.map_or(0, |ts| ts.get() as i64), value))
+        })
+        .collect();
```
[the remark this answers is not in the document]
tobz: I agree with the clanker here: we should exclude the metric entirely if it lacks any finite values/

bin/agent-data-plane/src/components/stateful_metrics/conversion.rs
```diff
@@ -0,0 +1,93 @@
...
+                    }
+                }
+            }
+            _ => {}
+        }
+    }
+    if let Some(device) = device {
+        let index = usize::from(metric.context().host().is_some_and(|host| !host.is_empty()));
+        resources.insert(index, MetricResource::new("device", device));
+    }
```
tobz: What's going on here? 🤔

lib/saluki-components/src/forwarders/queue.rs
```diff
@@ -0,0 +1,65 @@
+//! Shared priority scheduling and persistent retry storage for delivery components.
```
tobz: Seems like this should be under `common::datadog`?

releasenotes/notes/stateful-metrics-testing-ff678cfdeac5587d.yaml
```diff
@@ -0,0 +1,7 @@
+features:
+  - |
+    Add an opt-in plaintext gRPC path for testing stateful metric series delivery
+    with ``data_plane.stateful_metrics_endpoint``. Sketches and sets continue
+    over HTTP. Stateful retries use the existing low-priority queue and optional
+    disk storage; delivery does not automatically fall back to HTTP. This experimental path is disabled by default and is not intended
+    for production traffic.
```
tobz:
```suggestion
    Add experimental support for a new stateful transport that reduces the
    overhead of sending metrics to the Datadog backend. Currently, only
    non-distribution metric types (counts, rates, gauges) are supported. 
```

Basically, we don't intend for users to touch this at all yet, so we don't need to bother telling them if it's enabled/disabled by default, or the configuration setting to use, etc.... just that it exists.

lib/agent-data-plane-config-system/src/saluki_only.rs
```diff
@@ -239,6 +239,8 @@ pub struct SalukiOnly {
 #[derive(Clone, Debug, Default, Deserialize)]
 #[serde(default)]
 pub struct DataPlane {
+    /// Experimental plaintext gRPC metrics endpoint; unset disables stateful delivery.
+    pub stateful_metrics_endpoint: Option<String>,
```
tobz: We should stick this under `data_plane.experimental.stateful_metrics.endpoint`.

Raymond already added the `data_plane.experimental` bit for the configuration he added for metrics endpoint routing, so if you rebase on the latest `main`, you'll get that part for free.

tobz (changes requested):
Still reviewing this but throwing out some feedback early.

bin/agent-data-plane/src/cli/run.rs
aqian01: Ah this should be handled in future foldspace PRs, i'll make the changes there then just enable stateful + additional endpoints

bin/agent-data-plane/src/components/stateful_metrics/conversion.rs
aqian01: Oops mb missed this, AI slop, i think it got confused with v3.

PR #2632, opened 2026-09-16
fix(smp): restore v3 series encoding in benchmarks

rayz:
## Summary

Restores V3 series encoding in ADP SMP benchmarks.
[#2562](https://github.com/DataDog/saluki/pull/2562) removed the old experimental V3 setting, but SMP configs still referenced it. Since the mock intake is localhost, the ignored setting left series on V2 instead of the intended V3.

- Replace the retired variable with DD_USE_V3_API_SERIES_ENDPOINTS using the supported JSON map.
- Regenerate all 60 experiment configs.
- No runtime code changes

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Performance


jszwedko (approved):
Good catch!

webern (approved):
Thanks, I must have missed this regression in the schema/config update PR.

PR #2644, opened 2026-09-18
feat(traces): support probabilistic hash seeds and full trace IDs

### LINKED ISSUE #2406: Traces(otlp+dd): Support probabilistic hash seeds and full trace IDs
## Background

This sampler exists so ADP can run beside the OpenTelemetry probabilistic sampler processor and keep the same traces. A different hash seed or a different number of trace-ID bits puts the same trace in a different bucket, so the two components disagree and parts of a distributed trace go missing.

## Agent behavior

`NewProbabilisticSampler` writes `conf.ProbabilisticSamplerHashSeed` into four little-endian bytes and turns on `fullTraceIDMode` when `apm_config.features` contains `probabilistic_sampler_full_trace_id`. `SampleV1` (`probabilistic.go:92-116`) builds a 16-byte buffer: in legacy mode it copies the low 64 bits of the trace ID into the first eight bytes and leaves the rest zero, and in full mode it copies all sixteen. It then writes the seed and the buffer into an FNV-1a 32 hash and keeps the trace when the low 14 bits are below the scaled sampling percentage.

## ADP behavior

`ProbabilisticSampler::sample` (`probabilistic.rs:26-42`) already reproduces the Agent's legacy mode exactly: a zero four-byte seed, `[low64 || zeros]` as the hash input, FNV-1a 32, and the same 14-bit mask. It takes the trace ID as a `u64`, so there is no way to pass the high bits, and no configuration field carries a seed. `Trace` already stores `trace_id_high` alongside `trace_id_low`.

## Configuration keys

- [ ] apm_config.probabilistic_sampler.hash_seed

## Acceptance

- [ ] A configured `hash_seed` changes which traces are kept at a fixed rate
- [ ] Full-trace-ID mode hashes `trace_id_high` and `trace_id_low` as 16 bytes
- [ ] Legacy mode keeps today's `[low64 || zeros]` hash input
- [ ] The `probabilistic_sampler_full_trace_id` flag selects the mode

lucastemb:
## Summary

At different stages of a traces lifecycle, there is a predetermined probability that dictates the likelihood that a trace is retained or dropped. This PR ensures that of the `X%` of traces kept, the same traces are retained across all samplers. 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Performance

## How did you test this PR?

Unit tests

## References

- Closes: https://github.com/DataDog/saluki/issues/2406


lib/saluki-components/src/transforms/trace_sampler/mod.rs
```diff
@@ -53,7 +53,8 @@ const ERROR_SAMPLE_RATE: f64 = 1.0; // Default extra sample rate (matches agent'
 const KEY_SPAN_SAMPLING_MECHANISM: &str = "_dd.span_sampling.mechanism";
 const KEY_ANALYZED_SPANS: &str = "_dd.analyzed";
 
-// Decision maker values for `_dd.p.dm` (matching datadog-agent).
+/// Feature flag selecting full-trace-ID hashing in the probabilistic sampler.
+const FEATURE_PROBABILISTIC_SAMPLER_FULL_TRACE_ID: &str = "probabilistic_sampler_full_trace_id";
```
tobz: I realize this is a little bit pedantic, but, I think we should make this an enum.

Basically, over in `agent_data_plane_config::domains::traces`, add an enum called something like `ApmFeature` which has specific variants for the feature IDs we know about, and then a single variant called `Other` that just holds a `String` internally with the value. Make it implement `PartialEq` and what not.

The enum type should have a `From<&str>` conversion that maps the given string ID (like the one above) to the right variant _or_ returns an `Other` variant. At the DDA config/domain config boundary, we'd just use the aforementioned conversion impl to translate over.

Finally, in places like this file where we want to check if a feature is enabled, we just do `traces.features.contains(&ApmFeature::ProbabilisticSamplerFullTraceId)` instead. This ensures we have a strongly-typed set of features to utilize throughout the codebase _and_ that all of those usages are defined centrally in a single place.

tobz (changes requested):

lib/agent-data-plane-config/src/domains/traces.rs
```diff
@@ -41,13 +84,11 @@ pub struct Domain {
...
-    /// probabilistic sampler from hashing the low 64 bits of the trace ID to hashing the full
-    /// 128-bit ID. Empty by default. Values other than the recognized flag are ignored without a
-    /// warning: the flag inventory is not stable across versions, so unrecognized entries are
-    /// inert rather than errors. Enable the flag on every probabilistic sampler in the ingestion
-    /// path so they keep the same traces.
-    pub features: Vec<String>,
+    /// Empty by default. Unrecognized values are carried as `ApmFeature::Other` and ignored
+    /// without a warning: the flag inventory is not stable across versions. Enable
+    /// `ApmFeature::ProbabilisticSamplerFullTraceId` on every probabilistic sampler in the
+    /// ingestion path so they keep the same traces.
```
tobz, line 90: The portion of this comment starting with `Enable .... on every` seems overly specific to mention here, and should be removed... especially since the line right before it talks about how these aren't stable... and then we go on to give stable-ish sounding advice. 😅

tobz (approved):
One non-blocking nit, but otherwise LGTM.

PR #2654, opened 2026-09-21
fix(ddsketch): preserve counts and window placement in collapsing stores

### LINKED ISSUE #2596: Canonical DDSketch drops counts on large upward jumps and panics querying maximum quantile
A large upward value jump in canonical DDSketch's default `CollapsingLowestDenseStore` drops the new observation from its bins while incrementing the cached count. A subsequent maximum-quantile query panics with `rank out of bounds on non-empty sketch`.

Reproduced on current upstream HEAD `1b258af6cb167463d63dbe158d414d17be5214bd`. The reproduction uses an unmodified vendored `lib/ddsketch`; all 26 files under that directory were verified against the Git blob hashes at that commit.

```rust
use ddsketch::canonical::{DDSketch, store::Store};

#[test]
fn large_upward_jump_preserves_counts_and_maximum_quantile() {
    let mut h = DDSketch::with_relative_accuracy(1.0 / 1024.0).unwrap();
    h.add_n(1.0, 100);
    h.add(((1u64 << 30) - 1) as f64);
    assert_eq!(h.count(), 101);

    // Currently 100, despite the sketch reporting 101 observations.
    let stored_count: f64 = h.positive_store()
        .to_proto()
        .contiguousBinCounts
        .iter()
        .sum();
    assert_eq!(stored_count, 101.0);

    // If the preceding assertion is removed, this panics.
    assert!(h.quantile(1.0).is_some());
}
```

Observed independently in a benchmark contract test that asserts the existing failure: `count() == 101`, serialized contiguous bins sum to `100.0`, and `catch_unwind(|| h.quantile(1.0))` returns an error. The panic originates at `lib/ddsketch/src/canonical/sketch.rs:135`.

The apparent cause is in [collapsing_lowest.rs](https://github.com/DataDog/saluki/blob/1b258af6cb167463d63dbe158d414d17be5214bd/lib/ddsketch/src/canonical/store/collapsing_lowest.rs): `grow()` requests a collapse over more indices than currently allocated. `collapse_lowest()` clamps the shift to `bins.len() - 1`; with one allocated bin it does not advance the offset. Growth then caps the vector at the configured 2,048 bins while the new high index is still outside the range. `bin_index()` returns `None`, so `add()` skips storing the observation but still increments `self.count`. The maximum rank subsequently has no matching bin.

Expected: preserve all counts and answer valid quantiles without panicking. Loss of relative accuracy for collapsed low ranks is expected behavior; a cached/bin count mismatch and panic are separate correctness failures. Non-collapsing dense and sparse stores avoid this growth path in the benchmark audit.

Useful regression coverage would include large upward jumps from one/few allocated bins, reverse insertion order, bin-count conservation, and successful maximum-rank queries.

This report was prepared with AI assistance; the reproducer was executed and the upstream source identity checked before filing. Searches of open and closed issues for DDSketch, collapsing, and the exact panic found no matching report; #2063 concerns the separate agent/OTLP interpolation path.

tobz:
## Summary

This PR fixes an issue with the collapsing logic in the two canonical dense store variants in `ddsketch`.

At a high-level, the collapsing dense stores make an accuracy/size trade-off: when we get an observation that requires a bucket index that doesn't fit within the established index range we're handling, collapse the buckets on one end (either left or right for "lowest" or "highest", respectively) such that whatever bucket index needed for the incoming observation can fit without needing to store more buckets than the configured maximum.

During the original porting of `sketches-go` to `ddsketch`, we did this incorrectly: instead of simply calculating the new index that out-of-bounds buckets needed to be collapsed into, we calculated a "shift distance" -- how many bins need to be collapsed? -- and then passed that to the collapse function, which then _clamped_ that distance... reasonably so, since we know we have a bounded number of buckets. The problem with clamping is that it becomes invalid if the number of buckets being collapsed/shifted is greater than the maximum number of buckets for the store. Continuing from there, with our improperly-calculated index, we would fail to actually insert the observation _but_ we still updated the summary statistics to say the number of samples had grown by one. This all culminated in a call to `quantile` panicking because its logic utilizes the sample count and tallies up the values in all buckets... but if it tallies every single bucket and only observed N samples, when we are supposed to have N+1 samples... then that's considered an invalid state, and it fell through to an `unreachable!()` statement. Doh. 😅

In this PR, we've fixed this by simply doing what we alluded to above: calculate the new bounded index on whatever side of the store we're collapsing. This lets the collapsing code focus more directly on collapsing anything above/below that index, which inherently doesn't require clamping since it doesn't involve having to go past the _current_ bucket count. We've additionally updated the tests to match what is being tested in `sketches-go` since we had a false sense of security due to only testing non-collapsing inputs _or_ inputs that didn't cause us to have to collapse _enough_ buckets to trigger this bug.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance

## How did you test this PR?

New unit tests.

## References

DADP-2

Closes #2596


tobz:
* **#2654** <a href="https://app.graphite.com/github/pr/DataDog/saluki/2654?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/saluki/2654?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

aqian01 (approved):

